PR #43, opened 2012-10-31
FeatureCollection cleanup

jodygarnett:
This pull request is taken from the following change proposal:

http://docs.codehaus.org/display/GEOTOOLS/FeatureCollection+revolution

This is an API change for FeatureCollection:
- Remove CollectionListener methods (addListener, removeListener)
- Remove Iterator Methods (iterator, close, purge)
- Remove Collection Methods (add,adAll,clear,remove,removeAll,retainAll)

This has the consequence of hitting most of our FeatureCollectionImplementations.

This pull request should not be accepted until:
- pull request for GeoServer is in place
- pull request for uDig is in place

Warning:
- I could not cleanly fix PropertyValueCollection as implemented, as such ValueCollectionTypeBindingTest has an eXcluded test

Upgrade instructions are provided in the pull request (as part of update.rst) and the code examples reproduced on the wiki page for handy cut and paste reference when fixing GeoServer and uDig.


bradh:
@jodygarnett Looks like the PR went sour. Can this branch be deleted?

PR #197, opened 2013-05-27
GEOT-4472 - doing a search for primary key column name

jdeolive:
Rather than assuming 'fid' will work.

### DIFF
--- a/modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
+++ b/modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
@@ -1935,7 +1935,31 @@ public final class JDBCDataStore extends ContentDataStore
             }
         }
         
-        return createTableSQL(featureType.getTypeName(), columnNames, sqlTypeNames, nillable, "fid", featureType);
+        return createTableSQL(featureType.getTypeName(), columnNames, sqlTypeNames, nillable, 
+            findPrimaryKeyColumnName(featureType), featureType);
+    }
+
+    /*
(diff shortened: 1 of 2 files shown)


mrpiggi:
```findPrimaryKeyColumnName``` always returns ```"fid"``` after the first loop over ```featureType.getAttributeDescriptors()```. IMHO the method
```java
    protected String findPrimaryKeyColumnName(SimpleFeatureType featureType) {
        String[] suffix = new String[] {"", "_1", "_2"};
        String[] base = new String[] {"fid", "id", "gt_id", "ogc_fid"};

        for (String b : base) {
            O:
            for (String s : suffix) {
                String name = b + s;
                for (AttributeDescriptor ad : featureType.getAttributeDescriptors()) {
                    if (ad.getLocalName().equalsIgnoreCase(name)) {
                        continue O;
                    }
                }
                return name;
            }
        }
        // practically should never get here, but just fall back and fail later
        return "fid";
    }
```
should be changed to
```java
    protected String findPrimaryKeyColumnName(SimpleFeatureType featureType) {
        String[] suffix = new String[] {"", "_1", "_2"};
        String[] base = new String[] {"fid", "id", "gt_id", "ogc_fid"};

        for (String b : base) {
            O:
            for (String s : suffix) {
                String name = b + s;
                for (AttributeDescriptor ad : featureType.getAttributeDescriptors()) {
                    if (ad.getLocalName().equalsIgnoreCase(name) && !ad.isNillable()) {
                        return name;
                    }
                }
            }
        }
        // practically should never get here, but just fall back and fail later
        return "fid";
    }
```

PR #406, opened 2014-03-31
Complex feature parsing

riniangreani:
I am resurrecting Adam  Brown's unpushed work in https://github.com/Adam-Brown/geotools/tree/gml_client_lib. I think at that time he was waiting for Gabriel's wfs-ng branch to be pushed before he could proceed, but I can see the dependencies are all in master now. 

Proposal: http://docs.codehaus.org/display/GEOTOOLS/ComplexFeature+Parsing+and+Building+Support

Going through someone else's code is always a pain. There are some changes I don't quite understand, here are some relevant threads (very old):
http://osgeo-org.1560.x6.nabble.com/Migrate-ComplexFeature-Parsing-work-off-wfs-ng-td4991249.html
http://osgeo-org.1560.x6.nabble.com/Parsing-and-Building-Creating-a-complex-feature-type-out-of-a-XML-schema-td4998747.html
http://osgeo-org.1560.x6.nabble.com/Request-approval-for-changes-to-gt-wfs-org-geotools-data-wfs-WFSDataStoreFactory-java-tc4557246.html#a4557603

gt-main:
======
I got rid of some unnecessary changes, including "ComplexContentFeatureSource" and
"WFSContentComplexFeatureCollection"  based on Andrea's feedback: https://www.mail-archive.com/geotools-devel@lists.sourceforge.net/msg29109.html

gt-app-schema:
============
I've had a look at the app-schema changes; they are minor and they're fine. 

wfs-ng:
======
There are a lot of test failures on wfs-ng, but they are existing failures. There is only 1 new test class in wfs-ng relating to this, I think: XmlComplexFeatureParserTest, and it passes.


jodygarnett:
@riniangreani we can pick this up when you come back.

modules/library/main/src/main/java/org/geotools/data/AbstractDataAccessFactory.java
```diff
@@ -0,0 +1,189 @@
...
+import org.geotools.parameter.DefaultParameterDescriptorGroup;
+import org.geotools.parameter.FloatParameter;
+import org.geotools.parameter.Parameter;
+import org.opengis.parameter.ParameterDescriptorGroup;
+import org.opengis.parameter.ParameterValue;
+
+/**
+ * This implementation was lifted from AbstractDataStoreFactory.
+ * As there was nothing in the implementation that was Store-specific (i.e. a restriction on simple types)
+ * this was raised to allow the factory to be used for complex features.
```
jodygarnett: That sounds fine, if this class provides value I do not mind it being around. When rewriting the ContentDataStore tutorial we ended setting up a few utility methods (rather than an entire base class).

modules/library/main/src/main/java/org/geotools/data/AbstractDataAccessFactory.java
```diff
@@ -0,0 +1,189 @@
...
+     *     }
+     *     // example check
+     *     File file = (File) DIRECTORY.lookup( params ); // DIRECTORY is a param
+     *     return file.exists() && file.isDirectory();
+     * }
+     * </code></pre>
+     * @param params
+     * @return true if params is in agreement with getParametersInfo, override for additional checks.
+     */
+    public boolean canProcess( Map params ) {
```
jodygarnett: This functionality is the only good bit and is available here: http://docs.geotools.org/latest/javadocs/org/geotools/data/DataUtilities.html#canProcess(java.util.Map, org.geotools.data.DataAccessFactory.Param[])

modules/library/main/src/main/java/org/geotools/data/AbstractDataAccessFactory.java
```diff
@@ -0,0 +1,189 @@
...
+
+    /**
+     * Returns the implementation hints. The default implementation returns en empty map.
+     */
+    public Map<java.awt.RenderingHints.Key, ?> getImplementationHints() {
+        return Collections.EMPTY_MAP;
+    }
+}
+
+class ParamDescriptor extends DefaultParameterDescriptor {
```
jodygarnett: Not too sure about this, we do have:

* http://docs.geotools.org/latest/javadocs/org/geotools/data/DataAccessFactory.Param.html
* http://docs.geotools.org/latest/javadocs/org/geotools/data/Parameter.html

What would be the reason to use ParameterDescriptor and ParameterDescriptorGroup?

If we want to proceed this direction a change proposal will be needed (so we get a consistent vision and a migration plan).

jodygarnett:
@NielsCharlier were you going to have a look at this one?

NielsCharlier:
Continued in https://github.com/geotools/geotools/pull/841

PR #443, opened 2014-04-29
OGR native bindings - remove load library calls [GEOT-5323]

pmauduit:
The main idea behind this PR is to avoid calls to System.loadLibrary()

(This one was by mistake originally opened onto the fork from boundlessgeo, I apologize for the noise)

In some special but common cases, several different calls to System.loadLibrary using the same lib name as argument can lead to undesired behaviours (Raising an exception "java.lang.UnsatisfiedLinkError: library already loaded in classloader").

A typical case of specific classloading is when using the library in several different webapps nested into the same servlet container (tomcat). Even if the library is actually correctly configured and fully available, GeoTools reports it as unavailable (due to the exception thrown).

A PoC which tends to prove this issue has been provided here: https://github.com/pmauduit/jai-imageio-sample/tree/master/jai-imageio-gdal-sample-webapp

In this PoC webapp, a first approach makes use of the JNI GDAL/OGR bindings directly, another one is then attempted using GeoTools (class JniOGRDataStoreFactory). The first one (once tomcat is correctly configured) shows that the library is fully available, though it is reported as unavailable by GeoTools. Duplicating the webapp (so that we can have several contexts trying to use the same native lib) shows that the second one is also unable to access the features that GDAL provides.

The litterature I could read on the topic (classloading in Java seems to be one of the big mistery around the language, though) explains that we should avoid packaging webapps with jar files making use of the System.loadLibrary calls. These have to be shared: the jar file should be made available in the higher-level classloader (for tomcat under debian-like distros with libgdal-java installed, it means having a symlink from ${CATALINA_HOME}/lib/gdal.jar pointing to /usr/share/java/gdal.jar).

### DIFF
--- a/modules/plugin/ogr/ogr-jni/pom.xml
+++ b/modules/plugin/ogr/ogr-jni/pom.xml
@@ -16,7 +16,7 @@
   <parent>
     <groupId>org.geotools</groupId>
     <artifactId>gt-ogr</artifactId>
-    <version>12-remloadlibrary</version>
+    <version>12-SNAPSHOT</version>
   </parent>
 
 
--- a/modules/plugin/ogr/ogr-jni/pom.xml
+++ b/modules/plugin/ogr/ogr-jni/pom.xml
@@ -96,7 +96,7 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-surefire-plugin</artifactId>
-                <version>2.17</version>
+                <version>2.15</version>
                 <configuration>
                     <runOrder>alphabetical</runOrder>
                     <includes>
--- a/modules/plugin/ogr/ogr-jni/pom.xml
+++ b/modules/plugin/ogr/ogr-jni/pom.xml
@@ -16,7 +16,7 @@
   <parent>
     <groupId>org.geotools</groupId>
     <artifactId>gt-ogr</artifactId>
-    <version>12-SNAPSHOT</version>
+    <version>12-remloadlibrary</version>
   </parent>
 
 
@@ -77,6 +77,8 @@
       <groupId>org.gdal</groupId>
       <artifactId>gdal</artifactId>
       <version>${gdal.version}</version>
+      <scope>system</scope>
+      <systemPath>${env.GT_GDAL_JAR}</systemPath>
     </dependency>
     <dependency>
       <groupId>org.geotools</groupId>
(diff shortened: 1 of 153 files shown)


fvanderbiest:
Looking for feedback on this one, at least a thumbs up or down. 
Thanks.

aaime:
@jdeolive, what do you say?

modules/plugin/ogr/ogr-jni/pom.xml
```diff
@@ -108,15 +108,13 @@
                       <exclude>${test.exclude.pattern}</exclude>
                     </excludes>
                     <argLine>-Xmx${test.maxHeapSize} ${jvm.opts} -Djava.awt.headless=${java.awt.headless} -Djava.io.tmpdir="${java.io.tmpdir}" -Djava.library.path="${env.GT_GDAL}"</argLine>
-                    <environmentVariables>
-                        <PATH>${env.GT_GDAL}${path.separator}${env.PATH}</PATH>
-                        <DYLD_LIBRARY_PATH>${env.GT_GDAL}${path.separator}${env.DYLD_LIBRARY_PATH}</DYLD_LIBRARY_PATH>
-                        <LD_LIBRARY_PATH>${env.GT_GDAL}${path.separator}${env.LD_LIBRARY_PATH}</LD_LIBRARY_PATH>
-                    </environmentVariables>
```
jdeolive: I too need to set these environment variables to make things work. I usually do this in my shell settings anyways so I don't have a strong position on keeping them in the pom. I missed the motivation for taking them out?

jdeolive:
I think the new method to check if the library has already been loaded looks fine as long as it falls back cleanly to the existing behaviour (which it appears to). It would be nice if the "cruft" was cleaned from this patch. Only remaining question seems where the removal of the env settings from the pom is warranted.

aaime:
I see there are still issues with the removal of a system variable in this pull request, as well as changing the scope of a dependency. I will close this pull request within the month if both are not address in a _constructive_ way.

pmauduit:
> I see there are still issues with the removal of a system variable in this pull request, as well as changing the scope of a dependency. I will close this pull request within the month if both are not address in a constructive way.

As I tried to explain earlier several times  in this PR, I still disagree with the current approach. The `System.loadLibrary()` should be made by `gdal-jni` itself, it is still from my point of view outside of GeoTools' concerns to handle such a task (see https://github.com/OSGeo/gdal/blob/trunk/gdal/swig/include/java/gdal_java.i#L27).

The `gdal-java` jar generated upstream using SWIG corresponding to the actual native GDAL library used should be  shared in a common java classloader for each instance of the JVM, instead of trying to guess if the library has already been loaded JVM-wide at a GeoTools level, using classes that can be potentially unusable (in case of using a specific `SecurityManager`, as I mentioned it before). As it is done for JDBC drivers, a `Class.forName()` / `isAvailable()` should be sufficient to determine whether the setup is correct or not.

You can consider closing the  PR now if you think it is better for the GeoTools project, I'm not willing to go any further with a solution I am disagreeing with.

jodygarnett:
I will create a bug ticket for this, and close this pull request pending community member review.

PR #453, opened 2014-05-12
Log JAI messages [GEOT-4797]

jodygarnett:
See https://jira.codehaus.org/browse/GEOT-4797 for discussion

### DIFF
--- a/modules/library/metadata/src/main/java/org/geotools/factory/GeoTools.java
+++ b/modules/library/metadata/src/main/java/org/geotools/factory/GeoTools.java
@@ -389,50 +389,73 @@ public final class GeoTools {
     }
 
     /**
-     * Initializes GeoTools for use. This convenience method performs various tasks (more may
-     * be added in the future), including setting up the {@linkplain java.util.logging Java
-     * logging framework} in one of the following states:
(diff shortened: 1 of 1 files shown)


hastebrot:
Hi. I suspect this PR's code changes caused a very delicate situation. I'm stuck to GeoTools 10.5 because of this. 

When I try to use `FileDataStoreFilter` to load shapefiles with newer versions I get an `java.lang.NoClassDefFoundError: javax/media/jai/util/ImagingListener`. I guess this is because it does not find the dependency to `javax.media:jai-core:1.1.3`.

hastebrot:
It works, when the osgeo repository is before maven central (example for Gradle). Seems to be a licensing issue with `jai-core` in maven central.

~~~groovy
repositories {
    maven { url "http://download.osgeo.org/webdav/geotools/" }
    mavenCentral()
}

dependencies {
    compile "org.geotools:gt-main:12.2"
    compile "org.geotools:gt-shapefile:12.2"
}
~~~

There were some suggestions to add `javax.media:jai-core:1.1.3` or `org.geotools:gt-coverage:12.2`, but none of them were needed.

hastebrot:
Reproduction of the error can reduced from `FileDataStoreFinder.getDataStore(...)` to `GeoTools.init()`.

PR #557, opened 2014-10-04
[GEOT-4912] No escaping of object names used in DatabaseMetaData calls

sgraca:


bencaradocdavies:
These changes may be the cause of https://jira.codehaus.org/browse/GEOT-4968

These changes have not yet been backported to 12.x.

PR #604, opened 2014-11-24
KML22 - dump feature collections

pmauduit:
Following this thread on the ML from last week:

http://sourceforge.net/p/geotools/mailman/message/33052373/

This commit aims to add the write support of feature collections into KML2.2.

Manual tests, and a new test method has been added to check that the produced document is correct.

Note: Some other implementations around expect that the XML produced is not prefixed, and that the default namespace is the KML2.2 one. For example, the document generated in the added unit test won't be detected as valid with my current version of gdal/ogr, and won't work onto http://display-kml.appspot.com/ either. Removing the 'kml:' prefix on each elements and setting it as the default xml namespace fixed the "issue" (even though this is valid IMHO to prefix the whole document).

Feedbacks welcome (ping @jdeolive ;-))


pmauduit:
@jdeolive Our customer needed to have the `<kml:name>` xml tag as well (generated when dumping as KML if the feature has a `name` attribute), so we had to revamp the previously provided code a little bit.

Do you want us to open another PR so that we can contribute the changes ? Note that the name would be present as `<kml:name>` and in the `<kml:ExtendedData>` block as well, but our customer is ok with this behaviour.

jdeolive:
Hey Pierre, the original work was already merged so yes, if you could submit a new pull request that would be appreciated. Thanks!

pmauduit:
@jdeolive here it is: https://github.com/geotools/geotools/pull/683

PR #626, opened 2014-12-03
GEOT-3255: Fixes WFS 1.0.0 GetCapabilities parsing with empty OnlineResource

mtoonen:
Now parsing a WFS (1.0.0) GetCapabilities document with an empty OnlineResource doesn't fail anymore.
Fixes https://jira.codehaus.org/browse/GEOT-3255


jodygarnett:
This module is no longer under active development please use wfs-ng

PR #644, opened 2014-12-16
Geopackage: Make GeoPackage module dimension aware (GEOT-4631)

NielsCharlier:


modules/unsupported/geopkg/src/main/java/org/geotools/geopkg/geom/GeoPkgGeomWriter.java
```diff
@@ -28,13 +28,23 @@ public void setWriteEnvelope(boolean writeEnvelope) {
...
+        this(dim, new Configuration());
+    }
+    
     public GeoPkgGeomWriter(Configuration config) {
+        this(2, config);
+    }
+
+    public GeoPkgGeomWriter(int dim, Configuration config) {
         this.config = config;
+        this.dim = dim;
```
jodygarnett: Do you actually have code calling all these different constructors?

jodygarnett:
Manually merged (to rebase and sort out a conflict)

modules/unsupported/geopkg/src/main/java/org/geotools/geopkg/geom/GeoPkgGeomWriter.java
NielsCharlier: Hmm only the one without arguments and with all arguments are currently being used, but I wanted to allow some flexibility for the possible future with respect to the default constructor arguments (it isn't always all or nothing).

PR #651, opened 2014-12-18
GeoPackage: fix srs name (GEOT-4625)

NielsCharlier:


jodygarnett:
Okay so we have a problem with encoding of DefaultGeographic.WS84 constant? I know in some of the WMS client code I take care to refer to that one as "CRS:84" (as per the WMS spec). Do you think that would work for GeoPackage (which is also an OGC spec)

NielsCharlier:
re:cause, I really have no idea, I assumed the existing name encoding to be intentional to be honest. If you think it is abnormal perhaps that is something to look at
re:test, done as asked, see https://github.com/geotools/geotools/pull/700

PR #657, opened 2014-12-23
Csvdatastore update

travislbrundage:
This pull request is a combination of old #652 and #656 - due to the nature of the changes, neither of them will work separately (they will cause the build to break). As such I've made this pull request with each set of changes as one of the commits. The first commit is the backport of geoserver's updates to the gt-csv module. The second is an update to the ContentDataStore tutorial to reflect these new changes. A final update may be needed for licensing changes.


docs/src/main/java/org/geotools/test/FixtureUtilities.java
```diff
@@ -0,0 +1,148 @@
...
+/**
+ * Static methods to support the implementation of tests that use fixture configuration files. See
+ * {@link OnlineTestCase} and {@link OnlineTestSupport} for details.
+ * 
+ * @author Ben Caradoc-Davies, CSIRO Earth Science and Resource Engineering
+ *
+ *
+ * @source $URL$
+ */
+public class FixtureUtilities {
```
jodygarnett: Not sure why these three test support classes are being added to docs. The docs are just for example code, but perhaps it is useful to have a test case to confirm functionality?

These classes (FixtureUtilites, OnlineTestCase, OnlineTestSupport, TestData) should be available via a maven dependency (rather than duplicated). If we are extra tricky it can be a test dependency...

modules/unsupported/csv/src/test/java/org/geotools/data/csv/parse/CSVSpecifiedWKTStrategyTest.java
```diff
@@ -0,0 +1,73 @@
+/* (c) 2014 Open Source Geospatial Foundation - all rights reserved
+ * (c) 2001 - 2013 OpenPlans
+ * This code is licensed under the GPL 2.0 license, available at the root
```
jodygarnett: We will return this to public domain license, you asked for permission on geoserver devel to backport to GeoTools. A PMC member gave you permission etc...

Should also probably be (c) 2010-2013 OpenPlans as CSVDataStore was not yet written in 2001.

modules/unsupported/csv/src/test/java/org/geotools/data/csv/parse/CSVLatLonStrategyTest.java
```diff
@@ -0,0 +1,320 @@
+/* (c) 2014 Open Source Geospatial Foundation - all rights reserved
+ * (c) 2001 - 2013 OpenPlans
```
jodygarnett: Fix header as described below.

modules/unsupported/csv/src/test/java/org/geotools/data/csv/parse/CSVAttributesOnlyStrategyTest.java
```diff
@@ -0,0 +1,103 @@
+/* (c) 2014 Open Source Geospatial Foundation - all rights reserved
+ * (c) 2001 - 2013 OpenPlans
```
jodygarnett: Fix header as described below.

docs/user/tutorial/datastore/strategy.rst
```diff
@@ -0,0 +1,252 @@
+:Author: Travis Brundage
+:Thanks: geotools-devel list
+:Version: |release|
+:License: Creative Commons with attribution
+
+Strategy Patterns
```
jodygarnett: Thanks for adding this!

jodygarnett:
Do you want to fix this or shall I?

Does this commit depend on any of the ContentDataStore fixes? Or should we accept it first ...

travislbrundage:
I'm not sure on this one, and I think @tbarsballe would be better able to tell... If I remember right, anything without its own implementation of file locking may cause errors. I don't think the CSVDataStore has its own implementation, so it may cause problems, but I am not sure. Either way, I think the changes should be pretty separate, and I feel like this merge makes sense to come first.

tbarsballe:
Currently, the only ContentDataStore fixes that exist deal with getCount(). (There is a change in property name resolvine, but that doesn't directly alter ContentDataStore. So my changes should have no notable effect on the behavior of CSVDataStore.

jodygarnett:
Is there a Jira I should mark closed for this one?

PR #659, opened 2014-12-29
Fixed PropertyNameResolvingVisitor to return a blank name if the xpath is blank [GEOT-4972]

tbarsballe:
Additionial fixes required to avoid test failures:
Fix to SimpleFeatureTypeBuilder to exclude null values. 
Fix to ShapeFileFeatureSource.getReadSchema() to include geometry properties in a geometry query with no geometry properties included.

### DIFF
--- a/modules/library/main/src/main/java/org/geotools/feature/simple/SimpleFeatureTypeBuilder.java
+++ b/modules/library/main/src/main/java/org/geotools/feature/simple/SimpleFeatureTypeBuilder.java
@@ -1028,7 +1028,10 @@ public class SimpleFeatureTypeBuilder {
 
         // add attributes in order
         for (int i = 0; i < types.size(); i++) {
-            b.add(original.getDescriptor(types.get(i)));
+            AttributeDescriptor descriptor = original.getDescriptor(types.get(i));
+            if (descriptor != null) {
+                b.add(descriptor);
+            }
         }
         
         // handle default geometry
--- a/modules/library/main/src/main/java/org/geotools/filter/visitor/PropertyNameResolvingVisitor.java
+++ b/modules/library/main/src/main/java/org/geotools/filter/visitor/PropertyNameResolvingVisitor.java
@@ -49,6 +49,9 @@ public class PropertyNameResolvingVisitor extends DuplicatingFilterVisitor {
     }
     
     public Object visit(PropertyName expression, Object extraData) {
+        if ("".equals(expression.getPropertyName())) {
+            return super.visit(expression, extraData);
+        }
         AttributeDescriptor att = expression.evaluate( featureType, null);
         if ( att != null ) {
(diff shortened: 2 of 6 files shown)


modules/library/main/src/main/java/org/geotools/feature/simple/SimpleFeatureTypeBuilder.java
```diff
@@ -1028,7 +1028,10 @@ public static SimpleFeatureType retype(SimpleFeatureType original, List<String>
 
         // add attributes in order
         for (int i = 0; i < types.size(); i++) {
-            b.add(original.getDescriptor(types.get(i)));
+            AttributeDescriptor descriptor = original.getDescriptor(types.get(i));
+            if (descriptor != null) {
+                b.add(descriptor);
+            }
```
jodygarnett: So I understand protecting this method from referencing invalid properties - but I hate silent failure. How are invalid properties being selected? Is this a case where we pass in an empty string?

modules/library/main/src/main/java/org/geotools/filter/visitor/PropertyNameResolvingVisitor.java
```diff
@@ -49,6 +49,9 @@ public PropertyNameResolvingVisitor( SimpleFeatureType featureType ) {
     }
     
     public Object visit(PropertyName expression, Object extraData) {
+        if ("".equals(expression.getPropertyName())) {
+            return super.visit(expression, extraData);
```
jodygarnett: Okay so we are preserving the empty property name here.

modules/library/main/src/main/java/org/geotools/feature/simple/SimpleFeatureTypeBuilder.java
tbarsballe: Given that I only started getting nullPointerExceptions after I changed PropertyNameResolvingVisitor, I am fairly certain the null is being created from the empty string, I just don't know exactly where.

jodygarnett:
Going to grab this locally and do a careful review.

I am checking for a couple of things:
- That responsibilities still match up with each class, applying a fix in the wrong spot can make debugging confusing
- Checking bank (i.e. empty string) consistently references all properties (in the case of BBox)
- Checking null references no properties (in the case where we just want a set of FeatureIds returned)

modules/library/main/src/main/java/org/geotools/feature/simple/SimpleFeatureTypeBuilder.java
jodygarnett: Okay so this is called by ShapefileFeatureSource.getReadSchema - which already contains logic to deal with converting a blank attribute to the_geom.

Going to fix that workaround rather than make retype perform any logic.

tbarsballe:
Closed via 3d55568

PR #660, opened 2014-12-30
Modified getBoundsInternal to handle 3D CRS

tbarsballe:


modules/unsupported/property-ng/src/test/java/org/geotools/data/property/ng/PropertyDataStore2Test.java
```diff
@@ -49,6 +50,7 @@
  */
 public class PropertyDataStore2Test extends TestCase {
     PropertyDataStore store;
+    PropertyDataStore store3d;
```
jodygarnett: If you only use this in one of the tests no sense adding it to your fixture (i.e. startup / teardown).

PR #661, opened 2014-12-30
Let geopackage databases use specific geometry types for geometry columns

jericks:
Closes issue #GEOT-4970


NielsCharlier:
Approved.

PR #663, opened 2014-12-31
Added WKT reader to PropertyFeatureReader. 

tbarsballe:
Also added test cases for curves and linearization tolerance (copied from gt-property).


tbarsballe:
@jodygarnett Updated the pull request based on your feedback.

PR #665, opened 2015-01-05
ContentFeatureSource simplifies all filters within all passed queries.

tbarsballe:
DataUtilities.mixQueries modified to simplify filters.
DataUtilities.simplifyFilter(Query) method added to simplify the filter contained by a query.

### DIFF
--- a/docs/user/library/main/data.rst
+++ b/docs/user/library/main/data.rst
@@ -363,10 +363,18 @@ A common task with gt-main is preparing a Query against a FeatureSource. DataUti
 
 * DataUtilities.mixQueries(Query, Query, String)
   
-  Safely combines two queries in a sensible manner.
+  Safely combines two queries in a sensible manner. The provided string is used for the name of the new query.
+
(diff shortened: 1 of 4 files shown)


jodygarnett:
Two things:

a) test case (yeah I know it will be quick)
b) documentation - since this is a public api class I try for a code example (see http://docs.geotools.org/stable/userguide/library/main/data.html)

jodygarnett:
How is this one going?

tbarsballe:
Updated with docs and  test case. Some of the documentation on related DataUtilities methods was also updated to be a bit less sparse. 
I feel like a code example for this method is not especially usefull, since it boils down to giving the method a query - no point in cluttering the docs with this trivial of an example. (And there are other DataUtilities methods in the docs much more deserving of a code example that don't have one). A before/after filter example might be a good idea, but that would probably fit better in the Filter section of the docs, alongside an explanation of SimplifyingFilterVisitor. Thoughts?

PR #668, opened 2015-01-07
Fix typing on retype, previously could get different defaultGeometry values

tbarsballe:
https://jira.codehaus.org/browse/GEOT-4978

ContentFeatureSource.getSchema() would sometimes give a different value for default geometry than reader.getFeatureType().

This only occurs if the data store contains multiple geometries, with no geometry explicitly defined as the default geometry. It appears that getSchema was somehow getting a differently ordered list of attributes than the reader.  Because this error does not occur predictably, I am not sure that a viable test case for this issue is possible.

### DIFF
--- a/modules/library/main/src/main/java/org/geotools/feature/FeatureTypes.java
+++ b/modules/library/main/src/main/java/org/geotools/feature/FeatureTypes.java
@@ -471,8 +471,12 @@ public class FeatureTypes {
     static boolean equals( SimpleFeatureType typeA, SimpleFeatureType typeB, boolean compareUserMaps) {
         if (typeA == typeB)
             return true;
-
+        
         if (typeA == null || typeB == null) {
+            return false;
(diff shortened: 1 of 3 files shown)


modules/library/main/src/main/java/org/geotools/feature/FeatureTypes.java
```diff
@@ -471,11 +471,15 @@ public static boolean equalsExact( SimpleFeatureType typeA, SimpleFeatureType ty
...
-
+        
         if (typeA == null || typeB == null) {
             return false;
         }
         
+        if (!typeA.equals(typeB)) {
+            return false;
+        }
+
```
tbarsballe: This will return false if the default geometries are not equal.

jodygarnett:
This looks good I will create a Jira issue and merge this in.

PR #669, opened 2015-01-07
Fix getBounds for diff containing deleted features

tbarsballe:
Feature id is now obtained from the keySet so that it still works with deleted features.

### DIFF
--- a/modules/library/data/src/main/java/org/geotools/data/store/ContentFeatureSource.java
+++ b/modules/library/data/src/main/java/org/geotools/data/store/ContentFeatureSource.java
@@ -387,12 +387,12 @@ public abstract class ContentFeatureSource implements SimpleFeatureSource {
             Diff diff = state.getDiff();
             
             // don't compute the bounds of the features that are modified or removed in the diff
-            Iterator it = diff.getModified().values().iterator();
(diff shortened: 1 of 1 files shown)


modules/library/data/src/main/java/org/geotools/data/store/ContentFeatureSource.java
```diff
@@ -387,12 +387,12 @@ public final ReferencedEnvelope getBounds(Query query) throws IOException {
...
+            Iterator it = diff.getModified().keySet().iterator();
             FilterFactory2 ff = CommonFactoryFinder.getFilterFactory2();
             Set<FeatureId> modifiedFids = new HashSet<FeatureId>();
             while(it.hasNext()){
-                SimpleFeature feature = (SimpleFeature) it.next();
-                modifiedFids.add(ff.featureId(feature.getID()));
+                String featureId = (String)it.next();
+                modifiedFids.add(ff.featureId(featureId));
             }
             Id idFilter = ff.id(modifiedFids);
```
jodygarnett: This filter does not match what is described in the javadoc.

Expect it should be:

```
ff.not( ff.id( modifiedFids) )
```

I think we can construct a test case for this one.

PR #670, opened 2015-01-07
ContentFeatureStore writer flag for commit

tbarsballe:
Added diffWriter method to provide wrapper for ContentFeatureStore writer.
Fixed misc bugs:
 - Added transaction to store in DiffTransactionState.commit()
 - Brought locking wrapper outside of if statement in ContentFeatureStore.getWriter()
 - Tweaked ContentFeatureSourceEventsTest.java to match current results. It is now closer to expected behaviour (Although event handling still needs to be overhauled)

### DIFF
--- a/modules/library/data/src/main/java/org/geotools/data/store/ContentFeatureStore.java
+++ b/modules/library/data/src/main/java/org/geotools/data/store/ContentFeatureStore.java
@@ -155,12 +155,12 @@ public abstract class ContentFeatureStore extends ContentFeatureSource implement
                     }
                 }
             } else {
-                DiffTransactionState state = (DiffTransactionState) getTransaction().getState(
-                        getEntry());
+                DiffTransactionState state = (DiffTransactionState) getTransaction()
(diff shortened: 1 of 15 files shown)


modules/library/data/src/main/java/org/geotools/data/store/ContentFeatureStore.java
```diff
@@ -144,12 +145,25 @@ public ContentFeatureStore(ContentEntry entry,Query query) {
...
+                        writer = new FilteringFeatureWriter(writer, query.getFilter());
+                    }
+                }
+            } else {
+                DiffTransactionState state = (DiffTransactionState) getTransaction()
+                        .getState(getEntry());
+                // reader will take care of filtering
+                // DiffContentWriter takes care of events
+                FeatureReader<SimpleFeatureType, SimpleFeature> reader = getReader(query);
+                writer = state.diffWriter( this, reader );
```
jodygarnett: Thanks that looks great.

PR #671, opened 2015-01-08
Remove unnecessary System.err.println calls

mfarwell:
wfs-ng module

There are unnecessary System.err.println calls in org.geotools.data.wfs.internal.AbstractWFSStrategy.java.  The text content being written to standard error is also being written out to log files.

I'm deploying GeoTools as a Jboss module and these calls cause the post contents of all WFS requests to be duplicated in the server.log file.  The first time it's recorded as INFO, but the second call (System.err.println) records it as ERROR.

These changes resolve GEOT-4914.


jodygarnett:
:heavy_plus_sign: should of been caught during code review

PR #675, opened 2015-01-08
Updating bounds test for compatibility with PropertyFeatureSource (NG) and ContentFeatureSource

tbarsballe:
FeatureSource.getBounds is not guaranteed to return a result (eg. if a full feature scan is needed). Updated to use FeatureCollection.getBounds.


jodygarnett:
Fair enough, not sure why the old implementation was willing to do a full table scan.

PR #679, opened 2015-01-09
Migrating property-ng to property and property to property-old

tbarsballe:
Migrated plugin/property to unsupported/property-old and unsupported/property-ng to plugin/property. 
Updated poms. Refactored packages. Updated display names.

The diff is a bit wonky here due to all the rename/move operations.

The doc updates here are a bit of a necessary evil, as the AbstractDataStore Tutorial references classes in the old PropertyDataStore. Since we are (eventually) removing AbstractDataStore, it makes more sense to remove the tutorial than to update the doc references within the tutorial.


tbarsballe:
geoserver requires a fix to ContentFeatureStore here: https://github.com/geotools/geotools/pull/680

PR #681, opened 2015-01-11
Wfs-ng improvements: support getfeature requests to WFS 2.0 server

NielsCharlier:


modules/extension/xsd/xsd-fes/src/main/java/org/geotools/filter/v2_0/bindings/BBOXTypeBinding.java
```diff
@@ -16,10 +16,24 @@
  */
 package org.geotools.filter.v2_0.bindings;
 
+import java.util.ArrayList;
```
jodygarnett: header should be updated to 2015

modules/extension/xsd/xsd-fes/src/main/java/org/geotools/filter/v2_0/bindings/BBOXTypeBinding.java
```diff
@@ -52,4 +66,48 @@
...
+    public Class getType() {
+        return BBOX.class;
+    }
+
+    @Override
+    public Object getProperty(Object object, QName name) throws Exception {
+        BBOX box = (BBOX) object;
+        
+        if (FES.ValueReference.equals(name)) {
+            if (box.getExpression1() instanceof PropertyName) {
```
jodygarnett: I thought the assumption of property name was finally being lifted in filter 2.0? Let me check - seems to be handled by a value reference now.

Probably something to look at later...

modules/extension/xsd/xsd-fes/src/test/java/org/geotools/filter/v2_0/bindings/FilterTypeBindingTest.java
```diff
@@ -22,6 +22,7 @@
 import org.geotools.filter.v1_1.FilterMockData;
 import org.geotools.filter.v2_0.FES;
 import org.geotools.filter.v2_0.FESTestSupport;
```
jodygarnett: Update header when you modify a file:

```
(C) 2002-2015
```

modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/internal/v2_0/StrictWFS_2_0_Strategy.java
```diff
@@ -588,55 +512,230 @@ protected String getOperationURI(WFSOperationType operation, HttpMethod method)
...
 
     @Override
     public boolean supportsTransaction(QName typeName) {
-        // TODO Auto-generated method stub
-        return false;
+        try {
+            getFeatureTypeInfo(typeName);
+        } catch (IllegalArgumentException e) {
+            throw e;
+        }
```
jodygarnett: not sure what this method is doing? Just an excuse to test if the typeName is real? Do we need to force the side effect of parsing out a FeatureTypeInfo?

modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/internal/v2_0/StrictWFS_2_0_Strategy.java
```diff
@@ -588,55 +512,230 @@ protected String getOperationURI(WFSOperationType operation, HttpMethod method)
...
+                net.opengis.wfs.OperationType.INSERT_LITERAL,
+                net.opengis.wfs.OperationType.UPDATE_LITERAL,
+                net.opengis.wfs.OperationType.DELETE_LITERAL)) {
+
+            if (!operation.contains(required)) {
+                info("Transactions not supported since WFS didn't declare support for "
+                        + required.getName());
+                return false;
+            }
+        }*/
```
jodygarnett: The commented out code here makes a bit more sense to me, expect I am missing the point.

jodygarnett:
This pull request has received a thumbs up from sampo on the email list. There a some headers that could use correcting, but suspect my questions are based on the details with respect to the WFS 2.0 spec.

jodygarnett:
merged to master (after a rebase and local build)

modules/extension/xsd/xsd-fes/src/main/java/org/geotools/filter/v2_0/bindings/BBOXTypeBinding.java
NielsCharlier: Hmmm good point. I think the "if" can basically be removed without problems. This is a residue of older code that was altered. Will test with if removed.

modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/internal/v2_0/StrictWFS_2_0_Strategy.java
NielsCharlier: This is part of the work from last year so my memory isn't that fresh on this code. As I understand it is testing whether
1) the type is actually in the capabilities, otherwise an exception is generated, there is no side effect of parsing because it has already been parsed and it is just checking if it's in memory.
2) transactions itself are supported in the capabilties
3) commented: all three operations are supported. Not sure if this should really happen.

Transactions at the moment are not completely functional, there is still some additional work needed. I suggest I revisit this method then.

NielsCharlier: Actually, it is indeed creating a FeatureTypeInfoImpl that is immediately thrown away, but it doesn't involve additional parsing. Stylisticly very ugly, but not really a drain on resources. As I said, will revisit.

NielsCharlier:
Fixed headers and manually pushed https://github.com/geotools/geotools/commit/7b6d306fd4ed0b18798ca724b8e6b87c07169aaa

NielsCharlier:
After testing decided to remove if condition https://github.com/geotools/geotools/commit/ba335d5e212407bf576827fb5fab227db784864d

bradh:
@NielsCharlier Looks like there were commits on this branch post merge. Do you still need those?

NielsCharlier:
Wow this is from years ago. I have no idea why this happened, but I assume this work was finished somewhere else

PR #682, opened 2015-01-12
Correction for bug on directed graphs

brunodoamaral:
Correct list of related edges when a directed graph is used (must use getOutRelated insted of getRelated)

### DIFF
--- a/modules/extension/graph/src/main/java/org/geotools/graph/traverse/standard/DijkstraIterator.java
+++ b/modules/extension/graph/src/main/java/org/geotools/graph/traverse/standard/DijkstraIterator.java
@@ -266,7 +266,11 @@ public class DijkstraIterator extends SourceGraphIterator {
   }
   
   protected Iterator getRelated(Graphable current) {
-    return(current.getRelated());	
+    if ( current instanceof DirectedGraphable ) {
+      return ((DirectedGraphable)current).getOutRelated() ;
+    } else {
+      return (current.getRelated());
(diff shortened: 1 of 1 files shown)


brunodoamaral:
Added an missing import. Sorry!

jodygarnett:
Reviewing this before the feature freeze, since this is a bug fix it does not matter.

I need one thing from you before I can accept this change - a test case showing me (and importantly the build box) that your fix does something. Can you create a small directed graph (two nodes) and demonstrate that you can navigate one way, and not the other?

brunodoamaral:
Hi @jodygarnett ! I just found the correct way to iterate on a directed graph. I must have used the `DirectedDijkstraIterator` instead of `DijkstraIterator` to do so.
Anyway, the class `DijkstraShortestPathFinder` has a constructor that incorrectly instantiate a `DijkstraIterator` even if the graph is directed (lines 85-92):
```java
  public DijkstraShortestPathFinder(Graph graph, Graphable source, EdgeWeighter weighter, NodeWeighter nweighter) {
    m_graph = graph;
    m_iterator = new DijkstraIterator(weighter,nweighter);
    m_iterator.setSource(source);
    m_traversal = new BasicGraphTraversal(graph, this, m_iterator); 
  }
```
I could change it, but I suppose the gt-graph module is not so used... what do you think?

tbarsballe:
I have added test cases for this change here: #698

PR #686, opened 2015-01-13
Move memory datastore to gt-data (GEOT-4982)

jodygarnett:
The first step of implementing MemoryDataStore with ContentDataStore is moving it to the gt-data module.

In order to accomplish this several test cases and the UserLayer data structure have been updated to use new DataUtilities methods.

See: https://jira.codehaus.org/browse/GEOT-4982


tbarsballe:
Overall looks good, but there a few syntax and style issues here and there. I have added comments to the individial commits as it was easier to go through them in sequence rather than as a single mass of changes.

jodygarnett:
Thanks for the review @tbarsballe - I will take that DataStore out into a distinct class.

PR #687, opened 2015-01-13
Upgrading EPSG database to version 8.6.

jdeolive:
Includes a patch to DefaultCoordinateSystemAxis.getDirection() to account for
new names "northwest" and "northeast" axis directions.

DefaultFactoryTest.testValidArea() also upadted to reflect changes to area of
validity of EPSG:7400 coordinate system.


modules/library/referencing/src/main/java/org/geotools/referencing/cs/DefaultCoordinateSystemAxis.java
```diff
@@ -912,6 +912,18 @@ public static AxisDirection getDirection(String direction) throws NoSuchElementE
...
+        /*
+         * In latest version of the database north-east and north-west are now northeast and northwest respectively,
+         * and none of these heuristics pick that up, so we add a few special cases here.
+         */
+        if ("northeast".equalsIgnoreCase(direction)) {
+            return AxisDirection.NORTH_EAST;
+        }
+        if ("northwest".equalsIgnoreCase(direction)) {
+            return AxisDirection.NORTH_WEST;
+        }
```
jodygarnett: Seems fine (love how the standards change). We did have a formal alias system somewhere in here to handle things like non standard oracle and/or esri WKT.

tkunicki:
I think there may be an issue with the [DomainAliasTable.txt](https://github.com/geotools/geotools/blob/83e50a7ca63f11227d5840463c105ddf6a44330a/modules/library/referencing/src/main/resources/org/geotools/referencing/factory/DatumAliasesTable.txt)

In particular the EPSG entry "NAD83 (High Accuracy Regional Network)" was deprecated and has been replaced with "NAD83 (High Accuracy Reference Network)".  This is likely to affect shapefiles with this datum referenced in the proj file.   The change was due to ChangeID EPSG::2010.061 which is linked off 
the record for EPSG::6152 at http://www.epsg-registry.org/.  The old value is aliased but I don't think the GeoTools EPSG implementation handles datum aliases within an authority.

modules/plugin/epsg-hsql/src/test/java/org/geotools/referencing/factory/epsg/DefaultFactoryTest.java
```diff
@@ -503,20 +503,20 @@ public void testValidArea() throws FactoryException, TransformException {
...
+        assertEquals(-8.008, envelope.getMinimum(1), 1E-3);
         assertEquals( 6.548, envelope.getMaximum(1), 1E-3);
         assertNull(CRS.getEnvelope(null));
         final GeographicBoundingBox rep = new GeographicBoundingBoxImpl(envelope);
-        assertEquals(42.25, rep.getSouthBoundLatitude(), 1E-3);
-        assertEquals(51.10, rep.getNorthBoundLatitude(), 1E-3);
-        assertEquals(-5.20, rep.getWestBoundLongitude(), 1E-3);
+        assertEquals(42.33, rep.getSouthBoundLatitude(), 1E-3);
+        assertEquals(51.14, rep.getNorthBoundLatitude(), 1E-3);
+        assertEquals(-4.87, rep.getWestBoundLongitude(), 1E-3);
```
jodygarnett: So all just slightly different. I did not see a change EPSG: 7400 extent listed on the change list you sent out.

aaime:
One observation, I don't see epsg-wkt being updated. The epsg-hsql tests contain a property dumper class that, going by memory, does the job

jdeolive:
@tkunicki Afaik that change was incorporated the last time the epsg database was upgraded. Do you have a specific wkt representation that you are worried about?

jdeolive:
Thanks @aaime , epsg-wkt module updated now.

jdeolive:
Any more feedback on this one? Or ok to merge?

aaime:
Hey Justin, seems all ok to me.
Looking at a side by side comparison of the two WKT files I've noticed that all EPSG codes associated to the NAD83(HARN) datum have "lost" their TOWGS84 parameter, it had some value before, and now it's all zeroes instead.

Not sure if this is cause of concern, afaik libproj equates NAD83 and WGS84... but it's kind of surprising. Anyone has a clue?

Anyways, not a reason to stop the merge, go ahead.

jdeolive:
Thanks for the feedback guys.

PR #689, opened 2015-01-15
Fixed FeatureWriter next() and hasNext() implementations to conform to standard file handle behavior

tbarsballe:
QA consistency check:

* hasNext() returns false if the writer is closed
* next() throws an IOException if the writer is closed

A number of implementaitons set their internal reader/writer to null, then did not test this in next() or hasNext(), leading to NPEs. I have added checks for this.
Some other implementaitons would throw an IOException on hasNext() if the writer was closed. I have changed this to return false, for consistency.

The javadoc for the FeatureWriter interface has been updated to reflect these changes.
Test case(s) updated where applicable.


jodygarnett:
Thanks QA is always helpful, next time please update the headers with the correct date range (annoying I know).

PR #691, opened 2015-01-16
Migrate MemoryDataStore to extend ContentDataStore

tbarsballe:
Migrate MemoryDataStore to extend ContentDataStore. Seperated out FeatureSource and Reader/Writer as seperate files.
Updated tests to conform with new funtionality
Note: Events don't really work right now, event tests have been set to go against current behavior (This is a known issue of ContentDataStore)
Updated feature-pregeneralized tests to account for greater MemoryDataStore capabilities

### DIFF
--- a/modules/library/data/src/main/java/org/geotools/data/memory/MemoryDataStore.java
+++ b/modules/library/data/src/main/java/org/geotools/data/memory/MemoryDataStore.java
@@ -2,7 +2,7 @@
  *    GeoTools - The Open Source Java GIS Toolkit
  *    http://geotools.org
  * 
- *    (C) 2003-2008, Open Source Geospatial Foundation (OSGeo)
+ *    (C) 2003-2015, Open Source Geospatial Foundation (OSGeo)
  *    
  *    This library is free software; you can redistribute it and/or
  *    modify it under the terms of the GNU Lesser General Public
@@ -17,10 +17,12 @@
 package org.geotools.data.memory;
 
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.logging.Level;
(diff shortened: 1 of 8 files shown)


modules/library/data/src/main/java/org/geotools/data/memory/MemoryDataStore.java
```diff
@@ -316,46 +323,33 @@ private void addFeatureInternal(SimpleFeature feature) {
...
-            }
-                throw new SchemaNotFoundException(typeName);
-        }
+    
+    protected ContentFeatureSource createFeatureSource(ContentEntry entry) {
+        return createFeatureSource(entry, Query.ALL);
+    }
+    
+    protected ContentFeatureSource createFeatureSource(ContentEntry entry, Query query) {
+        return new MemoryFeatureStore(entry, query);
```
tbarsballe: This makes me a little nervous, but I can't think of any situation where we could not get a MemoryFeatureStore (rather than a MemoryFeatureSource)

modules/library/data/src/test/java/org/geotools/data/memory/MemoryDataStoreTest.java
```diff
@@ -911,11 +907,11 @@ public void testGetFeatureSourceRoad() throws IOException {
         assertEquals(rd12Bounds, some.getBounds());
         assertEquals(some.getSchema(), road.getSchema());
 
-        DefaultQuery query = new DefaultQuery("road", rd12Filter, new String[]{"name",});
+        DefaultQuery query = new DefaultQuery("road", rd12Filter, new String[]{"name", "geom"});
```
tbarsballe: I have no clue how this test managed to function before this change - without geom in the featureType we shouldn't expect to get geometries back, should we?

modules/library/data/src/main/java/org/geotools/data/memory/MemoryDataStore.java
jodygarnett: Memory is kind of wriable - so I am okay with FeatureStore all the time :)

If you really wanted you could make a constructor flag for read-only. We have DataUtilities.store( featureCollection ) class that can get you a read-only "memory" datastore for safely staging data.

modules/library/data/src/test/java/org/geotools/data/memory/MemoryDataStoreTest.java
jodygarnett: Perhaps it was a manditory field last time. In anycase the test as you have here makes sense.

PR #692, opened 2015-01-16
Updating CSVDataStore for write functionality [GEOT-4974]

travislbrundage:
This is an update of the CSVDataStore to give it read/write functionality.

Some design issues to note:
I merged the original CSVStrategy interface and AbstractCSVStrategy into one. They now reside in the abstract class, CSVStrategy. I did not see much sense in having both in this case.
I have opted to migrate all the methods from the CSVStrategySupport class into CSVStrategy. They are somewhat odd helper methods that could use some refactoring in the future, but for now I have put them there so that individual strategies can choose to override them and possibly make them more appropriate.
A few other smaller changes include making the lat/lon to x/y mapping consistent and storing the Point geometry where the LAT column resided.

Finally, for testing, I have included both the old CSVWriteTest to test basic file writing, and I have created a new CSVWriteStrategyTest which tests each strategy's implementation.


jodygarnett:
Is there is a jira associated with this?

travislbrundage:
Yes, sorry. Here is the jira ticket: https://jira.codehaus.org/browse/GEOT-4974

modules/unsupported/csv/src/main/java/org/geotools/data/csv/CSVDataStore.java
```diff
@@ -60,7 +60,11 @@ public Name getTypeName() {
 
     @Override
     protected ContentFeatureSource createFeatureSource(ContentEntry entry) throws IOException {
-        return new CSVFeatureSource(entry, Query.ALL);
+    	if (csvFileState.getFile().canWrite()) {
+    		return new CSVFeatureStore(csvStrategy, csvFileState, entry, Query.ALL);
+    	} else {
+    		return new CSVFeatureSource(entry, Query.ALL);
+    	}
```
jodygarnett: Looks about right :)

modules/unsupported/csv/src/main/java/org/geotools/data/csv/CSVFeatureStore.java
```diff
@@ -0,0 +1,141 @@
+/* GeoTools - The Open Source Java GIS Toolkit
+ * http://geotools.org
+ *
+ * (C) 2010-2014, Open Source Geospatial Foundation (OSGeo)
```
jodygarnett: This is a new file being added? So probably not 2010 ...

modules/unsupported/csv/src/main/java/org/geotools/data/csv/CSVFeatureWriter.java
```diff
@@ -0,0 +1,174 @@
...
+        }
+        if (this.currentFeature != null) {
+            this.write(); // the previous one was not written, so do it now.
+        }
+        // Step 1: Write out remaining contents (if applicable)
+        while (hasNext()) {
+            next();
+            write();
+        }
+        csvWriter.close();
```
jodygarnett: I sure hope this closes the FIleWriter we passed into it (during the constructor)

modules/unsupported/csv/src/main/java/org/geotools/data/csv/CSVFeatureWriter.java
```diff
@@ -0,0 +1,174 @@
...
+    public SimpleFeatureType getFeatureType() {
+        return this.featureType;
+    }
+    // featureType end
+
+    // hasNext start
+    @Override
+    public boolean hasNext() throws IOException {
+        if( csvWriter == null ){
+            throw new IOException("Writer has been closed");
```
jodygarnett: @tbarsballe has gone through and done a QA run to ensure closed writers return hasNext(): false (rather than thrown an IOException).

Lets stick with that for consistency (principle of least surprise)

modules/unsupported/csv/src/main/java/org/geotools/data/csv/CSVFeatureWriter.java
```diff
@@ -0,0 +1,174 @@
...
+        return iterator.hasNext();
+    }
+    // hasNext end
+    
+    // next start
+    @Override
+    public SimpleFeature next() throws IOException, IllegalArgumentException,
+            NoSuchElementException {
+        if(csvWriter == null){
+            throw new IOException("Writer has been closed");
```
jodygarnett: This one is fine :)

modules/unsupported/csv/src/main/java/org/geotools/data/csv/parse/CSVAttributesOnlyStrategy.java
```diff
@@ -17,25 +17,70 @@
...
+                String txt = value.toString();
+                csvRecord.add(txt);
+            }
+        }
+        return csvRecord.toArray(new String[csvRecord.size()-1]);
+    }
+    
     @Override
-    public SimpleFeature createFeature(String recordId, String[] csvRecord) {
+    public SimpleFeature decode(String recordId, String[] csvRecord) {
```
jodygarnett: thanks encode/decode is much more clear.

modules/unsupported/csv/src/main/java/org/geotools/data/csv/parse/CSVSpecifiedLatLngStrategy.java
```diff
@@ -68,20 +78,76 @@ protected SimpleFeatureType buildFeatureType() {
...
+    public void createSchema(SimpleFeatureType featureType) throws IOException {
+        List<String> header = new ArrayList<String>();
+        
+        GeometryDescriptor geometryDescrptor = featureType.getGeometryDescriptor();
+        if (geometryDescrptor != null
+                && CRS.equalsIgnoreMetadata(DefaultGeographicCRS.WGS84,
+                        geometryDescrptor.getCoordinateReferenceSystem())
+                && geometryDescrptor.getType().getBinding().isAssignableFrom(Point.class)) {
+            header.add("LAT");
+            header.add("LON");
```
jodygarnett: I find it odd that we are specifically asking the constructor for a custom "lat" and "lon" name ... and then ignoring it when making the schema?

If I was a user I would feel ignored :)

modules/unsupported/csv/src/main/java/org/geotools/data/csv/parse/CSVLatLonStrategy.java
```diff
@@ -65,26 +76,93 @@ protected SimpleFeatureType buildFeatureType() {
...
         if (validLat && validLon) {
-            builder.add(GEOMETRY_COLUMN, Point.class);
+        	List<String> csvHeaders = Arrays.asList(headers);
+        	int index = csvHeaders.indexOf(latSpelling);
+        	AttributeTypeBuilder builder2 = new AttributeTypeBuilder();
+        	builder2.setCRS(DefaultGeographicCRS.WGS84);
+        	builder2.binding(Point.class);       	
+        	AttributeDescriptor descriptor = builder2.buildDescriptor(GEOMETRY_COLUMN);
+        	builder.add(index, descriptor);
+
```
jodygarnett: Yeah sorry for the awkward workaround required here, I think we can improve the builder to be more helpful here.

How about this for a proposed API.
```
AttributeTypeBuilder builder2 = builder.get(GEOMETRY_COLUMN);
builder2.setCRS(DefaultGeographicCRS.WGS84);
builder2.binding(Point.class);
builder.set(index, builder2.buildDescriptor(GEOMETRY_COLUMN));
```

modules/unsupported/csv/src/main/java/org/geotools/data/csv/parse/CSVLatLonStrategy.java
```diff
@@ -65,26 +76,93 @@ protected SimpleFeatureType buildFeatureType() {
...
     }
 
+    private boolean isLatitude(String s) {
+        return "latitude".equalsIgnoreCase(s) || "lat".equalsIgnoreCase(s);
+    }
+
+    private boolean isLongitude(String s) {
+        return "lon".equalsIgnoreCase(s) || "lng".equalsIgnoreCase(s) || "long".equalsIgnoreCase(s)
+                || "longitude".equalsIgnoreCase(s);
+    }
```
jodygarnett: These two methods are getting duplicated a lot, could move them "up" and mark them protected ...

modules/unsupported/csv/src/main/java/org/geotools/data/csv/parse/CSVLatLonStrategy.java
```diff
@@ -65,26 +76,93 @@ protected SimpleFeatureType buildFeatureType() {
...
+    public void createSchema(SimpleFeatureType featureType) throws IOException {
+        List<String> header = new ArrayList<String>();
+        
+        GeometryDescriptor geometryDescrptor = featureType.getGeometryDescriptor();
+        if (geometryDescrptor != null
+                && CRS.equalsIgnoreMetadata(DefaultGeographicCRS.WGS84,
+                        geometryDescrptor.getCoordinateReferenceSystem())
+                && geometryDescrptor.getType().getBinding().isAssignableFrom(Point.class)) {
+            header.add("LAT");
+            header.add("LON");
```
jodygarnett: The hard coded LAT, LON again.

modules/unsupported/csv/src/main/java/org/geotools/data/csv/parse/CSVSpecifiedWKTStrategy.java
```diff
@@ -42,18 +52,68 @@ public CSVSpecifiedWKTStrategy(CSVFileState csvFileState, String wktField) {
...
+    @Override
+    public String[] encode(SimpleFeature feature) {
+    	List<String> csvRecord = new ArrayList<String>();
+        for (Property property : feature.getProperties()) {
+        	String name = property.getName().getLocalPart();
+            Object value = property.getValue();
+            if (value == null) {
+            	csvRecord.add("");
+            } else if (name.compareTo(wktField) == 0) {
+            	WKTWriter wkt = new WKTWriter();
```
jodygarnett: We probably have a custom GeoTools WKTWriter that can do curves ... this is just a nice to have.

modules/unsupported/csv/src/main/java/org/geotools/data/csv/parse/CSVStrategy.java
```diff
@@ -18,14 +18,135 @@
...
+        }
+        for (String col : headers) {
+            Class<?> type = typesFromData.get(col);
+            builder.add(col, type);
+        }
+        return builder;
+    }
+
+    /** Performs a full file scan attempting to guess the type of each column */
+    public static Map<String, Class<?>> findMostSpecificTypesFromData(CsvReader csvReader,
```
jodygarnett: Knock this down to protected (no need to define public api here)

modules/unsupported/csv/src/test/java/org/geotools/data/csv/CSVWriteTest.java
```diff
@@ -0,0 +1,339 @@
...
+            throw new IllegalStateException("Modification not supported");
+        }
+        SimpleFeatureStore featureStore = (SimpleFeatureStore) featureSource;
+
+        // featureStoreExample end
+        System.out.println("\nfeatureStoreExample end\n");
+    }
+
+    @Test
+    public void transactionExample() throws Exception {
```
jodygarnett: Future idea: May want to migrate the transaction test from PropertyDataStore and get a real test going on here.

modules/unsupported/csv/src/main/java/org/geotools/data/csv/CSVFeatureStore.java
travislbrundage: Looked back at this and I checked an [older snapshot of geoserver](https://github.com/travislbrundage/geoserver/tree/db287497cfdfbe7cff827c3c400be40113462461/src/extension/importer/core/src/main/java/org/geoserver/importer/csv), and this file was never present there.

What happened was the original gt-csv module was made for geotools and had this file.
Geoserver forked it and got rid of the write functionality, including removing this file.
It got backported into geotools again.
Now I am updating it to have write functionality again, this time with strategy objects, and now the code is here. So, I believe it is fine.

modules/unsupported/csv/src/main/java/org/geotools/data/csv/CSVFeatureWriter.java
travislbrundage: The [docs for CsvWriter](http://javacsv.sourceforge.net/com/csvreader/CsvWriter.html#close()) says it does, and I did a quick test, and confirmed. It does close whatever you pass it. :)

modules/unsupported/csv/src/main/java/org/geotools/data/csv/parse/CSVLatLonStrategy.java
travislbrundage: I don't see a duplication? I think the confusion is because I moved the location of the methods - so it's showing them as a "diff" even though they aren't changed, just moved.

travislbrundage:
Fixed the few issues that were remaining and addressed @jodygarnett 's feedback, so this branch should be ready to pull.

Noted future improvements/refactoring (probably won't make it into this release):
* Combine the LatLon strategies - there's no need for a LatLon and SpecifiedLatLon strategy. The current latlon strategy already parses for the specific lat/lon columns (ie LAT or lat or latitude) and uses them but doesn't save this info. We can just save it and use the fields the way the specifiedLatLon strategy does.
* Fix the builder API to be a bit cleaner as @jodygarnett suggested - this fix will be outside the gt-csv module, but then we can update it to be cleaner.
* Fix the ugly support methods in CSVStrategy. They can be overriden in specific strategies for now, but the general method is still confusing and could be made more readable and sensible.
* Make better tests for CSVWriteTest. Right now it doesn't even really "test" anything, it just prints things out. Fixing it up with some asserts or borrowing tests like PropertyDataStore's transcaction test would improve it.

PR #694, opened 2015-01-16
Added reprojection and forced CRS handling to ContentFeatureSource

tbarsballe:
API Clarifications to Query javadoc.
Added forced CRS handling to ContentFeatureSource getBounds and getReader.
Added reprojection handling to getBounds.
Added test cases for forced CRS handling and reprojection to MemoryDataStore (the ContentFeatureSource test mocks are insufficient to test the changes to ContentFeatureSource). Also updated DataTestCase to use EPSG:4326 for the the river data set, to allow for testing tranformations and pre-assigned CRSs (Verified that this does not break any other test cases).

### DIFF
--- a/modules/library/data/src/main/java/org/geotools/data/store/ContentFeatureSource.java
+++ b/modules/library/data/src/main/java/org/geotools/data/store/ContentFeatureSource.java
@@ -2,7 +2,7 @@
  *    GeoTools - The Open Source Java GIS Toolkit
  *    http://geotools.org
  * 
- *    (C) 2002-2008, Open Source Geospatial Foundation (OSGeo)
+ *    (C) 2002-2015, Open Source Geospatial Foundation (OSGeo)
  *
  *    This library is free software; you can redistribute it and/or
  *    modify it under the terms of the GNU Lesser General Public
@@ -432,6 +432,7 @@ public abstract class ContentFeatureSource implements SimpleFeatureSource {
             CoordinateReferenceSystem sourceCRS = query.getCoordinateSystem();
             CoordinateReferenceSystem targetCRS = query.getCoordinateSystemReproject();
             CoordinateReferenceSystem nativeCRS = getSchema().getCoordinateReferenceSystem();
+            
             if (sourceCRS != null && !sourceCRS.equals(nativeCRS)) {
                 //override native crs
                 bounds = new ReferencedEnvelope(bounds, sourceCRS);
@@ -670,7 +671,6 @@ public abstract class ContentFeatureSource implements SimpleFeatureSource {
             CoordinateReferenceSystem sourceCRS = query.getCoordinateSystem();
             CoordinateReferenceSystem targetCRS = query.getCoordinateSystemReproject();
             CoordinateReferenceSystem nativeCRS = reader.getFeatureType().getCoordinateReferenceSystem();
-
             
             if (sourceCRS != null && !sourceCRS.equals(nativeCRS)) {
                 //override the nativeCRS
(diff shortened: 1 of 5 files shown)


modules/library/data/src/main/java/org/geotools/data/store/ContentFeatureSource.java
```diff
@@ -425,6 +427,35 @@ public final ReferencedEnvelope getBounds(Query query) throws IOException {
...
+                if(sourceCRS == null) {
+                    throw new IOException("Cannot reproject data, the source CRS is not available");
+                } else if(!sourceCRS.equals(targetCRS)) {
+                    try {
+                        bounds = bounds.transform(targetCRS, true);
+                    } catch (Exception e) {
+                        if(e instanceof IOException)
+                            throw (IOException) e;
+                        else
+                            throw (IOException) new IOException("Error occurred trying to reproject data").initCause(e);
```
jodygarnett: Could probably use new IOException("Unable to reproject bounds",e)

modules/library/data/src/main/java/org/geotools/data/store/ContentFeatureSource.java
```diff
@@ -425,6 +427,35 @@ public final ReferencedEnvelope getBounds(Query query) throws IOException {
         } else {
             bounds = getBoundsInternal(query);
         }
+        // reprojection
+        if ( !canReproject() ) {
+            CoordinateReferenceSystem sourceCRS = query.getCoordinateSystem();
+            CoordinateReferenceSystem targetCRS = query.getCoordinateSystemReproject();
+            CoordinateReferenceSystem nativeCRS = getSchema().getCoordinateReferenceSystem();
```
jodygarnett: Nice clear code @tbarsballe, I am impressed you even updated the date on these files.

modules/library/data/src/test/java/org/geotools/data/memory/MemoryDataStoreBoundsTest.java
```diff
@@ -53,8 +56,9 @@ protected void setUp() throws Exception {
         data = new MemoryDataStore();
         data.addFeatures(roadFeatures);
         
-        SimpleFeatureType riverTypeWithCrs = SimpleFeatureTypeBuilder.retype(riverType, DefaultGeographicCRS.WGS84);
+        SimpleFeatureType riverTypeWithCrs = SimpleFeatureTypeBuilder.retype(riverType, CRS.decode("EPSG:4326"));
```
jodygarnett: DefaultGeographicCRS.WGS84 and EPSG:326 represent the same idea mathematically. We *may* get a slight difference based on axis order settings.

I got no problem using EPSG:4326 here as it is consistent with testing reprojection code.

jodygarnett:
Thanks @tbarsballe, quite impressed with your clarity on this one (great when reviewing a couple days before release).

modules/library/data/src/main/java/org/geotools/data/store/ContentFeatureSource.java
tbarsballe: Yeah, I just used the same structure and wording as in the reprojection section of getFeatures() for consistency.

modules/library/data/src/test/java/org/geotools/data/memory/MemoryDataStoreBoundsTest.java
tbarsballe: Some of the comparaison tests were failing with the previous configuration (Specifically assertEquals against the features and the CRSs). While they may be the same mathematically, they have slightly different WKT representations. Consequently, I decided to stick with EPSG:4326 for consistency.

PR #695, opened 2015-01-17
Always use english DecimalFormatSymbols for CQL numberFormat() function

sampov2:
Ensures server locale does not affect how numbers are converted to strings when evaluating CQL expressions.


jodygarnett:
Did we have a bug report for this one? It would be nice if it makes the release notes ...

PR #586, opened 2014-11-04
GEOT-4936 added support for instance instead of port & GEOT-4476 modify SQLServer datastore to use JTDS jdbc driver

ianturton:
* https://jira.codehaus.org/browse/GEOT-4936
* https://jira.codehaus.org/browse/GEOT-4476

PR #696, opened 2015-01-19
Jodygarnett sqlserver jtds

ianturton:
This adds support for the JTDS driver in addition to the Microsoft one, also allows users to use the INSTANCE rather than the PORT parameter if they have more than one DB installed. 

Replaces PR https://github.com/geotools/geotools/pull/586 .


jodygarnett:
I am confused on this one @ianturton, your pull requests contain many commits that have already been applied to master? I will take one shot at checking this out and rebasing it ...

PR #699, opened 2015-01-20
GEOT-4984 add INSTANCE parameter and GEOT-4476 allow use of JTDS driver

ianturton:
Allows users with more than one DB on their machine to use the instance parameter to pick the right one.

With dbtype parameter jtds-sqlserver selects the JTDS driver which we can distribute.

### DIFF
--- a/docs/user/library/jdbc/sqlserver.rst
+++ b/docs/user/library/jdbc/sqlserver.rst
@@ -6,6 +6,7 @@ Supports direct access to an SQL Server database.
 References:
 
 * http://www.microsoft.com/sqlserver/
+* http://jtds.sourceforge.net/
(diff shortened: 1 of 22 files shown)


aaime:
We were talking about this one during the meeting, and since there has been no movement, we were considering merging it on master after the freeze ends?

ianturton:
Hopefully I'll get a chance to add some docs by then, client is out of money so this back to my spare time

aaime:
Freeze is over, I guess it's safe to merge, but right now it conflicts with all the test class name changes...

ianturton:
I'll try to bring it up to date with those changes (I was afraid that would break),

PR #701, opened 2015-01-20
Fixed events implementation for ContentFeatureSource

tbarsballe:
Changed ContentState to issue events to only the current transaction on ROLLBACK, and only other transactions on COMMIT.

Fixed and improved ContentFeatureSourceEventsTest.java to test events properly. Added tests for CHANGED events and multiple changes in a batch event.

Fixed event tests for MemoryDataStore.

### DIFF
--- a/modules/library/data/src/test/java/org/geotools/data/memory/MemoryDataStoreTest.java
+++ b/modules/library/data/src/test/java/org/geotools/data/memory/MemoryDataStoreTest.java
@@ -1251,7 +1251,7 @@ public class MemoryDataStoreTest extends DataTestCase {
 
         store1.getTransaction().commit();
 
-        assertEquals(1, listener1.events.size()); //This is wrong (expect 0)
+        assertEquals(0, listener1.events.size());
 
         assertEquals(1, listener2.events.size());
         event = listener2.getEvent(0);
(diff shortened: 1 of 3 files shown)


modules/library/data/src/test/java/org/geotools/data/memory/MemoryDataStoreTest.java
```diff
@@ -1292,7 +1292,7 @@ FeatureEvent getEvent( int i ) {
         event = listener1.getEvent(0);
         assertEquals(feature.getBounds(), event.getBounds());
         assertEquals(FeatureEvent.Type.ADDED, event.getType());
-        assertEquals(1, listener2.events.size()); //this is wrong (expect 0)
+        assertEquals(1, listener2.events.size());
     }
```
tbarsballe: This is actually correct, since we are in the AUTO_COMMIT transaction here

jodygarnett:
Does this have a Jira?

jodygarnett:
Let me look:

* https://jira.codehaus.org/browse/GEOT-1327 Notification mechanism of feature events needs revisiting
* https://jira.codehaus.org/browse/GEOT-2832 H2 datastore cannot deal with multiple parallel transactions
* https://jira.codehaus.org/browse/GEOT-1113 Shapefile datastore should batch event notification
* https://jira.codehaus.org/browse/GEOT-3970 ContentEntry.notifiyFeatureEvent does not notify all listeners
* https://jira.codehaus.org/browse/GEOT-3956 ContentFeatureStore should provide event support to implementors
* https://jira.codehaus.org/browse/GEOT-681 WFS does not raise correct events when editing features

PR #705, opened 2015-01-25
GeoPackage databases should be initialized with default spatial references

jericks:
Solved issue #4995.

http://jira.codehaus.org/browse/GEOT-4995

Thanks!
Jared


jdeolive:
+1. Merge at your convenience Jared.

PR #707, opened 2015-01-27
GeoPackage should encode SRID with Geometry.

jericks:
Solves issue #4996.

http://jira.codehaus.org/browse/GEOT-4996

GeoPackage encodes the SRID with the geometry, but currently it is not set on the Geometry with setSRID. 

Thanks,
Jared


jdeolive:
+1, looks good Jared.

jericks:
Thanks Justin!

PR #708, opened 2015-01-27
Allow AnchorPoint to be optional when encoding SLD PointPlacement [GEOT-4997]

jodygarnett:
https://jira.codehaus.org/browse/GEOT-4997


aaime:
No objection to a new test, but why there is no code change?

jodygarnett:
It appears I lost it when squashing to one commit. Ideas is to introduce a null check:

```
            if (pp.getAnchorPoint() != null) {
                pp.getAnchorPoint().accept(this);
            }
```

jodygarnett:
I see something similar was introduced on master already.

PR #709, opened 2015-01-28
GEOT-4999: Adding support for bbox types to solr datastore.

jdeolive:
Includes introducing the abstraction called "SpatialStrategy" to encapsulate
the handling of different geometry types.

### DIFF
--- a/modules/unsupported/solr/src/main/java/org/geotools/data/solr/SolrFeatureReader.java
+++ b/modules/unsupported/solr/src/main/java/org/geotools/data/solr/SolrFeatureReader.java
@@ -23,6 +23,7 @@ import java.util.List;
 import java.util.NoSuchElementException;
 import java.util.logging.Level;
 
+import com.vividsolutions.jts.geom.GeometryFactory;
 import org.apache.commons.lang.StringUtils;
 import org.apache.solr.client.solrj.SolrQuery;
 import org.apache.solr.client.solrj.SolrServerException;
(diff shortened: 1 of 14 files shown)


modules/unsupported/solr/src/main/java/org/geotools/data/solr/SolrFeatureReader.java
```diff
@@ -168,7 +170,9 @@ public SimpleFeature next() throws IOException, IllegalArgumentException,
                 if (type instanceof GeometryDescriptor) {
                     GeometryDescriptor gatt = (GeometryDescriptor) type;
                     if (value != null) {
-                        Geometry geometry = wktReader.read(value.toString());
+                        SolrSpatialStrategy spatialStrategy = SolrSpatialStrategy.createStrategy(gatt);
+                        Geometry geometry = spatialStrategy.decode(value.toString());
```
aaime: Wondering if creating the strategy object at each feature read might be too much overhead.. maybe it should be created once when the reader is instantiated instead?

aaime:
Looking good, minus one potential performance issue in the reader.

jdeolive:
Cool, thanks Andrea. Fixed now and took it one step further making the two strategies singletons since they don't have be instantiated on an attribute by attribute basic, and are thread safe, etc... 

I'll go ahead and merge unless I hear anything back. Thanks for the review.

aaime:
+1

PR #710, opened 2015-01-30
GEOT-4998 improved check on empty intersection 

jdries:
Fix for GEOT-4998 to actually filter out empty intersections. The current code did not manage to filter out a 'POLYGON EMPTY' geometry in JTS.
I did not use the relate() or intersects() method provided by JTS as I was not sure of the performance impact. I did do some local benchmarking, which showed that the relate() method was definetly not faster, perhaps even a bit slower. Of course, anyone with more knowledge of JTS might contradict this, as my benchmark setup was not very scientific.


aaime:
Hi Jeroen, seems ok, two things:
* Any chance you can setup a test for this one? We might be reworking the footprint management soon, I'd hate to have a regression on this one
* Can you remove the TODO and add your explanation instead?

PR #711, opened 2015-01-30
GEOT-4992: fixing maven profile to skip image comparison tests

dromagnoli:


jodygarnett:
So why default to skipping tests?

simboss:
exact same question here.

aaime:
As far as I can see, the default value for the property is false, and there is the skip.image.tests profile that can turn it to true when enabled. What am I missing?

dromagnoli:
I confirm Andrea's feedback: The property is false by default (NO SKIP). You may skip tests by setting the skip.image.tests profile.

PR #713, opened 2015-01-30
GEOT-5003: fixing softValueHashMap import

dromagnoli:


bencaradocdavies:
Caused a compile failure in Eclipse (now fixed):
https://jira.codehaus.org/browse/GEOT-5011

PR #723, opened 2015-02-02
[GEOT-5006] ProjectionHandler should not intersect Geometries with wrong GeometryType

n-lagomarsini:
This pull request contains a fix for JIRA https://jira.codehaus.org/browse/GEOT-5006.

### DIFF
--- a/modules/library/render/src/main/java/org/geotools/renderer/crs/ProjectionHandler.java
+++ b/modules/library/render/src/main/java/org/geotools/renderer/crs/ProjectionHandler.java
@@ -441,12 +441,18 @@ public class ProjectionHandler {
                 // go piecewise, the JTS intersection can be pretty fragile in these cases
                 // and take a lot of time
                 List<Geometry> elements = new ArrayList<>();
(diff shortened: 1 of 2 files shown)


modules/library/render/src/main/java/org/geotools/renderer/crs/ProjectionHandler.java
```diff
@@ -441,12 +441,18 @@ private Geometry intersect(Geometry geometry, Geometry mask) {
...
+                String geometryType = geometry.getGeometryN(0).getGeometryType();
                 for (int i = 0; i < geometry.getNumGeometries(); i++) {
                     Geometry g = geometry.getGeometryN(i);
                     if (g.getEnvelopeInternal().intersects(mask.getEnvelopeInternal())) {
                         Geometry intersected = intersect(g, mask);
                         if (intersected != null) {
-                            elements.add(intersected);
+                            if(intersected.getGeometryType().equals(geometryType)){
+                                elements.add(intersected);
+                            } if(intersected instanceof GeometryCollection){
```
aaime: else if?

PR #726, opened 2015-02-04
GEOT-5000: Support for mapping solr index to single layer.

jdeolive:
Also added LayerMappingStrategy with an implementation for mapping to single layer.


jdeolive:
@aaime First cut at adding a different mapping to the solr datastore. Let me know what you think.

aaime:
Looks all reasonable to me.

jdeolive:
Great, thanks Andrea. I'll go ahead and merge. Fwiw I also updated the docs on the geoserver side as well. See geoserver/geoserver#911

PR #727, opened 2015-02-04
Using FieldInfo.getDocs() rather than query per attribute. GEOT-5009.

jdeolive:


jdeolive:
Couldn't really think up of a test for this one... other than mocking up the collaborates and asserting various method calls, etc... all tests do pass with this patch applied though, and run a little faster too :)

aaime:
Right, an optimization with no behavior changes cannot really be tested :-p

PR #729, opened 2015-02-07
[GEOT-5014] SimplifyingFilterVisitor breaks MatchAction on multivalued b...

bencaradocdavies:
...inary comparison operators

This fixes (some) failures in GeoServer app-schema-test online tests.
See: https://jira.codehaus.org/browse/GEOT-5014

@aaime your review would be much appreciated.

### DIFF
--- a/modules/library/main/src/main/java/org/geotools/filter/visitor/SimplifyingFilterVisitor.java
+++ b/modules/library/main/src/main/java/org/geotools/filter/visitor/SimplifyingFilterVisitor.java
@@ -31,6 +31,7 @@ import org.opengis.filter.BinaryLogicOperator;
 import org.opengis.filter.Filter;
 import org.opengis.filter.FilterFactory2;
 import org.opengis.filter.Id;
+import org.opengis.filter.MultiValuedFilter.MatchAction;
 import org.opengis.filter.Not;
 import org.opengis.filter.Or;
 import org.opengis.filter.PropertyIsBetween;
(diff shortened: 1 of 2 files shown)


modules/library/main/src/main/java/org/geotools/filter/visitor/SimplifyingFilterVisitor.java
```diff
@@ -388,42 +389,68 @@ public Object visit(Id filter, Object extraData) {
         return validIdFilter;
     }
 
+    /**
+     * Return true iff there is a MatchAction for the negation of a multivalued binary comparison operation.
+     */
+    private boolean isNegatableMatchAction(MatchAction matchAction) {
+        return matchAction == MatchAction.ANY || matchAction == MatchAction.ALL;
```
aaime: Are you completely sure filters cannot be built with a null matchaction? Maybe some by old code that predates the matchaction support?

modules/library/main/src/test/java/org/geotools/filter/visitor/SimplifyingFilterVisitorTest.java
```diff
@@ -224,8 +228,8 @@ public void testNegateBetween() {
         Literal l20 = ff.literal(20);
         Filter f = ff.not(ff.between(prop, l10, l20));
         Filter result = (Filter) f.accept(visitor, null);
-        assertEquals(ff.or(Arrays.asList((Filter) ff.less(prop, l10), ff.greater(prop, l20))),
-                result);
+        // *not* simplified
+        assertEquals(f, result);
```
aaime: Wait what? Why not simplified? This is a regression.

modules/library/main/src/main/java/org/geotools/filter/visitor/SimplifyingFilterVisitor.java
bencaradocdavies: @aaime I checked every implementation of MultiValuedFilter and all either hardcode getMatchAction() to return MatchAction.ANY or set MatchAction.ANY in all constructors that do not specify it. I am confident that there are no backward compatibility concerns. It is possible to set MatchAction to null but only by using a post-MatchAction constructor, and anyone doing so should know better. Cannot happen with old code.

modules/library/main/src/test/java/org/geotools/filter/visitor/SimplifyingFilterVisitorTest.java
bencaradocdavies: @aaime it is not a regression to remove a feature that does not work.

Suppose the property has values [5, 15, 25]. The pseudocode filter not(any-between(10, 20)) evaluates to false. Simplification converts this to or(any-less-than(10), any-greater-than(20)), which evaluates to true. Simplification is thus unvalid for multivalued properties.

The problem is that, to convert "between" to an "or" of "less-than" and "greater-than" requires consideration of *which* property values each child filter matches, which is not supported by the filter API. To do this would require something like a match-index array of booleans as the returned value, as in NumPy, and an element-wise "or".

I tried adding a (featureType instanceof SimpleFeatureType) guard but featureType is null for this test. I have not thought of a solution that works with the current filter API and does not break multivalued properties.

aaime: The first translation you made is wrong, let's do it step by step:
not(any-between(10, 20))  -> (expand between)
not(and(any-greater-equal(10), any-less-equal(20)) -> (apply De Morgan)
or(not(any-greater-equal(10), not(any-less-equal(20)) -> (apply your comparison inversions)
or(all-less-than(10), all-greater-than(20))

If we go back now to his example, this translation seems to work, any-between(10,20) returns true, negated to false.
all-less-than(10) is false, and all-greater-than(20) is false too, thus false, thus equivalent at least for this example.

However, it does not take much to prepare an example where the equivalence is not met, if the multi-valued property has values [5,25] then not(any-between(10,20)) return true, but or(all-less-than(10), all-greater-than(20)) still returns false.

This also has a severe implication, because the original filter could have been just have expressed as not(and(any-greater-equal(10), any-less-equal(20)), which
is something anyone can enter, no? At which point did we lose the equivalence?

For the [5,25] case:

not(any-between(10, 20))  -> true
not(and(any-greater-equal(10), any-less-equal(20)) -> true
or(not(any-greater-equal(10), not(any-less-equal(20)) -> true
or(all-less-than(10), all-greater-than(20)) -> false

And well, the issue is in the last step, if we just consider the inversion of not(any-greater-equal(10)) -> all-less-than(10) you can see that just negating the match-action does not do you any good, it is indeed not equivalent.

Long story short, the logic inversion here should just not be applied when we are dealing with a complex feature type, or we don't know the target feature type (well actually, if one wanted, a check of whether the properties involved are multi-valued or not could be performed, but I'll leave that as an excercise the reader, app-schema has much more serious issues than some filter simplification right now).

aaime:
Closing, replaced by #737 
Ben please review, I know that Jody has the finger on the trigger to cut RC1

PR #731, opened 2015-02-12
Use reflection to configure JAI logging [GEOT-5016]

jodygarnett:
Use reflection to redirect JAI ImageListener to our logging system without tripping up when running in environments where JAI is not available.

See https://jira.codehaus.org/browse/GEOT-5016


aaime:
I have the impression we are using more reflection than strictly necessary, we could refer to the classes directly after the class.forName, if that one fails the rest of the code won't be executed and thus no extra classloading attempts will be made, which should prevent errors just as well.
But no strong objection, -0

ischneider:
If the same can be accomplished without the extra complexity, agreed it would be good to remove. If not, document why?

jodygarnett:
Thanks for the feedback, squashed this down to a single commit for master and back port.

PR #733, opened 2015-02-12
Cache type names

jodygarnett:


modules/library/data/src/main/java/org/geotools/data/store/ContentDataStore.java
```diff
@@ -632,9 +637,21 @@ final protected void removeEntry(Name name) {
...
+        }
     }
 
     /**
+     * Check if {@link #createTypeNames()} needs to be called.
+     * <p>
+     * If typeNames have been added or removed 
+     * @return true if createTypeNames needs to be called.
+     */
+    protected boolean checkTypeNames(){
```
aaime: Code formatting --> checkTypeNames() {

modules/library/data/src/main/java/org/geotools/data/store/ContentDataStore.java
```diff
@@ -694,12 +711,10 @@ public SimpleFeatureSource getFeatureSource(Name typeName)
      * @see DataAccess#getNames()
      */
     public List<Name> getNames() throws IOException {
-        String[] typeNames = getTypeNames();
-        List<Name> names = new ArrayList<Name>(typeNames.length);
-        for (String typeName : typeNames) {
-            names.add(new NameImpl(typeName));
+        if( typeNames == null || checkTypeNames() ){
```
aaime: Likewise... Jody do you have the GeoTools coding conventions loaded?

aaime:
Looking good, only added a couple of minor comments. I guess something similar could be useful for the shapefile and directory data store (but of course not required to be in this pull request).

docs/user/tutorial/datastore/artifacts/pom.xml
```diff
@@ -13,7 +13,7 @@
 
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    <geotools.version>12-RC1</geotools.version>
+    <geotools.version>13-beta</geotools.version>
```
jodygarnett: Should be 13-RC1 - this should be updated as part of the release script.

bencaradocdavies:
@jodygarnett , @aaime  I do not like this new caching of typeNames.

Firstly, File.lastModified only has *second* precision on Linux before JDK8 so there is a huge race condition in the use of PropertyDataStore.watch: http://bugs.java.com/view_bug.do?bug_id=6939260

Secondly, there is no division of responsibility for updating typeNames; should this be done in ContentDataStore or its subclass? How can PropertyDataStore.watch be set without knowing if the caller of createTypeNames is storing the output in ContentDataStore.typeNames? One code path does not store the output, breaking GeoServer master tests. Try adding this to PropertyDataStoreTest to see the failure in GeoTools:

    public void testGetNamesAfterGetFeatureSource() throws IOException {
        store.getFeatureSource("road");
        testGetNames();
    }

aaime:
I think Jody will be on it next week, but since you are calling us out... the pull request has been out for almost a month before being merged... where were you then? :)

bencaradocdavies:
@aaime you know that I am too lazy to take notice until my beloved unit tests start failing.  ;-)

I am now testing a GT fix with a  full build of GT+GWC+GS+app-schema-online-test; if it passes I will push to GT master to let @jodygarnett off the hook and stop @NielsCharlier  from needlessly blaming himself for breaking the GS build.

Anyway, the one thing I will not fix is the use of File.lastModified when watching for property file changes. Are my Linux time resolution concerns justified? I do not have a better solution, and I do not know what has been backported to JDK7. @aaime your advice would be appreciated.

bencaradocdavies:
@aaime @jodygarnett I just ran a test and sure enough all File.lastModified of files in /usr/bin and my home directory were rounded *down* to a whole second. I am using:
openjdk-7-jdk:amd64 7u75-2.5.4-3 amd64 debian/sid

This means that any new types added in the second after the last update may be ignored.

I suspect that some logic using another field that compares against System.currentTimeMillis might be required. I do not know if there are any guarantees that network shares have a sensible time relationship with system time so I do not know if this is less bad.

bencaradocdavies:
@aaime @jodygarnett this attempted patch fixed the gs-main failure but there are still many other failures in gs-wfs:
https://github.com/bencaradocdavies/geotools/commit/05a769a8219c38528847e3f64060ffbabde12d5a

Even reverting the two substantive commits in this pull request, aec9ef830b374eca6988e60984c46041f2c078b0 and edbaea4d16841cdbd246917546848bc1ef52a82a, did not fix GeoServer. There are still gs-wps failures, so I did not push the reversions.

aaime:
Hi Ben, this thread seems to confirm that we are not going to be able to get precise file modification times in java7, and while we might get them out of java 8, there is no guarantee as that also depends on the underlying OS and libraries: http://stackoverflow.com/questions/24804618/get-file-mtime-with-millisecond-resolution-from-java

So... this caching makes sense at the human being level, and as a performance optimization in a system where you have other means of dropping the cache (like, forcing a refresh in GeoServer), but it seems as written it cause troubles. 

The gs-wps failures, do you have details? I'm travelling, doing a full build uses too much battery.

bencaradocdavies:
I have identified the (unrelated) commit causing the gs-wps-core failure and reverted it. I have also reverted the two commits aec9ef830b374eca6988e60984c46041f2c078b0 and edbaea4d16841cdbd246917546848bc1ef52a82a from this pull request to fix the geoserver build.

PR #736, opened 2015-02-14
[GEOT-4885] GeoTools stops rendering valid geometry elements after an invalid geometry element

ridethepenguin:
The issue involved malformed polygons containing either unclosed rings or rings with less than 4 points. Code in class `org.geotools.data.oracle.sdo.SDO` has been amended to make sure a `CoordinateSequence` forms a valid linear ring, before invoking `LinearRing`'s constructor.


aaime:
Thank you!

PR #737, opened 2015-02-14
[GEOT-5014] SimplifyingFilterVisitor breaks MatchAction on multivalued binary comparison operators

aaime:


aaime:
This pull request fixes the problem by avoiding the usage of certain types of simplifications against complex types, that might have multivalued properties (which is not easy to check btw, as the code works against generic expressions, non properties.

In order to retain the simplification advantage in most cases, it also makes sure the client code passes down the target feature type, which is almost always a simple one (besides one case in app-schema, and some cases in which it's simply not known).

aaime:
Ah, the change got a full build of both geotools and geoserver, with extensions included, on my box, both successes.

bencaradocdavies:
Andrea,

I can confirm that full geotools and geoserver builds pass using your geot-5014-v2 branch. I also ran all app-schema online tests against postgis and they also passed (I do not have oracle but the GEOT-5014 failure was seen in postgis as well). I have merged this pull request. Thanks very much for your prompt resolution of this issue.

My only remark from reviewing this pull request is that you have now disabled simplification for all complex feature types. In my view, with my MatchAction change, simplification of negation of multivalued binary comparison operators worked fine except in the case of is-between. However, I did not address range simplification. Your approach avoids a partial implementation and may be easier to maintain and describe.

Kind regards,
Ben.

PR #589, opened 2014-11-11
 ComplexDataTypeBinding encode override - GEOT-4934

afabiani:
https://jira.codehaus.org/browse/GEOT-4934

From Jira: The ComplexDataTypeBinding ignores completely the data attribute of hte Type.
Since the attribute data is not recognized as a property of the binding type, we need to envisage the override of the "encode" method.

PR #738, opened 2015-02-15
[GEOT-4934] ComplexDataTypeBinding does not encode its complex contents

aaime:


aaime:
Jody, Alessio, this pull request is meant to replace https://github.com/geotools/geotools/pull/589
For Jody, I've put some rationale of what we're trying to do here in the ticket: http://jira.codehaus.org/browse/GEOT-4934
For Alessio, I've simplified the previous code quite a bit by backporting the same approach used by the ComplexDataTypeBinding override that we have in GeoServer WPS (which might have become unecessary, but I'll have to check that later). Can you verify if this simpler approach still satisfies your requirements? It certainly makes the test you wrote pass, but I don't have the end to end case.

aaime:
Damn, there is a reformat in the binding, I'll fix it.

PR #740, opened 2015-02-16
[GEOT-4873] Provide useful error message in GeoTiffReader for file read errors.

lossyrob:
This addresses the issue of an unuseful error message (namely: "No input stream for the provided source") when the `GeoTiffReader` is given a file path that does not exist, is a directory, or when the user does not have permissions to the file.

The issue arises from `imageio-ext` code: [This code](https://github.com/geosolutions-it/imageio-ext/blob/master/library/streams/src/main/java/it/geosolutions/imageio/stream/input/spi/FileImageInputStreamExtImplSpi.java#L159) in imageio-ext's `FileImageInputStreamExtImplSpi` swallows any IO exceptions that are thrown from [this code](https://github.com/geosolutions-it/imageio-ext/blob/master/library/streams/src/main/java/it/geosolutions/imageio/stream/input/FileImageInputStreamExtImpl.java#L216) in `FileImageInputStreamExtImpl`. So instead of the IOExceptions bubbling up as [GEOT-4873] desires, we instead get logging of the error if the Level.FINE level is enabled, and get a generic exception thrown instead.

The solution proposed here is to do the File existance/permission check manually if the input stream creation fails for a `File` input. Though not ideal, since it is duplicating checks that already exist in the code path, it does side step the design choice made by the imageio-ext library to not have those useful exceptions bubble up to the client.

### DIFF
--- a/modules/plugin/geotiff/src/main/java/org/geotools/gce/geotiff/GeoTiffReader.java
+++ b/modules/plugin/geotiff/src/main/java/org/geotools/gce/geotiff/GeoTiffReader.java
@@ -184,6 +184,7 @@ public class GeoTiffReader extends AbstractGridCoverage2DReader implements GridC
 			}
 
 			closeMe = true;
+
 			// /////////////////////////////////////////////////////////////////////
 			//
 			// Get a stream in order to read from it for getting the basic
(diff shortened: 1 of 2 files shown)


dromagnoli:
Hi @lossyrob,
thanks for the fix.  I'm going to merge it. I have also opened a ticket on imageio-ext in order to check that on the lower level. Standard Oracle's ImageInputStreamSpi behave the same way. The Javadoc states that they throw an IOException in case the cache file can't be created.
We could consider revisiting the current behaviour on ImageIO-Ext subclasses.

https://github.com/geosolutions-it/imageio-ext/issues/92

dromagnoli:
I have also backported your fix to 13.x and 12.x

PR #743, opened 2015-02-19
[GEOT-4946] Fixes null geom when reading schema

williamvoor:
FeatureJSON#readFeatureCollectionSchema does not support null geometry.
It now checks if the geometry is null.
Also removed check that forced primite types to be always the same when parsing entire JSON.


williamvoor:
Will reopen these against 13.x

aaime:
Please reopen against master, that's the branch that first gets any change. Then we eventually backport to the stable branches.

PR #744, opened 2015-02-19
Added XML data support to JDBC and GML

tbarsballe:
Added Generic SQLXML type support to JDBC, and Native XML Support to jdbc-postgis.
Added XML Test cases, and postgis implementation
Added XML element support to GML, in the form of w3c.Node support with type xsd:anyType


tbarsballe:
For the GML support of XML data to show up properly in GeoServer, need to add `anyType` to `org/geoserver/wfs/xml/XSProfile.java`:
```java
public class XSProfile extends TypeMappingProfile {
        ...
        proper.add(name(XS.ANYTYPE));
        ...
}
```

modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/PostgisXMLDocumentConverterFactory.java
```diff
@@ -0,0 +1,34 @@
...
+import org.geotools.util.Converter;
+import org.geotools.util.XMLDocumentConverterFactory;
+import org.postgresql.util.PGobject;
+
+public class PostgisXMLDocumentConverterFactory extends
+        XMLDocumentConverterFactory {
+
+    @Override
+    public Converter createConverter(Class source, Class target, Hints hints) {
+        if (source.equals(PGobject.class)) {
```
aaime: Somewhat confused here... the base class is already registered in SPI, and available in the classpath, and also checks the source like this one. So, what exactly is this one adding that the base class is not already covering? For the PGObject to be converted by the base class, it has to be a Node no? It seems this class and check are un-necessary?

tbarsballe: PGObject is the SQL object, Node is the WC3 object. This case converts the PGObject from a string to a DocumentFragment (which extends Node).

aaime: If the source is a string... how can it be a PGObject at the same time? String is final

aaime: Ah ok, I did not notice that you're forcing the first argument to string, and then things work because the toString() of such pgObject happens to be the xml document itself.

This looks like a hack, generally speaking one cannot depend on toString() behavior, although in this particular case PGObject toString API state "the value of this object, in the syntax expected by org.postgresql"... so it might work fine, limited to PGObject.

I suggest you make a specific converter for PGObject instead of all these assumptions around the code that is supposed to be generic. Also, this would make it clearer what converter manages which conversions, right now you end up with the PostgisXMLDocumentConverterFactory being able to do the same conversions as the base class, which is also registered and available, plus one.

tbarsballe: I was ultimately trying to avoid too much code duplication, since the generic converter already handles string conversion, and the only useful aspect of PGObject in this context is toString (most of our existing code also casts PGObjects to strings where applicable).
But yes, I can definetly decouple these better. Probably still going to try and call as much code as possible from the generic class.

tbarsballe: And technically this can't really do the same conversions as base, since it overrides createConverter, which defines all of the conversions in the base class.

aaime: I would like Jody to have a look. To me, this is messy, but I'll trust his judgment on this one.

tbarsballe: @jodygarnett Could you take a look here?

aaime:
Hmm... two months since last change, has this one stalled?

tbarsballe:
Yes, this one has stalled.
At the point this has reached, XML extraction from PostGIS works, but there are still a few snags with certain output formats (mainly GML). More in-depth discussion on the geotools mailing list, "JDBCDataStore XML Support" (Last reply March 5). I have some work on a namespace fix from early March on my local machine, but it is failing a number of test cases. 

I don't see this progressing any further any time soon, as it is currently a low priority for me.

aaime:
Torbin, Jody, what shall we do with this one? Link from ticket and close until someone has time to complete the work?

tbarsballe:
I think that will work. I have created a geotools ticket for this one here: https://osgeo-org.atlassian.net/browse/GEOT-5081(There wasn't one yet).

PR #747, opened 2015-02-22
Fixes for GEOT-4946, GEOT-5026, and GEOT-5045

williamvoor:


aaime:
@jdeolive, how are we doing here?

williamvoor:
I have updated this patch as per Justin's suggestions.

jdeolive:
Looks great William! Thanks for the work on this one.

jdeolive:
@aaime What's the current state of back ports? Back port to 13.x and 12.x if stable enough?

aaime:
If they are bugfixes yep, you can backport without asking (unless we're doing a release like the next day.... but I don't think we are releasing from those branches until next month, don't we?)

PR #748, opened 2015-02-24
Add dependency to "batik-codec" artifact

snotling:
Adding "batik-codec" to the classpath automatically makes Batik support raster images in SVG output (through Base64 encoding).
Tested in GeoServer.

### DIFF
--- a/modules/plugin/svg/pom.xml
+++ b/modules/plugin/svg/pom.xml
@@ -85,6 +85,10 @@
       <groupId>org.apache.xmlgraphics</groupId>
       <artifactId>batik-transcoder</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.apache.xmlgraphics</groupId>
+      <artifactId>batik-codec</artifactId>
+    </dependency>
     <!-- Test dependencies -->
     <dependency>
       <groupId>org.geotools</groupId>


jodygarnett:
Should this be an optional dependency?

aaime:
This module is not meant for encoding, if this dependency is needed it should be added to GeoServer wms module, which has the SVG output format

snotling:
Indeed, I wasn't sure if it was more appropriate to add this dependency in GeoTools or GeoServer.

aaime:
Please add to GeoServer, and see if you can extend one of the tests to prove raster encoding works (and to make it break in case we ever remove that dependency... many people contributing over many years, the only truly reliable memory we have of any decision in the code is tests, it stays with the project as people come and go).

snotling:
Furthermore, the plot thickens: adding `batik-codec.jar` to GeoServer 2.5.x works fine but it breaks on GeoServer 2.6.x.
In this case, WMS GetMap requests as `image/svg` which happen to include a raster layer then throw an error:
```
<ServiceException>
      java.lang.IllegalArgumentException: null incompatible with Global antialiasing enable key
null incompatible with Global antialiasing enable key
</ServiceException>
```

With this stack trace:
```
java.lang.IllegalArgumentException: null incompatible with Global antialiasing enable key
        at java.awt.RenderingHints.put(RenderingHints.java:1075)
        at org.apache.batik.ext.awt.g2d.GraphicContext.setRenderingHint(GraphicContext.java:438)
        at org.apache.batik.ext.awt.g2d.AbstractGraphics2D.setRenderingHint(AbstractGraphics2D.java:1071)
        at org.geotools.renderer.label.LabelCacheImpl.end(LabelCacheImpl.java:417)
        at org.geotools.renderer.lite.StreamingRenderer.paint(StreamingRenderer.java:846)
        at org.geoserver.wms.svg.SVGBatikMapOutputFormat.createSVGMap(SVGBatikMapOutputFormat.java:167)
        at org.geoserver.wms.svg.SVGBatikMapOutputFormat.produceMap(SVGBatikMapOutputFormat.java:86)
        at org.geoserver.wms.svg.SVGBatikMapOutputFormat.produceMap(SVGBatikMapOutputFormat.java:42)
        at org.geoserver.wms.GetMap.executeInternal(GetMap.java:510)
        at org.geoserver.wms.GetMap.run(GetMap.java:254)
        at org.geoserver.wms.GetMap.run(GetMap.java:125)
        at org.geoserver.wms.DefaultWebMapService.getMap(DefaultWebMapService.java:354)
```

Hum, I'll have to further investigate...

PR #755, opened 2015-02-28
Resolved issue [GEOT-5021](https://jira.codehaus.org/browse/GEOT-5021);

Dortund:
For some reason readGeometryTaggedText in WKT2Reader used equals() instead of equalsIgnoreCase for evaluating "POINT". All other evaluations use equalsIgnoreCase ("LINESTRING", "LINEARRING", etc).

Added a simple test case to test case insensitivity in WKTReader2Test.


aaime:
Yep, works for me

aaime:
Backported on 13.x and 12.x as well

PR #757, opened 2015-02-28
Wfs-ng: tests: refactor, reorganise, complete, fix

NielsCharlier:
All of this was done with the purpose of creating test cases for the new WFS 2.0 functionality.
I thought the test directory in wfs-ng was such a terrible mess it needed serious work first.
A summary of what happened in this commit:

 1. Removed a bunch of duplicated test support reference data xml files,
    and put all of them in one single directory structure instead of being spread around.
 2. WFFTestData and DataTestSupport were almost duplicates from each other, 
    removed the older one and converted all tests to use the newer one.
 3. There were two separate geoserveronlinetests, one for wfs1.0 and one for both wfs1.0 and wfs1.1
     made one abstract online test class and one derived test class per server/version combination.
 4. Enabled all of the disabled online test methods and made them work with the geoserver tests (release data dir).
 5. Enabled most of the disabled test methods in the integration test and made them work for the wfs1.1 geoserver
    integration test case.
 6. Fixed issues with axis order implementation so the offline and online axis order tests work.
 7. Created a very clear package/directory structure for all tests so that you can very easily see 
    which server/wfs-version combinations have tests and which don't offline and online.
 8. Renamed some confusing class names and removed unnecessary duplicated stuff
    for example there was both a TestHttpClient and TestHTTPClient.
 9. Added pullparser test which tests the new parser (that works with GML3.2).

TODO NOW
 1. make online and offline WFS 2.0 tests. THis should now be rather easy with the existing abstract offline and online
    test classes available which only require you to insert the reference data.
 2. verify the existing online tests for other servers than geoserver


aaime:
Anyone with wfs client background can review this one?

sampov2:
I quickly went through the code and left a comment regarding re-using of WFS online test code. I didn't have a chance to compile & try out the code though.

modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/InvertAxisFilterVisitor.java
```diff
@@ -66,9 +67,17 @@ public Object visit(BBOX filter, Object extraData) {
...
                 Envelope geomEnvelope = geom.getEnvelopeInternal();
+                CoordinateReferenceSystem crs = null;
+                if (geom.getUserData() instanceof CoordinateReferenceSystem) {
+                    crs = (CoordinateReferenceSystem) geom.getUserData();
+                } else if (filter.getSRS() != null) {
+                    try {
+                        crs = CRS.decode(filter.getSRS());
+                    } catch (FactoryException e) {}
+                }
                 return ff.bbox(filter.getExpression1(), new ReferencedEnvelope(
```
jodygarnett: This appears to be an actual bug fix. No problem with the code as written...

modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/internal/WFSConfig.java
```diff
@@ -43,37 +46,37 @@
  */
 public class WFSConfig {
 
-    private String user;
+    protected String user;
```
jodygarnett: are these being made protected for a subclass, or for test case visibility?

modules/unsupported/wfs-ng/src/test/java/org/geotools/data/wfs/WFSTestData.java
```diff
@@ -98,26 +101,26 @@
     }
 
     public static final TestDataType GEOS_ARCHSITES_11 = new TestDataType("GeoServer_2.0/1.1.0",
-            new QName("http://www.openplans.org/spearfish", "archsites"), "sf:archsites",
+            new QName("http://www.openplans.org/spearfish", "archsites"), "sf_archsites",
```
jodygarnett: This really does not look right ... I would expect "archsites" as that is the name of the type in DescribeFeatureType?

The "sf" is only the prefix used to look up the http://www.openplans.org/spearfish namespace - why is it being part of the name.

modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/internal/v1_x/FeatureTypeInfoImpl.java
```diff
@@ -77,7 +77,7 @@ public String getDescription() {
 
     @Override
     public String getName() {
-        return eType.getName().getLocalPart();
+        return eType.getName().getPrefix() + "_" + eType.getName().getLocalPart();
```
jodygarnett: This seems odd to me as the prefix is used to line up with a namespace and not strictly speaking part of the name.

Looking at the API this information is available:

* ResourceInfo.getSchame(): provides the namespace
* ResourceInfo.getName() provides the Name

Do we have a solid reason to include the prefix here? This is "just" for the info object and should match up with what DataStore.getFeatureSource( String ) and getFeatureSource( Name ) provides.

In this case, since WFS is "namespace aware" I would expect a user to construct a new NameImpl( res.getSchema(), res.getName() ) to look up a FeatureSource.

jodygarnett:
See comments above, a bit of inconsistency on ResourceInfo.

modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/internal/WFSConfig.java
NielsCharlier: there is a test class that makes a 'mutablewfsconfig' subclass

modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/InvertAxisFilterVisitor.java
NielsCharlier: yeah that was to making the (disabled) tests regarding axis inversion work

modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/internal/v1_x/FeatureTypeInfoImpl.java
NielsCharlier: Well first, the system as it exists today was not created by me and wfs-ng has always worked like this. But let me explain this: the prefix in the name is _not_ the prefix of the namespace of the resource in the geotools/geoserver consuming the wfs data. It is the prefix of the namespace on the wfs server. 

So for example, you have a WFS server with two layers:
blah:mylayer
bleh:myotherlayer

in geotools/geoserver you now create a wfs-ng datastore that feeds from this namespace, and the datastore is inside namespace 'ns'. You will now be able to make two layers:

ns:blah_mylayer
ns:bleh_myotherlayer

If you say that the resourceinfo.getName must correspond with DataStore.getFeatureSource( String) then that means my change is correct within the current logic.

It has always worked like this afaik.

modules/unsupported/wfs-ng/src/test/java/org/geotools/data/wfs/WFSTestData.java
NielsCharlier: See my answer below.

NielsCharlier:
Is this still pending on something?

PR #759, opened 2015-03-04
PostGIS datastore returns the correct sequence value

Gnafu:
This fixes the incorrect behaviour of the PostGIS feature store: when INSERT queries trigger more than one sequence increase, the returned feature ID could be incorrect.

### DIFF
--- a/modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/PostGISDialect.java
+++ b/modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/PostGISDialect.java
@@ -708,7 +708,15 @@ public class PostGISDialect extends BasicSQLDialect {
         
         Statement st = cx.createStatement();
         try {
-            String sql = "SELECT lastval()";
+        
+            //Retrieve the sequence of the column
+            String sequenceName = getSequenceForColumn(schemaName,tableName,columnName, cx);
+            if(sequenceName == null){
+                // There is no sequence to get the value from
+                return null;
+            }
+            
+            String sql = "SELECT currval('"+sequenceName+"')";
             dataStore.getLogger().fine( sql);
             
             ResultSet rs = st.executeQuery( sql);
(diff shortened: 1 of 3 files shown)


aaime:
Looks good to me

aaime:
Merging, let's see how the build goes.

bjornharrtell:
This caused a regression in GeoServer where WFS inserts returns fid null if the sequence is not owned by the primary key column which perhaps is not the usual case but allowed in PostgreSQL.

More info on the subject of sequence ownership can be found at https://www.postgresql.org/docs/9.5/static/sql-altersequence.html in the section about OWNED BY.

Gnafu:
@bjornharrtell , do you have an example we can setup to reproduce the issue?

bjornharrtell:
A sequence and table created with the following will exhibit the issue:
```sql
CREATE SEQUENCE test4_id_seq;
CREATE TABLE test4
(
  id integer NOT NULL DEFAULT nextval('test4_id_seq'::regclass),
  geom geometry(Point),
  CONSTRAINT test4_pkey PRIMARY KEY (id)
);
```

As opposed to the following:
```
CREATE TABLE test5
(
  id serial,
  geom geometry(Point),
  CONSTRAINT test5_pkey PRIMARY KEY (id)
);
```

The above tables will function exactly the same but the sequence generated by the serial type will be associated with id column. The only difference AFAIK is that `select pg_get_serial_sequence('test5', 'id')` will return `"public.test5_id_seq"` but `select pg_get_serial_sequence('test4', 'id')` will return `NULL` causing the mentioned regression in GeoServer because of the changes introduced in this PR. I'm not saying the changes are wrong, but if you are hit by the regression it might be a hard problem to track down. I think it would make sense to disallow inserts in this case and instead return a service exception explaining the problem.

Gnafu:
The scripts you posted are not equivalent, as described in the PostgreSQL documentation: https://www.postgresql.org/docs/9.1/static/datatype-numeric.html

The first script needs an additional
```
ALTER SEQUENCE test4_id_seq OWNED BY test4.id;
```
so the pg_get_serial_sequence() can return the correct value.

I suggest you to convert this discusion into an actuall issue so the developers can discuss about it.
I agree with you about the exception part.

PR #761, opened 2015-03-04
Fix bug GEOT-5036

guidograzioli:
The first commit provides a regression test which exploits the bug against current code; the second provides a fix (and updated test).

### DIFF
--- a/modules/unsupported/geojson/pom.xml
+++ b/modules/unsupported/geojson/pom.xml
@@ -83,6 +83,10 @@
             <version>${project.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+          <groupId>commons-lang</groupId>
+          <artifactId>commons-lang</artifactId>
+        </dependency>
     </dependencies>
 
     <profiles>
--- a/modules/unsupported/geojson/src/main/java/org/geotools/geojson/GeoJSONUtil.java
+++ b/modules/unsupported/geojson/src/main/java/org/geotools/geojson/GeoJSONUtil.java
@@ -30,13 +30,12 @@ import java.io.Reader;
 import java.io.StringReader;
 import java.io.Writer;
 import java.lang.reflect.Proxy;
-import java.text.DateFormat;
-import java.text.SimpleDateFormat;
 import java.util.Date;
 import java.util.List;
 import java.util.Map;
 import java.util.TimeZone;
 
+import org.apache.commons.lang.time.FastDateFormat;
 import org.geotools.util.Converters;
 import org.json.simple.JSONObject;
 import org.json.simple.parser.JSONParser;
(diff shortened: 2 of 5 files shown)


jdeolive:
Thanks for doing this so quickly! I would say that perhaps the use of a thread local is a little bit overkill in this case, instead I would just create the date format on demand. I say that because in GeoTools and GeoServer we've had issues in the past with thread locals leading to resource leaks. That isn't the case here but still I think the "unofficial policy" is to only use them if we absolutely need them.

guidograzioli:
No worries, we are hitting this bug in production so we need quite a fast workaround; I am aware of the risks of leaking by using ThreadLocals, and I try myself to stay away from them in most cases, but in this particular case I'd say they are actually a best solution. The problem of making new instances on demand is that it is slow (probably even slower than synchronizing, but I'd want to see benchmarks to strongly support this) at least for our scenario which leverages threading a lot.
If you agree I would as well trade some readability for speed, by using some overloaded methods instead of the several uses of instanceof in this same class.

jdeolive:
That sounds great. If there is a performance benefit to be had with this approach then I'm all for it, but like you said I think it would be best to start simple and introduce the thread local if a benchmark shows it to be significant. Thanks!

ischneider:
Agreed that constructing `SimpleDateFormat` objects on the fly will result in performance degradation - in my testing, encoding 10000 features 1000 times, I saw the average speed drop from 20-22ms to 31-33ms. There were also more outliers (GC pauses one can assume).

The other approach would be to move the `DateFormat` to the FeatureJSON class as an instance member. Slightly more work but best of both worlds.

jdeolive:
Ok, a few ways we can skin this one but how about:

1. Create GeoJSONUtil.createDateFormat() that creates a new instance
2. Update DateAttributeIO (and anything else that needs an instance) to create an instance of date format as a member
3. Update GeoJSONUtil.entry and either overload it with a version that takes Date + DateFormat, or 
3b. Make callers of the method convert Date to String before hand using there cached and thread safe date format

A bit more work but I don't like the idea of using a thread local to address what was a foolish mistake on my part thinking DateFormat was thread safe (i always confuse it with Pattern).

guidograzioli:
I see the urge of avoiding ThreadLocal; but switching to instance based SimpleDateFormat would be a problem in case of big numbers of features ( we deal with millions in certain scenarios) and the short-lived instances of FeatureJSON would still cause a lot of instancing and garbage collecting. 

I found that the parent pom already has commons-lang 2.6 in dependencyManagement;
so let me propose this alternate solution, which uses the FastDateFormat provided by 
commons-lang (which is a threadsafe singleton, order of magnitude faster) for formatting,
and a disposable SimpleDateFormat for the (single) need of parsing Dates.

simboss:
I like the FastDateFormat option as I share guido's concerns over the performance of large extractions.

jdeolive:
Cool, let's go with the FastDateFormat option. If you could clean up the formatting cruft from the patch it would be greatly appreciated. If you don't have the time I will try to do it when I get some time. Thanks a lot for this contribution!

jdeolive:
Another comment, just food for thought. I generally try to avoid adding utility library dependencies in self contained modules like this unless they are widely used. Another option would be to copy the class

jdeolive:
Opps, sorry, didn't mean to close this.

aaime:
About the dependency, I guess we can have it for the moment, and then switch to this class once we upgrade to java 8 as the min jdk? https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html
Oracle is going to drop support for jdk 7 soon, so I guess we'll have to do the switch to java 8 in the next year?

aaime:
So Guido actually removed the use of thread locals (and I did not notice). Seems good to merge, opinions?

ischneider:
None here.

jdeolive:
works for me.

aaime:
Merged. Sorry it took so long, too much stuff to do, not enough time.

guidograzioli:
No worries; thanks for applying the patch.

PR #762, opened 2015-03-04
GEOT-5039: Fixing gt-propery pom, using wrong gt version

dromagnoli:


bencaradocdavies:
Thanks, @dromagnoli. This bug was introduced in 89f806bb1e8ef03e4fea9630b9f5ac05d1fbca7d, and was one of two errors in https://github.com/geotools/geotools/pull/699, the other being unmerged content in the gt-jdbc-sqlserver pom.xml that broke the build (I fixed that one). Thanks for the fix! Kind regards, Ben.

PR #764, opened 2015-03-06
[GEOT-5041] ProjectionHandler fails with empty GeometryCollection

n-lagomarsini:
Please look at the following JIRA: https://jira.codehaus.org/browse/GEOT-5041.


aaime:
+1

PR #767, opened 2015-03-06
GEOT-5040 - Apply default timeouts to HTTPURIHandler

smithkm:
Fix for http://jira.codehaus.org/browse/GEOT-5040

Added unit tests.  The XSD module didn't have any mocking libraries so I pulled in EasyMock from the root POM.  I can drop the unit tests if this is a problem.

I noticed an oddity in the POM, while adding EasyMock , the JUnit dependancy is a different version from the root POM and has a different scope (compile rather than test)

### DIFF
--- a/modules/extension/xsd/xsd-core/pom.xml
+++ b/modules/extension/xsd/xsd-core/pom.xml
@@ -143,6 +143,16 @@
       <scope>compile</scope>
       <optional>true</optional>
     </dependency>
+    <dependency>
+      <groupId>org.easymock</groupId>
+      <artifactId>easymock</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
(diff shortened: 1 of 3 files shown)


aaime:
Looks good to me

smithkm:
One odd thing I noticed but left alone before and, which seems to be causing trouble with a backport.  The pom for xsd-core specifies a JUnit dependency with a different version (4.4) from that of the root pom (4.11), and as a compile scoped optional dependency rather than a required test scoped one.  JUnit Rules were added in 4.7.  I can replace the rule with try/fail/catch/assertThrown or just remove it as it's a trivial test but this dependency seems very odd.

bencaradocdavies:
@aaime, this breaks gt-xsd SchemaCacheOnlineTest. See:
http://jira.codehaus.org/browse/GEOT-5040

PR #769, opened 2015-03-11
[GEOT-5044] A less expensive WFSFeatureCollection.getCount()

gerson721:
This version uses a less expensive way of counting the number of features inside
the FeatureCollection document if numberOfFeatures attribute is not provided. This
also caches the bounds once a forced bounds calculation has been performed.

### DIFF
--- a/modules/unsupported/wfs/src/main/java/org/geotools/data/wfs/v1_1_0/WFSFeatureCollection.java
+++ b/modules/unsupported/wfs/src/main/java/org/geotools/data/wfs/v1_1_0/WFSFeatureCollection.java
@@ -86,7 +86,7 @@ class WFSFeatureCollection extends DataFeatureCollection {
      * Calculates and returns the aggregated bounds of the collection contents, potentially doing a
      * full scan.
      * <p>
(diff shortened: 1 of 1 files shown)


aaime:
Hi, I guess the fix is ok (did not check deeply), there is however a couple of issues. One is that I don't know if the exists tests are covering WFSFeatureCollection count, if there are not, at least should be added.... but this is probably not too hard.

The other issue is that the wfs module is going to be dropped soon, replaced by wfs-ng, so I was wondering if it would not be better to check if the new wfs client is affected by the same problem instead? 
https://github.com/geotools/geotools/tree/master/modules/unsupported/wfs-ng

gerson721:
Hi @aaime, I can verify that the WFSFeatureCollection.getCount() --> doForceFeatureCount() is covered by existing tests (eg . TinyOwsTest and AxisOrderTest) and the tests/assertions still pass. As for the wfs-ng, I haven't used that module before but after doing a quick browse through, I couldn't find any similar implementation which might suffer from the same issue as the one I'm trying to fix here.

aaime:
Merged

PR #771, opened 2015-03-11
GEOT-5040 - Apply default timeouts to HTTPURIHandler, Backport to 13.x

smithkm:
Backport of #767

This is last minute but this should be in 13.0 as it prevents an indefinite wait in a critical section over a utility class. (Schemas)


bencaradocdavies:
@aaime, this breaks gt-xsd SchemaCacheOnlineTest. See:
http://jira.codehaus.org/browse/GEOT-5040

PR #772, opened 2015-03-13
Changed the (E)CQL filter-to-string so that "!=" becomes "<>" in agreeme...

cne1x:
Change filter-to-string so that "!=" becomes "<>" in agreement with the spec.

https://jira.codehaus.org/browse/GEOT-5051

### DIFF
--- a/modules/library/cql/src/test/java/org/geotools/filter/text/cql2/FilterToCQLTest.java
+++ b/modules/library/cql/src/test/java/org/geotools/filter/text/cql2/FilterToCQLTest.java
@@ -83,6 +83,26 @@ public class FilterToCQLTest{
     }
 
     @Test
+    public void testNotEqualTo() throws Exception{
+        String cql = "ATTR1 <> 'foo'";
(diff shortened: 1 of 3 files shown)


modules/library/cql/src/test/java/org/geotools/filter/text/cql2/FilterToCQLTest.java
```diff
@@ -83,6 +83,26 @@ public void testLike() throws Exception {
...
+        
+        FilterToCQL toCQL = new FilterToCQL();
+        String output = filter.accept( toCQL, null ).toString();
+        Assert.assertEquals( output2, output );        
+    }
+    
+    @Test (expected=CQLException.class)
+    public void testFailNotEqualTo() throws Exception{
+    	
+    	cqlTest( "ATTR1 != 'foo'" );
```
jodygarnett: Thanks for the round trip test ...

jodygarnett:
Need to back port this to 13.x :)

PR #777, opened 2015-03-15
Tile Client

utaddei:
Hello GeoTools Devels,

please consider this contribution as proposed on the mailing list.

### DIFF
--- a/modules/unsupported/tile-client/src/test/java/resources/logging.properties
+++ b/modules/unsupported/tile-client/src/test/java/resources/logging.properties
@@ -1,0 +1,5 @@
+java.util.logging.ConsoleHandler.level = FINE
+java.util.logging.ConsoleHandler.formatter = java.util.logging.SimpleFormatter
+java.util.logging.ConsoleHandler.encoding = Cp850
+
+org.geotools.tile = FINE
--- a/modules/unsupported/tile-client/pom.xml
+++ b/modules/unsupported/tile-client/pom.xml
@@ -1,0 +1,168 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- ======================================================================= 
+  Maven Project Configuration File The Geotools Project http://www.geotools.org/ 
+  Version: $Id$ ======================================================================= -->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 
+                               http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.geotools</groupId>
+    <artifactId>geotools</artifactId>
+    <version>13-SNAPSHOT</version>
+  </parent>
+
+
+
+  <!-- =========================================================== -->
+  <!-- Properties and Profiles -->
+  <!-- explanation of their effect. -->
+  <!-- -->
+  <!-- Available profiles are: -->
+  <!-- -->
+  <!-- nameOfprofile Explantion of what the profile does -->
+  <!-- for the user. -->
+  <!-- -->
+  <!-- Example: mvn -PnameOfprofile install -->
+  <!-- -->
+  <!-- Note ... -->
+  <!-- =========================================================== -->
+  <!-- Default setting of properties -->
+  <!-- <properties> <nameOfproperty>valueOfsetting</nameOfproperty> </properties> -->
+
+  <!-- Profiles set on the command-line to override default properties. -->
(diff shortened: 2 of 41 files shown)


aaime:
Hi, sorry it took a while to give you feedback, we had FOSS4G-NA and the 13.0 release keeping us busy.
I had a quick look, the code is in some parts basic but looks ok to me, and it can be improved later (e.g., opening a url directly to read the tile is not exactly efficient nor testable, we have a pluggable HTTPClient interface in GeoTools that can be used to solve both issues).

The only issue that is bothering me a bit is that tests are not enabled when building with maven, even if most of them do not look like online tests, and some are failing. Again, for an unsupported module it's ok to have no tests, but since they are there and most of them are working, it would be good if at least those are enabled by default.

One more note, I remember udig having more tile providers, do you have a pointer to that code, I cannot find it right now (and yes, it's fine to have only the ones you're proposing, I was just wondering where the rest would be).

Finally, I guess it would be nice to get a review/ok from Jody, since the code was originally part of udig.

utaddei:
Hi Andreas,

thanks for the input. Yes, the code is basic, as it's the very nature of tiling systems: uncover the algorithm and you're off to building URLs. Further, I already had enough to handle with porting the code, but point me the directions and I'd be happy to use the pluggable HttpClient.

I don't think I have online tests. The problem with OSM is to avoid being blacklisted, and the one with Bings is providing a developer license. I didn't wat wo check my own license in to avoid misuse. But, there are tests! Quite a few of them, and indeed testing stuff (albeit perhaps not everything and not in a way you require). I can fix the ones failing.

There are more tile providers, yes. But I haven't ported those because I wanted to make sure I got the basics right and a code review, before I carried on. All the OSM tiles are now handled by a single provider familiy. I think missing are: JGrass TMS folder, Nasa WW and Tile WMS. The original code is somewhere on the net, on udig community github, and on my harddisk.

aaime:
I see some stuff here:
https://github.com/uDig-Community/udig-community/tree/master/jody/plugins/net.refractions.udig.community.jody.tile/src/org/geotools/tile

Funny, I though I did see more providers, as you are mentioning, WMTS for example

aaime:
Ciao Ugo, one more thing, did you sign the osgeo contribution agreement? See: http://docs.geotools.org/latest/developer/procedures/contribute.html

aaime:
Btw, found the contribution agreement. So we just need to have the tests actually run and work

aaime:
Hi Ugo, did you ever find time to fix the tests situation?

aaime:
Ugo, any news?

utaddei:
Hi Andrea,

Sorry mate, haven't yet. I will. Just had no time for it.

aaime:
Well, good to hear you're still around :-)
Do you think you'll get some time in the next month or so?

utaddei:
I am. Just that I code this on my spare time. Had other priorities.
Certainly. It's in my interest to get this work accepted. I'll contact you
again when the thing is fixed and also for a better http client.

aaime:
Thank you Ugo. If I may add a suggestion, do prioritize the tests, the better http client is also going to be nice, but say for example we have a TMS hierarchy on disk, we won't be able to use it there. Just saying it's not going to be a simple "replace URL usage with http client"

utaddei:
Hi,

Had a quick look and fixed two broken tests. Travis is saying tests are OK
now. Please have a look and let me know if I can help further.

You're obviously right about the http client. But a few tile providers are
http only. Original code includes a in-memory caching mechanism for the
tile, and another basic one I wrote for testing. On the udig repos you'll
find interfaces worth considering for extending the tile client work in
this direction.

jodygarnett:
I would like to review this before it is merged, would like to tag @smithkm in to look at how tiles are stored on disk.

utaddei:
Go for it, but bear in mind that tile caching/storing was not in scope of
this. The old code uses an ObjectCache, which I left in the original class
(which *does* need some re-write).

modules/unsupported/tile-client/src/test/java/resources/logging.properties
```diff
@@ -0,0 +1,5 @@
+java.util.logging.ConsoleHandler.level = FINE
+java.util.logging.ConsoleHandler.formatter = java.util.logging.SimpleFormatter
+java.util.logging.ConsoleHandler.encoding = Cp850
+
+org.geotools.tile = FINE
```
jodygarnett: Not sure if you intended to commit this, we generally leave logging turned down during builds.

jodygarnett:
merged to master, added -Ptile profile

modules/unsupported/tile-client/src/test/java/resources/logging.properties
utaddei: Nope. Added by accident. Feel free to remove.

PR #782, opened 2015-03-17
[GEOT-5050] Deprecate Views Management classes

n-lagomarsini:
Please look at the following JIRA: https://jira.codehaus.org/browse/GEOT-5050


jodygarnett:
This looks fine, I will make a note in the docs that these interfaces are being retired. 

Is there a replacement workflow that you recommend users migrate to?

n-lagomarsini:
Hi Jody,
since these classes/methods will be dropped down, no replacement workflow is present.

jodygarnett:
Merging to 13.x, you will probably want to cherry pick this for master?

n-lagomarsini:
Hi Jody,
thank you for merging the PR.

These classes/methods will be directly removed with the next PR (for master) related to the ViewsManagement removal so I don't think the cherry-pick should be needed.

PR #783, opened 2015-03-18
GEOT-5038: Allow warp-affine reduction in advanced projection handling

dromagnoli:


bencaradocdavies:
@aaime, @dromagnoli, @n-lagomarsini, this breaks the GeoServer build in gs-wcs1_1 GeoCoverageTest raster filter tests for OpenJDK 7 only:
https://jira.codehaus.org/browse/GEOS-6955

Reverting d3f0801da5fc1b8608f70a178f386c254380d7c1 on GeoTools master fixes the GeoServer master build OpenJDK 7. (I also reverted adaaf98eacf8da5a9a8bd48ac4775e1997f0096f for GEOT-5056 but only to avoid conlicts.)

PR #786, opened 2015-03-20
GEOT-5052: added cql expression support for ColorMapEntry attributes (co...

mbarto:
...lor, quantity, opacity)


aaime:
The changes look ok to me, in line with the dynamic symbolizer practices.
Only one observation, during the build test vms and threads are reused, to make sure you clearn the env map on teardown

mbarto:
Good advice. Cleanup code added.

PR #787, opened 2015-03-22
[Geot-5053]: Fixed FeatureTypeStyleImpl copying constructor does not copy some properties

wvaandrager:
Fix for Issue [Geot-5053]((https://jira.codehaus.org/browse/GEOT-5053)).


mbarto:
Hi, thank you for the PR. The issue has already been solved by https://github.com/geotools/geotools/pull/788

aaime:
Hey I've checked, this is a different solution to the same ticket, but in fact the patch is valid, the copy constructor is still not copying everything, Mauro's patch acts in the visitor instead. So I'd keep it anyways.
Two missing bits:
* Since you modified more than one file, we need a signed contribution agreement to OSGeo, see http://docs.geotools.org/latest/developer/procedures/contribute.html
* You should squash all commits into one

wvaandrager:
Thanks for checking the commit @aaime! I submitted the contribution license and squashed the three commits into one. I hope it is in order now.

aaime:
Merged. I see you broke an interface (by adding a method), so the change cannot be backported to the "stable" series. Well.. unless the project management committee decides otherwise, you may want to ask on geotools-devel

PR #792, opened 2015-03-24
Added support for Nvarchar filtering in SQL Server [GEOT-5587]

mwronowski:
See - https://osgeo-org.atlassian.net/browse/GEOT-5587

Change made because there was no option to filter SQL Server columns that contains special/national characters as a query generated by this SQL Server jdbc plugin was not using nvarchar type comparison with N' prefix. I think it is necesary to have nvarchar type of comparison because when using geotools in geoserver there is need to create complicated shapes, layers with different type of data and characters. Without this change GeoServer is useless for national data sets

### DIFF
--- a/modules/plugin/jdbc/jdbc-sqlserver/src/main/java/org/geotools/data/sqlserver/SQLServerFilterToSQL.java
+++ b/modules/plugin/jdbc/jdbc-sqlserver/src/main/java/org/geotools/data/sqlserver/SQLServerFilterToSQL.java
@@ -17,12 +17,15 @@
 package org.geotools.data.sqlserver;
 
 import java.io.IOException;
+import java.io.StringWriter;
 import java.text.SimpleDateFormat;
 import java.util.Date;
 
 import org.geotools.data.jdbc.FilterToSQL;
+import org.geotools.data.jdbc.FilterToSQLException;
 import org.geotools.filter.FilterCapabilities;
 import org.geotools.jdbc.SQLDialect;
+import org.opengis.filter.Filter;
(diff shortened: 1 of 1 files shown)


aaime:
What happens if the column in question is not a nvarchar?
Plus, for each fix, we need a test, you can extend the existing ones, if you need help getting them to run ask on the geotools-devel mailing list.

mwronowski:
Hi Andrea,

Thanks for a quick reply. Nothing should go wrong and it is even safer to
use N prefix when filtering in SQL Server. N prefix means that all
characters will be converted to Unicode.
Please se article
http://stackoverflow.com/questions/7703192/sql-server-nvarchar-n-prefix-for-varchar-fields
It is important for unicode characters. I am not sure if I am able to
produce some tests as I am .Net developer but I will try

Marek

2015-03-24 15:03 GMT+01:00 Andrea Aime <notifications@github.com>:

aaime:
Marek, it's ok if you cannot create the test, but we'll have to wait for someone that does add one that fails without the change, and succeeds with your change, before merging this pull request.

jodygarnett:
@mwronowski I am afraid this pull request is not being merged in time for the upcoming 14-beta. Not a lot of volunteers have access to SQL Server to test against.

@ianturton you may be able to review this one? Not sure ...

ianturton:
I'm afraid I'm completely linux based now and no longer have access to an SQLServer instance to test against.

bencaradocdavies:
Is this PR still active or does it need need to updated?

jodygarnett:
We may need to create a bug report, link to this ticket and close the pull request.

PR #794, opened 2015-03-27
Updated pom for new MS SQL Server jar

tbarsballe:
Added documentation and sqlserver profile activation parameter (based on Oracle plugin), and updated driver version to 4.1 (from 4.0).


tbarsballe:
At this point, this is primarily intended to support running online tests (Although the version update is probably a good idea anyway)

PR #796, opened 2015-03-30
[GEOT-5063] Crop operation fails to render ROI correctly in openJDK7

n-lagomarsini:
Please look at the following JIRA: https://jira.codehaus.org/browse/GEOT-5063


bencaradocdavies:
@n-lagomarsini thanks very much. I can confirm that this fixes [GEOS-6955] Build failure in gs-wcs1_1 GeoCoverageTest raster filter tests for OpenJDK 7 only

This fix should also be backported to 13.x and 12.x in a few days.

I will update GEOT-6955 on the new Atlassian JIra. Other watchers: @aaime @dromagnoli

PR #800, opened 2015-03-31
Fix test failure on Windows

travislbrundage:
I don't know if my way of checking the OS is the correct way to go about it, but I believe this fixes it. We only care to change the test when the OS is Windows, in which case we need \r\n instead of just \n.

### DIFF
--- a/modules/unsupported/csv/src/test/java/org/geotools/data/csv/CSVWriteTest.java
+++ b/modules/unsupported/csv/src/test/java/org/geotools/data/csv/CSVWriteTest.java
@@ -250,7 +250,9 @@ public class CSVWriteTest {
         SimpleFeatureStore featureStore = (SimpleFeatureStore) store.getFeatureSource("locations");
         assertEquals("featureStore should only have the one feature we created", 
                 1, featureStore.getFeatures().size());
-        String contents = "LAT,LON,CITY,NUMBER,YEAR\nPOINT (45.52 -122.681944),Portland,800,2014,";
+        final String newline = System.lineSeparator();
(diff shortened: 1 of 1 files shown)


modules/unsupported/csv/src/test/java/org/geotools/data/csv/CSVWriteTest.java
```diff
@@ -250,11 +250,20 @@ public void replaceAll() throws Exception {
         SimpleFeatureStore featureStore = (SimpleFeatureStore) store.getFeatureSource("locations");
         assertEquals("featureStore should only have the one feature we created", 
                 1, featureStore.getFeatures().size());
-        String contents = "LAT,LON,CITY,NUMBER,YEAR\nPOINT (45.52 -122.681944),Portland,800,2014,";
+        String contents;
+        if (isWindows() == true) {
+        	contents = "LAT,LON,CITY,NUMBER,YEAR\r\nPOINT (45.52 -122.681944),Portland,800,2014,";
+        } else {
+        	contents = "LAT,LON,CITY,NUMBER,YEAR\nPOINT (45.52 -122.681944),Portland,800,2014,";
+        }
```
jodygarnett: Use the following rather than isWindows()

````
String contents = "LAT,LON,CITY,NUMBER,YEAR"+System.getProperty("line.separator")"+"POINT (45.52 -122.681944),Portland,800,2014,";
```

travislbrundage: Didn't realize that existed - thanks for the feedback.

PR #801, opened 2015-03-31
Delimit CQL idenifiers that are reserved words or contain reserved characters

smithkm:
Fixes issues when properties have names like `point` or `hyphens-are-cool`.

Should produce correct output for names containing spaces or double quotes although the parser doesn't handle these going the other way. (see the ignored unit tests)

### DIFF
--- a/modules/library/cql/src/main/java/org/geotools/filter/text/commons/ExpressionToText.java
+++ b/modules/library/cql/src/main/java/org/geotools/filter/text/commons/ExpressionToText.java
@@ -18,13 +18,19 @@ package org.geotools.filter.text.commons;
 
 import java.text.DateFormat;
 import java.text.SimpleDateFormat;
+import java.util.Arrays;
 import java.util.Calendar;
+import java.util.Collections;
 import java.util.Date;
 import java.util.GregorianCalendar;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Set;
 import java.util.SimpleTimeZone;
 import java.util.TimeZone;
+import java.util.TreeMap;
+import java.util.TreeSet;
+import java.util.regex.Pattern;
 
 import org.opengis.filter.expression.Add;
 import org.opengis.filter.expression.Divide;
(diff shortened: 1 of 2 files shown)


modules/library/cql/src/test/java/org/geotools/filter/text/cql2/FilterToCQLTest.java
```diff
@@ -283,6 +286,49 @@ public void testContains() throws Exception {
...
+    }
+    
+    @Test
+    public void testAttributeContainsOperator() throws Exception {
+        cqlTest("INTERSECTS(\"the-geom\", POINT (1 2))");
+    }
+    
+    @Test
+    public void testAttributeContainsComma() throws Exception {
+        cqlTest("INTERSECTS(\"the,geom\", POINT (1 2))");
```
jodygarnett: For whatever reason I usually delimiters with single quotes (probably just so the java string representation reads better.

```
cqlTest("INTERSECTS('the,geom', POINT (1 2))");
```

I am sure this is just what I am used to.

modules/library/cql/src/main/java/org/geotools/filter/text/commons/ExpressionToText.java
```diff
@@ -214,11 +220,45 @@ public Object visit(Multiply expression, Object extraData) {
 	@Override
 	public Object visit(PropertyName expression, Object extraData) {
 
-        StringBuilder output = asStringBuilder(extraData);        
-        output.append( expression.getPropertyName() );
+        StringBuilder output = asStringBuilder(extraData);
+        if(propertyNeedsDelimiters(expression)) {
+            output.append('"');
+            output.append( expression.getPropertyName().replace("\"", "\"\"") );
+            output.append('"');
```
jodygarnett: Looks good, I am used to delimiting with single quotes:

```
+            output.append('\'');
+            output.append( expression.getPropertyName().replace("\"", "\"\"") );
+            output.append('\'');
```

It makes java code examples easier to read:

```
CQL.filter("INTERSECTS('the-geom', POINT (1 2))");
```

I guess I am confused about how to tell property name reference from a string? How does this work with functions:

```
concatenate( "the_geom","the geom")
```

modules/library/cql/src/test/java/org/geotools/filter/text/cql2/FilterToCQLTest.java
smithkm: Single quotes are for strings, double are for delimiting identifiers.

modules/library/cql/src/main/java/org/geotools/filter/text/commons/ExpressionToText.java
aaime: I can confirm what Kevin suggests, if you use single quotes that should be treated as a string, so if what you suggested works, that might be considered another bug (or an excess of leniency)

jodygarnett:
I stand corrected, sure hope I do not have to rewrite documentation examples now :(

smithkm:
https://osgeo-org.atlassian.net/browse/GEOT-5066

PR #805, opened 2015-04-02
JDBCDataStore custom tables

ebocher:
Dear geotools community,

A new method has been added to the SQLDialec class in order to custom the filtered tables.
This method is used in JDBCDataStore.
Each geotools JDBC dialects can now propose its own list of tables (PostGIS, MySQL and so on). This change is useful for example to support with the H2 database the linked table type.

We use it to connect H2GIS with geoserver and for example display a linked shapefile in geoserver (see http://www.h2gis.org/docs/dev/FILE_TABLE/).

We are working here (https://github.com/ebocher/geoserver-h2gis)  on an extension to do that. Note it's still under active development and uses a GPL 3 license due to the H2GIS compatibility.


aaime:
The pull request needs to be redone, it contains several changes that are unrelated (I guess you really wanted to only push the changes in the jdbc module).
Also check the contribution guide for copyright assignment modules and other procedures (e.g., adding a test for each change): http://docs.geotools.org/latest/developer/procedures/contribute.html

ebocher:
Because this PR introduces minor change and has no effect on unit tests I will send a mail on the developers list to explain my need. Regarding the contribution guide, it seems the best method. 
Sorry for the pollution.

PR #808, opened 2015-04-05
GEOT-5047: Make sure we have stable date/time literal conversion to sql.

ianturton:
First attempt to fix GEOT-5047 Make sure we have stable date/time literal conversion to sql.

I now know far too much about how broken Java.util.Date and Java.sql.Date are :-( it seems that the only place we can fix this is at the database level. This PR currently improves the tests to try out a variety of timezones across the world (so that we see time zone related failures) and applies a fix to the H2 db by adding an explicit GMT/UTC timezone to the dateformat statement generated for date filters in the SQL dialect.

### DIFF
--- a/modules/plugin/jdbc/jdbc-h2/src/main/java/org/geotools/data/h2/H2FilterToSQL.java
+++ b/modules/plugin/jdbc/jdbc-h2/src/main/java/org/geotools/data/h2/H2FilterToSQL.java
@@ -19,6 +19,7 @@ package org.geotools.data.h2;
 import java.io.IOException;
 import java.text.SimpleDateFormat;
 import java.util.Date;
+import java.util.Locale;
 import java.util.TimeZone;
 
 import org.geotools.data.jdbc.FilterToSQL;
@@ -217,7 +218,7 @@ public class H2FilterToSQL extends FilterToSQL {
         if (literal instanceof Date) {
             out.write("PARSEDATETIME(");
             if (literal instanceof java.sql.Date) {
-                out.write("'" + DATE_FORMAT.format(literal) + "', 'yyyy-MM-dd'");
+                out.write("'" + DATE_FORMAT.format(literal) + "', 'yyyy-MM-dd', '"+Locale.getDefault()+"', 'GMT'");
             }
             else {
                 out.write("'" + DATETIME_FORMAT.format(literal) + "', 'yyyy-MM-dd HH:mm:ss.SSSZ'");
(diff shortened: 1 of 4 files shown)


modules/plugin/jdbc/jdbc-h2/src/main/java/org/geotools/data/h2/H2FilterToSQL.java
```diff
@@ -217,7 +218,7 @@ protected void writeLiteral(Object literal) throws IOException {
         if (literal instanceof Date) {
             out.write("PARSEDATETIME(");
             if (literal instanceof java.sql.Date) {
-                out.write("'" + DATE_FORMAT.format(literal) + "', 'yyyy-MM-dd'");
+                out.write("'" + DATE_FORMAT.format(literal) + "', 'yyyy-MM-dd', '"+Locale.getDefault()+"', 'GMT'");
```
jodygarnett: Does the PARSEDATETIME function record in GMT or in the current Locale? Trying to see if the databases are portable or are specific to a timezone

aaime:
HI Ian, I've already committed a patch to try and solve this issue, https://github.com/geotools/geotools/commit/f65d9d563eddb98bcdbbef8781227ba568ce35c7

Could you describe what still does not work, in particular in the H2 case?

modules/plugin/jdbc/jdbc-h2/src/main/java/org/geotools/data/h2/H2FilterToSQL.java
ianturton: Parsedatetime was using the JVM timezone by default - this change forces it to use GMT (which is what we claim to store.

ianturton:
If you look at the new tests you'll see that they fail for any timezone before GMT - previously it depended on which timezone the tests were run in, see https://github.com/geotools/geotools/pull/808/files#diff-23365cff76cbac15102642b8dea1698eR41

modules/plugin/jdbc/jdbc-h2/src/main/java/org/geotools/data/h2/H2FilterToSQL.java
ianturton: Reading that more carefully - I'm not sure what the locale does there, in my local it just writes the date out in GMT, because sqlDates have a timezone (sigh).

ianturton: seems the locale is passed to SimpleDateFormat and controls locale for which the date format symbols should be found..

aaime:
Ian, I am in CEST and the test passes for me?
Also, we never claimed to handle data in GMT only, the WFS might do that conversion (and it might be an implementation accident), but this does not say anything about what GeoTools should do

aaime:
Ian, wondering, is this one superceded by your latest work? If so, could you close it?

aaime:
@ianturton , has this been superceded by your latest work?

PR #809, opened 2015-04-07
Add a method to custom the desired tables

ebocher:
This PR shall cancel and replace the #805.

The code contribution licence file has been send by e-mail.


ebocher:
Is there any chance you could accept this pull request ?
Thanks

aaime:
Looking at it, I see a behavior regression, the old code was making sure we were asking for supported table types, the new one does not anymore. Please fix.

ebocher:
I don't understand the comment. Do you have a unit test , a fail ?
Did I miss something but the code support the same list of table types ?

aaime:
I did bother running the tests, the regression is evident. The old code starts with a list of candidate values, then checks against the values supported by the database, and performs an intersection between the two, and then runs the query with the subset that is both desired and supported, now the code just runs over the list of desired tables, without checking if they are supported.
JDBCDataStore is a framework, not a single database implementation, the framework must work for all databases and continue to work as people upgrade the dbms and the drivers, so the checks against what table types are really supported must stay there.

aaime:
Checking the history, the current code was added by @jodygarnett while working on "[GEOT-4486] JDBCDataStore allow synonyms". The reason is evident, not all databases have a notion of synonym, and the getTables javadoc is also pretty clear: "types - a list of table types, which must be from the list of table types returned from getTableTypes(),to include; null returns all types"

aaime:
Yep, that works better. I assume the H2 driver reports the new file based table type, right?

ebocher:
Yes. There is an update here https://github.com/ebocher/geoserver-h2gis/blob/H2gis_linkedTable/src/main/java/org/orbisgis/geoserver/h2gis/datastore/H2GISDialect.java.

```java
@Override
    public String[] getDesiredTablesType() {
        return new String[]{"TABLE", "VIEW", "MATERIALIZED VIEW", "SYNONYM", "TABLE LINK", "EXTERNAL"};
    }
```

But the geotools - h2 driver must updated also with the last H2 database jar.

aaime:
The update of H2 is difficult, see discussion on the mailing list. We basically have to upgrade geotools, gwc and geoserver in a single shot, preserving backwards compatibility (e.g., a data directory using the old h2 files must keep on working)

ebocher:
Ok I understand.

ebocher:
Before merging I'll add a test in H2GIS-GS to be sure it works well.

ebocher:
Doesn't work with H2GIS-GS. I need to investigate.

ebocher:
Finally this PR is OK.  It requires a small  fix on H2 database to support its new  external file table. I sent a PR to solve this second issue : https://github.com/h2database/h2database/pull/2.

aaime:
Could you ping me back when the pull request against H2 is merged?
I'll follow up with this one (just don't want to merge the current solution, if the H2 one does not get merged, you might have to rework the GeoTools side in a different way).

ebocher:
Dear Andreas,

Just to inform you that the H2 patch has been integrated. See https://github.com/h2database/h2database/commit/b5b920171a2fd76434731c0f86ea27c1a2ca3db8.

Best regards

PR #813, opened 2015-04-09
Fixed outstanding test failures for the vecmath transition.

jnh5y:
* Updated a constructor to call setIdentity.
* Tweaked copySubMatrix.
* Fixed sub.

Additionally, changed an expected exception in org/geotools/referencing/operation/transform/ProjectiveTransform.java.

### DIFF
--- a/modules/library/referencing/src/main/java/org/geotools/referencing/operation/matrix/GeneralMatrix.java
+++ b/modules/library/referencing/src/main/java/org/geotools/referencing/operation/matrix/GeneralMatrix.java
@@ -68,6 +68,7 @@ public class GeneralMatrix implements XMatrix {
      */
     public GeneralMatrix(final int size) {
         mat = new DenseMatrix64F(size, size);
+        setIdentity();
     }
 
     /**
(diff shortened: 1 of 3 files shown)


aaime:
I guess this needs a review from Jody?

jnh5y:
@aaime I'm happy to take comments and suggestions from anyone.  @jodygarnett had merged my PRs into the matrix branch during the code sprint, but I imagine this work will need general approval to get merged.

aaime:
@jnh5y, the thing is, I don't know squat about the classes you're changing so... it would take me some time (I guess around a day) to first get a grip on what we are doing with vecmath and the behavior of the classes involved. But I don't have that much time handy (not in a single shot, it's probably two dedicated weekends worth of work, given that I'm left with more or less 4 hours per we).

jnh5y:
@aaime That's fine, no worries.  I think merging the matrix branch in is likely a low priority.  When I was looking for general approval, I actually meant for switching vecmath to the EJML.  

The most controversial thing in this PR is TODOs in place of some of the more pedantic test cases one could imagine.  If that's a problem, I can remove them or try and find some time to implement a few more tests.  

Anyhow, from the mailing list, Jody indicated that the switch is unlikely to be backported to the 13.x series.  In that case, based on the previous year's release schedule, it looks like this entire branch should get wrapped up before July 18th or so for a 14-beta.

If we could get your help at any point, I think it may be useful to get some suggestions for specific downstream tests or verification steps in GeoServer to make sure that there are no bugs or speed losses.

aaime:
That's why it's best to get this merged asap, low level library changes end up affecting lots of code, they are not suitable for last minute changes.

jodygarnett:
Merging now, I am going to try and pick up Tylers work (of removing the JAI dependency now that we have Matrix working).

If it goes well we can co-ordinate during todays meeting.

PR #814, opened 2015-04-10
added Timezones to JDBC date tests

ianturton:
By explicitly setting the timezone of the JVM when running the tests we can check that the date doesn't change when the user is East of Greenwich, i.e. their TimeZone is a GMT+X one.

This test will/should fail for CEST and beyond.


aaime:
More tests are good... but these break the build, I guess they should come with a solution too.
(I completely understand you're just making an existing situation evident, not introducing the issue yourself)

jodygarnett:
With #808 merged in doe these tests now work @ianturton / @aaime ?

PR #815, opened 2015-04-11
Add Geobuf DataStore

jericks:
Pull request for issue #GEOT-5070.

Discussion here: http://osgeo-org.1560.x6.nabble.com/Geobuf-DataStore-td5200324.html

Thanks!
Jared


modules/unsupported/geobuf/pom.xml
```diff
@@ -0,0 +1,72 @@
...
+    </dependencies>
+    <build>
+        <plugins>
+            <!--
+            Commented out to make building easier (no dependency on protoc executable.
+            To regenerate the org.geotools.data.geobuf.Geobuf Java file, place the new geobuf.proto
+            in src/main/protobuf and add package org.geotools.data.geobuf; to the first line. Then
+            uncomment the following plugin and run mvn clean install.  This will create a new Geobuf
+            Java file in target/generated-sources.
+            -->
```
jodygarnett: Rather than this manual procedure consider using a profile.

Or you can do what we do for oracle and use -D property to activate one profile when true, and another when false.

modules/unsupported/geobuf/pom.xml
```diff
@@ -0,0 +1,72 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.geotools</groupId>
+        <artifactId>unsupported</artifactId>
+        <version>14-SNAPSHOT</version>
```
jodygarnett: Add a profile that will include this module to unsupported/pom.xml (there are plenty of other examples there). This makes it easier for people to build geotools *with* your module, rather than having to cd into this directory.

modules/unsupported/geobuf/src/main/java/org/geotools/data/geobuf/GeobufDirectoryDataStore.java
```diff
@@ -0,0 +1,78 @@
...
+import org.geotools.data.store.ContentFeatureSource;
+import org.geotools.feature.NameImpl;
+import org.opengis.feature.simple.SimpleFeatureType;
+import org.opengis.feature.type.Name;
+
+import java.io.*;
+import java.util.ArrayList;
+import java.util.List;
+
+public class GeobufDirectoryDataStore extends ContentDataStore {
```
aaime: Not a problem and no need to rework this, but I'd suggest you also have a look at ShapefileDirectoryFactory and DirectoryDataStore.

PR #817, opened 2015-04-14
AddConst is calling the wrong super constructor

jericks:
See http://osgeo-org.1560.x6.nabble.com/SubtractFromConst-and-AddConst-errors-td5200933.html for discussion.  This solves issue #5073.

Thanks!
Jared


n-lagomarsini:
Thank you @jericks for your fix.

PR #818, opened 2015-04-15
Add Gnomonic projection and new codes in AUTO namespace for Gnomonic and...

SimonSchafer:
... Stereographic projections

The Gnomonic projection is new, the Stereographic is existing code however I need to call it from a WMS so have also added a AUTO namespace code.
The AUTO namespace codes are 97001 and 97002 after discussions with Andrea Aime.
These new codes are not part of any WMS standard.  They are documented in the source code however I do not know if they should be documented at some higher level.


modules/library/referencing/src/main/java/org/geotools/referencing/factory/wms/Auto97001.java
```diff
@@ -0,0 +1,99 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ * 
+ *    (C) 2004-2008, Open Source Geospatial Foundation (OSGeo)
```
aaime: The file has been added in 2015, so this line should be:
``(C) 2015, Open Source Geospatial Foundation (OSGeo)``

modules/library/referencing/src/main/java/org/geotools/referencing/factory/wms/Auto97002.java
```diff
@@ -0,0 +1,108 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ * 
+ *    (C) 2004-2008, Open Source Geospatial Foundation (OSGeo)
```
aaime: The file has been added in 2015, so this line should be:
``(C) 2015, Open Source Geospatial Foundation (OSGeo)``

aaime:
Hi, besides the two minor issues found, everything else looks good.
Am I correct in assuming you want this for GeoServer WMS? If so, it would be nice, but not a requirement, to also get a pull request documenting how the two new AUTO projections are supposed to be used (since one cannot refer people to the WMS spec for those two).

SimonSchafer:
Thanks for that.

Yes it is for GeoServer WMS.  I was wondering how it should be documented.
Can you by any chance point me towards where the documentation ought to go?

Simon

aaime:
Hi Simon, I'd add it around here, in its own new page I guess: http://docs.geoserver.org/stable/en/user/services/wms/index.html
In the source tree, that would be here: https://github.com/geoserver/geoserver/tree/master/doc/en/user/source/services/wms

SimonSchafer:
New page added. Pull request

Add documentation for non standard auto CRS namespace in WMS #1063

PR #825, opened 2015-04-21
[GEOT-5082] Speed up simple feature GML encoding with XSD-GML

aaime:


modules/extension/xsd/xsd-gml2/src/main/java/org/geotools/gml2/bindings/GML2EncodingUtils.java
```diff
@@ -266,5 +269,42 @@ public static List GeometryPropertyType_getProperties(Geometry geometry) {
...
+        if (e == null || e.isNull() || e.getCoordinateReferenceSystem() == null) {
+            return null;
+        }
+
+        // check if srsDimension is turned off
+        if (configuration.hasProperty(GMLConfiguration.NO_SRS_DIMENSION)) {
+            return null;
+        }
+
+        return e.getCoordinateReferenceSystem().getCoordinateSystem().getDimension();
```
jodygarnett: This appears to be common logic can we take it into a CRS utility method?

modules/extension/xsd/xsd-gml2/src/main/java/org/geotools/gml2/bindings/GMLEncodingUtils.java
```diff
@@ -260,31 +276,39 @@ public XSDTypeDefinition createXmlTypeFromFeatureType(SimpleFeatureType featureT
             if ( attribute instanceof GeometryDescriptor ) {
                 Class binding = attribute.getType().getBinding();
                 if ( Point.class.isAssignableFrom( binding ) ) {
```
jodygarnett: Is it possible to use Java 7 support for String switch statements here combined with Geometry.getGeometryType() or similar?

modules/extension/xsd/xsd-gml2/src/main/java/org/geotools/gml2/bindings/GMLEncodingUtils.java
```diff
@@ -260,31 +276,39 @@ public XSDTypeDefinition createXmlTypeFromFeatureType(SimpleFeatureType featureT
...
                 }
                 else if ( MultiPolygon.class.isAssignableFrom( binding) ) {
-                    element.setTypeDefinition( schemaIndex.getTypeDefinition(gml.qName("MultiPolygonPropertyType")));
+                    element.setTypeDefinition(schemaIndex.getTypeDefinition(gml
+                            .qName("MultiPolygonPropertyType")));
                 }
                 else {
-                    element.setTypeDefinition( schemaIndex.getTypeDefinition(gml.qName("GeometryPropertyType")));
+                    element.setTypeDefinition(schemaIndex.getTypeDefinition(gml
+                            .qName("GeometryPropertyType")));
```
jodygarnett: I see no support for curves, is that because this is GML2 encoder?

aaime: Indeed, curves are supported only starting GML3. And besides that, the way they have been implemented makes them mingle with linear geometries rather seamlessly.

modules/extension/xsd/xsd-gml2/src/main/java/org/geotools/gml2/bindings/GMLEncodingUtils.java
aaime: No, we need inheritance checks here, not simple equality (otherwise I would have used it instead of isAssignableFrom).

modules/extension/xsd/xsd-gml2/src/main/java/org/geotools/gml2/bindings/GML2EncodingUtils.java
aaime: CRS does not know about ReferencedEnvelope (but we could use opengis Envelope), more importantly, does not know about Configuration (a gt-xsd class)

modules/extension/xsd/xsd-gml2/src/main/java/org/geotools/gml2/simple/FeatureCollectionEncoderDelegate.java
```diff
@@ -0,0 +1,444 @@
...
+ *    This library is free software; you can redistribute it and/or
+ *    modify it under the terms of the GNU Lesser General Public
+ *    License as published by the Free Software Foundation;
+ *    version 2.1 of the License.
+ *
+ *    This library is distributed in the hope that it will be useful,
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ *    Lesser General Public License for more details.
+ */package org.geotools.gml2.simple;
```
jodygarnett: usually package starts on its own line

modules/library/xml/src/main/java/org/geotools/GML.java
```diff
@@ -296,7 +310,9 @@ public void encode(OutputStream out, SimpleFeatureCollection collection) throws
             }
         }
         if (version == Version.WFS1_0) {
-            Encoder e = new Encoder(new org.geotools.wfs.v1_0.WFSConfiguration());
+            org.geotools.wfs.v1_0.WFSConfiguration configuration = new org.geotools.wfs.v1_0.WFSConfiguration();
+            configuration.getProperties().add(GMLConfiguration.OPTIMIZED_ENCODING);
```
jodygarnett: Thanks for not forgetting the GML class :)

jodygarnett:
Finished reviewing, interested to see how it performs. Merge when ready

modules/library/xml/src/main/java/org/geotools/GML.java
aaime: As your class says in another point, cannot encode a GML2 collection (there is no such a thing)

aaime:
The ticket is mentioned in the commit message, here is a direct link: https://osgeo-org.atlassian.net/browse/GEOT-5082

Checked the docs, they do use the GML class which has been updated, so nothing to add there, they should get the speed up transparently.

modules/extension/xsd/xsd-gml2/src/main/java/org/geotools/gml2/simple/FeatureCollectionEncoderDelegate.java
aaime: Fixed

PR #826, opened 2015-04-21
A slight update to the JAI Logging fix.

jnh5y:
This relates to GEOT-5016 (https://osgeo-org.atlassian.net/browse/GEOT-5016).


jodygarnett:
squashed into a single commit, does not look like we can back port to 13.x

PR #828, opened 2015-04-24
GEOT-5085 Workaround for slow performance of SDE Table Stats on Oracle

dr-jts:
On SDE on Oracle (and maybe other platforms) WFS queries against layers containing large numbers of features is very slow. This is because WFS requires computing the size of the query result before the query is executed to retrieve data. The SDE API method SEQuery.calculateTableStatistics currently used to do this is very slow (on Oracle at least).
A workaround is to run the query twice, once to count the features and once to fetch the data.
Note that due to an undocumented quirk of the SDE API, it is also necessary to call SEQuery.prepareQueryInfo before SEQuery.setSpatialConstraints. Otherwise, the spatial filter will not apply.

See https://osgeo-org.atlassian.net/browse/GEOT-5085

### DIFF
--- a/modules/plugin/arcsde/datastore/src/main/java/org/geotools/arcsde/data/ArcSDEQuery.java
+++ b/modules/plugin/arcsde/datastore/src/main/java/org/geotools/arcsde/data/ArcSDEQuery.java
@@ -591,69 +591,77 @@ class ArcSDEQuery {
             colNames.add(geomCol);
         }
         
-        final SeQueryInfo qInfo = filters.getQueryInfo(colNames.toArray(new String[colNames.size()]));
-
+        final SeQueryInfo queryInfo = filters.getQueryInfo(colNames.toArray(new String[colNames.size()]));
         final SeFilter[] spatialFilters = filters.getSpatialFilters();
+        final boolean hasSpatialFilter = spatialFilters != null && spatialFilters.length > 0;
 
-        final Command<Integer> countCmd = new Command<Integer>() {
-            @Override
-            public Integer execute(ISession session, SeConnection connection) throws SeException,
-                    IOException {
+    final Command<Integer> countCmd = new Command<Integer>() {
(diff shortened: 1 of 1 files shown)


groldan:
Patch looks good to me

modules/plugin/arcsde/datastore/src/main/java/org/geotools/arcsde/data/ArcSDEQuery.java
```diff
@@ -591,69 +591,77 @@ public int calculateResultCount() throws IOException, UnsupportedOperationExcept
...
+                versioningHandler.setUpStream(session, query);
+
+                if (hasSpatialFilter) {
+                    /*
+                     * Despite the FeatureSource.getCount() contract saying it's
+                     * ok to return -1 if count is too expensive to calculate,
+                     * the GeoServer codebase is plagued of
+                     * FeatureCollection.size() calls depending on actual result
+                     * counts or some operations don't work at all. return -1;
+                     */
```
aaime: Mind, there is no contract breackage here. FeatureSource.getCount() can return -1, featureCollection.size() cannot, it must scroll over the features in case a fast count cannot be performed. 
So, if you experienced problems, there must be some feature collection that is not abiding to its contract.
I'd thus remove this comment.

dr-jts: That comment was in the previous code.  I didn't try and see if the count
could be returned as -1 (which would improve performance even more - but
sounded like it might break things in ways which would not show up during
brief testing).  So I just left the comment in there.

It's easy to remove the comment.  I might even be able to try just
returning -1.  But I'm worried that might work for me, but leave a time
bomb ticking...

aaime:
Ok then, merged :)

dr-jts:
Confirmed that GeoServer master now includes this patch and has good performance for WFS against SDE Oracle.

PR #829, opened 2015-04-29
Change from vecmath to ejml [GEOT-22]

jodygarnett:
For details see https://github.com/geotools/geotools/wiki/Change-from-vecmath-to-EJML

### DIFF
--- a/modules/unsupported/wfs-ng/src/test/java/org/geotools/data/wfs/integration/v1_1/GeoServerIntegrationTest.java
+++ b/modules/unsupported/wfs-ng/src/test/java/org/geotools/data/wfs/integration/v1_1/GeoServerIntegrationTest.java
@@ -24,7 +24,6 @@ import java.io.IOException;
 import org.geotools.data.DataStore;
 import org.geotools.data.DataUtilities;
 import org.geotools.data.ResourceInfo;
-import org.geotools.data.simple.SimpleFeatureSource;
 import org.geotools.data.simple.SimpleFeatureStore;
 import org.geotools.data.wfs.WFSDataStore;
 import org.geotools.data.wfs.integration.AbstractIntegrationTest;
@@ -113,7 +112,7 @@ public class GeoServerIntegrationTest extends AbstractIntegrationTest {
         assertTrue(info.getKeywords().contains("sfRoads roads"));
         assertEquals("roads_Type", info.getTitle());
         assertEquals("http://www.openplans.org/spearfish", info.getSchema().toString());
-        assertEquals(589242.03, info.getBounds().getMinX(),0.01);
+        assertEquals(589275.24, info.getBounds().getMinX(), 0.01);
         assertEquals("EPSG:NAD27 / UTM zone 13N", info.getCRS().getName().toString());
     }
     
--- a/modules/library/referencing/src/main/java/org/geotools/referencing/datum/BursaWolfParameters.java
+++ b/modules/library/referencing/src/main/java/org/geotools/referencing/datum/BursaWolfParameters.java
@@ -19,12 +19,9 @@
  */
 package org.geotools.referencing.datum;
 
-import java.io.Serializable;
 import static java.lang.Double.doubleToLongBits;
 
-import org.opengis.referencing.datum.GeodeticDatum;
-import org.opengis.referencing.operation.Matrix;
-import org.opengis.util.Cloneable;
+import java.io.Serializable;
 
 import org.geotools.referencing.operation.matrix.Matrix4;
 import org.geotools.referencing.operation.matrix.XMatrix;
@@ -33,6 +30,9 @@ import org.geotools.referencing.wkt.Formatter;
 import org.geotools.resources.i18n.ErrorKeys;
 import org.geotools.resources.i18n.Errors;
 import org.geotools.util.Utilities;
+import org.opengis.referencing.datum.GeodeticDatum;
+import org.opengis.referencing.operation.Matrix;
+import org.opengis.util.Cloneable;
 
 
 /**
@@ -161,7 +161,8 @@ public class BursaWolfParameters extends Formattable implements Cloneable, Seria
             throw new IllegalArgumentException("Illegal matrix size.");
         }
         for (int i=0; i<4; i++) {
-            if (matrix.getElement(3,i) != (i==3 ? 1 : 0)) {
+            double difference = matrix.getElement(3, i) - (i == 3 ? 1 : 0);
+            if (Math.abs(difference) > eps) {
                 throw new IllegalArgumentException(Errors.format(ErrorKeys.NON_AFFINE_TRANSFORM));
             }
         }
--- a/modules/plugin/epsg-hsql/src/test/java/org/geotools/referencing/factory/epsg/OperationFactoryTest.java
+++ b/modules/plugin/epsg-hsql/src/test/java/org/geotools/referencing/factory/epsg/OperationFactoryTest.java
@@ -23,26 +23,26 @@ import junit.framework.Test;
 import junit.framework.TestCase;
 import junit.framework.TestSuite;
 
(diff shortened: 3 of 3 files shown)


aaime:
Jody, what's happened to the "one ticket, one commit" practice?

modules/library/metadata/src/main/java/org/geotools/math/Plane.java
```diff
@@ -45,6 +44,86 @@
...
+    
+    /** Point3d data structure used to define Plane. */
+    public static class Point3d {
+        double x,y,z;
+        public Point3d( double x, double y, double z){
+            this.x=x;
+            this.y=y;
+            this.z=z;
+        }
+        public Point3d( double array[]) throws IllegalArgumentException {
```
jnh5y: (picky) Inconsistent spacing around arguments for the constructors.

modules/library/metadata/src/main/java/org/geotools/math/Plane.java
```diff
@@ -45,6 +44,86 @@
...
+            else if( array.length == 3 ){
+                this.x = array[0];
+                this.y = array[1];
+                this.z = array[2];
+            }
+            else {
+                throw new IllegalArgumentException("Vector array (length " + array.length +") is not expected length of 3");
+            }
+        }
+        public Point3d( Point2D point ){
```
jnh5y: If this Point3d class is *only* for defining planes, then constructors which take x,y data and assume Double.NaN for z are likely unnecessary.  

In IntelliJ, this constructor is unused.

modules/library/metadata/src/main/java/org/geotools/math/Plane.java
jnh5y: This constructor is unused.

modules/library/referencing/src/main/java/org/geotools/referencing/operation/matrix/GeneralMatrix.java
```diff
@@ -94,11 +94,11 @@ public GeneralMatrix(final int numRow, final int numCol) {
      *
      * @param numRow Number of rows.
      * @param numCol Number of columns.
-     * @param matrix Initial values.
+     * @param matrix Initial values in row order
      */
     public GeneralMatrix(final int numRow, final int numCol, final double[] matrix) {
-        super(numRow, numCol, matrix);
-        if (numRow*numCol != matrix.length) {
+        mat = new DenseMatrix64F(numRow, numCol, true, matrix);
```
jnh5y: Perhaps we should put the check/exception before the assignment to mat?

modules/library/referencing/src/main/java/org/geotools/referencing/operation/matrix/GeneralMatrix.java
```diff
@@ -404,6 +411,106 @@ public final boolean isAffine() {
     }
 
     /**
+     *  Changes the sign of each element in the matrix.
+     */
+    @Override
+    public void negate() {
+        // JNH: This seems the most aggressive approach.
```
jnh5y: My comment can go.

modules/library/referencing/src/main/java/org/geotools/referencing/operation/matrix/Matrix2.java
```diff
@@ -27,11 +26,12 @@
  * A matrix of fixed {@value #SIZE}&times;{@value #SIZE} size.
  *
  * @since 2.2
- *
+ * @version 13.0
```
jnh5y: Should this be version 14.0?

modules/library/referencing/src/main/java/org/geotools/referencing/operation/matrix/Matrix3.java
```diff
@@ -16,29 +16,27 @@
...
-
 /**
- * A matrix of fixed {@value #SIZE}&times;{@value #SIZE} size. This specialized matrix provides
- * better accuracy than {@link GeneralMatrix} for matrix inversion and multiplication.
+ * A matrix of fixed {@value #SIZE}&times;{@value #SIZE} size.
  *
  * @since 2.2
- *
- *
+ * @version 13.0
```
jnh5y: Ditto re: version.

modules/library/referencing/src/main/java/org/geotools/referencing/operation/matrix/Matrix3.java
```diff
@@ -171,9 +134,9 @@ public void setMatrix(final AffineTransform transform) {
      * @since 2.3
      */
     public boolean equalsAffine(final AffineTransform transform) {
-        return m00==transform.getScaleX() && m01==transform.getShearX() && m02==transform.getTranslateX() &&
-               m10==transform.getShearY() && m11==transform.getScaleY() && m12==transform.getTranslateY() &&
-               m20==0                     && m21==0                     && m22==1;
+        return mat.data[0]==transform.getScaleX() && mat.data[1]==transform.getShearX() && mat.data[2]==transform.getTranslateX() &&
+               mat.data[3]==transform.getShearY() && mat.data[4]==transform.getScaleY() && mat.data[5]==transform.getTranslateY() &&
+               mat.data[6]==0                     && mat.data[7]==0                     && mat.data[8]==1;
```
jnh5y: As a note, clone() doesn't work.
scala> val m = new Matrix3(1.0, 1.0, 1.0,1.0, 1.0, 1.0,1.0, 1.0, 1.0)
scala> m.clone
java.lang.ClassCastException: org.geotools.referencing.operation.matrix.GeneralMatrix cannot be cast to org.geotools.referencing.operation.matrix.Matrix3
	at org.geotools.referencing.operation.matrix.Matrix3.clone(Matrix3.java:164)

modules/library/referencing/src/test/java/org/geotools/referencing/operation/matrix/GeneralMatrixTest.java
```diff
@@ -0,0 +1,323 @@
...
+ *    Lesser General Public License for more details.
+ */
+
+/**
+ * Tests {@link org.geotools.referencing.operation.matrix.GeneralMatrix}.
+ *
+ *
+ *
+ * @source $URL$
+ * @version $Id$
```
jnh5y: Ooops.  I didn't know what to put for URL and Id.

modules/unsupported/wfs-ng/src/test/java/org/geotools/data/wfs/integration/v1_1/GeoServerIntegrationTest.java
```diff
@@ -113,7 +113,7 @@ public void testInfo() throws IOException {
         assertTrue(info.getKeywords().contains("sfRoads roads"));
         assertEquals("roads_Type", info.getTitle());
         assertEquals("http://www.openplans.org/spearfish", info.getSchema().toString());
-        assertEquals(589275.24, info.getBounds().getMinX(),0.01);
+        assertEquals(589242.03, info.getBounds().getMinX(),0.01);
```
jnh5y: Is this in meters/feet?  Hmm...  I'm not sure what to do suggest regarding the change in precision.

jnh5y:
@aaime, I think Jody was trying to preserve the git history to show that I helped out as well.  

That said, I accidentally, committed a few changes with under the wrong name (shared laptop).  I'm fine with whatever way this gets handled.  It'd be nice to remove mention of Alex, but we can do whatever is easiest.

jnh5y:
@jodygarnett, I've made my first pass through.  There are tiny style things and I noted a few places where there is unused or functions which won't work.  Let me know how we should go from here.

aaime:
Already merged? Damm... wanted to actually review the code, seeing changes to the tests to just "make it work" (replacing new value with old value like in the wfs test) was an alarm bell...
Also, I think I did mention about squashing commits?

PR #836, opened 2015-05-11
Revised matrix work [GEOT-22]

jodygarnett:
The following pull request is based on Andrea's review.

I have one commit in the wings that removes Matrix3 and Matrix4 from the MatrixFactory. I am holding off applying it while I consider implementing Matrix3 in a more direct fashion.

TODO:

* [x] XMatrix method tests for Matrix1, Matrix2, Matrix3, GeneralMatrix

Consider:

* [ ] There are two XMatrix methods (normalize and normalize(Matrix) ) that I have not added - they are described as using SVD Normalization.

### DIFF
--- a/modules/library/metadata/src/main/java/org/geotools/math/Line.java
+++ b/modules/library/metadata/src/main/java/org/geotools/math/Line.java
@@ -2,7 +2,7 @@
  *    GeoTools - The Open Source Java GIS Toolkit
  *    http://geotools.org
  * 
- *    (C) 1998-2008, Open Source Geospatial Foundation (OSGeo)
+ *    (C) 1998-2015, Open Source Geospatial Foundation (OSGeo)
  *
  *    This library is free software; you can redistribute it and/or
  *    modify it under the terms of the GNU Lesser General Public
--- a/modules/library/metadata/src/main/java/org/geotools/math/Plane.java
+++ b/modules/library/metadata/src/main/java/org/geotools/math/Plane.java
@@ -2,7 +2,7 @@
  *    GeoTools - The Open Source Java GIS Toolkit
  *    http://geotools.org
  *
- *    (C) 1998-2008, Open Source Geospatial Foundation (OSGeo)
+ *    (C) 1998-2015, Open Source Geospatial Foundation (OSGeo)
  *
  *    This library is free software; you can redistribute it and/or
  *    modify it under the terms of the GNU Lesser General Public
@@ -44,86 +44,7 @@ import org.opengis.util.Cloneable;
  * @author Howard Freeland, for algorithmic inspiration
  */
 public class Plane implements Cloneable, Serializable {
-    
-    /** Point3d data structure used to define Plane. */
-    public static class Point3d {
-        double x,y,z;
-        public Point3d( double x, double y, double z){
-            this.x=x;
-            this.y=y;
-            this.z=z;
-        }
-        public Point3d( double array[]) throws IllegalArgumentException {
-            if( array == null){
-                throw new NullPointerException("Vector array required");
-            }
-            if( array.length == 2 ){
-                this.x = array[0];
-                this.y = array[1];
-                this.z = Double.NaN;
-            }
-            else if( array.length == 3 ){
-                this.x = array[0];
-                this.y = array[1];
-                this.z = array[2];
-            }
-            else {
-                throw new IllegalArgumentException("Vector array (length " + array.length +") is not expected length of 3");
-            }
-        }
-        public Point3d( Point2D point ){
-            this.x = point.getX();
-            this.y = point.getY();
-            this.z = Double.NaN;
-        }
-        @Override
-        public int hashCode() {
-            final int prime = 31;
-            int result = 1;
-            long temp;
-            temp = Double.doubleToLongBits(x);
-            result = prime * result + (int) (temp ^ (temp >>> 32));
-            temp = Double.doubleToLongBits(y);
-            result = prime * result + (int) (temp ^ (temp >>> 32));
-            temp = Double.doubleToLongBits(z);
-            result = prime * result + (int) (temp ^ (temp >>> 32));
-            return result;
-        }
-        @Override
-        public boolean equals(Object obj) {
-            if (this == obj)
-                return true;
-            if (obj == null)
-                return false;
-            if (getClass() != obj.getClass())
-                return false;
-            Point3d other = (Point3d) obj;
-            if (Double.doubleToLongBits(x) != Double.doubleToLongBits(other.x))
-                return false;
-            if (Double.doubleToLongBits(y) != Double.doubleToLongBits(other.y))
-                return false;
-            if (Double.doubleToLongBits(z) != Double.doubleToLongBits(other.z))
-                return false;
-            return true;
-        }
-        @Override
-        public String toString() {
-            return "Point3d [x=" + x + ", y=" + y + ", z=" + z + "]";
-        }
-        /**
-         * Cartesian distance between two points.
-         * 
-         * @param pt second point
-         * @return cartesian distance between two points
-         */
-        public double distance(Point3d pt) {
-            double dX = Math.abs(this.x-pt.x);
-            double dY = Math.abs(this.y-pt.y);
-            double dZ = Math.abs(this.z-pt.z);
-            
-            return Math.sqrt( dX*dX + dY*dY + dZ*dZ);
-        }
-    }
+
     /**
      * Serial number for compatibility with different versions.
      */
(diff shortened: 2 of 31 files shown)


modules/library/metadata/src/main/java/org/geotools/math/Plane.java
```diff
@@ -44,86 +44,7 @@
...
-         */
-        public double distance(Point3d pt) {
-            double dX = Math.abs(this.x-pt.x);
-            double dY = Math.abs(this.y-pt.y);
-            double dZ = Math.abs(this.z-pt.z);
-            
-            return Math.sqrt( dX*dX + dY*dY + dZ*dZ);
-        }
-    }
+
```
jodygarnett: > The Point3D class are amassed without vertical spacing (coding convention breakage), none of public methods have javadocs. Unsure why the class is nested, the class seems general and it is being used without a Plane around

Parameter class was used to stage data for the constructor, data now provided using double[][] instead.

modules/library/referencing/src/main/java/org/geotools/referencing/operation/matrix/GeneralMatrix.java
```diff
@@ -569,28 +630,22 @@ static boolean isIdentity(final Matrix matrix, double tolerance) {
...
+        if (mat == null) {
             return prime * result;
         }
         result = prime * result + mat.numRows;
         result = prime * result + mat.numCols;
-        result = prime * result + (int) mat.data[0];
+        for (double d : mat.data) {
+            long bits = Double.doubleToRawLongBits(d);
+            result = prime * ((int)(bits ^ (bits >>> 32)));
+        }
```
jodygarnett: > Confused about the new hashcode business in GeneralMatrix:
> Issues:
> * This does not seem like a valid duplicate of the original hash code of vecmath, that used all matrix elements (this one will result in a accumulation of all matrices having the same top/left element in the same bucket)
> * The code is duplicated in a static method a few lines below:
> This static method is then called to build the hashcode of Matrix3, but I don't see the reason, aren't we inheriting the GeneralMatrix one anyways? Same goes for equals with tolerance in Matrix3, seems that we should be inheriting the base class one

Hashcode implementation was generated by hand without reference to vecmath. I have had a second attempt at implementing now that I know we would like to reference every element.

The remaining duplication was introduced by having Matrix3 extend GeneralMatrix and has now been resolved.

modules/library/metadata/src/main/java/org/geotools/math/Line.java
```diff
@@ -2,7 +2,7 @@
  *    GeoTools - The Open Source Java GIS Toolkit
  *    http://geotools.org
  * 
- *    (C) 1998-2008, Open Source Geospatial Foundation (OSGeo)
+ *    (C) 1998-2015, Open Source Geospatial Foundation (OSGeo)
```
jodygarnett: > I've seen that many of the classes changed did not receive copyright header updates (some, like GeneralMatrix, did instead).

Headers updated

modules/library/referencing/src/main/java/org/geotools/referencing/operation/matrix/GeneralMatrix.java
```diff
@@ -836,17 +911,41 @@ public final void mul(GeneralMatrix matrix){
...
+        DenseMatrix64F b = internal(matrix2);
+        if( a == mat || b == mat ){
+            mat = new DenseMatrix64F(a.numRows, b.numCols );
+        }
+        else {
+            mat.reshape(a.numRows, b.numCols, false);
         }
+        CommonOps.mult(a, b, mat);
+    }
+
```
jodygarnett: > Here I seem to see an optimization for nothing
> It seems the original intention was to leverage an in-place operation, to avoid instantiating another DenseMatrix64F, but then the implementation does create one regardless and replaces mat (so the original test has been done for nothing, or not?)

Agreed, was able to capture the sprit of the thing by checking if I can use reshape (to avoid creating a new DenseMatrix64F for the result.

modules/library/referencing/src/main/java/org/geotools/referencing/operation/matrix/GeneralMatrix.java
```diff
@@ -873,12 +972,38 @@ public void setSize(int numRows, int numCols) {
         }
     }
 
-    public void sub(GeneralMatrix matrix) {
-        CommonOps.subtract(mat, matrix.mat, mat);
+    @Override
```
jodygarnett: > The sub public methods lack a javadoc

Darn you are correct, the javadocs were provided by the former superclass.

Adding javadocs to XMatrix, aiming for method compatibility (should help prevent code casting to GeneralMatrix as well).

modules/library/referencing/src/main/java/org/geotools/referencing/operation/matrix/Matrix2.java
```diff
@@ -31,7 +32,6 @@
  * @source $URL$
  * @version $Id$
  * @author Martin Desruisseaux (IRD)
- * @deprecated Use GeneralMatrix
  */
```
jodygarnett: > Matrix2 has been marked as deprecated, yet Eclipse still finds 22 references to it. If you deprecate a class it would be good to clean the source code of its usage, no?
> Same goes for Matrix3 and Matrix4... wondering if the deprecation notes should just be removed?
> Otherwise we are just getting into the Filter business again, years and years of deprecated classes without a (funded) plan to remove them.

Matrix2 is still holding its own weight (as an optimized implementation) and I suspect we will do the same for Matrix3. Marking them as non deprecated for now.

Interestingly I found the "fast" implementation we could be using behind this class - FixedMatrix2x2_64F. Looks like there is a host of auto generated implementations for fixed matrixes that we could be using instead of DenseMatrix64F.

modules/library/referencing/src/main/java/org/geotools/referencing/operation/matrix/Matrix3.java
```diff
@@ -31,7 +31,6 @@
  * @source $URL$
  * @version $Id$
  * @author Martin Desruisseaux (IRD)
- * @deprecated Use GeneralMatrix
  */
```
jodygarnett: > About Matrix3 and Matrix4, it seems that vecmath had special high performance classes for the 3 and 4 cases, while
here we have replaced them with a general construct instead, so we are going to pay array checks for all operations.
Not awfully worried, the cases in which we do a ton of computations are for coverage reprojection, where we tend to replace the full math with a localized affine equivalent that's coded in JAI, but ... I thought you were going to expect a speedup, by what I see, maybe we'll get a small slowdown instead?

Interestingly I found the "fast" implementation we could be using behind this class - FixedMatrix2x2_64F. Looks like there is a host of auto generated implementations for fixed matrixes that we could be using instead of DenseMatrix64F.

modules/library/metadata/src/main/java/org/geotools/math/Plane.java
aaime: My comment suggested to make Point3D public and stand-alone, but this works too.

modules/library/referencing/src/main/java/org/geotools/referencing/operation/matrix/Matrix1.java
```diff
@@ -155,14 +164,30 @@ public final void transpose() {
...
-        m00 = 1/m00;
+        m00 = 1.0/m00;
+    }
+
+    @Override
+    public void invert(Matrix matrix) throws SingularMatrixException {
+        if (matrix.getNumRow()!=SIZE || matrix.getNumCol()!=SIZE) {
+            throw new IllegalArgumentException(Errors.format(ErrorKeys.ILLEGAL_MATRIX_SIZE));
+        }
+        m00 = 1.0/matrix.getElement(0, 0);
```
aaime: Does not check for singularity

modules/library/referencing/src/main/java/org/geotools/referencing/operation/matrix/Matrix1.java
```diff
@@ -224,4 +249,137 @@ public Matrix1 clone() {
...
+        if (matrix1.getNumRow() != SIZE || matrix1.getNumCol() != SIZE) {
+            throw new IllegalArgumentException(Errors.format(ErrorKeys.ILLEGAL_MATRIX_SIZE));
+        }
+        if (matrix2.getNumRow() != SIZE || matrix2.getNumCol() != SIZE) {
+            throw new IllegalArgumentException(Errors.format(ErrorKeys.ILLEGAL_MATRIX_SIZE));
+        }
+        m00 = matrix1.getElement(0, 0) * matrix2.getElement(0, 0);
+    }
+
+//    @Override
```
aaime: Commented out code

modules/library/referencing/src/main/java/org/geotools/referencing/operation/matrix/Matrix2.java
aaime: Matrix4 is still deprecated

modules/library/referencing/src/main/java/org/geotools/referencing/operation/matrix/Matrix2.java
```diff
@@ -293,4 +321,158 @@ public Matrix2 clone() {
             throw new AssertionError(e);
         }
     }
+
+    @Override
+    public void setRow(int row, double... values) {
```
aaime: Lots of new methods, are they tested?

modules/library/referencing/src/main/java/org/geotools/referencing/operation/matrix/XMatrix.java
```diff
@@ -112,6 +136,48 @@
     void invert() throws SingularMatrixException;
 
     /**
+     * Set to the inverse of the provided matrix.
+     * 
+     * @param matrix The matrix that is to be inverted. Not modified.
+     * @throws SingularMatrixException if this matrix is not invertible.
+     */
+    void invert( Matrix matrix ) throws SingularMatrixException;
```
aaime: A bit confused here, why are all these methods in XMatrix now? I guess because you have to make sure code expecting GeneralMatrix works also with MatrixN (N=1,2,3,4?)

pom.xml
```diff
@@ -231,14 +231,15 @@
   <!-- =========================================================== -->
   <issueManagement>
     <system>JIRA</system>
-    <url>http://jira.codehaus.org/browse/GEOT</url>
+    <url>https://osgeo-org.atlassian.net/projects/GEOS</url>
```
aaime: This is not GeoServer :-)

pom.xml
```diff
@@ -231,14 +231,15 @@
...
   </issueManagement>
 
   <!-- =========================================================== -->
   <!--    Continuous Integration                                   -->
   <!-- =========================================================== -->
   <ciManagement>
-    <system>continuum</system>
+    <system>jenkins</system>
+    <url>http://ares.opengeo.org/jenkins/view/geoserver/</url>
```
aaime: Likewise, we have a view for GeoTools too

modules/library/metadata/src/main/java/org/geotools/math/Plane.java
jodygarnett: Yeah it was not holding its weight, no matter what we did we had to break API.

modules/library/referencing/src/main/java/org/geotools/referencing/operation/matrix/XMatrix.java
jodygarnett: The methods were previously defined by superclasses, adding them to a common interface so we are method compatible.

Aside: this does introduce duplication with mul and multiple which I would like to address.

modules/library/referencing/src/main/java/org/geotools/referencing/operation/matrix/Matrix1.java
jodygarnett: Does now.

modules/library/referencing/src/main/java/org/geotools/referencing/operation/matrix/Matrix1.java
jodygarnett: Removed references to normalize and normalize( matrix ) we can implement later if needed.

modules/library/referencing/src/main/java/org/geotools/referencing/operation/matrix/Matrix2.java
jodygarnett: Added MatrixTest, may need to add additional tests over time.

pom.xml
jodygarnett: fixed

pom.xml
jodygarnett: fixed

jodygarnett:
If you are content anime I can merge this tomorrow.

modules/library/referencing/src/main/java/org/geotools/referencing/operation/matrix/Matrix3.java
```diff
@@ -135,23 +167,234 @@ public void setMatrix(final AffineTransform transform) {
...
+
+    @Override
+    public int hashCode() {
+        final int prime = 31;
+        int result = 1;
+        if (mat == null) {
+            return prime * result;
+        }
+        result = prime * result + SIZE; // for hashCode compatibility with GeneralMatrix
+        result = prime * result + SIZE;
```
jnh5y: Repeated line?

modules/library/referencing/src/main/java/org/geotools/referencing/operation/matrix/Matrix3.java
```diff
@@ -135,23 +167,234 @@ public void setMatrix(final AffineTransform transform) {
...
     /**
+     * Returns an affine transform for this matrix.
+     * This is a convenience method for interoperability with Java2D.
+     *
+     * @return The affine transform for this matrix.
+     * @throws IllegalStateException if this matrix is not 3&times;3,
+     *         or if the last row is not {@code [0 0 1]}.
+     */
+    public final AffineTransform toAffineTransform2D() throws IllegalStateException {
+        int check;
```
jnh5y: Is check used?

PR #840, opened 2015-05-14
[GEOT-5067]: Adding support for NetCDF CRS parsing through CF GridMap…

dromagnoli:
…ping and SPATIAL_REF Attribute


aaime:
I see quite a bit of changes that have to do wtih dimension handling, nodata, attribute types, and so on, that are not related to this ticket.

aaime:
The general approach looks fine, did not do a line by line review though

dromagnoli:
Several changes involving attribute type, unit and dimensions handling are related to the projection setup. Projection coordinates can came up expressed in km so we need to adjust them to meters. This require proper dimensions parsing and unit retrieval. NoData is used to deal with CRS reprojection.

aaime:
I see. I have nothing to add then :)

PR #842, opened 2015-05-16
[GEOT-5119] Filter function to fetch a feature's default geometry

aaime:


modules/library/main/src/main/java/org/geotools/filter/function/FilterFunction_geometry.java
```diff
@@ -0,0 +1,74 @@
...
+
+import com.vividsolutions.jts.geom.Geometry;
+
+/**
+ * Function the returns the default geometry of a feature, or null if there is none, or it's not a
+ * JTS geometry
+ * 
+ * @author Andrea Aime - GeoSolutions
+ */
+public class FilterFunction_geometry  extends FunctionExpressionImpl implements VolatileFunction {
```
jodygarnett: This naming convention is based on a code generator, for functions we make by hand GeometryFunction would be a better choice.

jodygarnett:
Merged as https://github.com/geotools/geotools/commit/244d23654fa9e5807906b08b624256bad77c75a0 based on above review.

PR #844, opened 2015-05-17
GEOT-5120 Avoid querying areas on the other side of the dateline, whenn possible and wrapping is disabled

aaime:


modules/library/render/src/main/java/org/geotools/renderer/crs/ProjectionHandlerFinder.java
```diff
@@ -84,7 +84,7 @@ public void setWrapLimit(int wrapLimit) {
     public static ProjectionHandler getHandler(ReferencedEnvelope renderingArea, CoordinateReferenceSystem sourceCrs, boolean wrap) throws FactoryException {
         if (renderingArea.getCoordinateReferenceSystem() == null)
             return null;
-
+
```
tbarsballe: Bit of a nitpick, but this file only contains whitespace changes and should not be included in the commit.

jodygarnett:
Merged to master as c9ea50f5c279711e2f69b1313535a173c3d37fc2

jodygarnett:
Have to reopen this one, it caused failures in GeoSever org.geoserver.wcs2_0.xml.GetCoverageTest.testDatelineCrossingMercatorPDC due to generated envelope not match what was expected.

PR #847, opened 2015-05-19
[GEOT-4547] Support for nonProxyHosts for MultithreadedHttpClient

awaterme:
PR as bugfix for [GEOT-4547](https://osgeo-org.atlassian.net/browse/GEOT-4547).

I think in the long run an update of the HTTP-Client for v3 to v4 will be the better solution, as it supports the JRE proxy settings out-of-the-box.
For now the submitted solution just works.

Also created test case.
Best regards,
Andreas


jodygarnett:
merged to master for milestone release

PR #855, opened 2015-05-22
Return empty bounds on empty shapefiles

HendrikPeilke:
As talked about in the mailing list (topic: bounds of empty shape file), this makes the shapefile datastore return empty bounds on a shapefile without data instead of returning the bounds stated in the header (like [0:0,0:0]). This makes it easier to calculate the bounds of a whole map by using globalBounds.expandToInclude(layerBounds), where the special case of no features inside a shapefile layer had to be handled to get the correct bounds.

The pull request includes the fix, a test and updated documentation.


jodygarnett:
merged to master

jodygarnett:
Have been forced to revert this commit due to a failure in ImageMoasic code as called by GeoServer:

StructuredCoverageStoresTests

Any one of the StructuredCoverageStoresTests will work ... but the second one will fail.

The first time SystemTestData sets up water temp for use - everything is okay:

```
>$ ls ./target/default5670765278288903012data/watertemp
	datastore.properties_h2
	datastore.properties_postgis
	elevationregex.properties
	NCOM_wattemp_000_20081031T0000000_12.tiff
	NCOM_wattemp_100_20081031T0000000_12.tiff
	NCOM_wattemp_000_20081101T0000000_12.tiff
	NCOM_wattemp_100_20081101T0000000_12.tiff
	indexer.properties
	sample_image
	timeregex.properties
```

This works and GridFormatFinder.findFormat( file ) is able to return an ImageMosaicFormat that produces a reader.

The second time SystemTestData runs:

```
$ ls ./target/default5670765278288903012data/watertemp
	datastore.properties_h2
	datastore.properties_postgis	
	elevationregex.properties
	NCOM_wattemp_000_20081031T0000000_12.tiff
	NCOM_wattemp_000_20081101T0000000_12.tiff
	NCOM_wattemp_100_20081031T0000000_12.tiff
	NCOM_wattemp_100_20081101T0000000_12.tiff	indexer.properties
	sample_image
	timeregex.properties
	watertemp.properties
	watertemp.dbf
	watertemp.shp
	watertemp.fix
	watertemp.shx
	watertemp.prj
```

This works and GridFormatFinder.findFormat( file ) is able to return an ImageMosaicFormat ... that produces a null reader. Digging into the details it seems "java.lang.IllegalArgumentException: Cannot create a mosaic out of an empty index".

Digging further GTDataStoreGranuleCatalog getBounds( typeName ) is grabbing an empty envelope out of its cache.

The first time it calculates from the ShapeFileDataStore: ReferencedEnvelope[0.23722068851276978 : 14.592757149389236, 40.562080748421806 : 44.55808294568743]

Second time the ShapeFileDataStore comes up empty: ReferencedEnvelope[0.0 : -1.0, 0.0 : -1.0]

Looking at the generated shapefile it is indeed empty:

- feature count 0
- DBF: location,C,254 ingestion,D elevation,N,9,0

jodygarnett:
I hope that we can test the ImageMosaicFormat code as a JUnit test (rather than downstream in GeoServer).

HendrikPeilke:
I have not really understood why the test is failing. Is the ImageMosaic shapefile supposed to be empty? Or is that a bug? If it is supposed to be empty, it looks to me like something inside the ImageMosaic plugin has to be fixed (returning a reader with no entries on an empty shape file) and I will be happy to do this, because I have worked with the ImageMosaic plugin code already.

aaime:
Hum... not sure, mosaics in empty state are not supposed to be working properly, not right now at the moment. As far as I remember, empty it's a state that was completely not working up until some time ago, and that was allowed to work in initialization, but not for general requests... but I might be wrong, let's see if @dromagnoli can help us here.

HendrikPeilke:
Looking at the problem more closely there seems to be an interesting property of the imagemosaic datastore: Working with empty mosaics is possible as long as the bounds are not empty (but they should be by definition). 
To Fix this problem there are several solutions inside the imagemosaic plugin:

- make the getBounds(...) methods in the several catalog implementations inside the imagemosaic plugin return a [0:0,0:0] envelope if bounds are empty. This is the easiest and at the same time the imho the most incorrect fix
- look at the two places inside the imagemosaic plugin where the catalog bounds are queried (ImageMosaicReader.setGridGeometry and RasterManager.initialize)
  - just remove the exception. But this is the most time consuming (but most correct?) fix, because then every impact has to be tracked down (looking at AbstractGridCoverage2DReader I doubt that there is any CoverageReader supporting the use case of empty bounds, since originalEnvelope is used as if this case does not have to be handled). This solution (done properly) would allow any CoverageReader to not work around the use case of empty bounds since it is supported by the base library.
  - instead of throwing an exception handle the case carefully inside these methods like in option 1 (has no need to change every catalog implementation)

Option 1 and 2.2 seem to me as intermediate solutions to handle the problem, the best fix to me would be 2.1, but it does not seem possible this close to a new version release, does it?

HendrikPeilke:
After some more investigation, I am convinced that removing the thrown exception is going to work. I did a full GeoTools build and on top of that a full GeoServer build, just to make sure. The pull request should work out fine now.

aaime:
We'll need @dromagnoli  to review, and we probably need some more tests against an empty shapefile... I mean, in theory we can create an empty mosaic, but not read a coverage out of it yet (we don't know what the specifics for the raster are, resolution, sample model, color model, until we have at least one granule in the mosaic).

HendrikPeilke:
Added some tests and thereby found some more places to fix. Also added a workaround for the envelope conversion eating up empty envelopes (see last commit), which may need to be reverted.

Because of the last code changes, the StructuredCoverageStoresTests of GeoServer fail again (and in my opinion must!). In org\geoserver\catalog\CatalogBuilder.java the case of empty coverages is explicitly denied (line 975, if (gc == null)). Therefor the exception is thrown when setting up test data (line 67 in StructuredCoverageStoresTest.java). 

GeoServer does not seem to clean its files correctly: When calling the setup to unzip watertemp2.zip it holds on to the old shape files inside the directory built by the imagemosaic plugin. This confuses the imagemosaic plugin. So if I placed a shapefile with the right name and format in the target directory, where watertem2.zip gets unzipped, I could prevent the mosaic from being read (because the shapefile may have completely different source files, which may not exist anymore, because its a leftover).

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/RasterManager.java
```diff
@@ -204,11 +204,10 @@ private void prepareCoverageSpatialElements() throws TransformException, Factory
                 final MathTransform transform = CRS.findMathTransform(coverageCRS,
                         (CoordinateReferenceSystem) coverageCRS2D);
                 final GeneralEnvelope bbox = CRS.transform(transform, coverageEnvelope);
-                bbox.setCoordinateReferenceSystem(coverageCRS2D);
-                coverageBBox = new ReferencedEnvelope(bbox);
+                coverageBBox = ReferencedEnvelope.create(bbox,coverageCRS2D);
```
HendrikPeilke: safe conversion of empty envelope

dromagnoli:
In general, it looks good.
I would also consider squashing multiple commits into a single one.
I will locally test these changes to make sure there isn't any regression (also dealing with empty mosaic creation)

dromagnoli:
I did a local test and I didn't notice any regression.

dromagnoli:
Hi @HendrikPeilke. As suggested before, a pull request should be made by minimal set of commits. Usually we do pull requests with 1 single commit unless they contain totally uncorrelated changes (in the latter case user should consider splitting the work into different pull requests too).
Could you please consider squashing your commits?

If you need some assistance, you could take a look to this:
http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

HendrikPeilke:
I will do this until this weekend. Still the GeoServer won't build with this pull request (see Jodys comment from May 26 and my comment from June 8). This would require a subsequent fix in GeoServer tests or code.

jodygarnett:
If we have changed image moasic functionality then downstream in geoserver should work? I am willing to do a build locally to confirm if this patch is otherwise ready...

HendrikPeilke:
I tried my best on rebasing. Did I do this correctly? 

With this pull request merged the geoserver build will probably fail in the REST test setup (if I remember my last test build correctly). You are right, that it will not be the same place as the initial one (your comment from May 26) but it will fail after all (see my comment from June 8). This needs further investigation before merging the pull request.

aaime:
That's not how it works... if a commit breaks the geoserver build, it either gets fixed quickly, or it gets reverted

aaime:
Fixed the build issue in GeoServer... then it took me 2 hours to untangle the history of this messy pull request in a single commit, and fix some coding convention and header issues. Replaced by https://github.com/geotools/geotools/pull/941

PR #857, opened 2015-05-25
tone down wfs-ng use of system.err, warnings and info messages

jodygarnett:
Reducing the logs during our unit tests:

* downgrade common xml binding warnings to level fine
* makes Transaction AutoClosable (GEOT-5129), and closes transactions used during unit testing
* Clean up use of System.err and System.out.


smithkm:
Looks good to me.

PR #858, opened 2015-05-26
Fix issues with dates being coerced into GMT [GEOT-5047]

ianturton:
This PR adds improved JDBC tests to make sure we can add and filter by dates regardless of the TimeZone we are in. It also modifies the temporal convertor to keep track of the TimeZone of the source object (if it has one) so that the GML code can continue to be fixed in GMT(at least in the tests).


ianturton:
Tested GeoServer (on ubuntu) and all tests pass too

bencaradocdavies:
Breaks online tests for gt-jdbc-oracle and gt-jdbc-sqlserver on Boundless Jenkins:
http://ares.boundlessgeo.com/jenkins/view/geotools/job/geotools-master-online/134/consoleText

Breaks gt-jdbc-postgis for me locally against postgis 9.4:
PostgisUDTOnlineTest>OnlineTestCase.run:123->testRead:81 expected:<2004-10-30 1[7]:30:00.0> but was:<2004-10-30 1[8]:30:00.0>

PR #860, opened 2015-05-28
removed static dateformats as they are not thread safe

ianturton:


modules/library/coverage/src/test/java/org/geotools/image/ImageWorkerTest.java
```diff
@@ -924,6 +925,7 @@ public void testOpacityAlphaRGBDirect() {
     }
     
     @Test
+    @Ignore ("Due to LibCMS Bug")
```
aaime: This actually only happens in openjdk under some linux distros (not all, 12.04 for example is not affected), maybe make this as "assume" test, checking if the jdk is a openjdk?

modules/library/main/src/main/java/org/geotools/util/CommonsConverterFactory.java
```diff
@@ -220,8 +220,8 @@
     }
     
     static class DateConverter implements Converter {
-        private static SimpleDateFormat format1 = new SimpleDateFormat( "yyyy-MM-dd HH:mm:ss.S a" );
-        private static SimpleDateFormat format2 = new SimpleDateFormat( "yyyy-MM-dd HH:mm:ssa" );
+        private SimpleDateFormat format1 = new SimpleDateFormat( "yyyy-MM-dd HH:mm:ss.S a" );
```
aaime: Hum... interesting, the converters are created per conversion, or are always there?

modules/plugin/jdbc/jdbc-db2/src/main/java/org/geotools/data/db2/DB2FilterToSQL.java
```diff
@@ -435,17 +435,12 @@ public void setLooseBBOXEnabled(boolean looseBBOXEnabled) {
...
-        // Set DATE_FORMAT time zone to GMT, as Date's are always in GMT internaly. Otherwise we'll
-        // get a local timezone encoding regardless of the actual Date value        
-        DATE_FORMAT.setTimeZone(TimeZone.getTimeZone("GMT"));
-    }
-
-    static SimpleDateFormat DATETIME_FORMAT = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss.SSS");
 
     @Override
     protected void writeLiteral(Object literal) throws IOException {
+        SimpleDateFormat DATE_FORMAT = new SimpleDateFormat("yyyy-MM-dd");
```
aaime: Maybe in DB2 dates are actually in GMT in this change might break stuff? @mcrmcr  should have a look

modules/unsupported/geopkg/src/main/java/org/geotools/geopkg/GeoPackage.java
```diff
@@ -829,6 +826,8 @@ FeatureEntry createFeatureEntry(ResultSet rs) throws SQLException, IOException {
     }
 
     void addGeoPackageContentsEntry(Entry e) throws IOException {
+        final SimpleDateFormat DATE_FORMAT = new SimpleDateFormat("yyyy-mm-dd'T'HH:MM:ss.SSS'Z'");
+//            DATE_FORMAT.setTimeZone(TimeZone.getTimeZone("GMT"));
```
aaime: Likewise for geopackage, does the spec say anything about dates? They might have to be in GMT

modules/library/coverage/src/test/java/org/geotools/image/ImageWorkerTest.java
ianturton: Sorry I thought I backed that out as it should only be on my machine (which is affected)

modules/unsupported/geopkg/src/main/java/org/geotools/geopkg/GeoPackage.java
ianturton: I scanned it and couldn't see anything - tests all still pass.

modules/plugin/jdbc/jdbc-db2/src/main/java/org/geotools/data/db2/DB2FilterToSQL.java
ianturton: It looked like a simple copy & paste to me

modules/library/main/src/main/java/org/geotools/util/CommonsConverterFactory.java
ianturton: They were always there but since they are in a static method I suspect they only get built once?

aaime: Just checked, these converters are created once and saved into a static hash map, from which they are grabbed and used. So... moving the converters from static field to field does not hlpe, the DateConverter instance itself is static. Believe we need to have them created inside the conversion method instead.

aaime:
Hi @ianturton, if we cannot get a review from @mcrmcr  I'd say to move on, but I'd still fix the issue with the format being a field of a singleton class in CommonsConverterFactory (create it when you need it instead).

mcrmcr:
Hi
The change  in DB2FilterToSQL looks ok  to me.
Cheers

aaime:
Hi @ianturton , shall we merge?

jodygarnett:
Looks like we need to ask @ianturton to resolve conflicts before merging.

modules/library/main/src/main/java/org/geotools/util/CommonsConverterFactory.java
jodygarnett: Moved them into the method, and only create as needed. I still don't like the overhead here as this converter is likely called a fair bit. Lets see how it does on master and we can always do a "quick" regex check before having to create this single threaded SimpleDateFormat.

jodygarnett:
This pull request has been closed due to merge conflicts.

A replacement pull request #934 is available for review.

PR #861, opened 2015-05-31
[GEOT-5139] CSV datastore won't honor the configured namespace URI

aaime:


tbarsballe:
Backported to 13.x so that https://github.com/geoserver/geoserver/pull/1247 can be backported to 2.7.x (Required for importer to pass tests, specifically testImportCSV()).

PR #865, opened 2015-06-02
fix paths for windows users

jodygarnett:
The change to file markup gobbled up file separator characters making the quick start tutorial harder to use.

Also added additional foss4g conferences to locations.csv


docs/user/tutorial/quickstart/eclipse.rst
```diff
@@ -424,24 +424,33 @@ generating eclipse :file:`.project` and :file:`.classpath` files.
 
 1. Download Maven from http://maven.apache.org/download.html 
    
-   The last version we tested with was: Maven 2.2.1
+   The last version we tested with was: Maven 3.3.3
```
jodygarnett: Just to double check that maven 3.3.3 is available.

PR #867, opened 2015-06-05
fast bounds using SDO_GEOM_METADATA

HendrikPeilke:
As discussed about two years ago in the geotools users mailing list (see http://sourceforge.net/p/geotools/mailman/message/30415309/) this enables fast bounds calculation for oracle database layers containing a lot of data (assumption: the access to the SDO_GEOM_METADATA views is not slower than SDO.TUNE). We integrated this into our applications and would like to share it.

The pull request includes:
- updated source code: A switch, as Andrea suggested, to turn the behaviour on (off by default)
- a test case
- updated documentation

### DIFF
--- a/modules/plugin/jdbc/jdbc-oracle/src/main/java/org/geotools/data/oracle/OracleNGDataStoreFactory.java
+++ b/modules/plugin/jdbc/jdbc-oracle/src/main/java/org/geotools/data/oracle/OracleNGDataStoreFactory.java
@@ -24,6 +24,7 @@ import java.util.Map;
 import oracle.jdbc.OracleConnection;
 
 import org.geotools.data.Transaction;
+import org.geotools.data.DataAccessFactory.Param;
 import org.geotools.jdbc.JDBCDataStore;
 import org.geotools.jdbc.JDBCDataStoreFactory;
 import org.geotools.jdbc.SQLDialect;
@@ -33,7 +34,7 @@ import org.geotools.jdbc.SQLDialect;
  * 
  * @author Justin Deoliveira, OpenGEO
  * @author Andrea Aime, OpenGEO
- *
+ * @author Hendrik Peilke
  *
  *
  * @source $URL$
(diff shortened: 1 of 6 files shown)


modules/plugin/jdbc/jdbc-oracle/src/main/java/org/geotools/data/oracle/OracleNGDataStoreFactory.java
```diff
@@ -63,6 +64,9 @@
     public static final Param GEOMETRY_METADATA_TABLE = new Param("Geometry metadata table", String.class,
             "The optional table containing geometry metadata (geometry type and srid). Can be expressed as 'schema.name' or just 'name'", false);
     
+    /** parameter for getting bbox from MDSYS.USER_SDO_GEOM_METADATA or MDSYS.ALL_SDO_GEOM_METADATA table */
+    public static final Param METADATA_BBOX = new Param("Metadata bbox", Boolean.class, "Get data bounds quickly from MDSYS.USER_SDO_GEOM_METADATA or MDSYS.ALL_SDO_GEOM_METADATA table", false, Boolean.FALSE);
```
aaime: This parameter should be advertised also in the JNDI and OCI factory

HendrikPeilke: Doesn't it get advertised by adding it to setupParameters, which has been done?

aaime: As far as I know, it does not, see how OracleNGJNDIDataStoreFactory and OracleNGOCIDataStoreFactory handle setupParameters on their own

HendrikPeilke: That is what I thought, I added it to their methods already, you probably missed it (shown below in the files overview).

aaime: Ah yes, I did miss it indeed. 1 minute review in between two actual work tasks, happens.

aaime:
Merged and added docs to GeoServer.

PR #870, opened 2015-06-09
General Oblique Transform (Rotated Pole Coordinates) [GEOT-5190]

mfilocha:
Add new transformation used quite often by Numerical Weather Forecasting community to support "Rotated Pole" coordinates. That transformation is also available in proj.4 as "+proj=ob_tran".

For initial idea of this pull request, see:
http://sourceforge.net/p/geoserver/mailman/message/34095134/

### DIFF
--- a/modules/library/referencing/src/main/java/org/geotools/referencing/operation/projection/GeneralOblique.java
+++ b/modules/library/referencing/src/main/java/org/geotools/referencing/operation/projection/GeneralOblique.java
@@ -1,0 +1,192 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ *
+ *    (C) 1999-2008, Open Source Geospatial Foundation (OSGeo)
+ *
+ *    This library is free software; you can redistribute it and/or
+ *    modify it under the terms of the GNU Lesser General Public
+ *    License as published by the Free Software Foundation;
+ *    version 2.1 of the License.
+ *
+ *    This library is distributed in the hope that it will be useful,
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ *    Lesser General Public License for more details.
+ *
+ *    This package contains formulas from the PROJ package of USGS.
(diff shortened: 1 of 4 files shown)


modules/library/referencing/src/main/java/org/geotools/referencing/operation/projection/GeneralOblique.java
```diff
@@ -0,0 +1,192 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ *
+ *    (C) 1999-2008, Open Source Geospatial Foundation (OSGeo)
```
aaime: The header should just say "2015"

modules/library/referencing/src/test/java/org/geotools/referencing/ScriptTest.java
```diff
@@ -268,7 +268,16 @@ public void testMollweide() throws Exception {
     public void testWagnerIV() throws Exception {
         runScript("scripts/WagnerIV.txt");
     }
-    
+
```
aaime: The header of this file should be updated to have 2015 as the last modified year

modules/library/referencing/src/test/resources/org/geotools/referencing/test-data/scripts/GeneralOblique.txt
```diff
@@ -0,0 +1,77 @@
+// Test points for the GeneralOblique projection
+//
+// (C) 2014, Open Source Geospatial Foundation (OSGeo)
```
aaime: 2015

aaime:
Hi, things look in orfder, minus some issue with the headers that must be fixed before committing. The other thing missing is the OSGeo contributor agreement, see the procedure here (we cannot accept your code otherwise):  http://docs.geotools.org/latest/developer/procedures/contribution_license.html

mfilocha:
Contributor License Agreement signed and sent.

jodygarnett:
Thanks, in the process of merging now. Do we have a Jira issue for this pull request (i.e. for our release notes).

jodygarnett:
merged to master, thanks for contributing to GeoTools!

PR #871, opened 2015-06-10
Safe envelope conversion

HendrikPeilke:
As discussed on the geotools-devel mailing list (topic "Envelope conversion loses emptyness"), this pull request tries to extend ReferencedEvenlope implementation, such that emptiness property survives the conversion between different envelope formats.

This pull request includes:
- the changed source code
- some minor fixes in source code (spelling mistakes, elimination of doubly defined methods, unused variables)
- a test

GeoTools and GeoServer build fine on my machine

### DIFF
--- a/modules/library/api/src/main/java/org/geotools/geometry/jts/ReferencedEnvelope.java
+++ b/modules/library/api/src/main/java/org/geotools/geometry/jts/ReferencedEnvelope.java
@@ -21,13 +21,10 @@ import java.awt.geom.Rectangle2D;
 import org.geotools.geometry.DirectPosition2D;
 import org.geotools.geometry.GeneralEnvelope;
 import org.geotools.referencing.CRS;
-import org.geotools.referencing.operation.TransformPathNotFoundException;
 import org.geotools.resources.Classes;
 import org.geotools.resources.i18n.ErrorKeys;
 import org.geotools.resources.i18n.Errors;
-import org.geotools.util.Utilities;
 import org.opengis.geometry.BoundingBox;
-import org.opengis.geometry.BoundingBox3D;
 import org.opengis.geometry.DirectPosition;
 import org.opengis.geometry.MismatchedDimensionException;
 import org.opengis.geometry.MismatchedReferenceSystemException;
@@ -181,6 +178,11 @@ public class ReferencedEnvelope extends Envelope implements org.opengis.geometry
 
     /**
      * Creates an envelope for a Java2D rectangle.
+     * 
+     * <p> NOTE: if the rectangle is empty, the resulting ReferencedEnvelope will not be. In 
+     *           case this is needed use {@link #create(Rectangle2D, 
+     *           CoordinateReferenceSystem) ReferencedEnvelope.create(rectangle,
+     *           crs)}
      *
      * @param rectangle The rectangle.
      * @param crs The coordinate reference system.
@@ -211,10 +213,12 @@ public class ReferencedEnvelope extends Envelope implements org.opengis.geometry
 
     /**
      * Creates a new envelope from an existing bounding box.
-     *
+     * <p> NOTE: if the bounding box is empty, the resulting ReferencedEnvelope will not be. In 
+     *           case this is needed use {@link #create(org.opengis.geometry.Envelope, 
+     *           CoordinateReferenceSystem) ReferencedEnvelope.create(bbox,
+     *           bbox.getCoordinateReferenceSystem())}
      * @param bbox The bounding box to initialize from.
      * @throws MismatchedDimensionException if the CRS dimension is not valid.
-     *
      * @since 2.4
      */
     public ReferencedEnvelope(final BoundingBox bbox) throws MismatchedDimensionException {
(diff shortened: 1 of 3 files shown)


modules/library/api/src/main/java/org/geotools/geometry/jts/ReferencedEnvelope.java
```diff
@@ -776,14 +786,14 @@ public String toString() {
     /**
```
HendrikPeilke: just corrected a spelling mistake down here

modules/library/api/src/main/java/org/geotools/geometry/jts/ReferencedEnvelope.java
```diff
@@ -875,21 +897,7 @@ public static ReferencedEnvelope reference(Envelope e) {
             return new ReferencedEnvelope(e, null);
         }
     }
-
```
HendrikPeilke: method is a duplicate to reference(org.opengis.geometry.Envelope env), which is not needed

modules/library/api/src/main/java/org/geotools/geometry/jts/ReferencedEnvelope.java
```diff
@@ -925,10 +935,36 @@ public static ReferencedEnvelope reference(org.opengis.geometry.Envelope env) {
             return (ReferencedEnvelope) env;
         }
     
-        if (env.getDimension() == 3) {
```
HendrikPeilke: changed to >= as in create methods

modules/library/api/src/main/java/org/geotools/geometry/jts/ReferencedEnvelope.java
```diff
@@ -925,10 +935,36 @@ public static ReferencedEnvelope reference(org.opengis.geometry.Envelope env) {
...
+            }
         }
     
+        // emptiness test according to com.vividsolutions.jts.geom.Envelope
+        if(env.getMaximum(0) < env.getMinimum(0))
+            return new ReferencedEnvelope(env.getCoordinateReferenceSystem());
+        
         return new ReferencedEnvelope(env);
     }
+
```
HendrikPeilke: added a method for a Rectangle2D, don't really know if this is needed

modules/library/api/src/test/java/org/geotools/geometry/jts/ReferencedEnvelopeTest.java
```diff
@@ -142,7 +151,7 @@ public void testBoundsEquals2D() {
     @Test
     public void testFactoryMethod() throws Exception {
         try {
-            ReferencedEnvelope bounds = new ReferencedEnvelope( DefaultGeographicCRS.WGS84_3D );
```
HendrikPeilke: eliminated an unused variable

modules/library/api/src/main/java/org/geotools/geometry/jts/ReferencedEnvelope.java
```diff
@@ -21,13 +21,10 @@
 import org.geotools.geometry.DirectPosition2D;
```
jodygarnett: Please update header to (C) 2005-2015

modules/library/api/src/main/java/org/geotools/geometry/jts/ReferencedEnvelope.java
jodygarnett: I thought having a specific BoundingBox method helped one of the compilers not freak out, requiring client code to add casting.

modules/library/api/src/main/java/org/geotools/geometry/jts/ReferencedEnvelope.java
jodygarnett: Can't hurt.

jodygarnett:
Thanks @HendrikPeilke - that is a pretty nice code review you have provided :)

I checked and we do have some contributions from you, can I double check that you have sent OSGeo the paperwork described in [CONTRIBUTING.md](https://github.com/geotools/geotools/blob/master/CONTRIBUTING.md)? It can be sent via email these days which is nice.

modules/library/api/src/main/java/org/geotools/geometry/jts/ReferencedEnvelope.java
HendrikPeilke: did it

modules/library/api/src/main/java/org/geotools/geometry/jts/ReferencedEnvelope.java
HendrikPeilke: that was the case for the methods considering a ReferencedEnvelope as a parameter (which is an OGC and a JTS envelope), but the bounding box is only an OGC envelope, so there was no problem.

HendrikPeilke:
Thanks Jody. Yes, I signed and sent the OSGeo paperwork. I got an email back from "Jachym Cepicky <secretary@osgeo.org>" saying "your CLA was filed, thank you". I filled in GeoTools as a "notify project" on the form, but apparantly that did not work as I expected. :)

modules/library/api/src/main/java/org/geotools/geometry/jts/ReferencedEnvelope.java
HendrikPeilke: Yes, it is. The constructor is also used in some places, where raster coordinates are transformed to screen coordinates (I don't know if an empty rectangle resulting is really a possible case). I have not problem with removing this method (too much API makes a library more complex and leads towards misunderstandings, that result in workarounds/solutions, which do not use the preferred way of doing things)

aaime:
Hi Jody, had a look, this one seems in order. Do you have further feedback?

aaime:
No more feedback, merging. Just one note for the future, make sure your commit messages are referring to a jira ticket.

PR #874, opened 2015-06-12
[GEOT-5145] ContentDataStore inefficient verification of available co…

aaime:
From the original discussion about this one on the ml:

> Would caching getTypeNames() alleviate this performance issue?

Yes, it would, but would have side effects too, the method would stop returning fresh information (and
we don't have a quick way to determine is something changed in a database), and the change
Ben reverted had issues in the filesystem cases, where it seemed like we had a fast way instead so...
not sure how we can effectively bring it back?

The change I'm proposing has less side effects, createTypeNames() would still return fresh values,
ensureEntry() may be fooled by the cache we setup at the first time we cannot find an entry in
case a table got removed, but the feature source created that way would fail at the first data access attempt... so, it just ends up failing a bit later down the road.
And it would not fail in case a table got added, which would instead be another failure case for createTypeNames() caching.

> Yep, please proceed and I can review.

See: https://osgeo-org.atlassian.net/browse/GEOT-5145

### DIFF
--- a/modules/library/data/src/main/java/org/geotools/data/store/ContentDataStore.java
+++ b/modules/library/data/src/main/java/org/geotools/data/store/ContentDataStore.java
@@ -560,8 +560,8 @@ public abstract class ContentDataStore implements DataStore {
         ContentEntry entry = null;
 
         boolean found = entries.containsKey(name);
-        if (!found && name.getNamespaceURI() == null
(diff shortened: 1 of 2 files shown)


jodygarnett:
Merged, cleaned up headers on these and other files in a subsequent commit.

PR #876, opened 2015-06-12
[GEOT-5146] Deprecate SOLR data store per attribute layer generation

aaime:


aaime:
Been out for 18 days with no comments, merging it

PR #879, opened 2015-06-15
Fixed issue: CQL LIKE filter not working on MongoDB DataStore

ridethepenguin:
In general, POST filters require that attributes they operate on be retrieved from the datastore and this was fixed in MongoFeatureSource.toCursor() method. However, a PropertyIsLike filter can actually be processed server-side by MongoDB itself, so it was added to the FilterCapabilities instance created by MongoDataStore.
Other changes are small fixes needed to make the online tests run (tested on Mongo v. 2.6.10 and 3.0.4).

### DIFF
--- a/modules/unsupported/mongodb/src/main/java/org/geotools/data/mongodb/FilterToMongo.java
+++ b/modules/unsupported/mongodb/src/main/java/org/geotools/data/mongodb/FilterToMongo.java
@@ -262,13 +262,13 @@ public class FilterToMongo implements FilterVisitor, ExpressionVisitor {
     @Override
     public Object visit(PropertyIsLike filter, Object extraData) {
         BasicDBObject output = asDBObject(extraData);
-        String expr = convert(filter.accept(this, null), String.class);
+        String expr = convert(filter.getExpression().accept(this, null), String.class);
 
         String multi = filter.getWildCard();
         String single = filter.getSingleChar();
         int flags = (filter.isMatchingCase()) ? 0 : Pattern.CASE_INSENSITIVE;
 
-        String regex = filter.getLiteral().replaceAll(multi, ".*").replaceAll(single, ".");
+        String regex = filter.getLiteral().replace(multi, ".*").replace(single, ".");
         Pattern p = Pattern.compile(regex, flags);
         output.put((String) expr, p);
 
--- a/modules/unsupported/mongodb/src/main/java/org/geotools/data/mongodb/FilterToMongo.java
+++ b/modules/unsupported/mongodb/src/main/java/org/geotools/data/mongodb/FilterToMongo.java
@@ -23,12 +23,16 @@ import com.mongodb.BasicDBObjectBuilder;
 import com.mongodb.DBObject;
 import com.vividsolutions.jts.geom.Envelope;
 import com.vividsolutions.jts.geom.Geometry;
+
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Set;
 import java.util.regex.Pattern;
+
 import org.bson.types.ObjectId;
+
 import static org.geotools.util.Converters.convert;
+
 import org.opengis.filter.And;
 import org.opengis.filter.BinaryComparisonOperator;
 import org.opengis.filter.ExcludeFilter;
@@ -50,6 +54,7 @@ import org.opengis.filter.PropertyIsNotEqualTo;
 import org.opengis.filter.PropertyIsNull;
 import org.opengis.filter.expression.Add;
 import org.opengis.filter.expression.Divide;
+import org.opengis.filter.expression.Expression;
 import org.opengis.filter.expression.ExpressionVisitor;
 import org.opengis.filter.expression.Function;
 import org.opengis.filter.expression.Literal;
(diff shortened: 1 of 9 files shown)


jodygarnett:
You may want to factor out the change to make the online tests run; as I see a similar fix in #880?

modules/unsupported/mongodb/src/main/java/org/geotools/data/mongodb/FilterToMongo.java
```diff
@@ -262,13 +262,13 @@ public Object visit(PropertyIsLessThanOrEqualTo filter, Object extraData) {
     @Override
     public Object visit(PropertyIsLike filter, Object extraData) {
         BasicDBObject output = asDBObject(extraData);
-        String expr = convert(filter.accept(this, null), String.class);
+        String expr = convert(filter.getExpression().accept(this, null), String.class);
```
jodygarnett: I expect this makes sense ... since you are handling the remaining parts of the filter below.

Hold on - do you want to check if expression is a Literal first? If it is a literal you have succeeded in encoding it as a mongo query.

If it is not a literal (say a function that generates the pattern used to match against) then your encoding will not work ...

modules/unsupported/mongodb/src/test/java/org/geotools/data/mongodb/FilterToMongoTest.java
```diff
@@ -54,4 +55,13 @@ public void testBBOX() throws Exception {
         assertNotNull(obj);
         System.out.println(obj);
     }
+
+    public void testLike() throws Exception {
+        PropertyIsLike like = ff.like(ff.property("stringProperty"), "on%", "%", "_", "\\");
+        BasicDBObject obj = (BasicDBObject) like.accept(filterToMongo, null);
+
+        assertNotNull(obj);
+        System.out.println(obj);
```
jodygarnett: Please do not include system outs in test cases.

You may wish to add a second test here to try out an expression that is not a propertyName.

modules/unsupported/mongodb/src/main/java/org/geotools/data/mongodb/FilterToMongo.java
aaime: If concat is not declared as supported in the filter capabiliities, it won't be part of the pre-filter. Unless there are other bugs, a FilterToSql subclass will only ever have to deal with bits that are declared in the store filter capabilities

ridethepenguin: As pointed out by Andrea, no function is listed in the filter capabilities, so `filter.getExpression()` should never return an instance of `Function`. In any case, I added a check to make sure the filter expression is a `PropertyName` instance, as Mongo's $regex operator cannot deal with anything other than a field name.

modules/unsupported/mongodb/src/test/java/org/geotools/data/mongodb/FilterToMongoTest.java
ridethepenguin: Done.

ridethepenguin:
I added another commit containing two main improvements, inspired by `JDBCFeatureSource`:

1. When a post-filter requires that additional attributes be returned from the datastore because they are used in the filter, wrap the returned `FeatureReader` in a `ReTypeFeatureReader`, to get rid of those after the filter has been applied
2. When count can't be performed in the backing datastore, do it "manually", i.e. grab a `FeatureReader` and iterate on it

I wonder why 2. was missing in the original plugin: perhaps it was deemed too expensive, since the number of features in a Mongo collection is likely very high? Maybe Tom has something to say on this.

jodygarnett:
um, you have moved on to solving two bits of work in one pull request, making this much harder to review/merge manage.

As for getCountInternal you will have to read the API contract to see if it is expected to return a result always; or if it is expected to return -1 when a full table scan is required. 

Checking getCountInternal and getCount looks like -1 can be returned - but I should do a better job of the javadocs.
```
     * It is possible that this method will return {@code -1} if the calculation
     * of number of features is judged to be too costly by the implementing class.
     * In this case, you might call <code>getFeatures(query).size()</code>
     * instead.
````

So if featureSouce.getCount() returns -1 clients are expected to do use a reader and do a full table scan.

File: modules/unsupported/mongodb/src/main/java/org/geotools/data/mongodb/FilterToMongo.java [hunk 3]
```diff
@@ -262,13 +267,19 @@ public Object visit(PropertyIsLessThanOrEqualTo filter, Object extraData) {
     @Override
     public Object visit(PropertyIsLike filter, Object extraData) {
         BasicDBObject output = asDBObject(extraData);
-        String expr = convert(filter.accept(this, null), String.class);
+        Expression filterExpression = filter.getExpression();
+        // Mongo's $regex operator only works on fields
+        if (!(filterExpression instanceof PropertyName)) {
+            throw new UnsupportedOperationException();
+        }
```
jodygarnett: This restrictions should be communicated in the UnsuppoertedOperationMessage....

And the splitting visitor need to check this restriction and only ask MongoDB to encode like expressions using a property name.

I am going to go hack up an example, but the MongoDBDataStore uses deprecated PostPreProcessFilterSplittingVisitor and should change to CapabilitiesFilterSplitter

jodygarnett:
resolved conflicts and merged, see email list for discussion.

PR #880, opened 2015-06-15
MongoDB plugin ignores arrays when generating a schema

tbarsballe:
Port of https://github.com/boundlessgeo/geoserver-exts/pull/85 to GeoTools

### DIFF
--- a/modules/unsupported/mongodb/src/main/java/org/geotools/data/mongodb/MongoDataStore.java
+++ b/modules/unsupported/mongodb/src/main/java/org/geotools/data/mongodb/MongoDataStore.java
@@ -219,8 +219,7 @@ public class MongoDataStore extends ContentDataStore {
         incoming.getUserData().put(KEY_collection, incoming.getTypeName());
         
         // Collection needs to exist (with index) so that it's returned with createTypeNames()
-        dataStoreDB.createCollection(incoming.getTypeName(), new BasicDBObject())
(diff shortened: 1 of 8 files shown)


modules/unsupported/mongodb/src/main/java/org/geotools/data/mongodb/MongoDataStore.java
```diff
@@ -219,8 +219,7 @@ public void createSchema(SimpleFeatureType incoming) throws IOException {
         incoming.getUserData().put(KEY_collection, incoming.getTypeName());
         
         // Collection needs to exist (with index) so that it's returned with createTypeNames()
-        dataStoreDB.createCollection(incoming.getTypeName(), new BasicDBObject())
-                .ensureIndex(new BasicDBObject(geometryMapping, "2dsphere"));
+        createCollection(incoming.getTypeName(), new BasicDBObject()).ensureIndex(new BasicDBObject(geometryMapping, "2dsphere"));
```
jodygarnett: This method does not exist, so I cannot merge this work? Did you forget some code ...

tbarsballe: Yup, I forgot something when I was fixing the GeoTools version of this branch - {{dataStoreDB.createCollection( .... )}} is correct

PR #881, opened 2015-06-16
GEOT-5147: NetCDF Reprojection coordinates issue

dromagnoli:
We are using the same approach used in ImageMosaic as part of GEOT-5038.

https://github.com/geotools/geotools/pull/783/files#diff-508b14536906839ca4a1ab8124b79195L1241


mbarto:
Just fix the copyright header year in SpatialRequestHelper.java. In general it looks good.

dromagnoli:
Thx... header updated

PR #884, opened 2015-06-17
Backport mongodb community module

ridethepenguin:
Brings the new version of the mongodb community plugin (based on code from Boundless) to the 13.x branch.


modules/unsupported/mongodb/README.rst
```diff
@@ -2,17 +2,128 @@ MongoDB DataStore
...
+Multiple schemas, or views, can be created for a single MongoDB document collection by creating a new, unique, typeName and specifying the collection under the root-level userData object.
+
+Implementation Notes
+--------------------
+
+* Bounding box calculation makes use of a full table scan.
+
+* Multigeometry support requires MongoDB versions 2.5 and newer
+
+* Self-intersecting polygons is a common data problem preventing MongoDBDataStore from functioning.
```
tkunicki: This might need some clarification to ease debugging of problematic datasets.  As MongoDB relies on data in WGS84 ingested data (e.g. through the importer) need to be transformed to WGS84.  While self-intersecting polygons may not exist in the original data in its native CRS, self-intersection may be introduced with a transform to WGS84.  This was observed early in testing against shapefiles representing detailed polygons in HARN projections.

modules/unsupported/mongodb/README.rst
```diff
@@ -2,17 +2,128 @@ MongoDB DataStore
...
 
-The MongoDB plugin depends on a JavaScript MapReduce script to calculate schema. The script (MetaDataCompute.js), and installation specifics (INSTALL.txt), can be found in modules/unsupported/mongodb/src/main/javascript/. Install this script on your MongoDB server.
+To publish a collection:
+
+* A compliant collection must have one or more spatial indexes of type 2dsphere. Any collection without an indexed field of type 2dsphere will be ignored.
+
+* When examining compliant collections, the schema store will be queried for an existing schema by using the collection name as the schema name. If a schema exists it is utilized otherwise it is inferred from the collection and cached to the schema store for reuse.
+
+* A collection must have its underlying coordinate system set to EPSG:4326 (WGS84, latitude/longitude order).
+
```
tkunicki: The wording here is a little confusing...  At first read I interpreted this to be a MongoDB collection property when I believe it refers to the GeoTools DataStore instance property for CRS (when I worked on this WGS84 for MongoDB 2dsphere indices was required and implicit so there wasn't a MongoDB collection level property).  Also confusing is mention of "latitude/longitude" as the coordinate when the ordering required by GeoJSON geometries is "longitude/latitude" (ignore that GeoJSON can use OGC urn for CRS and *not* respect authority defined axis ordering...)

jodygarnett:
Can I confirm that the LIKE filter is handled correctly (both when a propertyName is used and when an expression is used)? While I outlined the code I did not write a test case as I did not have an environment to work against.

Do a search through this back port I could not find exactly where the LIKE filter is tested?

ridethepenguin:
It's tested [here](https://github.com/geotools/geotools/pull/884/files#diff-e2e5cae94b8dcc2684bb93158af6f3e5R58) and [here](https://github.com/geotools/geotools/pull/884/files#diff-e2e5cae94b8dcc2684bb93158af6f3e5R65). Also [here](https://github.com/geotools/geotools/pull/884/files#diff-6279664c738cf4cd619afd63f0baa594R81) and [here](https://github.com/geotools/geotools/pull/884/files#diff-6279664c738cf4cd619afd63f0baa594R102), but these require a working MongoDB installation. Do you think it's enough?

I'll be looking into the `embedmongo-maven-plugin` mentioned by Tom to have automated integration tests back in place.

modules/unsupported/mongodb/README.rst
```diff
@@ -2,17 +2,128 @@ MongoDB DataStore
...
+Implementation Notes
+--------------------
+
+* Bounding box calculation makes use of a full table scan.
+
+* Multigeometry support requires MongoDB versions 2.5 and newer
+
+* Self-intersecting polygons is a common data problem preventing MongoDBDataStore from functioning. Please note that self-intersections may arise due to the transformation to WGS84 coordinates (which is a necessary preliminary step for importing data into MongoDB), even if they didn't exist in the original dataset.
+
+* All 2dsphere indexes and spatial operations assume the WGS84 ellipsoid. All indexed GeoJSON data stored in a MongoDB document collection is assumed to be referenced with the WGS84 coordinate reference system.
```
tkunicki: Subtle detail I noticed earlier in use of "ellipsoid" versus "datum".  I decided to skip commenting on this earlier but subsequent commits introduced the term "datum".  I would think "datum" is more correct as this includes the prime meridian definition and hence the 0 reference point for longitude.  Yes, a minor detail...

modules/unsupported/mongodb/src/test/java/org/geotools/data/mongodb/FilterToMongoTest.java
```diff
@@ -0,0 +1,86 @@
...
+        BasicDBObject obj = (BasicDBObject) like.accept(filterToMongo, null);
+
+        assertNotNull(obj);
+    }
+
+    public void testLikeUnsupported() throws Exception {
+        PropertyIsLike likeLiteral = ff.like(ff.literal("once upon a time"), "on%", "%", "_", "\\");
+        PropertyIsLike likeFunction = ff.like(
+                ff.function("Concatenate", ff.property("stringProperty"),
+                        ff.literal("test")), "on%", "%", "_", "\\");
```
jodygarnett: To test the pre/post splitter we need to try this query as part of the MongoDataStore test below. The splitter should isolate the likeFunction for post processing so we should get the same answer for both:

post (ie java processing)
````
ff.like(ff.function("Concatenate", ff.property("stringProperty")), "on%", "%", "_", "\\");
```

And pre (ie mongodb processing)
````
ff.like(ff.property("stringProperty"), "on%", "%", "_", "\\");
```

modules/unsupported/mongodb/src/test/java/org/geotools/data/mongodb/MongoFeatureSourceTest.java
```diff
@@ -0,0 +1,142 @@
...
+        // filter a post-filter
+        PropertyIsLike f = ff.like(ff.function("Concatenate",
+                ff.property("properties.stringProperty"), ff.literal("test")),
+                "on%", "%", "_", "\\");
+
+        SimpleFeatureSource source = dataStore.getFeatureSource("ft1");
+        Query q = new Query("ft1", f, new String[] { "geometry" });
+
+        // filter should match just one feature
+        assertEquals(1, source.getFeatures(q).size());
```
jodygarnett: Thanks @ridethepenguin (and @tbarsballe for checking) - this was the test I was looking for.

PR #885, opened 2015-06-17
Replaced old Codehaus/JIRA links with the new Atlassin counterparts

brett-walker:


brett-walker:
Is a quick comment about the history of JIRA need. Old issues look very different to new issues due to the migration from Codehaus to Atlassin

aaime:
Thank you

PR #894, opened 2015-06-29
fix event delegation

julian1:
Needed for https://github.com/geoserver/geoserver/pull/1125


aaime:
I think the change is fixing one bug and introducing another. The super.started() method should be called only if the sub-listener starts from 0, but not if it's, say, the third sub-listener in a list of sub-listeners all attached to the same master listener.

Also,  a  simple test should be added in the commit: we cannot merge changes that do not have tests, unless it's impossible or way too hard to write one.

PR #896, opened 2015-07-01
Geopackage: File Parameter

NielsCharlier:
Registers the geopackage "database" attribute (which is a local file path) as a File parameter, in order to support file upload in rest config.


NielsCharlier:
See also https://github.com/geoserver/geoserver/pull/1127

aaime:
The module is in unsupported land with no active maintainer (afaik), don't think you actually need a review on this one

NielsCharlier:
I think you may regard me as the maintainer of geopkg

PR #901, opened 2015-07-03
[GEOT-5162] Fixed issue with null antialiasing value being set when using Batik.

ianturton:
Fix for https://osgeo-org.atlassian.net/browse/GEOT-5162


aaime:
Wondering if a test could be written... maybe in the svg module, where you should have the dependencies you need?

ianturton:
I could try that - I hadn't added a test as I didn't have the dependencies in main

ianturton:
interestingly the issues seems to be in using SVGGraphics2D rather than an image as the rest of the tests do.

aaime:
Hi Ian, substance looks good, formally you forgot to update the licence headers (last modified year, 2015) and the if added does not follow the coding conventions (spacing around the constructs)

PR #904, opened 2015-07-03
Added WCS 1.1 schema from geoserver

tbarsballe:
Port of schemas and WCS from https://github.com/geoserver/geoserver/tree/master/src/wcs1_1, to be consistent with WCS 1.0 and WCS 2.0.
Added test cases based on the GeoTools WCS 2.0 test cases that were readily backportable to WCS 1.1.
All GeoServer WCS 1.1 test cases were preserved upream in GeoServer.


modules/extension/xsd/xsd-wcs/src/main/java/org/geotools/wcs/v1_1/WCS.java
```diff
@@ -0,0 +1,221 @@
+/* (c) 2014 - 2015 Open Source Geospatial Foundation - all rights reserved
+ * (c) 2001 - 2013 OpenPlans
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
```
jodygarnett: You will need to change to the GeoTools header. The change from GPL --> LGPL is normally fraught with peril (i.e. rewriting) but since we have code contributions in place a GeoServer PSC member can give you permission to port the code and relicense under LGPL.

So send an email to geoserver-devel asking to back port to GeoTools and I can give you a +1 in public.

modules/extension/xsd/xsd-wcs/src/main/java/org/geotools/wcs/v1_1/WCSConfiguration.java
```diff
@@ -0,0 +1,72 @@
+/* (c) 2014 - 2015 Open Source Geospatial Foundation - all rights reserved
+ * (c) 2001 - 2013 OpenPlans
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
```
jodygarnett: code is not GPL

modules/extension/xsd/xsd-wcs/src/main/resources/org/geotools/wcs/v1_1/gml4wcs.xsd
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<schema targetNamespace="http://www.opengis.net/gml" 
+xmlns:gml="http://www.opengis.net/gml"
+xmlns="http://www.w3.org/2001/XMLSchema" 
+elementFormDefault="qualified" version="1.1.1">
+	<annotation>
+		<appinfo source="urn:opengis:specification:gml:schema-xsd:gml:3.1.1">gml.xsd</appinfo>
+		<documentation>GML profile for WCS 1.2. Primary editor: Arliss Whiteside. Last updated 2007-06-06
+		Copyright © 2007 Open Geospatial Consortium, Inc. All Rights Reserved. </documentation>
+	</annotation>
```
jodygarnett: These are the actual schemas from the OGC, we track external (non LGPL) stuff like this in a review.MD file in the root of the module directory. Yeah I know the developer guide says /site/resources/review.txt but we moved to git hub.

Here is an example:

* https://github.com/geotools/geotools/blob/master/modules/extension/graph/REVIEW.md

jodygarnett:
A bit of feedback, please adjust pull request and squish the commits so we can cleanly back port.

modules/extension/xsd/xsd-wcs/src/main/resources/org/geotools/wcs/v1_1/gml4wcs.xsd
tbarsballe: FYI, this is actually not mentioned in the developer guide. Some quick google seaches show that the only place an IP check is mentioned is http://docs.geotools.org/latest/developer/procedures/check.html, and even there the IP check process is not described.

tbarsballe: One other note - All the other preexisting schemas (WCS 1.0 and WCS 2.0) in the xsd-wcs module are also licensed by the OGC. Should I include this in the REVIEW.md file as well?

tbarsballe:
Updated headers, copied and updated {{src/site/apt/review.apt}} to {{REVIEW.md}}, squahed commits.

aaime:
+1

aaime:
Merging this one for starters

PR #905, opened 2015-07-06
Speed up initialization of Oracle DataStores

sampov2:
Oracle JDBC DataStore has used SQL queries where parameter values are concatenated into the query. This approach is unsafe (consider strings containing control characters) and causes performance issues. The database needs to create a separate execution plan for each query with different values and in complex database setups, creating the execution plan might take multiple seconds. This patch has been proven to increase GeoServer startup time by a factory of 7x in a configuration where most features (~70 features) are backed by the Oracle JDBC DataStore.

### DIFF
--- a/modules/library/jdbc/src/main/java/org/geotools/jdbc/MetadataTablePrimaryKeyFinder.java
+++ b/modules/library/jdbc/src/main/java/org/geotools/jdbc/MetadataTablePrimaryKeyFinder.java
@@ -2,7 +2,7 @@
  *    GeoTools - The Open Source Java GIS Toolkit
  *    http://geotools.org
  *
- *    (C) 2002-2008, Open Source Geospatial Foundation (OSGeo)
+ *    (C) 2002-2015, Open Source Geospatial Foundation (OSGeo)
  *
  *    This library is free software; you can redistribute it and/or
  *    modify it under the terms of the GNU Lesser General Public
@@ -18,9 +18,9 @@ package org.geotools.jdbc;
 
 import java.sql.Connection;
 import java.sql.DatabaseMetaData;
+import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
-import java.sql.Statement;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashSet;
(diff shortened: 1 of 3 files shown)


aaime:
Interesting. The fun thing is that for actual spatial data queries prepared statements are a complete disaster, as the plan changes completely depending on the size of the bbox (if the bbox is small, index access is to be preferred, if larger than 10% of the data, sequential scan is actually faster).
But here we are working on metadata, so more similar to business-like queries that always extract small amount of data, so it makes sense.

sampov2:
Interesting info about PreparedStatements and spatial data. Makes sense, but it does pretty much sound like a bug in the database. The database should be clever enough to handle this case via, for example, creating alternate strategies depending on the size of the envelope of the parametrized spatial query and then select the correct strategy from these options.

Anyhow, I hope we can work this pull request into the next release.

ronnieLind:
There is an Oracle database parameter CURSOR_SHARING that you can set to FORCE (default=EXACT) so that cursors are reused. 
From Oracle 11g on there is "adapative cursor sharing" which means that there can be different execution plans for different binding values. I did not test if this also works for spatial indexes.
At least you should have done "compute statistics".

aaime:
Checked, looks good, but you must update the licence headers on all files to report the last year they got changed is 2015. Please do so and squash the commits, the push using --force to update the pull request.

aaime:
@ronnieLind I read a bit about cursor sharing, I was wondering, since the store uses prepared statements for data access, and the queries are all machine generated, what difference would it make if we push the cursor sharing to "exact"? As far as I can tell, we are generating exactly the same sql over and over for, say, most GetMap queries, if it changes, it normally means that we are trying to execute an entirely different query.

ronnieLind:
When you use bind variables, you are right.
In the logging I can see you use them for the sdo_filter bbox queries.
But you don't use bind variables for all metadata queries, PK-column checks, SRID checks etc.
If I understand correctly, this is going to be changed? Good idea. In which version?

aaime:
Yes, this pull request fixes the metadata queries. It is going to be changed in 2.8.0, if we manage to merge this pull request before August 18th (feature freeze date)

ronnieLind:
In which GeoTools version?

aaime:
14.0, also released in September 2015

aaime:
@sampov2 it's getting a bit tight for the feature freeze (Aug 18th), are you going to update the headers?

sampov2:
Hi. Sorry about the delay. The headers are now updated.

aaime:
Thanks, merged

aaime:
Hum... do we have a ticket in Jira that can be closed? If not, could you please add one? We generate the changelogs from Jira at release time.

sampov2:
There we go: https://osgeo-org.atlassian.net/browse/GEOT-5178

PR #909, opened 2015-07-07
Using ConcurrentHashMap for performance in FeatureTypeCache class

hgryoo:


modules/extension/xsd/xsd-gml2/src/main/java/org/geotools/gml2/FeatureTypeCache.java
```diff
@@ -29,28 +29,19 @@
...
-                }
-
-                return;
-            }
-
-            map.put(type.getName(), type);
-        }
+    	FeatureType other = map.putIfAbsent(type.getName(), type);
+    	if(other != null) {
+    		if(!other.equals(type)) {
```
aaime: Could you fix the extra indentation here? Does not adhere to the coding conventions.
Also, please squash all the commits into one (you can do a push --force to update your pull request after the squash)

hgryoo:
I fixed indentation problem and squashed commits too.

aaime:
Looking good, thank you!

PR #910, opened 2015-07-08
[GEOT-5166] add IN filter function taking variable number of args

ridethepenguin:


modules/library/main/src/main/java/org/geotools/filter/function/FilterFunction_in.java
```diff
@@ -0,0 +1,81 @@
+package org.geotools.filter.function;
+
+import java.util.List;
+
+import org.geotools.filter.FunctionExpressionImpl;
+import org.opengis.filter.capability.FunctionName;
+import org.opengis.filter.expression.Expression;
+
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
```
aaime: Header should be above package and imports

modules/library/main/src/test/java/org/geotools/filter/function/FilterFunction_inTest.java
```diff
@@ -0,0 +1,138 @@
+package org.geotools.filter.function;
+
+import org.opengis.feature.simple.SimpleFeature;
+import org.opengis.filter.expression.Expression;
+import org.opengis.filter.expression.Function;
+import org.opengis.filter.expression.PropertyName;
+
+/*
```
aaime: Same as the other file

modules/library/main/src/main/java/org/geotools/filter/function/FilterFunction_in.java
ridethepenguin: Fixed it.

modules/library/main/src/main/java/org/geotools/filter/function/FilterFunction_in.java
```diff
@@ -0,0 +1,81 @@
...
+ * </p>
+ * 
+ * @author Stefano Costa, GeoSolutions
+ * 
+ * 
+ * 
+ * @source $URL$
+ *
+ */
+public class FilterFunction_in extends FunctionExpressionImpl {
```
jodygarnett: Just call it InFunction, the FilterFunction_ naming convention is from a code generator that Dave Blasby wrote.

modules/library/main/src/main/java/org/geotools/filter/function/FilterFunction_in.java
```diff
@@ -0,0 +1,81 @@
...
+
+    public static FunctionName NAME = functionName("in", "result:Boolean", "candidate:Object:1,1",
+            "v:Object:1,");
```
jodygarnett: Excellent, great use of variable length parameter :)

modules/library/main/src/main/java/org/geotools/filter/function/FilterFunction_in.java
```diff
@@ -0,0 +1,81 @@
...
+    public Object evaluate(Object feature) {
+        Object candidate = getExpression(0).evaluate(feature);
+
+        boolean result = false;
+
+        if (candidate != null) {
+            List<Expression> valuesToTest = (List<Expression>) getParameters().subList(1,
+                    getParameters().size());
+            for (Expression expression : valuesToTest) {
+                Object value = expression.evaluate(feature);
```
jodygarnett: This part is difficult as you want to use the two argument form: evaluate( feature, type ) where "type" is is candidate.getClass().

This will let you preserve (as best we can) the CQL Expressions being typless.

modules/library/main/src/main/java/org/geotools/filter/function/FilterFunction_in.java
```diff
@@ -0,0 +1,81 @@
...
+        return NAME.getName();
+    }
+
+    public int getArgCount() {
+        return NAME.getArgumentCount();
+    }
+
+    public Object evaluate(Object feature) {
+        Object candidate = getExpression(0).evaluate(feature);
+
```
jodygarnett: You may want to do a null check here, if candidate is null you could use the isNull function to evaluate each value in turn...

modules/library/main/src/test/java/org/geotools/filter/function/FilterFunction_inTest.java
```diff
@@ -0,0 +1,138 @@
...
+        Object result = func.evaluate(feature);
+        assertEquals(true, result);
+    }
+
+    public void testDoubleMissing() {
+        PropertyName exp = ff.property("bar");
+        Function func = ff.function(FUNCTION_NAME, exp, ff.literal(2.499999), ff.literal(2.500001));
+        Object result = func.evaluate(feature);
+        assertEquals(false, result);
+    }
```
jodygarnett: To be clear, we need a test of:

Function func = ff.function(FUNCTION_NAME, exp, ff.literal(2.499999), ff.literal("2.500001"),ff.literal(2));

modules/library/main/src/main/java/org/geotools/filter/function/FilterFunction_in.java
ridethepenguin: Done.

modules/library/main/src/main/java/org/geotools/filter/function/FilterFunction_in.java
ridethepenguin: Thanks, but I just copied what you did [here](https://github.com/geotools/geotools/blob/master/modules/library/main/src/main/java/org/geotools/filter/function/string/ConcatenateFunction.java#L52) :-)

modules/library/main/src/main/java/org/geotools/filter/function/FilterFunction_in.java
ridethepenguin: I'm not sure I understand your point here... However, the function I'm using to compare values, `StaticGeometry.equalsTo()`, strives to do typeless comparisons (uses `toString()` to compare values of different type). I added a test to verify that int 4 is considered equal to string "4" and double 4.0. It's not perfect though, for example int 4 would not be equal to string "4.0".

modules/library/main/src/main/java/org/geotools/filter/function/FilterFunction_in.java
ridethepenguin: Done. I do believe that using isNull to compare values when candidate is null is the correct behavior, but it's not how the in* functions behave (they always return false if either value in the comparison is null): I'm worried this inconsistency may confuse users. What do you think? Should we leave everything as is or update the old functions as well?

PR #913, opened 2015-07-08
Fix example URL in WMS doc

ath0mas:


aaime:
@jodygarnett I guess you might want to have a look at this one?

PR #915, opened 2015-07-09
change import junit.framework.Assert to org.junit.Assert.

hgryoo:
junit.framework.Assert class is deprecated.


hgryoo:
What about applying this to other test classes which are importing deprecated "Assert" class?

aaime:
As long as the tests still work, no objection. There is a lot of classes that are still using junit3, only the ones written in recent years use junit4 syntax and annotations.

hgryoo:
Oh, I see. Thanks.

PR #916, opened 2015-07-10
GEOT-5167: introduced new flag in PostGIS DataStore to optimize some …

mbarto:
…edge case bbox filter usage with small bbox and big geometries

### DIFF
--- a/modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/FilterToSqlHelper.java
+++ b/modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/FilterToSqlHelper.java
@@ -2,7 +2,7 @@
  *    GeoTools - The Open Source Java GIS Toolkit
  *    http://geotools.org
  *
- *    (C) 2002-2009, Open Source Geospatial Foundation (OSGeo)
+ *    (C) 2002-2015, Open Source Geospatial Foundation (OSGeo)
  *
  *    This library is free software; you can redistribute it and/or
  *    modify it under the terms of the GNU Lesser General Public
@@ -115,6 +115,7 @@ class FilterToSqlHelper {
     FilterToSQL delegate;
     Writer out;
     boolean looseBBOXEnabled;
+    boolean encodeBBOXFilterAsEnvelope;
 
     public FilterToSqlHelper(FilterToSQL delegate) {
         this.delegate = delegate;
@@ -263,8 +264,13 @@ class FilterToSqlHelper {
         
         // add && filter if possible
         if(!(filter instanceof Disjoint)) {
-            
+            if(encodeBBOXFilterAsEnvelope && !isCurrentGeography()) {
+                out.write("ST_envelope(");
+            }
             property.accept(delegate, extraData);
+            if(encodeBBOXFilterAsEnvelope && !isCurrentGeography()) {
+                out.write(")");
+            }
             out.write(" && ");
             geometry.accept(delegate, extraData);
     
--- a/modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/PostgisNGDataStoreFactory.java
+++ b/modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/PostgisNGDataStoreFactory.java
@@ -85,13 +85,6 @@ public class PostgisNGDataStoreFactory extends JDBCDataStoreFactory {
     public static final Param SIMPLIFY = new Param("Support on the fly geometry simplification", Boolean.class, 
             "When enabled, operations such as map rendering will pass a hint that will enable the usage of ST_Simplify", false, Boolean.TRUE);
     
-    /**
-     * Enables usage of ST_Envelope to wrap geometries on non disjoint spatial filters, to speed up queries against large geometries (GEOT-5167)
-     */
-    public static final Param LARGE_GEOMETRIES_OPTIMIZATION = new Param("large geometries optimization", Boolean.class, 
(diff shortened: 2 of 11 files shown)


modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/FilterToSqlHelper.java
```diff
@@ -263,8 +264,13 @@ void visitComparisonSpatialOperator(BinarySpatialOperator filter,
         
         // add && filter if possible
         if(!(filter instanceof Disjoint)) {
-            
+            if(encodeBBOXFilterAsEnvelope && !isCurrentGeography()) {
+                out.write("ST_envelope(");
```
jodygarnett: if we are using ST_enveolope and &&...

Can we drop down to _ST_enveolope and && ?

modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/PostgisNGDataStoreFactory.java
```diff
@@ -85,6 +85,12 @@
     public static final Param SIMPLIFY = new Param("Support on the fly geometry simplification", Boolean.class, 
             "When enabled, operations such as map rendering will pass a hint that will enable the usage of ST_Simplify", false, Boolean.TRUE);
     
+    /**
+     * Enables usage of ST_Envelope to wrap geometries on Overlap filters, to speed up queries (GEOT-5167)
+     */
+    public static final Param ENCODE_BBOX_FILTER_USING_ENVELOPE = new Param("Wrap geometries with Envelope function in Overlap filters", Boolean.class, 
+            "When enabled, ST_Envelope is wrapper around geometries in overlap queries, to speed-up some edge use cases", false, Boolean.FALSE);
+
```
jodygarnett: The "some edge use cases" is very developer focused, this description is used in user interfaces shown to real. people. The "key" used in a parameter map is verbose for a map key ...

* KEY: Wrap geometries with Envelope function in Overlap filters <-- more a 
* DESCRIPTION: "When enabled, ST_Envelope is wrapper around geometries in overlap queries, to speed-up some edge use cases"

Propose:

* key: "overlap_optimisation"
* description: "Enables use of ST_Envelope to wrap geometries on Overlap filter"

modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/PostgisNGDataStoreFactory.java
```diff
@@ -156,6 +162,11 @@ protected JDBCDataStore createDataStoreInternal(JDBCDataStore dataStore, Map par
         // check geometry simplification (on by default)
         Boolean simplify = (Boolean) SIMPLIFY.lookUp(params);
         dialect.setSimplifyEnabled(simplify == null || simplify);
+        
+        // encode BBOX filter with wrapping ST_Envelope (GEOT-5167)
+        Boolean encodeBBOXAsEnvelope = (Boolean) ENCODE_BBOX_FILTER_USING_ENVELOPE.lookUp(params);
```
jodygarnett: Add note of parameter [PostGIS DataStore](http://docs.geotools.org/latest/userguide/library/jdbc/postgis.html) doc page.

modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/FilterToSqlHelper.java
mbarto: Not sure what you mean. Can you explain, please?

modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/PostgisNGDataStoreFactory.java
aaime: "Overlap" might be understood as the OGC Overlaps filter, but what we're doing is really wrapping the attribute in ST_Envelope for all non disjoint filters (bbox, intersect).  Maybe we should also state, to speed up queries against large geometries.

mbarto:
I added the new parameter documentation and changed related name and description following your suggestions. Thanks

modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/FilterToSqlHelper.java
jodygarnett: This may be an internal detail of postgis? Perhaps I have my naming convention wrong ...

jodygarnett:
I am happy to merge this, can we make this the default?

mbarto:
Currently it is a datastore configuration flag. Would you add a system variable in addition or to replace the configuration flag?
Should the system property used to force a behaviour (so setting it to true or false force the related behaviour for any postgis datastore)?

modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/FilterToSqlHelper.java
mbarto: Still not clear to me: this is the PostGIS own FilterToSqlHelper, so anything is in this class is a postgis detail.

aaime:
I'd go with the system variable indeed, and have the ST_Envelope be used by default

mbarto:
Trying to summarize:
 * introduce a system variable INSTEAD of the datastore parameter (so I will remove the parameter and related documentation)
 * use ST_Envelope by default, removing it only if the system variable is set to true

PR #918, opened 2015-07-21
[GEOT-5170] A visitor to optimize filters for in memory evaluation

aaime:
See: https://osgeo-org.atlassian.net/browse/GEOT-5170

### DIFF
--- a/modules/library/main/src/main/java/org/geotools/filter/visitor/BindingFilterVisitor.java
+++ b/modules/library/main/src/main/java/org/geotools/filter/visitor/BindingFilterVisitor.java
@@ -1,0 +1,322 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ * 
+ *    (C) 2015, Open Source Geospatial Foundation (OSGeo)
+ *    
+ *    This library is free software; you can redistribute it and/or
+ *    modify it under the terms of the GNU Lesser General Public
+ *    License as published by the Free Software Foundation;
+ *    version 2.1 of the License.
+ *
+ *    This library is distributed in the hope that it will be useful,
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ *    Lesser General Public License for more details.
+ */
+package org.geotools.filter.visitor;
+
+import java.util.Date;
+import java.util.List;
+
+import org.opengis.feature.type.FeatureType;
+import org.opengis.filter.PropertyIsBetween;
+import org.opengis.filter.PropertyIsEqualTo;
+import org.opengis.filter.PropertyIsGreaterThan;
+import org.opengis.filter.PropertyIsGreaterThanOrEqualTo;
+import org.opengis.filter.PropertyIsLessThan;
+import org.opengis.filter.PropertyIsLessThanOrEqualTo;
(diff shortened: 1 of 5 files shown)


jodygarnett:
I will review and merge, need to add a note to the docs.

bencaradocdavies:
Hmm, I cherry-picked this onto master and see failures in gt-jdbc-postgis:

Failed tests: 
  PostgisTemporalFilterOnlineTest>OnlineTestCase.run:123->JDBCTemporalFilterOnlineTest.testTContains:178->JDBCTemporalFilterOnlineTest.assertDatesMatch:70->JDBCTemporalFilterOnlineTest.assertDatesMatch:76 expected:<2> but was:<0>
  PostgisTemporalFilterOnlineTest>OnlineTestCase.run:123->JDBCTemporalFilterOnlineTest.testTContains:178->JDBCTemporalFilterOnlineTest.assertDatesMatch:70->JDBCTemporalFilterOnlineTest.assertDatesMatch:76 expected:<2> but was:<0>

bencaradocdavies:
I can confirm that the gt-jdbc-postgis test fails on master if and only if this change is applied.

I have not yet tested gs-app-schema-test (offline or online).

bencaradocdavies:
@aaime, @jodygarnett, the good news is that, with this change on master, a full GeoServer build passes (-Prelease), and GeoServer app-schema online tests pass against postgis. (We have previously had some complex feature fun and games after filter optimisation (SimplifyingFilterVisitor MatchAction), so I thought I should check.)

The only failure I could find is the gt-jdbc-postgis online test failure described above.

jodygarnett:
Given that @aaime is doing way too much this week are you in position to look at the failure reported above @bencaradocdavies ?

I am still trying to sort out where this goes in our docs, looking at (https://osgeo-org.atlassian.net/browse/GEOT-5170) and the code it looks like this class is intended to be extended as an optimization.  Sound like it should be covered in the section on the different visitor implementations, possibly with a code example (probably for Date which always seems to be the problem when sorting).

bencaradocdavies:
@jodygarnett I had a look but could not see the problem. Some type conversion problem that occurs in jdbc or postgis, perhaps? I have not yet stepped through it in the debugger and may not have time today.

@ianturton if you have a moment could you please advise? You may have seen similar jdbc or postgis date handling problems in your recent fixes.

modules/library/main/src/main/java/org/geotools/filter/visitor/BindingFilterVisitor.java
```diff
@@ -0,0 +1,322 @@
...
+        Expression expr2 = optimizeTime(overlappedBy.getExpression2(), extraData);
+
+        return getFactory(extraData).overlappedBy(expr1, expr2, overlappedBy.getMatchAction());
+    }
+
+    public Object visit(TContains contains, Object extraData) {
+        Expression expr1 = optimizeTime(contains.getExpression1(), extraData);
+        Expression expr2 = optimizeTime(contains.getExpression2(), extraData);
+
+        return getFactory(extraData).contains(expr1, expr2, contains.getMatchAction());
```
bencaradocdavies: Aha! Found it. This statement should use the method "tcontains" not "contains". Changing this method to "tcontains" fixes the build failure in gt-jdbc-postgis online tests.

bencaradocdavies:
@aaime @jodygarnett please see #935 for a new pull request that should replace this one. #935 is Andrea's commit (cherry-picked) and my tcontains fix.

PR #920, opened 2015-07-27
[GEOT-5174] ImageMosaic won't work correctly against a index store with multiple geometries

aaime:


dromagnoli:
Looks good to me. +1

PR #926, opened 2015-08-07
[GEOT-5183] direct access TextSymbolizer fonts list

jodygarnett:
API change covered by https://github.com/geotools/geotools/wiki/Text-Symbolizer-Multi-font-support proposal.

For additional details see https://osgeo-org.atlassian.net/browse/GEOT-5183


aaime:
In my branch I had to fix duplicating style visitor too. Generally speaking, everything that's using getFont() setFont() directory kills multifont support, I just stumbled into the duplicating* family as it was eating away the fonts during the copy, so they were not getting into the renderer.
SLDParser needs to be fixed as well (which I also did in my branch).

Trouble is, my branch is based on the array methods, so I have to throw away all of that and handle the conflicts (I also already changed TextSymbolizerImpl)...

jodygarnett:
Looks like https://github.com/geotools/geotools/wiki/Text-Symbolizer-Multi-font-support is a go.

Sorry about the merge hell, was trying to act quick to get you a running start :(

jodygarnett:
This pull request has been replaced by #928

PR #929, opened 2015-08-08
GEOT-5186 Add support for DB2 selectivity clause

mcrmcr:
Pull request includes

Core Code
Adapted Test
Documentation

See https://osgeo-org.atlassian.net/browse/GEOT-5186

### DIFF
--- a/docs/user/library/jdbc/db2.rst
+++ b/docs/user/library/jdbc/db2.rst
@@ -69,6 +69,25 @@ For more information check the java docs for:
 * DB2NGJNDIDataStoreFactory
 * DB2NGDataStoreFactory
 
+Advanced Geotools Parameters
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
++-------------------+------------------------------------------+
+| Parameter         | Description                              |
++===================+==========================================+
+| "loose bbox"      | Flag controlling loose bbox comparisons, |
+|                   | default is true                          |
++-------------------+------------------------------------------+
+| "use selectivity" | Flag to raise the priority of spatial    |
+|                   | indices for the db2 optimizer,           |
+|                   | default is false                         |
++-------------------+------------------------------------------+
+
+Example use::
+  
+  params.put(DB2NGDataStoreFactory.LOOSEBBOX, true );
+  params.put(DB2BNGDataStoreFactory.USE_SELECTIVITY, true );
+
 Registering spatial columns
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
--- a/modules/plugin/jdbc/jdbc-db2/src/main/java/org/geotools/data/db2/DB2FilterToSQL.java
+++ b/modules/plugin/jdbc/jdbc-db2/src/main/java/org/geotools/data/db2/DB2FilterToSQL.java
@@ -2,7 +2,7 @@
  *    GeoTools - The Open Source Java GIS Toolkit
  *    http://geotools.org
  *
- *    (C) 2011-2012, Open Source Geospatial Foundation (OSGeo)
+ *    (C) 2011-2015, Open Source Geospatial Foundation (OSGeo)
  *
  *    This library is free software; you can redistribute it and/or
  *    modify it under the terms of the GNU Lesser General Public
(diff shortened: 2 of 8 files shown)


docs/user/library/jdbc/db2.rst
```diff
@@ -69,6 +69,25 @@ For more information check the java docs for:
...
++-------------------+------------------------------------------+
+| Parameter         | Description                              |
++===================+==========================================+
+| "loose bbox"      | Flag controlling loose bbox comparisons, |
+|                   | default is true                          |
++-------------------+------------------------------------------+
+| "use selectivity" | Flag to raise the priority of spatial    |
+|                   | indices for the db2 optimizer,           |
+|                   | default is false                         |
++-------------------+------------------------------------------+
```
jodygarnett: yay documentation! I am still a bit confused on how it is worded ...

If this is just changing a priority would you do better to use a word here (rather than just a flag).

jodygarnett:
This pull request seems fine, I am still a bit confused about what is going on and would like a second opinion.

docs/user/library/jdbc/db2.rst
mcrmcr: Hi Jody

Feel free to change the wording. I will try to explain in my words.

DB2 uses a cost based optimizer based on distribution statistics. As an example, if you have an index on a  column with  values 'Y' or 'N' and 50% of the records have a value of 'Y' and 50% have a value of 'N', DB2 will ignore the index and will perform a table scan. In this case we have

SELECTIVITY 0.5

This value is calculated from DB2, no need to add it to the select statement.


For geometries (stored as BLOBs) it is not possible to calculate a distribution statistic. The optimizer prefers all other indices before using a spatial index.

Adding SELECTIVITY 0.00001 is a hint to the optimizer. (simulating a distribution statistic). 

Hope this helps
Cheers
Christian

aaime:
Looks ok, but I have a question, why was the selectivity applied to distance based filters, and bbox, but not to the other ones hanlded by visitBinarySpatialOperator?

mcrmcr:
Hi Andrea

The call to addSelectivity() is already part of visitBinarySpatialOperator. I added addSelectivity() as a dummy method in earlier days, now I really want to activate it.

mcrmcr:
Merged the commit manually

docs/user/library/jdbc/db2.rst
jodygarnett: It does help. If SELECTIVITY means something to DB2 users (which it looks like it does since you have it in an SQL example) then I am happy to keep the word.

The description though is about an effect, SELECTIVITY seems to be a detail on how it is implemented?

How about: SELECTIVITY: Value of true will inject ``SELECTIVITY 0.0001`` hint to the optimizer to raise the priority of the spatial index. Default is false.

mcrmcr: HI Jody

This correct, SELECTIVITY 0.0001 is a hint for the optimizer, but the
default is false for backward compatibility.

Cheers
Christian

On Tue, Aug 18, 2015 at 6:24 AM, Jody Garnett <notifications@github.com>
wrote:

> In docs/user/library/jdbc/db2.rst
> <https://github.com/geotools/geotools/pull/929#discussion_r37263195>:
>
> > @@ -69,6 +69,25 @@ For more information check the java docs for:
> >  * DB2NGJNDIDataStoreFactory
> >  * DB2NGDataStoreFactory
> >
> > +Advanced Geotools Parameters
> > +^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> > +
> > ++-------------------+------------------------------------------+
> > +| Parameter         | Description                              |
> > ++===================+==========================================+
> > +| "loose bbox"      | Flag controlling loose bbox comparisons, |
> > +|                   | default is true                          |
> > ++-------------------+------------------------------------------+
> > +| "use selectivity" | Flag to raise the priority of spatial    |
> > +|                   | indices for the db2 optimizer,           |
> > +|                   | default is false                         |
> > ++-------------------+------------------------------------------+
>
> It does help. If SELECTIVITY means something to DB2 users (which it looks
> like it does since you have it in an SQL example) then I am happy to keep
> the word.
>
> The description though is about an effect, SELECTIVITY seems to be a
> detail on how it is implemented?
>
> How about: SELECTIVITY: Value of true will inject SELECTIVITY 0.0001 hint
> to the optimizer to raise the priority of the spatial index. Default is
> false.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/geotools/geotools/pull/929/files#r37263195>.
>



-- 
DI Christian Mueller MSc (GIS), MSc (IT-Security)
OSS Open Source Solutions GmbH

PR #930, opened 2015-08-09
GEOT-5179, single layer z order control

aaime:
Single layer portion of GEOT-5179, and https://github.com/geotools/geotools/wiki/Z-order-control-during-map-production


aaime:
Btw, please review this one first, I won't make a pull request for cross layer z ordering until this one, much simpler and with little/no refactoring required, is merged.

modules/library/main/src/main/java/org/geotools/filter/SortByImpl.java
```diff
@@ -52,4 +52,55 @@ public void setSortOrder(SortOrder sortOrder) {
...
+
+    /*
+     * (non-Javadoc)
+     * 
+     * @see java.lang.Object#toString()
+     */
+    @Override
+    public String toString() {
+        return "SortByImpl [propertyName=" + propertyName + ", sortOrder=" + sortOrder + "]";
+    }
```
jodygarnett: Thanks, toString() methods help so much when debugging

modules/library/render/src/test/java/org/geotools/renderer/lite/ZOrderTest.java
```diff
@@ -0,0 +1,88 @@
+package org.geotools.renderer.lite;
```
jodygarnett: Will add/fix headers during merge

aaime: No problem, just tell me where the problems are, I'll add the header

jodygarnett:
merged, with docs, to master

PR #931, opened 2015-08-09
GEOT-5188 SQLDialect ignores Schema for index (create statement)

mcrmcr:
https://osgeo-org.atlassian.net/browse/GEOT-5188

DB2 tests are working now. If the fix is ok I will  backport to 13.x.

### DIFF
--- a/modules/library/jdbc/src/main/java/org/geotools/jdbc/SQLDialect.java
+++ b/modules/library/jdbc/src/main/java/org/geotools/jdbc/SQLDialect.java
@@ -2,7 +2,7 @@
  *    GeoTools - The Open Source Java GIS Toolkit
  *    http://geotools.org
  *
- *    (C) 2002-2008, Open Source Geospatial Foundation (OSGeo)
+ *    (C) 2002-2015, Open Source Geospatial Foundation (OSGeo)
  *
  *    This library is free software; you can redistribute it and/or
  *    modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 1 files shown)


modules/library/jdbc/src/main/java/org/geotools/jdbc/SQLDialect.java
```diff
@@ -1247,6 +1247,10 @@ public void createIndex(Connection cx, SimpleFeatureType schema, String database
             sql.append("UNIQUE ");
         }
         sql.append("INDEX ");
+        if (databaseSchema != null) {
+            encodeSchemaName(databaseSchema, sql);
+            sql.append(".");
+        }        
         sql.append(escape).append(index.getIndexName()).append(escape);
```
jodygarnett: Please update the header to  (C) 2002-2015 when making a change.

This change does not have a test case so I cannot determine if it is "ok". Is it already covered by a DB2TestCase - if so which one?

mcrmcr:
Hi Jody

My last comment is lost, no idea why, lets do it again.

The tests are 
org.geotools.jdbc.JDBCDataStoreOnlineTest.testMultiColumnIndex()
org.geotools.jdbc.JDBCDataStoreOnlineTest.testSimpleIndex()

I want to merge and backport because the DB2 plugin runs 321 tests, about 20 of them are failing.  I want  to get all tests running.

Cheers
Christian

aaime:
Works for me, +1

aaime:
Btw @mcrmcr you still haven't updated the copyright headers year. Please do so before merging.

PR #932, opened 2015-08-10
[GEOT-5189] getNames() and getFeatureSource() incosistency in AppSchemaDataAccess

ridethepenguin:
Basically, I merged the logic from `getMappingByName()` into `getMappingByElement()`, and renamed it to `getMappingElementByNameOrElement()`.

Technically, this is an API change, because the `getMappingByElement()` method is public; however, the JavaDocs state _"this method is public just for unit testing purposes"_, so I guess it's safe to rename it. Also, the `getMappingByName()` method is now redundant and could be removed, but I didn't want to go that far (it's also public _"for unit testing purposes"_).

### DIFF
--- a/modules/extension/app-schema/app-schema/src/main/java/org/geotools/data/complex/AppSchemaDataAccess.java
+++ b/modules/extension/app-schema/app-schema/src/main/java/org/geotools/data/complex/AppSchemaDataAccess.java
@@ -2,7 +2,7 @@
  *    GeoTools - The Open Source Java GIS Toolkit
  *    http://geotools.org
  *
- *    (C) 2008-2011, Open Source Geospatial Foundation (OSGeo)
+ *    (C) 2008-2015, Open Source Geospatial Foundation (OSGeo)
  *
  *    This library is free software; you can redistribute it and/or
  *    modify it under the terms of the GNU Lesser General Public
@@ -163,7 +163,7 @@ public class AppSchemaDataAccess implements DataAccess<FeatureType, Feature> {
      * list of FeatureType mappings and returns it.
      */
     public FeatureType getSchema(Name typeName) throws IOException {
-        return (FeatureType) getMappingByElement(typeName).getTargetFeature().getType();
+        return (FeatureType) getMappingByNameOrElement(typeName).getTargetFeature().getType();
     }
 
     /**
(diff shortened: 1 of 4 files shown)


modules/extension/app-schema/app-schema/src/main/java/org/geotools/data/complex/AppSchemaDataAccess.java
```diff
@@ -197,17 +198,23 @@ public FeatureTypeMapping getMappingByName(Name typeName) throws IOException {
...
+            return mapping;
         }
-        ArrayList<String> availables = new ArrayList<String>();
-        for (FeatureTypeMapping mapping : mappings.values()) {
-            availables.add(mapping.getTargetFeature().getName().toString());
+
+        // lookup by mapping name failed, try to lookup by target element
+        // NOTE: in this case, there is a risk of ambiguity. E.g. consider a (questionable) mapping configuration where multiple mappings of the
+        // same type have been specified, each with its own mappingName: they would have the same targetElement, and we wouldn't know which one to
+        // pick. The result would be unpredictable in such a situation.
```
jodygarnett: If you like you could detect that inconsistency when setting up the mapping - rather than leave the confusion until here.

jodygarnett:
This looks fine, thanks for the fix, will do a local build and merge shortly.

modules/extension/app-schema/app-schema/src/main/java/org/geotools/data/complex/AppSchemaDataAccess.java
ridethepenguin: The problem I see in forbidding such a mapping configuration, is that it's neither wrong nor ambiguous, it's just more verbose.

The ambiguity is a consequence of using the element name to lookup the mapping, which in this case is not unique. I could throw an exception when the ambiguity cannot be resolved, but picking the first match has been the behavior  of the `getMappingByElement()` method so far and I'm afraid that changing it could break existing clients.

bencaradocdavies:
@jodygarnett as this is an API change, and may, as @ridethepenguin indicated above, break existing clients, I advise against backporting this change.

bencaradocdavies:
I am pleased to confirm that with this change on master, full GT+GWC+GS builds pass, as do GS app-schema online tests against postgis. Great work @ridethepenguin!

PR #933, opened 2015-08-10
GEOT-3776 Implementation of dasharray as List<Expression>

igorvolk:
Tracker: https://osgeo-org.atlassian.net/browse/GEOT-3776

Related proposal from Nuno Oliveira: 
https://github.com/geotools/geotools/wiki/stroke-dasharray-as-an-expression

Discussion: http://osgeo-org.1560.x6.nabble.com/SLD-Stroke-Dasharray-Property-td5083512.html

I have implemented some solution for the problem of dasharray expression. In last variant of my solution I have created a new interface org.geotools.styling.Stroke2, which is inherited from interface org.geotools.styling.Stroke. In this new interface were added 2 methods: Expression[] getDashExpressionArray() and void setDashExpressionArray(Expression[] value). An implementation of Stroke2 was added to class org.geotools.styling.StrokImpl. Also was created an interface StyleFactory3 with methods of Stroke creation, using Expression[] for dasharray. An implementation of StyleFactory3 was added to StyleFactoryImpl. The key idea of the Expression[] support for dasharray in SLD is the using of some filter function. In fact, no matter what function to use. What we need is not the result of a function but a list of its arguments. Since it is desirable that the value of CssParameter for stroke-dasharray was meaningful, I chose to use in it the function Concatenate. It is a controversial issue whether to use space delimiters in the argument list of this function. (Such delimiters can be represented only as CDATA's, what is not aesthetically pleasing, although there using is naturally enough). I chose to use them. Modified SLDTransformer inserts such delimiters and modified SLDParser eliminates them.
For example, stroke2.setDashExpressionArray(new Expression[]{ff.property(dash_size), ff.property(blang_size) }) will be transformed by SLDTransformer to

&lt;CssParameter name="stroke-dasharray"&gt;
&lt;Function name="Concatenate"&gt;
&lt;PropertyName&gt;dash_size&lt;/PropertyName&gt;
&lt;Literal&gt;&lt;![CDATA[ ]]&gt;&lt;/Literal&gt;
&lt;PropertyName&gt;blank_size&lt;/PropertyName&gt;
&lt;/Function&gt;
&lt;/CssParameter&gt;

SLDParser accepts from such CssParameter only Concatenate function or literal. Literal is interpreted by SLDParser as float[] and Concatenate function is interpreted as Expression[], containing all arguments of this function except CDATAs.
Also was executed modification of visitors using dasharray (DuplicatingStyleVisitor, UomRescaleStyleVisitor and UomRescaleHelper).
Necessary tests were added.


jodygarnett:
Travis failed with:

```
/home/travis/build/geotools/geotools/modules/library/render/src/main/java/org/geotools/renderer/style/SLDStyleFactory.java:[1561,34] error: cannot find symbol
````

Investigating.

nmco:
Igor sorry for insisting, but I really don't like the idea of using a function (in the example the concat function) just to access is arguments. A preferable way will be something like what is done for the label property. Take a look a the function 'parseCssParameter' in the SLDParser.

jodygarnett:
@nmco have a look at the revised proposal https://github.com/geotools/geotools/wiki/stroke-dasharray-as-an-expression

Any discussion on geotools-devel would help as I would like to ask PMC members to vote on the API change.

igorvolk:
In my last commit I have implemented the proposals of Jody Garnett, concerning the interface org.geotools.styling.Stroke. I have left interface  org.opengis.style.Stroke unmodified. I have some doubts concerning the implementation of legacy method StrokeImpl.getDashArray().
Look at my implementation.

 public float[] getDashArray() {
        if (dashArray.isEmpty()){
        	return Stroke.DEFAULT.getDashArray();
        }  else {
            int floatLiteralCount = 0;
            for (Expression expr: dashArray){
                if (expr instanceof Literal){
                    Float dash = expr.evaluate(null, Float.class);
                    if (dash != null){
                      floatLiteralCount++;
                    }
                }
            }
            if (floatLiteralCount != dashArray.size()){
                return Stroke.NULL.getDashArray();
            } else {
                float[] result = new float[floatLiteralCount];
                int j = 0;
                for (Expression expr: dashArray){
                    Float dash = expr.evaluate(null, Float.class);
                    result[j] = dash;
                    j++;
                }
                return result;
            }
        }
    }

May be it's all right. But may be I need to swap the returns of Stroke.DEFAULT.getDashArray() and Stroke.NULL.getDashArray() or to change somethig else. I am waiting for your opinion.

I abandoned my old ideas about the use in SLDParser of filter function  for extraction of dash expressions.
For SLDParser I chose to use the excellent code of Nuno Oliveira (after fixing small bug in it). A lot of thanks to him!!!
I suggested to Nuno to finish this task because he initiated it. But he has not answerd yet. So I continue to work with dashArray (But I am ready at any moment to give him way) .
I believe that there is no need to make changes in classes 
library/opengis/src/main/java/org/opengis/style/StyleFactory.java
library/api/src/main/java/org/geotools/styling/StyleFactory.java
library/main/src/main/java/org/geotools/styling/AbstractStyleFactory.java
library/main/src/main/java/org/geotools/styling/StyleFactoryImpl.java
library/main/src/main/java/org/geotools/styling/StyleFactoryImpl2.java
because the use of method List<E[pression>dashArray() assumes that the add expressions after the creation of Stroke instance.
But class extension/brewer/src/main/java/org/geotools/styling/builder/StrokeBuilder I have modified by changing the implementation of method dashArray(float... dashArray) and adding of method dash(Expession dash).

Regarding the failure of Travis while building of my commit, I am sure that I'm sure the reason is the merge conflict with latest changes if SLDParser. To avoid repeating this failure, I tried to upgrade my local copy of geotools from upstream/master following the instructions at https://help.github.com/articles/syncing-a-fork/  . But I guess I did something wrong, because I have got in SLDParser some references on methods of ContrastEnhancement (setNormalize();
,m .setHistogram(), setLogarithmic(), setExponential), that are absent in ContrastEnhancement interface. At geotools/geotolls those methods are absent too. I'm confused and don't know what to do.

igorvolk:
In my last message I said that I'm confused about the conflict in SLDParser. Later I have pulled original SLDParser from up upstream/master and added Nuno's code to it. Now SLDParser compiles normally, but SLDParserTest fails on testContrastEnhancement. Original upstream/master passes all tests successfully. So conflict remains somewhere. Then I merged my brunch with geotools/master, but it did not help. Should I do something else to restore testContrastEnhancement?
Now I can see, that merging with upstream/master was not a good idea. But how can I avoid conflicts in Travis than? I wish to create new pool request for anather item. From what point should I start? Should I create new fork from geotools/master? Or it is enough to start from my original master, created from previouse fork, and merge it with current geotools/master? Where is right place for such questions?

igorvolk:
In order to clarify the essence of my doubts about the expected behavior of the method float[] Stroke.getDashArray() I have add new test in SLDStyleTest:

        Stroke stroke = sf.createStroke(null, null);
        assertNull(stroke.getDashArray());

        List<Expression> dashArray = stroke.dashArray();
        dashArray.add(ff.literal("5"));
        dashArray.add(ff.literal("10"));
        assertTrue(Arrays.equals(new float[]{5, 10},  stroke.getDashArray()));

        dashArray.add(ff.property("dash_size"));
        assertTrue(Arrays.equals(new float[]{},  stroke.getDashArray()));

        Stroke stroke2 = sf.createStroke(null, null, null,
                null, null, new float[]{5, 10}, null,
                null, null);
        assertTrue(Arrays.equals(new float[]{5, 10},  stroke2.getDashArray()));

Is this the behavior that users of method getDashArray() can expect?

aaime:
Closing, there is a new pull request implementing this

PR #935, opened 2015-08-11
[GEOT-5170] A visitor to optimize filters for in memory evaluation

bencaradocdavies:


bencaradocdavies:
@aaime @jodygarnett here is the new PR with the tcontains fix. This should replace #918.

bencaradocdavies:
Thanks, @aaime. Local full GT (online), GWC, GS, and GS app-schema online test builds passed with this change. Boundless Jenkins still running.

PR #936, opened 2015-08-11
fixes worldToGrid function by fully applying TOL

apc3n:


jodygarnett:
This seems okay, is there a bug report or something so I can figure out what is going on?

apc3n:
no @jodygarnett  I don't have a bug report (beyond my first writeup on SO: http://stackoverflow.com/questions/31595565/gridgeometry2ds-world-to-grid-function-translating-wrong). I'm honestly not sure what's exactly going on (if there were some documentation for what TOL means it might help).

apc3n:
My guess is that some strange floating point conversion goes screwy and the TOL adjustment needs to occur for the first row and column not just the last row and column (aka the currently existing two if checks that adjust the location using TOL).

aaime:
@apc3n, the commit message should refer a ticket in Jira, please create one and rewrite the commit message to be "[TICKET-NUMBER] Ticket title".

The copyright year in the files you have modified should look like "startYear - 2015".

The change looks otherwise fine, thanks for adding a test too.

modules/library/coverage/src/main/java/org/geotools/coverage/grid/GridGeometry2D.java
```diff
@@ -2,7 +2,7 @@
  *    GeoTools - The Open Source Java GIS Toolkit
  *    http://geotools.org
  *
- *    (C) 2001-2008, Open Source Geospatial Foundation (OSGeo)
+ *    (C) 2001-2015, Open Source Geospatial Foundation (OSGeo)
```
apc3n: @aaime updated with a proper ticket and date changes

aaime:
Merged, the OSX build box failure is not an issue, that build box is proving to be unreliable...

PR #940, opened 2015-08-14
Cross layer z order control in SLD

aaime:
Split in 3 commits:
* The first is preparation in the StreamingRenderer, removing state kept in StreamingRenderer fields, support for bean collections (which has not been operational in years), and other cleanups
* The second refactors the sort/merge reader code to extract some reusable functionality
* The third is the actual cross layer implementation


jodygarnett:
I assume we can do docs when you come back, want to get the API change in for the beta.

aaime:
Hi Jody, I only see one comment about lack of docs. Is there anything else?

aaime:
So is this one good to go, under the notion that docs will be added in the next few weeks?

jodygarnett:
Sorry Andrea, juggling too many things was only 75% through the review. Additional feedback incoming.

modules/library/main/src/main/java/org/geotools/data/sort/SimpleFeatureIO.java
```diff
@@ -0,0 +1,273 @@
...
+import com.vividsolutions.jts.io.ParseException;
+import com.vividsolutions.jts.io.WKBReader;
+import com.vividsolutions.jts.io.WKBWriter;
+
+/**
+ * Allows writing and reading features to/from the given file
+ * 
+ * @author Andrea Aime - GeoSolutions
+ */
+public class SimpleFeatureIO {
```
jodygarnett: So this seems to be the main cool new feature, a random access file to dump features into.

This is public API - do you expect other datastore implementors to be making use of it? I ask so I can figure out if it is worth a tutorial, or addition to the ContentDataStore tutorial.

modules/library/render/src/main/java/org/geotools/renderer/lite/LiteFeatureTypeStyle.java
```diff
@@ -100,8 +104,9 @@ public LiteFeatureTypeStyle(BufferedImage image, AffineTransform at, List ruleLi
      * @param ruleList
      * @param elseRuleList
      */
-    public LiteFeatureTypeStyle(Graphics2D graphics, List ruleList, List elseRuleList, Expression transformation) {
-        this.myImage = null;
+    public LiteFeatureTypeStyle(Layer layer, Graphics2D graphics, List ruleList, List elseRuleList,
+            Expression transformation) {
```
jodygarnett: got a few more parameters here to javadoc

modules/library/render/src/main/java/org/geotools/renderer/lite/MarkFeatureIterator.java
```diff
@@ -0,0 +1,244 @@
...
+     * should be stopped using {@link ProgressListener#isCanceled()}
+     * 
+     * @param fc
+     * @param maxFeaturesInMemory
+     * @return
+     * @throws IOException
+     */
+    public static MarkFeatureIterator getMarkFeatureIterator(FeatureCollection fc,
+            int maxFeaturesInMemory, ProgressListener listener)
+            throws IOException {
```
jodygarnett: Factory method (switching between private implementations). Often I name factory methods with create (by convention), in this case since it is a static method:

```
MarkFactoryIterator iterator = MarkFactoryIterator.create( features, 10000, null )
```

modules/library/render/src/main/java/org/geotools/renderer/lite/SortKey.java
```diff
@@ -0,0 +1,164 @@
...
+        return "SortKey [components=" + Arrays.toString(components) + "]";
+    }
+
+    /**
+     * Builds a SortKey Comparator from a SortBy array
+     * 
+     * @param sortBy
+     * @return
+     */
+    static Comparator buildSortKeyComparator(SortBy[] sortBy) {
```
jodygarnett: Once again we can shorten up method names since static methods are a bit self-documenting:

```
SortKey.buildComparator( sortBy )
```

Since this is package visible you may not want to put in the effort.

modules/library/render/src/main/java/org/geotools/renderer/lite/StreamingRenderer.java
```diff
@@ -60,7 +59,7 @@
 import org.geotools.data.DataUtilities;
 import org.geotools.data.FeatureSource;
 import org.geotools.data.Query;
-import org.geotools.data.memory.CollectionSource;
```
jodygarnett: Goodbye CollectionSource - remind me to do a pass through the documentation in case we have any lurking references to it.

modules/library/render/src/main/java/org/geotools/renderer/lite/StreamingRenderer.java
```diff
@@ -1185,6 +1146,26 @@ Query getStyleQuery(FeatureSource<FeatureType, Feature> source,
         
         return query;
     }
+
+    private void setMetaBuffer(List<LiteFeatureTypeStyle> styleList, int metaBuffer) {
+        for (LiteFeatureTypeStyle fts : styleList) {
+            fts.metaBuffer = metaBuffer;
+        }
+
```
jodygarnett: I am not used to blank lines before a closing brace, seeing it a couple times in this patch and wanted to ask if it was on purpose.

modules/library/render/src/main/java/org/geotools/renderer/lite/StreamingRenderer.java
```diff
@@ -1193,7 +1174,7 @@ Query getStyleQuery(FeatureSource<FeatureType, Feature> source,
      * @param styles
      * @return
      */
-    private SortBy[] getSortByFromLiteStyles(LiteFeatureTypeStyle[] styles) {
+    private SortBy[] getSortByFromLiteStyles(List<LiteFeatureTypeStyle> styles) {
         for (LiteFeatureTypeStyle fts : styles) {
             if (fts.sortBy != null) {
                 return fts.sortBy;
```
jodygarnett: I may be missing the logic here, are we just returning the value from the first FeatureTypestyle to have a sortBy (like do we have a danger of missing information if two FeatureTypeStyles contain sortBy information.

If this is intensional I expect method name, or javadoc can clarify.

modules/library/render/src/main/java/org/geotools/renderer/lite/StreamingRenderer.java
```diff
@@ -1996,114 +1892,68 @@ public int getMaxBackBufferMemory(int width, int height) {
...
-            CoordinateReferenceSystem destinationCrs, Envelope mapArea,
-            Rectangle screenSize, String layerId) throws Exception {
-        /*
-         * DJB: changed this a wee bit so that it now does the layer query AFTER
-         * it has evaluated the rules for scale inclusion. This makes it so that
-         * geometry columns (and other columns) will not be queried unless they
-         * are actually going to be required. see geos-469
-         */
+    private void processStylers(final Graphics2D graphics, final Layer layer, String layerId)
+            throws Exception {
```
jodygarnett: javadocs no longer match the method signature. Yeah I know it is a private method but it is an important one.

```
      * @param graphics Target graphics for rendering
      * @param layer Layer styled
      * @param layerId Handle used to identify the layer
```

modules/library/render/src/main/java/org/geotools/renderer/lite/StreamingRenderer.java
```diff
@@ -1996,114 +1892,68 @@ public int getMaxBackBufferMemory(int width, int height) {
...
-        final CoordinateReferenceSystem sourceCrs;
-        final NumberRange scaleRange = NumberRange.create(scaleDenominator,scaleDenominator);
-        final ArrayList<LiteFeatureTypeStyle> lfts ;
-
-        if ( featureSource != null ) {
-
-            final FeatureType schema = featureSource.getSchema();
+        final FeatureSource featureSource = layer.getFeatureSource();
+        if (featureSource == null) {
+            throw new IllegalArgumentException("The layer does not contain a feature source");
```
jodygarnett: Does it help to identify the layerId in this exception?

modules/library/main/src/main/java/org/geotools/data/sort/SimpleFeatureIO.java
aaime: It is public API out of necessity, since there is two users in two different packages for it. Wish we could mark classes as not exported outside of org.geotools, which might be coming in JDK 9.
I would not advertise it much right now.

modules/library/render/src/main/java/org/geotools/renderer/lite/StreamingRenderer.java
aaime: Does not look like it's documented anywhere:

````
aaime@hydra ~/devel/git-gt/docs $ find . -name "*.rst" | xargs grep CollectionSource
aaime@hydra ~/devel/git-gt/docs $ find . -name "*.java" | xargs grep CollectionSource
`````

aaime: But... I did not remove CollectionSource. Do you want to do it?

modules/library/render/src/main/java/org/geotools/renderer/lite/StreamingRenderer.java
aaime: By accident

modules/library/render/src/main/java/org/geotools/renderer/lite/StreamingRenderer.java
aaime: The javadoc already states:

"Returns the sort-by from the list of feature type styles for a given layer. The code assumes the styles have already been classified and are uniform in sorting clauses".

So no changes needed

modules/library/render/src/main/java/org/geotools/renderer/lite/StreamingRenderer.java
aaime: The layerId is just a counter, used to generate unique ids for the LabelCache. Going to use the layer title instead.

PR #943, opened 2015-08-20
GEOT-3776: Make dasharray property support expressions

nmco:
This pull request is related to proposal https://github.com/geotools/geotools/wiki/stroke-dasharray-as-an-expression.


aaime:
All modified files headers should have "beginYear - 2015", e.g., Stroke.java should switch from "2002-2008" to "2002-2015", and so on for all others

modules/library/api/src/main/java/org/geotools/styling/Stroke.java
```diff
@@ -246,8 +256,20 @@ public Graphic getGraphicStroke() {
     void setLineCap(Expression lineCap);
 
     /**
-     * This parameter encodes the dash pattern as a seqeuence of floats.<br>
-     * The first number gives the length in pixels of the dash to draw, the
+     * Shortcut to retrieve dash array in the case where all expressions
+     * are literal numbers. Return the default value if one of the expressions
+     * is not a literal.
+     */
+    float[] getDashArray();
```
aaime: I would just deprecate this one and ask people to switch to the Expression based version

modules/library/api/src/main/java/org/geotools/styling/Stroke.java
```diff
@@ -246,8 +256,20 @@ public Graphic getGraphicStroke() {
...
+
+    /**
+     * Shortcut to define dash array using literal numbers.
+     */
+    void setDashArray(float[] dashArray);
```
aaime: Same here

modules/library/main/src/main/java/org/geotools/styling/visitor/DpiRescaleStyleVisitor.java
```diff
@@ -62,15 +62,6 @@ protected Expression rescale(Expression expr) {
         return RescalingMode.Pixels.rescaleToExpression(scale, v);
     }
     
-    @Override
```
aaime: Shouldn't we have a method working on the list of expressions, with the same logic here?
If the build is not failing, I guess I'm either missing something, or we lack a test covering this functionality.

aaime:
There is a lot of unrelated whitespace changes. I was tempted to let it go and review with a diff tool that can ignored whitespace changes, but thinking about it a bit more, all these changes will make it a nightmare backport changes to the same classes in the year to come. Please amend the pull request and remove _all_ pure whitespace changes so that we can backport bugfixes in the year to come without pulling hairs.

modules/library/main/src/main/java/org/geotools/styling/SLDParser.java
```diff
@@ -2019,6 +1997,56 @@ protected Stroke parseStroke(Node root) {
...
+            }
+        }
+        return expressions;
+    }
+
+    private void handleDashArrayText(String text, List<Expression> expressions) {
+        if (text == null || text.isEmpty()) {
+            return;
+        }
+        for (String textPart : text.split("\\s+")) {
```
aaime: Thank you, this is a more robust and cleaner way to tokenize a string separated than spaces, compared to the original code

modules/library/main/src/main/java/org/geotools/styling/SLDTransformer.java
```diff
@@ -375,22 +368,20 @@ public void visit(Stroke stroke) {
...
-                encodeCssParam("stroke-dasharray", ff.literal(sb.toString()));
+            end("Stroke");
+        }
 
+        private void encodeStrokeDasharray( List<Expression> expressions) {
+            if (expressions == null) return;
+            AttributesImpl attributes = new AttributesImpl();
+            attributes.addAttribute("", "name", "name", "", "stroke-dasharray");
+            start("CssParameter", attributes);
+            for (Expression expression : expressions) {
```
aaime: This one seems to be problematic. If one starts with a SLD standard dasharray as text, like "10 10" the encoder should return the same, and roll the mixed mode expressions only if the expressions are not literals.
The idea is that if we get a OGC compliant SLD in, it should be compliant when we rewrite it, too

aaime: Btw, you should add tests for this case

modules/library/main/src/test/java/org/geotools/styling/SLDStyleTest.java
```diff
@@ -193,11 +196,27 @@ public void testDashArray2() throws Exception {
...
-        validateDashArrayStyle(sld);
+
+        Stroke stroke = validateDashArrayStyle(sld);
+        assertTrue(Arrays.equals(new float[] {2.0f, 1.0f, 4.0f, 1.0f}, stroke.getDashArray()));
     }
 
-    private void validateDashArrayStyle(StyledLayerDescriptor sld) {
+    public void testDashArray3() throws Exception {
+        // using expressions in the dasharray
+        java.net.URL surl = TestData.getResource(this, "dasharray3.sld");
```
aaime: Nice, good example with all the cases

aaime:
There is no rendering test, you should add at least one. You could for example extend LineTest.java

aaime:
Please fix as suggested asap, if this does not get in by Thursday, it's going to be delayed to the next release

nmco:
start working on this thanks for the warning

modules/library/main/src/main/java/org/geotools/styling/visitor/DpiRescaleStyleVisitor.java
nmco: It will use the rescale method inherited from the RescaleStyleVisitor class.

modules/library/main/src/main/java/org/geotools/styling/SLDTransformer.java
nmco: good catch ...

nmco:
I have fixed the suggest issues. Thanks for review an sorry for the white-spaces :)

aaime:
The white spaces might be a blocker... I'm not happy about them, but want to hear other developers.
Devs, are you ok with getting conflict every time you backport some commit involving these files on the 13.x branch, for the next six months?

aaime:
Actually wait... did you also fix the white space changes? By your comment it seemed like you did not.

nmco:
Yes I fixed the white spaces.

nmco:
Originally they were not intentional, I just realized too late that my editor was removing the trailing spaces of edited files. I was just apologising for  the white spaces in the first commit.

aaime:
Hi Nuno, all good... but I don't see a OSGEO contribution agreement from you or your company... could you sign and send one ASAP (add in in CC please, so that I can see it directly without waiting for the OSGeo officer to update the contributors database)? 
The feature freeze is tomorrow, if we don't have the paperwork in time I cannot merge this, and it will become part of GeoTools 15.x instead of 14.x

nmco:
Done.

aaime:
Got the mail with the license agreement (with OSGeo cc'ed), merged

nmco:
Thanks.

PR #944, opened 2015-08-22
Docs update

ianturton:
bumped the GeoTools version from 12-r1 to 13.2 and removed instructions about jars that no longer ship in the eclipse (no maven) instructions. 
Added a call to preDispose to the dispose method of MapContent to get rid of an annoying warning in Layer that predispose hadn't been called.
Tested both the Eclipse and Maven quickstarts


jodygarnett:
Thanks @ianturton - wonder if it is worth making a seperate page for the "no maven" instructions so it is easier to find for those troubled souls with and aversion to maven.

PR #945, opened 2015-08-23
GEOT-5191: Use GeographicLib-Java for geodesic calculations.

cffk:
The affected classes are GeodeticCalculator and DefaultEllipsoid.  Some
of the tests were wrong and these needed to be fixed.  Now you can
calculate the distance between Wellington, 41.32S 174.81E, and
Salamanca, 40.96N 5.50W.

The latitude must lie in [-90,90] as before.  But now, any finite
longitude, azimuth, and distance is accepted by the GeodeticCalculator.
The calculated longitude and azimuth is in the range [-180,180), except
that the longitudes returned by getGeodeticPath are "unrolled" and so
vary continuously (this is governed by the GeodesicMask.LONG_UNROLL
flag).

### DIFF
--- a/modules/library/referencing/src/main/java/org/geotools/referencing/GeodeticCalculator.java
+++ b/modules/library/referencing/src/main/java/org/geotools/referencing/GeodeticCalculator.java
@@ -13,15 +13,8 @@
  *    but WITHOUT ANY WARRANTY; without even the implied warranty of
  *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
  *    Lesser General Public License for more details.
- *
- *    Portions of this file is adapted from Fortran code provided by NOAA.
- *    Programmed for CDC-6600 by LCDR L.Pfeifer NGS ROCKVILLE MD 18FEB75
- *    Modified for IBM SYSTEM 360 by John G.Gergen NGS ROCKVILLE MD 7507
- *    Source: ftp://ftp.ngs.noaa.gov/pub/pcsoft/for_inv.3d/source/
  */
 package org.geotools.referencing;
-
-import static java.lang.Math.*;
 
 import java.awt.Shape;
 import java.awt.geom.GeneralPath;
@@ -64,7 +57,7 @@ import org.opengis.referencing.datum.Datum;
 import org.opengis.referencing.datum.Ellipsoid;
 import org.opengis.referencing.datum.GeodeticDatum;
 import org.opengis.referencing.operation.TransformException;
-
+import net.sf.geographiclib.*;
 
 /**
  * Performs geodetic calculations on an {@linkplain Ellipsoid ellipsoid}. This class encapsulates
@@ -75,7 +68,7 @@ import org.opengis.referencing.operation.TransformException;
  *   <li>Point located at a given distance and azimuth from an other point.</li>
  * </ul>
  * <p>
- * The calculation use the following informations:
+ * The calculation uses the following information:
  * <p>
  * <ul>
  *   <li>The {@linkplain #setStartingPosition starting position}, which is always considered valid.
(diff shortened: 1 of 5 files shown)


cffk:
The build failure seems to be unrelated to the patch I made.  Can someone check?

aaime:
Yes, it's clearly not related. The new OSX build boxes from Travis are not being exactly super-reliable...

ianturton:
@cffk I've pushed another test on to your repo - also have you signed the code contributor licence (http://docs.geotools.org/latest/developer/roles/contributor.html)?

cffk:
I just sent in an osgeo contribution license to info@osgeo.org.

For your information the failure cases you grabbed from Wikipedia apply
to the latest NGS geodesic program

  http://www.ngs.noaa.gov/TOOLS/Inv_Fwd/Inv_Fwd.html

which includes Vincenty's "fix" for nearly antipodal points (and show
that the fix doesn't always work).  The existing geotools implementation
of Vincenty does not include this fix and thus fails with a much bigger
set of inputs.

aaime:
The copyright headers of the files modified should be updated to have 2015 as the second year, e.g., "2004-2008" -> "2004-2015"
Ian, do you have time to add the extra tests?

jodygarnett:
Was able to fix headers and grab Ian's additional test. The only odd thing was restoring the following:

 ````
 *    Portions of this file is adapted from Fortran code provided by NOAA.
 *    Programmed for CDC-6600 by LCDR L.Pfeifer NGS ROCKVILLE MD 18FEB75
 *    Modified for IBM SYSTEM 360 by John G.Gergen NGS ROCKVILLE MD 7507
 *    Source: ftp://ftp.ngs.noaa.gov/pub/pcsoft/for_inv.3d/source/
 ````

I was unsure why this was removed?

modules/library/referencing/src/main/java/org/geotools/referencing/GeodeticCalculator.java
```diff
@@ -13,16 +13,9 @@
  *    but WITHOUT ANY WARRANTY; without even the implied warranty of
  *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
  *    Lesser General Public License for more details.
- *
- *    Portions of this file is adapted from Fortran code provided by NOAA.
- *    Programmed for CDC-6600 by LCDR L.Pfeifer NGS ROCKVILLE MD 18FEB75
- *    Modified for IBM SYSTEM 360 by John G.Gergen NGS ROCKVILLE MD 7507
- *    Source: ftp://ftp.ngs.noaa.gov/pub/pcsoft/for_inv.3d/source/
```
jodygarnett: Just wanted to ask why this bit was removed?

cffk: The code that this refers to was removed.

jodygarnett:
Just to confirm, this pull requests add a dependency (net.sf.geographiclib) that is available under an MIT license.

While this pull request fixes a problem, it does so with a hammer (making use of another library). Is it possible to fix the problem in GeoTools (or retire code that is no longer used).

It may also to just fork over a couple classes or calculations rather than test on an entire dependency.

cffk:
GeographicLib-Java just does geodesic calculations for an ellipsoid of
revolution.  So it comes with hardly any "baggage".  Possibly you could
skip the class which calculates the area of geodesic polygons.
However, that's something that geotools might need to offer in the
future.

ianturton:
I'd rather add the dependency on a new library that has been written and maintained by someone who knows what they are doing than us continuing to rely on unmaintained code in the middle of gt-core that, to be honest, none of the remaining contributors understand.

aaime:
Ian, Jody has merged it already, just did it manually  :-)

PR #946, opened 2015-08-23
[GEOT-5194] types defined in imported schemas not found

ridethepenguin:
Forces parsing of imported schemas (if any) when importing schema has no element / type declaration.

See [related JIRA ticket](https://osgeo-org.atlassian.net/browse/GEOT-5194) for a description of the issue fixed by this PR.

### DIFF
--- a/modules/extension/xsd/xsd-core/src/test/java/org/geotools/xml/SchemasTest.java
+++ b/modules/extension/xsd/xsd-core/src/test/java/org/geotools/xml/SchemasTest.java
@@ -2,7 +2,7 @@
  *    GeoTools - The Open Source Java GIS Toolkit
  *    http://geotools.org
  *
- *    (C) 2007-2008, Open Source Geospatial Foundation (OSGeo)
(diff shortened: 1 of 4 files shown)


modules/extension/xsd/xsd-core/src/test/java/org/geotools/xml/SchemasTest.java
```diff
@@ -95,7 +95,8 @@ protected void setUp() throws Exception {
                 "targetNamespace='http://geotools.org/test'> " + 
             "</xsd:schema>";
         write( f, xsd );
-            
+
+        System.setProperty(Schemas.FORCE_SCHEMA_IMPORT, "false");
```
bencaradocdavies: @ridethepenguin please also add this System.setProperty statement to tearDownl test method order is undefined and, if testForcedSchemaImport runs last, the system property setting would leak into other tests because one JVM is used to run all tests in this module.

ridethepenguin: @bencaradocdavies right, now it's there.

PR #947, opened 2015-08-25
Meteosat Second Generation imagery projection

mfilocha:
Conversion of image coordinates (pixel column and row) into the corresponding geographical coordinates of all Meteosat Second Generation (MSG) Satellites (Meteosat-8, Meteosat-9 and Meteosat-10) Level 1.5 VIS/IR data.

Example usage provided in annotated code.

### DIFF
--- a/modules/library/referencing/src/main/java/org/geotools/referencing/operation/projection/MeteosatSG.java
+++ b/modules/library/referencing/src/main/java/org/geotools/referencing/operation/projection/MeteosatSG.java
@@ -1,0 +1,423 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ *
+ *    (C) 1999-2015, Open Source Geospatial Foundation (OSGeo)
+ *
+ *    This library is free software; you can redistribute it and/or
+ *    modify it under the terms of the GNU Lesser General Public
+ *    License as published by the Free Software Foundation;
+ *    version 2.1 of the License.
+ *
+ *    This library is distributed in the hope that it will be useful,
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ *    Lesser General Public License for more details.
+ *
+ *    This package contains formulas from the PROJ package of USGS.
+ *    USGS's work is fully acknowledged here. This derived work has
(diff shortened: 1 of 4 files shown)


modules/library/referencing/src/main/java/org/geotools/referencing/operation/projection/MeteosatSG.java
```diff
@@ -0,0 +1,414 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ *
+ *    (C) 1999-2015, Open Source Geospatial Foundation (OSGeo)
```
aaime: This should just be 2015

modules/library/referencing/src/test/resources/org/geotools/referencing/test-data/scripts/MeteosatSG.txt
```diff
@@ -0,0 +1,69 @@
...
+//print crs
+
+// col, row
+// lon, lat
+
+source pt = (0,0)
+target pt = (1856, 1856)
+//print pts
+
+source pt = (4.0244517481909288,44.701975477556189)
```
aaime: Hi, wondering, what was used to compute the reference target values?

aaime:
Looks good minus one detail, can you fix it?

modules/library/referencing/src/main/java/org/geotools/referencing/operation/projection/MeteosatSG.java
mfilocha: Fixed.

aaime:
Nice, will wait for the Travis build checks, just to make sure, and then merge

modules/library/referencing/src/test/resources/org/geotools/referencing/test-data/scripts/MeteosatSG.txt
mfilocha: As stated in lines 5-10 of the MeteosatSG.txt file:

The following test points were calculated with the reference software
"MSG_navigation" v1.02 provided by EUMETSAT on their "Support Software & 
Tools" web page:
http://www.eumetsat.int/website/home/Data/DataDelivery/SupportSoftwareandTools/index.html
//
Example (for non-HRV image coordinates):
  ./MSG_navigation_v1.02 1116 3062 0

aaime:
Merging, the Travis OSX build failure is an unrelated issue

PR #949, opened 2015-08-25
ProjectiveTransform2D.inverse() throws class cast exception.

highsource:
This is a fix for https://osgeo-org.atlassian.net/browse/GEOT-5198.

As the test case demonstrates, inversing `ProjectiveTransform2D` throws a class cast exception. Subclasses should use the provided `createInverse(...)` method when creating the inverse transformation.

(I've signed and submitted the contributor agreement/license today.)

### DIFF
--- a/modules/library/referencing/src/main/java/org/geotools/referencing/operation/transform/GeocentricTranslation.java
+++ b/modules/library/referencing/src/main/java/org/geotools/referencing/operation/transform/GeocentricTranslation.java
@@ -17,10 +17,18 @@
 package org.geotools.referencing.operation.transform;
 
 import java.util.Collections;
-
-import javax.measure.unit.NonSI;
 import javax.measure.unit.SI;
+import javax.measure.unit.NonSI;
 
+import org.opengis.parameter.ParameterDescriptor;
+import org.opengis.parameter.ParameterDescriptorGroup;
+import org.opengis.parameter.ParameterNotFoundException;
+import org.opengis.parameter.ParameterValueGroup;
+import org.opengis.referencing.operation.Matrix;
+import org.opengis.referencing.operation.MathTransform;
+import org.opengis.referencing.operation.Transformation;
+
+import org.geotools.util.Utilities;
 import org.geotools.measure.Units;
 import org.geotools.metadata.iso.citation.Citations;
 import org.geotools.parameter.DefaultParameterDescriptor;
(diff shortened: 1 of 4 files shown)


modules/library/referencing/src/main/java/org/geotools/referencing/operation/transform/GeocentricTranslation.java
```diff
@@ -154,7 +154,7 @@ public ParameterValueGroup getParameterValues() {
      * Creates an inverse transform using the specified matrix.
      */
     @Override
-    MathTransform createInverse(final Matrix matrix) {
+    GeocentricTranslation createInverse(final Matrix matrix) {
```
jodygarnett: I am a little uncomfortable changing the return type in a public API. 

Is GeocentricTranslation a kind of MathTransform?

modules/library/referencing/src/test/java/org/geotools/referencing/operation/transform/ProjectiveTransform2DTest.java
```diff
@@ -0,0 +1,45 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ *
+ *    (C) 2001-2015, Open Source Geospatial Foundation (OSGeo)
```
jodygarnett: As this is a new file the header can be simply (C) 2015.

BTW: Thanks for sending in the code contribution agreement - makes this easier.

jodygarnett:
This looks nice and self contained, was worried it would turn out to be a side effect of transition to EJML.

I am going to try and merge locally / squash the commit history / see if we can make the fix without changing public API.

highsource:
Also possible without changing the public API but would require to cast the `MathTransform` to `ProjectiveTransform`.
Or makeing `inverse` to `MathTransform` instead and then something like:

```
if (inverse instanceOf ProjectiveTransform)
{
   ((ProjectiveTransfrom) inverse).inverse = this;
}
```

I was considering all these options but thought that changing method signatures would be better.

modules/library/referencing/src/main/java/org/geotools/referencing/operation/transform/GeocentricTranslation.java
aaime: Yes, it is, no harm done here

aaime:
Looks good to me, besides the wrong copyright header. I believe we can merge once that is fixed (squash the commit and push using --force to update the pull request)

modules/library/referencing/src/main/java/org/geotools/referencing/operation/transform/GeocentricTranslation.java
jodygarnett: And I see this  is only package visible, rather than public.

Building locally and then I will merge/commit.

highsource:
Thanks for the fast reaction. :)

PR #951, opened 2015-08-26
GEOS-7152: wfs-ng geoscript compatibility

NielsCharlier:
* Replace xpp3 dependency and replace by xpp3_min (and stop using XmlPullParserFactory).
xpp3 is unfortunately not compatibility with groovy! (LinkageError)

* implement getCapabilitiesURL so that geoscript-groovy can be ported without code changes


jodygarnett:
Related question @NielsCharlier - is wgs-ng ready to be promoted to supported status?

NielsCharlier:
Yes, as far as I am concerned

PR #959, opened 2015-09-11
unhandled exceptions

waooog:
Two unhandled exceptions are reported by our automatic tool, details are as follows:

In method dispose() of class AbstractGridCoverageWriter ( geotools-master\modules\library\coverage\src\main\java\org\geotools\coverage\grid\io\AbstractGridCoverageWriter.java)
The catch block caught an IOException, while without further actions, to fix this, we need add more code to handle this.
             dispose(){
              ...
              try {
		outStream.flush();			
	} catch (IOException e) {
         ...
}

Another case is in method getQueryInfo() of class ArcSDEQuery (geotools-master\modules\plugin\arcsde\datastore\src\main\java\org\geotools\arcsde\data\ArcSDEQuery.java)

The catch block caught an NullPointerException, while without further actions, to fix this, we need add more code to handle this.
getQueryInfo(){
...
try {
     byClause = definitionQuery.getByClause();
 } catch (NullPointerException e) {
     // no-op 
  }
...
}
These two cases are reported by our automatic tool, they might be two unhandled exceptions, I report them, could some of you double check them? Thanks.


aaime:
Please redo the pull request so that it contains only your commit, and not 250+ others :)

waooog:
sure, I will create a new one, thanks. ;)

waooog:
Hi aaime, 
I just found that I have not modified related code snippets to create new branchs, so a new pull request will also wrap a lot of other commits, and  I reopen this request again. 
These odd/buggy code snippets are reported by my on-going automatic code analysis tool, your responses to the results definitely will help me improve the tool :). Thanks.

aaime:
Closing, still contains a lot of unrelated commits. Please try to make a pull request with a single change and a single commit.

PR #960, opened 2015-09-12
MemoryDataStore Thread safety

fgdrf:
If a Reader is iterating through the collection and a Writer (e.g. from Datastream) uses a Transaction to write SimpleFeatures the reader can get ConcurrentModificationException if not finished meanwhile. That can break rendering (BasicFeatureRenderer in uDig).

Implications : The reader might work on a slightly out-of-date DataSet, depending on the amount of data has been written in between.


aaime:
Looks ok to me, but would like @jodygarnett to double check it

aaime:
Hi @jodygarnett , are you good with these changes? :-)

PR #964, opened 2015-09-19
Remove JAI workaround for floating point warp transform when using JAI-Ext

tkunicki:


simboss:
I did check quickly but it looks like there is a compilation error with this change. 

Can you double check the travis builds failing?

simboss:
I will close this PR and will open a different one.

PR #974, opened 2015-09-26
[GEOT-5228] Correct LGPL -> GPL guidance

bradh:


docs/user/welcome/faq.rst
```diff
@@ -205,10 +205,9 @@ project so the changes can be incorporated into the core source code.
 
 .. note::
 
-   This means you can use GeoTools as a library but you cannot
-   incorporate GeoTools code directly into your GLP application. Legally, the
-   latter amounts to re-licensing GeoTools under a new license and you do not have
-   the right to do so.
+   You can also incorporate GeoTools code directly into your GPL application. Legally, the
+   latter amounts to re-licensing GeoTools under the GPL, which is specifically allowed
+   by the LGPL. This re-licensing is one-way, and requires specific actions - see the LGPL.
```
tbarsballe: We should clarify that this is for cutting and pasting code into your GPL app - if you do this the license terms allow it and the result is GPL.

PR #978, opened 2015-09-27
[GEOT-5216] Labels are not getting painted when direct layers are inc…

ianturton:
…luded in the map content
This fixes an issue with the previous fix.


aaime:
Which means the previous fix test was weak. Can you think of ways to make it intercept this case too?

aaime:
Cough cough :)

PR #970, opened 2015-09-24
[GEOT-5224] ArcGrid extension only supports ASCII GRID

tbarsballe:
For [GEOT-5224](https://osgeo-org.atlassian.net/browse/GEOT-5224).
Update the ArcGrid format description and documentation to indicate that only the ARC/INFO ASCII GRID format is supported (not the binary format).

PR #979, opened 2015-09-28
[GEOT-5224] ArcGrid extension only supports ASCII GRID

tbarsballe:
Fixed #970 to support WCS 1.0 DescribeCoverage response (changed name back to ArcGrid, ASCII limitation mentioned in description)


aaime:
Looks good now. Backport as you see fit (you do have direct commit access on geotools, right?)

tbarsballe:
Yes, I have commit access. Backported to 14.x and 13.x.

PR #988, opened 2015-10-07
test property files are now created within target directory

wumpz:
Compiling module gt-property actual version 15-SNAPSHOT does generate files and directories within the project directory. I relocated these files into the target directory. Now git as well does not find changes anymore.

### DIFF
--- a/modules/plugin/property/src/test/java/org/geotools/data/property/PropertyDataStore2Test.java
+++ b/modules/plugin/property/src/test/java/org/geotools/data/property/PropertyDataStore2Test.java
@@ -16,21 +16,16 @@
  */
 package org.geotools.data.property;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertTrue;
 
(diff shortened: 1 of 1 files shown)


aaime:
I'm not getting any local change regardless... were the tests failing to complete and cleanup in your case? That said, I agree it's bad practice to write test files in the project root.

wumpz:
The local changes were the new created files within the project root. So I described it not clear enough. Sorry.

-------- Ursprüngliche Nachricht --------
Von: Andrea Aime <notifications@github.com> 
Datum: 11.10.2015  09:47  (GMT+01:00) 
An: geotools/geotools <geotools@noreply.github.com> 
Cc: Tobias <t.warneke@gmx.net> 
Betreff: Re: [geotools] test property files are now created within target
  directory (#988) 

I'm not getting any local change regardless... were the tests failing to complete and cleanup in your case? That said, I agree it's bad practice to write test files in the project root.

PR #989, opened 2015-10-07
CSS Additions

wumpz:
Add the ability to set style title and style abstract within a css file (directives **styleTitle** and **styleAbstract**).
Implemented translation mode **Flat** to avoid rule combining ala CSS to get more control over the generated style and therefore the generated legend.

### DIFF
--- a/modules/unsupported/css/src/main/java/org/geotools/styling/css/CssTranslator.java
+++ b/modules/unsupported/css/src/main/java/org/geotools/styling/css/CssTranslator.java
@@ -124,6 +124,10 @@ public class CssTranslator {
     static final String DIRECTIVE_AUTO_THRESHOLD = "autoThreshold";
 
     static final String DIRECTIVE_TRANSLATION_MODE = "mode";
+    
+    static final String DIRECTIVE_STYLE_TITLE = "styleTitle";
+    
+    static final String DIRECTIVE_STYLE_ABSTRACT = "styleAbstract";
 
     static final int MAX_OUTPUT_RULES_DEFAULT = Integer
             .valueOf(System.getProperty("org.geotools.css." + DIRECTIVE_MAX_OUTPUT_RULES, "10000"));
@@ -285,6 +289,8 @@ public class CssTranslator {
         // prepare the full SLD builder
         StyleBuilder styleBuilder = new StyleBuilder();
         styleBuilder.name("Default Styler");
+        styleBuilder.title(stylesheet.getDirectiveValue(DIRECTIVE_STYLE_TITLE));
+        styleBuilder.styleAbstract(stylesheet.getDirectiveValue(DIRECTIVE_STYLE_ABSTRACT));
 
         // split rules by index and typename, then build the power set for each group and
         // generate the rules and symbolizers
--- a/modules/unsupported/css/src/main/java/org/geotools/styling/css/CssTranslator.java
+++ b/modules/unsupported/css/src/main/java/org/geotools/styling/css/CssTranslator.java
@@ -362,12 +362,24 @@ public class CssTranslator {
                             return super.buildResult(rules);
                         }
 
+                    private boolean haveSameSelector(List<CssRule> rules) {
+                        Selector selector = null;
+                        for (CssRule rule : rules) {
+                            if (selector == null) {
+                                selector = rule.selector;
+                            } else if (!selector.equals(rule.selector)) {
+                                return false;
+                            }
+                        }
+                        return true;
+                    }
+
                     @Override
                     protected boolean accept(List<CssRule> rules) {
-                        if (mode == TranslationMode.Flat && rules.size() > 1) {
+                        if (mode == TranslationMode.Flat && rules.size() > 1 && !haveSameSelector(rules)) {
                             return false;
                         }
-                        return super.accept(rules); 
+                        return super.accept(rules);
                     }
 
                     @Override
--- a/modules/unsupported/css/src/main/java/org/geotools/styling/css/CssTranslator.java
+++ b/modules/unsupported/css/src/main/java/org/geotools/styling/css/CssTranslator.java
@@ -77,6 +77,7 @@ import org.geotools.styling.css.selector.Selector;
 import org.geotools.styling.css.selector.TypeName;
 import org.geotools.styling.css.util.FeatureTypeGuesser;
 import org.geotools.styling.css.util.OgcFilterBuilder;
+import org.geotools.styling.css.util.PseudoClassRemover;
 import org.geotools.styling.css.util.ScaleRangeExtractor;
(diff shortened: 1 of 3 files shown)


aaime:
Hi, don't have time to review right now but I'm wondering, did you check that ":mark" pseudo-rules are still working as intended? By themselves they do nothing, they need to be mixed with other rules.

wumpz:
Good hint. I did not look for pseudo rules. Let me build a test to verify or correct my patch.

wumpz:
I modified my pull request to handle pseudorules as hopefully intended and added specific tests. So I hope I did use your methods in the right way. 

The main difference between normal mode and flat is that the domain rules (I think thats the name you used) are not combined. So you were so kind ;) to use the FilteredPowerSetBuilder, which made it easy to block the combination of rules simply by using the already provided accept (overridden within CssTranslator) method. The pseudorules were mixed in by using another instance of RulePowerSetBuilder, which is not "flat" anymore (no overridden accept method). Hence it should work, shouldn't it?

Tobias

aaime:
Hmm... unsure... in a flat mode, where no cascading is allowed, I'd think the reasonable approach would be to merge the pseudo-selector rules only if they are exactly following the current rule.
The implementation you're providing instead does mixes in a hard to predict order, and won't work if there are multiple pseudo matchers.

So, ideally, get the first non pseudo-matcher rule, see if there is any pseudo matcher following it, if there are, mix all the ones that you can find before the next true rule, then move to the next true rule, and move on, rinse and repeat

wumpz:
Could you provide an example for multiple pseudo matchers. I thought I would follow the path you described. I don't unterstand the problem.

wumpz:
Now I understand your problem (and mine). After building some pseudorule tests I see, what you mean with "unpredictable". I changed the patch to mixin only pseudo rules if the filter is exactly the same. So I introduces a filter possibility within `RulePowerSetBuilder`.

aaime:
Hi, I'd break this pull request in two, the title support is fine, no issues, can go in right now.
The flat mode is still problematic though: it should really not be implemented using the PowerSetBuilder.
The issue is that the PSB will try all possible combinations, in whatever order, and "flat" is anything but that. To make an example, what if the user is trying to specify multiple rules with the same filter, but making sure a different symbolizer is painted in the middle of the two? The PSB will just ignore that and treat the two rules with the same filters as overrides/additions.

wumpz:
IMHO at the moment I am guessing that the actual implementation is a huge step forward in useability. Isn't the mode 'flat' some theoretical beast, to get 1:1 to sld rules, that could never or only very hard reached? The implementation with your additions were what I had in mind to smooth my workflow. Do you have a problem with this implementation, with this mode or only with its naming? 
Nevertheless I will try to construct an example of your problem case. Or could you give me a simple CSS style of that? 
I sense that my unterstanding of how cascading, inheritance, ordering and selection for this style building is somehow incomplete.

aaime:
The name is fine, the implementation is not (as explained above).
I'll try to come up with an example as spare time allows.

wumpz:
So this is my example. Hope that is what you meant:
~~~
@mode "Flat";
[value1=1] { stroke: green; stroke-width:2px;}
[value2=2] { stroke: green; stroke-width:2px;}
[value1=1] { stroke: blue; stroke-width:10px;}
~~~

Therefore I assume we have a different view of what 'Flat' mode really means. To me it is a strict subset of PSBs result. For you its not. PSB will reduce first and third rule to third. Third overwrites first rule completely.
First difference of Flat mode is that its using the first rule and never the third. That is corrected within the next patch, to follow my understanding of flat mode, but not part of our problem.

To achieve what you want, one has to specify a z-index. (each rule a different one). After that normal mode and flat mode deliver the one to one relation. Isn't that enough.

So it burns down to:
"flat" you:
~~~
   for (rule : rules) 
              PSB(rule);
~~~
"flat" mine:
~~~
  PSB(rules) {
       accept(subrules) {
             return same_selector
       }
  }
~~~ 
Right?

So you have the flatter ;) mode. I merge indeed rules with the same selector. So maybe I call my flat mode "flat filter mode" and yours become the "real flat" ;) mode.

aaime:
Yes, we have different visions on that. What you describe seems a "partially cascaded" mode which I have a very strong premonition won't satisfy anybody.
Those that need to do complex styles in a compact way should use the normal CSS cascading rules (several guides online), those who just want SLD with a lighter syntax should use "flat" mode, which should be completely and absolutely flat. I'm not going to support a compromise in the middle, but if you want you can create your own css-partially-cascaded module and maintain it ;-)

wumpz:
So if I switch to your mode regarding the "main" rules and process pseudorules like my implementation does at this time, your fine? So to say I would really like you to merge this (or a modified version of this) flat mode into main :). So what else to make you happy with it?
You are right in naming my implementation a "partially cascaded" version. My version merges same selectors but does not generate a superset of all available selectors.

aaime:
A flat mode should be ... flat :-)
That is, the order of the CSS rules is maintained, there is no notion of priority, only the rules with pseudo-matchers are merged with the full rules preceding them. One can literally read it as a simplified SLD.

This requires a completely different merging algorithm, not a variation of the power set builder, but one that decides if a rule has a pseudo-selector or not, if not, it checks the following rules for pseudo matchers, if found merges them, and then moves to the next rule without pseudo-matchers. Pretty simple to implement.

The pseudo matcher thing is annoying, it would be better to have the ability to just nest those rules (LESS/SASS style) but that requires larger changes to the parser... eventually we'll get there, but probably not today ;-)

wumpz:
So you would leave z-index behind or not? Or would you order the "non pseudo" rules by it and then search from the original position for following pseudo rules? 
Within the algorithm you described you do not mentioned this "preordering".

wumpz:
Any comments to my last comment? Handling z-index seems mandatory for me.

aaime:
Hi Tobias, I'm pretty busy with work during the week, I'll try to follow up during the weekend

aaime:
Hi Tobias, z-index is indeed a source of problems. Suggestion: first merge all pseudo-matches into their corresponding rules, getting a list of rules that do no need any further merging, then order by z-order and emit in different feature type styles.

aaime:
Btw, mandatory clarification: the Project Management Committee rules GeoTools. 
But I am the CSS module maintainer, so I have to be comfortable with decisions that will affect the future of the module (since I'm the one that has to maintain it long term).

wumpz:
Added (hopefully) flat mode. I am not sure if flattening of rules is necessary in this kind of context. The implementation follows your suggestion to build up rules and then doing the z-ordering.

wumpz:
Removed "my" flat filter mode. So now there is only one flat mode. ;)

aaime:
Replaced by #989, added an integration tests, made commits refer to tickets, have one commit per ticket

PR #991, opened 2015-10-09
Fix OracleDialect.unwrap() to work when wrapped by LifeCycleConnection.

wmstoval:
See JIRA GEOT-5239. The unwrap code would succeed only for a single
level of wrapping.  If the DataStore is configured with startup-sql
the connection will be additionally wrapped by LifeCycleConnection
which makes the unwrap throw an exception as it fails to obtain
OracleConnection.

Modified OracleDialect to loop thru multiple levels of unwrapping when
necessary instead of assuming one level. Given potential multiple
levels of wrap, the code now maintains a Map of UnWrappers to preserve
optimized (minimal) access to DataSourceFinder.getUnWrapper().

Also updated the JDBCConnectionLifecycleOnlineTest to include a new
testLifeCycleDoubleUnwrap() member.  The test simply created a DataStore
with startup SQL. This alone makes the (unfixed) code fail (without
actually reading any SDO geometry) since the OracleNGDataStore
explcitly invokes OracleDialect during construction of the DataStore.

### DIFF
--- a/modules/library/jdbc/src/test/java/org/geotools/jdbc/JDBCConnectionLifecycleOnlineTest.java
+++ b/modules/library/jdbc/src/test/java/org/geotools/jdbc/JDBCConnectionLifecycleOnlineTest.java
@@ -19,11 +19,7 @@ package org.geotools.jdbc;
 import java.io.IOException;
 import java.sql.Connection;
 import java.sql.SQLException;
-import java.util.HashMap;
-import java.util.Properties;
 
-import org.geotools.data.DataStore;
-import org.geotools.data.DataStoreFinder;
 import org.geotools.data.DefaultTransaction;
 import org.geotools.data.Query;
 import org.geotools.data.Transaction;
@@ -105,23 +101,6 @@ public abstract class JDBCConnectionLifecycleOnlineTest extends JDBCTestSupport 
             // we don't actually expect an exception to percolate up since it's happening
             // on the closeSafe method, that swallows exceptions
             featureStore.getCount(Query.ALL);
-        }
-    }
-    
-    public void testLifeCycleDoubleUnwrap() {
-        try {
-            // Use startup SQL when connecting so the connection is
-            // doubly wrapped (adding LifeCycleConnection).
-            // That tests ability of OracleDialect to unwrap properly.
-            Properties addStartupSql = (Properties) fixture.clone();
-            addStartupSql.setProperty(JDBCDataStoreFactory.SQL_ON_BORROW.key,
-                    "select sysdate from dual");
-            HashMap params = createDataStoreFactoryParams();
-            params.putAll(addStartupSql);
-            DataStore withWrap = (JDBCDataStore) DataStoreFinder.getDataStore(params);
-            withWrap.dispose();
-        } catch (Exception e) {
-            throw new RuntimeException("Connection unwrap test failed!", e);
         }
     }
     
--- a/modules/library/jdbc/src/test/java/org/geotools/jdbc/JDBCConnectionLifecycleOnlineTest.java
+++ b/modules/library/jdbc/src/test/java/org/geotools/jdbc/JDBCConnectionLifecycleOnlineTest.java
@@ -2,7 +2,7 @@
  *    GeoTools - The Open Source Java GIS Toolkit
  *    http://geotools.org
  *
- *    (C) 2002-2008, Open Source Geospatial Foundation (OSGeo)
+ *    (C) 2002-2015, Open Source Geospatial Foundation (OSGeo)
  *
  *    This library is free software; you can redistribute it and/or
  *    modify it under the terms of the GNU Lesser General Public
--- a/modules/library/jdbc/src/test/java/org/geotools/jdbc/JDBCConnectionLifecycleOnlineTest.java
+++ b/modules/library/jdbc/src/test/java/org/geotools/jdbc/JDBCConnectionLifecycleOnlineTest.java
@@ -2,7 +2,7 @@
  *    GeoTools - The Open Source Java GIS Toolkit
  *    http://geotools.org
  *
- *    (C) 2002-2015, Open Source Geospatial Foundation (OSGeo)
+ *    (C) 2002-2008, Open Source Geospatial Foundation (OSGeo)
  *
  *    This library is free software; you can redistribute it and/or
  *    modify it under the terms of the GNU Lesser General Public
--- a/modules/library/jdbc/src/test/java/org/geotools/jdbc/JDBCConnectionLifecycleOnlineTest.java
+++ b/modules/library/jdbc/src/test/java/org/geotools/jdbc/JDBCConnectionLifecycleOnlineTest.java
@@ -19,7 +19,11 @@ package org.geotools.jdbc;
 import java.io.IOException;
 import java.sql.Connection;
 import java.sql.SQLException;
+import java.util.HashMap;
+import java.util.Properties;
 
+import org.geotools.data.DataStore;
+import org.geotools.data.DataStoreFinder;
 import org.geotools.data.DefaultTransaction;
 import org.geotools.data.Query;
 import org.geotools.data.Transaction;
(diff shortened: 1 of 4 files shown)


wmstoval:
Honestly, I don't know if the Java 6 unwrapping code belongs in the while loop.  I didn't fully understand that code.  I saw that control would fall there only when no Unwrapper can be found for the Connection.  I don't know when that actually happens.

But your suggestion might be right on the mark - but it also begs the question of how to force that case to happen (for testing).  I could of course force the condition, but to properly test it I'm not sure what to do.  I may need to research that more.  Do you know?

wmstoval:
I'm looking into the Java 6 code/while loop code...

wmstoval:
Actually that was a great suggestion Jody - I moved the code into the while loop and initial testing looks good and I am in fact hitting that code and seeing it unwrap successfully.  I'll push a fix up shortly.

aaime:
Hi, I have no time to check right now, but have a look at the git history in that section of the code, I remember there were some issues related to java 6 unwapper flooding the logs in some situations.

wmstoval:
Thanks aaime - I see reference to the log flooding.  Looking into it...

wmstoval:
I've reconsidered moving the Java 6 code into the while loop.  I did some testing like that, with the Java 6 code successfully being used at any stage of the unwrap if calling the Unwrapper.unwrap did not yield an Oracle connection - and that worked for me.

But having reviewed the log-flooding issue Andre referred to (rev 1566148 in my history) I no longer find that a reasonable approach.

Before the fix for log-flooding Java 6 was the first choice for an unwrap strategy.  But that fix made it the last choice, with preference going to the use of the Unwrapper interface.

The code I pushed has the same characteristics - i.e. use the Unwrapper first.  If an Unwrapper can't be found the while loop will fall thru and Java 6 still gets tried.  To try it at each iteration of the while loop would be inappropriate if we can in fact find an Unwrapper to use IMO.

wmstoval:
I see that I made the build fail.  I tested the fix but see I failed to use -Dall on the build.  My test incorrectly impacted H2 or other JDBC data sources where startup-sql may not be applicable.  I'm correcting the problem and will push a fix to my wmstoval repository shortly.

wmstoval:
I pushed new changes to wmstoval/geotools that correct junit tests.  Please excuse my lax testing.  I was not using -Dall as I should have.

wmstoval:
Based on discussion with Andrea I have some concern about the fix to CSVWriteTest because it's not clear that will work correctly under OSX.  Andrea suggests I could ask other developers with OSX access to try it out.

modules/library/jdbc/src/test/java/org/geotools/jdbc/JDBCConnectionLifecycleOnlineTest.java
```diff
@@ -2,7 +2,7 @@
  *    GeoTools - The Open Source Java GIS Toolkit
  *    http://geotools.org
  *
- *    (C) 2002-2008, Open Source Geospatial Foundation (OSGeo)
+ *    (C) 2002-2015, Open Source Geospatial Foundation (OSGeo)
```
aaime: No need to update the copyright dates anymore, there is no other change

modules/plugin/jdbc/jdbc-oracle/src/main/java/org/geotools/data/oracle/OracleDialect.java
```diff
@@ -127,7 +127,12 @@ public boolean canUnwrap(Connection conn) {
      */
     public static final String GEODETIC = "geodetic";
     
-    UnWrapper uw;
+    /**
+     * Map of <code>UnWrapper</code> objects keyed by the class of <code>Connection</code>
+     * it is an unwrapper for. This avoids the overhead of at each unwrap.
```
aaime: The overhead of... what? :-)

modules/plugin/jdbc/jdbc-oracle/src/main/java/org/geotools/data/oracle/OracleDialect.java
```diff
@@ -617,23 +622,30 @@ OracleConnection unwrapConnection( Connection cx ) throws SQLException {
...
+                    }
+                    uwMap.put(testCon.getClass(), unwrapper);
                 }
-            }
-            if (uw != null && uw != UNWRAPPER_NOT_FOUND) {
-                Connection uwcx = uw.unwrap( cx );
-                if ( uwcx != null && uwcx instanceof OracleConnection ) {
-                    return (OracleConnection) uwcx;
+                if (unwrapper == UNWRAPPER_NOT_FOUND) {
+                    // give up and try legacy approaches below
```
aaime: The java 6 unwrapper is actually the "new" approach (the unwrappers were created when we were still java 4 compatible, and we never know which interfaces the env we're running into implements)

modules/plugin/jdbc/jdbc-oracle/src/main/java/org/geotools/data/oracle/OracleDialect.java
```diff
@@ -617,23 +622,30 @@ OracleConnection unwrapConnection( Connection cx ) throws SQLException {
...
+                    break;
                 }
-            } else if (cx instanceof Wrapper) {
+                testCon = unwrapper.unwrap(testCon);
+                if (testCon instanceof OracleConnection) {
+                    return (OracleConnection) testCon;
+                }
+            } while (testCon != null);
```
aaime: I'm somewhat worried about this one, if we stumble into a badly written un-wrapper (one that in some case returns the connection its given) we might never get out of the loop.
We might want to guard against that too

aaime:
Looks good to me besides a few minor issues to address

modules/library/jdbc/src/test/java/org/geotools/jdbc/JDBCConnectionLifecycleOnlineTest.java
wmstoval: OK - I'll push a change back to the original 2002-2008 copyright.

modules/plugin/jdbc/jdbc-oracle/src/main/java/org/geotools/data/oracle/OracleDialect.java
wmstoval: Woops - should be mention of DataSourceFinder service registry...will fix.

modules/plugin/jdbc/jdbc-oracle/src/main/java/org/geotools/data/oracle/OracleDialect.java
wmstoval: Thanks, misunderstood.

modules/plugin/jdbc/jdbc-oracle/src/main/java/org/geotools/data/oracle/OracleDialect.java
wmstoval: Thanks, I'll add a check for that.

wmstoval:
I've pushed changes to wmstoval/geotools to address the review comments above.

wmstoval:
Thank you for the review – I’ve addressed each point and pushed new code for the pull request.

Walter Stovall

jodygarnett:
Thanks for the dedication @wmstoval and @aaime, working with both of you (and such a clear line of communication makes it much easier when we are split across timezones, and a holiday weekend here in Canada)

modules/unsupported/csv/src/test/java/org/geotools/data/csv/CSVWriteTest.java
```diff
@@ -307,8 +306,11 @@ public void appendContent() throws Exception {
         String contents2 = new String(baos2.toByteArray(), StandardCharsets.UTF_8);
 
         // Making sure whitespace doesn't cause problems
+        // and adjust for potential windows line endings.
         contents1 = contents1.replace(" ", "").trim();
         contents2 = contents2.replace(" ", "").trim();
+        contents1 = contents1.replace("\r", "");
+        contents2 = contents2.replace("\r", "");
```
jodygarnett: I think I would make an assertEqualsIgnoreWhitespace method, that takes care of any whitespace (not just special casing " "  and "\n") and then we can use it more generally.

Rather than hold up the show let me do that when merging.

wmstoval: That sounds reasonable to me Jody.  I appreciate your help resolving this.  Thanks.

wmstoval:
Would it be possible to get these changes merged into the 2.8.x branch of geoserver so its geotools code includes this fix?  We're making a near-term release where we need this functionality in geoserver.  Will we need to build our own geotools to do that or can this be merged soon?  Please let me know if I can help with the process in any way.  Thanks.

aaime:
Closing as it has been manually merged. Jody, do you want to do the backport? I noticed you did not squash the commits into one which makes the backport somewhat harder.

PR #994, opened 2015-10-13

ridethepenguin:
Adds support for complex types with arbitrary contents (i.e. derived from `xs:anyType` by extension). Simple content is added to a special property called `unrestrictedContent` and encoded as text.

See [related discussion](http://osgeo-org.1560.x6.nabble.com/App-schema-cannot-set-string-value-to-swe-DataArray-swe-values-element-td5228781.html) on the mailing list.


ridethepenguin:
Superseded by #1000 .

PR #997, opened 2015-10-14
[GEOT-5258] Add compatability for negative date and XDate type in JDBC datastore

travislbrundage:
This pull request pulls in the changes made by Ian for handling new date types in the jdbc-postgis plugin. Addresses [GEOT-5258](https://osgeo-org.atlassian.net/browse/GEOT-5258)

### DIFF
--- a/modules/library/api/src/main/java/org/geotools/util/Converters.java
+++ b/modules/library/api/src/main/java/org/geotools/util/Converters.java
@@ -16,9 +16,11 @@
  */
 package org.geotools.util;
 
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Set;
 import java.util.logging.Level;
 import java.util.logging.Logger;
@@ -54,6 +56,14 @@ public final class Converters {
      * Will be initialized only when first needed.
      */
     private static FactoryRegistry registry;
+    
+    static final Converter IDENTITY = new Converter() {
+
+        @Override
+        public <T> T convert(Object source, Class<T> target) throws Exception {
+            return (T) source;
+        }
+    };
 
     /**
      * Returns the service registry. The registry will be created the first
@@ -183,6 +193,32 @@ public final class Converters {
             return target.cast( source.toString() );
         }
         return null;
+    }
+    
+    public static Converter getConverter(Class source, Class target, final Hints hints) {
+        if (target.isAssignableFrom(source)) {
+            return IDENTITY;
+        }
+        Set<ConverterFactory> factories = getConverterFactories(source, target);
+        if (factories.isEmpty()) {
+            return null;
+        }
+        final Converter[] converters = new Converter[factories.size()];
+        int i = 0;
+        for (ConverterFactory f: factories) {
+            converters[i] = f.createConverter(source, target, hints);
+        }
+        return new Converter() {
+
+            @Override
+            public <T> T convert(Object source, Class<T> target) throws Exception {
+                T result = null;
+                for (int i = 0; i < converters.length && result != null; i++) {
+                    result = converters[i].convert(source, target);
+                }
+                return result;
+            }
+        };
     }
 
     /**
--- a/modules/library/jdbc/src/main/java/org/geotools/data/jdbc/FilterToSQL.java
+++ b/modules/library/jdbc/src/main/java/org/geotools/data/jdbc/FilterToSQL.java
@@ -109,6 +109,7 @@ import org.opengis.filter.temporal.TOverlaps;
 import org.opengis.temporal.Period;
 
 import com.vividsolutions.jts.geom.Geometry;
+import java.util.*;
 
 /**
  * Encodes a filter into a SQL WHERE statement.  It should hopefully be generic
@@ -1415,11 +1416,22 @@ public class FilterToSQL implements FilterVisitor, ExpressionVisitor {
         } else if(literal instanceof java.sql.Date || literal instanceof java.sql.Timestamp) {
             // java.sql.date toString declares to always format to yyyy-mm-dd 
             // (and TimeStamp uses a similar approach)
-            out.write("'" + literal + "'");
+            // BC - better to statically compute this than use a calendar here
+        	out.write("'" + literal);
+        	java.util.Date d = (java.util.Date) literal;
+            if (d.getTime() < -62135769600000L) {
+                out.write(" BC");
+            }
+            out.write("'");
         } else if(literal instanceof java.util.Date) {
             // get back to the previous case
             Timestamp ts = new java.sql.Timestamp(((Date) literal).getTime());
-            out.write("'" + ts + "'");
+            out.write("'" + ts);
+        	java.util.Date d = (java.util.Date) literal;
+            if (d.getTime() < -62135769600000L) {
+                out.write(" BC");
+            }
+            out.write("'");
         } else {
             // we don't know the type...just convert back to a string
             String encoding = (String) Converters.convert(literal,
@@ -1433,7 +1445,7 @@ public class FilterToSQL implements FilterVisitor, ExpressionVisitor {
             String escaped = encoding.replaceAll("'", "''");
             out.write("'" + escaped + "'");
         }
-    } 
+    }
 
     /**
      * Subclasses must implement this method in order to encode geometry
--- a/modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
+++ b/modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
@@ -101,6 +101,7 @@ import org.opengis.referencing.operation.TransformException;
 import com.vividsolutions.jts.geom.Envelope;
 import com.vividsolutions.jts.geom.Geometry;
 import com.vividsolutions.jts.geom.Point;
+import org.geotools.util.Converter;
 
 
 /**
@@ -1448,9 +1449,17 @@ public final class JDBCDataStore extends ContentDataStore
                     st.setFetchSize(fetchSize);
                     rs = st.executeQuery( sql );
                 }
-             
+                Class resultType = getMapping(rs.getMetaData().getColumnType(1));
+                Class outputType = att == null ? resultType : att.getType().getBinding();
+                final Converter converter = Converters.getConverter( resultType, outputType, null);
+                
                 while(rs.next()) {
                     Object value = rs.getObject(1);
+                    try {
+                        value = converter.convert(value, outputType);
+                    } catch (Exception ex) {
+                        throw new IOException("Error converting " + value + " to " + outputType);
+                    }
                     result = value;
                     results.add(value);
                 }
@@ -3049,7 +3058,7 @@ public final class JDBCDataStore extends ContentDataStore
          *        <LI><B>NUM_PREC_RADIX</B> int => usually 2 or 10
          */
         ResultSet types = metaData.getTypeInfo();
-
+        SQLDialect dialect = getSQLDialect();
         try {
             while (types.next()) {
                 int sqlType = types.getInt("DATA_TYPE");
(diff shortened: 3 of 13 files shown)


aaime:
One second check reveals the usual suspects: 
* Single commit with reference to a ticket please
* Update the copyright headers as needed
* Travis build failed (on all configurations), looks like we have a compile failure

travislbrundage:
Thanks for your diligent feedback, Andrea, it is much appreciated. I've squashed it to a single commit which references ticket GEOT-5258, and the build failure is fixed.

travislbrundage:
Looks like this actually has some failures in the online tests, so I'll be fixing that up next.

travislbrundage:
The online tests which were failing before are now passing.

modules/library/api/src/main/java/org/geotools/util/Converters.java
```diff
@@ -16,9 +16,11 @@
  */
```
jodygarnett: Please update header - kevin has a script:
```
(C) 2002-2015, Open Source Geospatial Foundation (OSGeo)
```

modules/library/api/src/main/java/org/geotools/util/Converters.java
```diff
@@ -184,6 +194,32 @@ private static FactoryRegistry getServiceRegistry() {
         }
         return null;
     }
+    
+    public static Converter getConverter(Class source, Class target, final Hints hints) {
```
jodygarnett: need a comment for a public API method.

Q: should you call this factory method from convert( source, target, hints ).

modules/library/jdbc/src/test/java/org/geotools/jdbc/JDBCDateOnlineTest.java
```diff
@@ -116,4 +128,48 @@ public void testFilterByTime() throws Exception {
...
+        if (!testNegativeDates) {
+            return;
+        }
+        
+//        Logger.getLogger("").setLevel(Level.ALL);
+//        Logger.getLogger("").getHandlers()[0].setLevel(Level.ALL);
+//        Logging.getLogger(JDBCDataStore.class).setLevel(Level.ALL);
+        
+        // make our negative year, reset hours/minutes/seconds
+        // to allow test of date and datetime
```
jodygarnett: please remove commented out codes

modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/PostGISDialect.java
```diff
@@ -453,6 +455,10 @@ public Envelope decodeGeometryEnvelope(ResultSet rs, int column,
         if("citext".equalsIgnoreCase(typeName)) {
     	    return String.class;
         }
+
+        if ("bigdate".equalsIgnoreCase(typeName)) {
+            return XDate.class;
```
jodygarnett: Is XDate a new dependency? I have not heard of this one before.

modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/PostGISDialect.java
```diff
@@ -1175,5 +1191,17 @@ protected void addSupportedHints(Set<Hints.Key> hints) {
             hints.add(Hints.GEOMETRY_SIMPLIFICATION);
         }
     }
+
+    public static class XDate extends java.util.Date {
```
jodygarnett: If this is a public class we will need javadocs, and probably an entry in the user guide.

jodygarnett:
A few headers and comments to clean up. I am not quite sure how the XDate class is being used? Perhaps you could explain.

modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/PostGISDialect.java
```diff
@@ -1083,7 +1095,11 @@ public void encodeValue(Object value, Class type, StringBuffer sql) {
             else {
                 encodeByteArrayAsEscape(input, sql);
             }
-
+        } else if (type == PostGISDialect.XDate.class) {
```
jodygarnett: So I can kind of see how this works, XDate class is "just" a placeholder to mark "bigdate" and is being used to add in this special case.

modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/PostgisFilterToSQL.java
```diff
@@ -92,6 +94,21 @@ protected void visitLiteralGeometry(Literal expression) throws IOException {
     }
 
     @Override
+    protected Object evaluateLiteral(Literal expression, Class target) {
+        Date converted = null;
+        if (target == PostGISDialect.XDate.class) {
+            converted = Converters.convert(expression.evaluate(null), Date.class);
```
jodygarnett: And over here the "bigdata" placeholder is being used as a bit of a trick (violating the API contract of evaulateLiteral since a XDate is not actually being returned.

modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
```diff
@@ -3049,7 +3058,7 @@ private Integer findVarcharColumnLength(AttributeDescriptor att) {
          *        <LI><B>NUM_PREC_RADIX</B> int => usually 2 or 10
          */
         ResultSet types = metaData.getTypeInfo();
-
+        SQLDialect dialect = getSQLDialect();
```
travislbrundage: This is never used...

modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/PostgisFilterToSQL.java
travislbrundage: Thanks for providing that perspective. I didn't understand what was going on before, but now it makes sense - in a disconcerting, hackish way.

modules/library/api/src/main/java/org/geotools/util/Converters.java
aaime: Agreed, we should try to avoid duplication

modules/library/jdbc/src/main/java/org/geotools/data/jdbc/FilterToSQL.java
```diff
@@ -1415,11 +1416,22 @@ protected void writeLiteral(Object literal) throws IOException {
         } else if(literal instanceof java.sql.Date || literal instanceof java.sql.Timestamp) {
             // java.sql.date toString declares to always format to yyyy-mm-dd 
             // (and TimeStamp uses a similar approach)
-            out.write("'" + literal + "'");
+            // BC - better to statically compute this than use a calendar here
+        	out.write("'" + literal);
```
aaime: Formatting issue (extra indentation)

modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
```diff
@@ -1448,9 +1449,17 @@ protected Object getAggregateValue(FeatureVisitor visitor, SimpleFeatureType fea
...
                 }
-             
+                Class resultType = getMapping(rs.getMetaData().getColumnType(1));
+                Class outputType = att == null ? resultType : att.getType().getBinding();
+                final Converter converter = Converters.getConverter( resultType, outputType, null);
+                
                 while(rs.next()) {
                     Object value = rs.getObject(1);
+                    try {
+                        value = converter.convert(value, outputType);
```
aaime: Wondering if this could cause regressions, like, today code dealing with dimensions happily takes a java.util.Date subclass, but with this conversion attempt we might fail to convert a custom java.util.Date subclass to the expected target type. Also, this method can freely return null (see the javadoc), and this is a problem too, as null will be returned when a non null value was returned from database. 
Finally, if this is the only point using Converters.getConverter, we can just drop the new API and use the existing Converters.convert

modules/library/main/src/main/java/org/geotools/util/TemporalConverterFactory.java
```diff
@@ -135,8 +135,48 @@ public Object convert(Object source, Class target) throws Exception {
...
+                    public <T> T convert(Object source, Class<T> target) throws Exception {
+                        Date date = (Date) source;
+                        if (date == null) {
+                            return null;
+                        }
+                        return target.cast(new Long(date.getTime()));
+                    }
+                };
+            }
+            // if ( String.class.equals( target ) ) {
```
aaime: Commented out code

modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/PostGISDialect.java
aaime: Can we make it package private instead?

modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/PostgisFilterToSQL.java
```diff
@@ -92,6 +94,21 @@ protected void visitLiteralGeometry(Literal expression) throws IOException {
...
+        }
+        if (converted != null) {
+            return converted.getTime();
```
aaime: Confused here, if it's a date, why are we extracting the time as a Long?

modules/plugin/jdbc/jdbc-postgis/src/test/java/org/geotools/data/postgis/PostgisUDTOnlineTest.java
```diff
@@ -38,6 +49,49 @@
...
+        
+        SimpleFeatureType schema = dataStore.getSchema(tname("bigdates"));
+        SimpleFeatureType newFeatureType = FeatureTypes.newFeatureType(schema.getAttributeDescriptors().toArray(new AttributeDescriptor[0]), "bigdates2");
+        dataStore.createSchema(newFeatureType);
+        schema = dataStore.getSchema(tname("bigdates2"));
+        assertEquals(XDate.class, schema.getType("d").getBinding());
+        
+        // make sure the string query works
+        FilterFactory ff = dataStore.getFilterFactory();
+        // @todo revisit - not sure if string query will ever get passed in this formatting...
```
aaime: Commented out code

modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/PostgisFilterToSQL.java
```diff
@@ -92,6 +94,21 @@ protected void visitLiteralGeometry(Literal expression) throws IOException {
...
+        } else {
+            return super.evaluateLiteral(expression, target);
```
travislbrundage: As a note, in the PostgisUDTOnlineTest's testBigDates(), it always goes to this line. It never runs the XDate part of the code...

modules/library/main/src/main/java/org/geotools/util/TemporalConverterFactory.java
```diff
@@ -135,8 +135,48 @@ public Object convert(Object source, Class target) throws Exception {
                 };
             }
             
+            if (Long.class.isAssignableFrom(target)) {
```
travislbrundage: As far as I can tell, the code added here never is true and never gets executed either...

aaime:
Hum... checking this one, it seems that quite a bit of the feedback still needs to be addressed, right?

travislbrundage:
Yeah, I have updates that I haven't pushed, but I haven't finished addressing all the things yet, primarily related to how the XDate class is used for big dates. I'm still working at it. Thanks for checking in. :)

jodygarnett:
I would like to reject this as violating the api contract.

PR #1000, opened 2015-10-16

ridethepenguin:
Adds support for complex types with arbitrary contents (i.e. derived from `xs:anyType` by extension). Simple content is converted to text and added to the special property called `simpleContent`.

See [related discussion](http://osgeo-org.1560.x6.nabble.com/App-schema-cannot-set-string-value-to-swe-DataArray-swe-values-element-td5228781.html) on the mailing list.


bencaradocdavies:
@ridethepenguin that looks great. Excellent test coverage!

PR #1001, opened 2015-10-18
[GEOT-2933] add geostationary statellite projection support

tkunicki:
This PR is in progress, my unit tests need to be ported to the GeoTools projection test harness.

### DIFF
--- a/modules/library/referencing/src/main/java/org/geotools/referencing/operation/projection/GeostationarySatellite.java
+++ b/modules/library/referencing/src/main/java/org/geotools/referencing/operation/projection/GeostationarySatellite.java
@@ -1,0 +1,346 @@
+package org.geotools.referencing.operation.projection;
+
+import org.geotools.geometry.DirectPosition2D;
+import org.geotools.geometry.Envelope2D;
+import org.geotools.metadata.iso.citation.Citations;
+import org.geotools.referencing.CRS;
+import org.geotools.referencing.NamedIdentifier;
+import org.opengis.parameter.*;
+import org.opengis.referencing.FactoryException;
+import org.opengis.referencing.crs.CoordinateReferenceSystem;
+import org.opengis.referencing.operation.MathTransform;
+import org.opengis.referencing.operation.TransformException;
+
+import javax.measure.unit.SI;
+
+import java.awt.geom.Point2D;
+import java.util.Collection;
+
+/**
+ * The Geostationary Satellite Projection
+ * <p>
+ * Adapted from https://github.com/OSGeo/proj.4/blob/4.9/src/PJ_geos.c
+ * <p>
+ * NOTE: Not all valid coordinates in this projection will transform to valid
+ * terrestrial coordinates, this is especially true of "Full Disk" earth
+ * coverages.  If one must deal with coverages in this projection with
+ * generalized code which requires the coverage bounding-box coordinates
+ * to transform to valid terrestrial values consider clipping to a rectangle
+ * inscribing the ellipsoid.
+ *
+ * @author Tom Kunicki
+ */
+public abstract class GeostationarySatellite extends MapProjection {
+
+    private static final long serialVersionUID = 1L;
+
+    final double h;
+    final double radius_g;
+    final double radius_g_1;
+    final double C;
+
+    public GeostationarySatellite(ParameterValueGroup parameters) throws ParameterNotFoundException {
+        super(parameters);
+
+        final Collection<GeneralParameterDescriptor> expected = getParameterDescriptors().descriptors();
+
+        // from https://github.com/OSGeo/proj.4/blob/4.9/src/projects.h
+        //  a,  /* major axis or radius if es==0 */
+
+        final double a = semiMajor;
+        h = doubleValue(expected, Provider.SATELLITE_HEIGHT, parameters);
+
+        // from https://github.com/OSGeo/proj.4/blob/4.9/src/PJ_geos.c
+        //  P->radius_g_1 = P->h / P->a;
+        //  P->radius_g = 1. + P->radius_g_1;
+        //  P->C  = P->radius_g * P->radius_g - 1.0;
+        radius_g_1 = h / a;
+        radius_g = 1d + radius_g_1;
+        C = radius_g * radius_g - 1d;
+    }
+
+    @Override
+    public ParameterDescriptorGroup getParameterDescriptors() {
+        return Provider.PARAMETERS;
+    }
+
+    @Override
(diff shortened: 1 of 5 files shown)


tkunicki:
@aaime : as discussed at FOSS4G here is a Proj4 port of the [Geostationary_Satellite](http://www.remotesensing.org/geotiff/proj_list/geos.html) projection

modules/library/referencing/src/main/java/org/geotools/referencing/operation/projection/GeostationarySatellite.java
```diff
@@ -0,0 +1,346 @@
+package org.geotools.referencing.operation.projection;
```
aaime: Lacks copyright header (just to OSGeo, only 2015)

modules/library/referencing/src/main/java/org/geotools/referencing/operation/projection/GeostationarySatellite.java
```diff
@@ -0,0 +1,346 @@
...
+/**
+ * The Geostationary Satellite Projection
+ * <p>
+ * Adapted from https://github.com/OSGeo/proj.4/blob/4.9/src/PJ_geos.c
+ * <p>
+ * NOTE: Not all valid coordinates in this projection will transform to valid
+ * terrestrial coordinates, this is especially true of "Full Disk" earth
+ * coverages.  If one must deal with coverages in this projection with
+ * generalized code which requires the coverage bounding-box coordinates
+ * to transform to valid terrestrial values consider clipping to a rectangle
```
aaime: Just a note, no need to act on it, but at least rendering wise this could be achieved by writing a ProjectionHandler

modules/library/referencing/src/main/java/org/geotools/referencing/operation/projection/GeostationarySatellite.java
```diff
@@ -0,0 +1,346 @@
...
+            double phi = Math.atan(Vz * Math.cos(lambda) / Vx);
+            phi = Math.atan(radius_p_inv2 * Math.tan(phi));
+            p2d.setLocation(lambda, phi);
+
+            return p2d;
+        }
+    }
+
+    // from https://github.com/OSGeo/proj.4/blob/4.9/src/hypot.c
+    static double hypot(double x, double y) {
```
aaime: Wondering if this might be equivalent to Math.hypot

modules/library/referencing/src/test/java/org/geotools/referencing/ScriptTest.java
```diff
@@ -314,7 +314,18 @@ public void testSinusoidal() throws Exception {
...
     }
-    
+
+    /**
+     * Run "GEOS.txt"
+     *
+     * @throws Exception If a test failed.
+     */
+    @Test
+    @Ignore
```
aaime: Err... why ignored?

modules/library/referencing/src/test/java/org/geotools/referencing/operation/projection/GeostationarySatelliteTest.java
```diff
@@ -0,0 +1,354 @@
...
+        geogToEllipsoidalGeos.transform(p, p);
+        assertThat(p, is(notNullValue()));
+
+        p.setLocation(inscribed.getMinX(), inscribed.getMinY());
+        ellipsoidalGeosToGeog.transform(p, p);
+        assertThat(p, is(notNullValue()));
+        geogToEllipsoidalGeos.transform(p, p);
+        assertThat(p, is(notNullValue()));
+
+        // Inscribed rectangle is smaller than largest inscribing rectangle, hence ESTIMATE
```
aaime: We try not to have commented out code. But since it's here I have to ask, why was it commented out?

aaime:
Looks good and thoroughly tested, minus some observations I've added in the diff

modules/library/referencing/src/test/java/org/geotools/referencing/ScriptTest.java
tkunicki: Ooooh, an oversight (and why we do code reviews!).  Initially I copied the last entry from bottom of file, later to realize those were all ignored.  I moved my modified copy behind last unignored test but never removed `@ignored`.  I didn't catch this because I run the script individually.

modules/library/referencing/src/test/java/org/geotools/referencing/operation/projection/GeostationarySatelliteTest.java
tkunicki: I'd intended the method to return the maximum-area inscribed rectangle from the visible ellipse in a full disk image. The tests were to verify that condition. The reality is that the projection math is a little more complicated and what is returned is the almost-maximum area inscribed rectangle.  It's also possible that an ellipse or the ellipse axes i am using are not the appropriate. I would like to address this eventually...

This method is intended to aid in coverage clipping, resampling and tiling where it's problematic if coverage bounding box can't be transformed.  This work is on my road map.

modules/library/referencing/src/main/java/org/geotools/referencing/operation/projection/GeostationarySatellite.java
tkunicki: Unit tests pass w/ conversion to `Math.hypot`.  Will modify and push...

tkunicki:
@aaime changes made per your review.  Can squash if you'd like.

aaime:
All good, manually merged here: https://github.com/geotools/geotools/commit/766c2e39ac7de4ef8f54ed25fb25ee66d9e249b9

Do you need it on 14.x? If so we can ask on the devel list for a backport, eventually waiting a month

mfilocha:
@tkunicki Are you aware of https://osgeo-org.atlassian.net/browse/GEOT-5199 (already on 14.0)?

tkunicki:
I did look at that.  The projection algorithm implemented for 5199 looks different, possibly designed solely for meteostat (this was from a cursory examination, but it didn't seem consistent with Proj.4 geos algorithm. e.g. What are the projection units for the 5199 impl?).  This PR is a general, paramaterized, implementation that's a 1:1 port of the Proj.4 geos implementation.  We need support for the gamut of geostationary satellites as a basis for global coverage generation.

mfilocha:
You're right, my port is a particular case of geostationary satellite projection, based directly on the reference code provided by Eumetsat. Units are pixels. Algorithms are consistent (modulo origin offset and units, of course) .

PR #1003, opened 2015-10-19
[GEOT-5256] added removeSchema Support for MemoryDataStore

fgdrf:
To remove Schemas from MemoryDataStore removeSchema-methods should be overriden by impelemtations of ContenDataStore otherwise UnsupportedOperationException are thrown.

fixes https://osgeo-org.atlassian.net/browse/GEOT-5256


fgdrf:
Like discussed during last PMC meeting this is just an implementation within MemoryDataStrore class that supports schema removal. If I understood @aaime, @jodygarnett and others correctly it can be easy backported to 14.x which is the final goal. 
Since the travis builds finished successfully and the test passed, I merged it into master

PR #1009, opened 2015-10-21
[GEOT-5264] Batch inserts

pvalsecc:
Those commits changes geotools to group the features to be inserted by batch and uses JDBC batch insert to speed up big insertions. It improves as well the way we get the primary key values (avoid doing on select per primary key field and per feature inserted).

I ran the online tests for those DBs:    
* Postgis
* H2
* Mysql
* Oracle

Improvements I've mesured with a geogig Oracle export of a table with 17000 records (4 simple attributes and a single point geometry) without sequence is run in:
* 640 seconds without my modifs
* 13 seconds with my modifs


pvalsecc:
Working on fixing GeoPackage...

modules/library/data/src/main/java/org/geotools/data/store/ContentFeatureStore.java
```diff
@@ -283,37 +278,41 @@ public ContentFeatureStore(ContentEntry entry,Query query) {
...
+        toWrite.getUserData().putAll(feature.getUserData());
+
         // pass through the fid if the user asked so
         boolean useExisting = Boolean.TRUE.equals(feature.getUserData().get(Hints.USE_PROVIDED_FID));
         if(getQueryCapabilities().isUseProvidedFIDSupported() && useExisting) {
             ((FeatureIdImpl) toWrite.getIdentifier()).setID(feature.getID());
         }
-        
+
+        //PV: I'm not very happy about this hack. But that's the only way I've found, with the
```
aaime: Please avoid this type of comment (with initials), if everybody did them this way the sources would be a mess.

modules/library/data/src/main/java/org/geotools/data/store/ContentFeatureStore.java
```diff
@@ -283,37 +278,41 @@ public ContentFeatureStore(ContentEntry entry,Query query) {
...
+        //    current infra, to have the feature's ID to be updated in case of delayed write
```
aaime: "infra"?

modules/library/data/src/main/java/org/geotools/data/store/ContentFeatureStore.java
```diff
@@ -283,37 +278,41 @@ public ContentFeatureStore(ContentEntry entry,Query query) {
...
+        //    (see JDBCInsertFeatureWriter)
+        toWrite.getUserData().put("_original_", feature);
```
aaime: Please use a constant for "_original_"

modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
```diff
@@ -1549,70 +1547,201 @@ protected void insert(SimpleFeature feature, SimpleFeatureType featureType, Conn
...
-                        PreparedStatement ps = insertSQLPS( featureType, feature, keyValues, cx );
-                        try {
-                            ((PreparedStatementSQLDialect)dialect).onInsert(ps, cx, featureType);
-                            ps.execute();
-                        } finally {
-                            closeSafe( ps );
+                    Class binding = att.getType().getBinding();
+
+                    Object value = feature.getAttribute(colName);
+                    /*if (value == null) {
```
aaime: Do or don't, there is no try (don't add commented out sections of code, just uncomment and throw the exception

pvalsecc: That is old code I've moved without touching it. I wanted to avoid changing 
things too much. But no problem, I'll uncomment it and see if I don't break 
anything.

How should I move forward with your comments? Amend the existing commits or 
add new commits?

Envoyé avec AquaMail pour Android
http://www.aqua-mail.com

aaime: Well, in theory you should end up with a single commit with a message in form "[GEOT-XYZW] title of the ticket"

modules/library/data/src/main/java/org/geotools/data/store/ContentFeatureStore.java
pvalsecc: Infrastructure.

I spent a long time trying to understand why the features are copied in the JDBCInsertFeatureWriter before inserting them. There is, somewhere a big comment explaining that the feature's structure may not match the DB structure, but why would that matter? If they don't match too much it won't work anyway.

That comment was more a note for the reviewer to pay attention to the hack and make sure it doesn't go un-noticed. Mission accomplished  ;-)

Now, do you see a better way? If not and you are OK with it, I'll remove the negativity and the personal touch from this comment.

By the way, thanks for the review.

modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
```diff
@@ -1549,70 +1547,201 @@ protected void insert(SimpleFeature feature, SimpleFeatureType featureType, Conn
...
+            }
+            int[] inserts = ps.executeBatch();
+            checkAllInserted(inserts, features.size());
+            keysFetcher.postInsert(featureType, features, ps);
+        } finally {
+            closeSafe(ps);
+        }
+    }
+
+    private static void checkAllInserted(int[] inserts, int size) throws IOException {
```
aaime: Why static, only this class seems to be using it

modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
pvalsecc: OK, I'll squash them all together.

PS: I thought, for the review, it would have been be nicer to have the changes split by "theme".

modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
```diff
@@ -1878,7 +2007,7 @@ protected String encodeFID( PrimaryKey pkey, ResultSet rs, int offset ) throws S
         return encodeFID( keyValues );
     }
 
-    protected String encodeFID( List<Object> keyValues ) {
+    protected static String encodeFID( List<Object> keyValues ) {
```
aaime: Normally classes that need services from JDBCDataStore take a reference to it (see how the various closeSafe methods are used)

modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
```diff
@@ -3840,9 +3848,9 @@ protected String insertSQL(SimpleFeatureType featureType, SimpleFeature feature,
             Object value = feature.getAttribute(colName);
 
             if (value == null) {
-                if (!att.isNillable()) {
+                /*if (!att.isNillable()) {
```
aaime: Same as above

modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
```diff
@@ -3863,23 +3871,8 @@ protected String insertSQL(SimpleFeatureType featureType, SimpleFeature feature,
...
-                    sql.append( "," );
-                } 
-                catch( Exception e ) {
-                    throw new RuntimeException( e );
-                }
-            }
-        }
-        sql.setLength(sql.length() - 1);
+        keysFetcher.setKeyValues(this, cx, featureType, feature, sql);
+        sql.setLength(sql.length() - 1);  //remove last coma
```
aaime: I guess it's a comma, not a coma :-p

modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCInsertFeatureWriter.java
```diff
@@ -20,39 +20,55 @@
...
  * @source $URL$
  */
 public class JDBCInsertFeatureWriter extends JDBCFeatureReader implements FeatureWriter<SimpleFeatureType, SimpleFeature> {
-
-    
-    ResultSetFeature last;
+    /**
+     * Grouping elements together in order to have a decent batch size.
+     */
+    private ResultSetFeature[] buffer = new ResultSetFeature[100];
```
aaime: Hard coded buffer size... we might want to have a store parameter, a system variable, or ... something :-)

modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCInsertFeatureWriter.java
```diff
@@ -71,28 +88,65 @@ public void remove() throws IOException {
...
+                final SimpleFeature orig = (SimpleFeature)cur.getUserData().get("_original_");
+                if (orig != null) {
+                    ((FeatureIdImpl)orig.getIdentifier()).setID(fid);
+                    orig.getUserData().putAll(cur.getUserData());
+                    orig.getUserData().remove("_original_");
+                }
+
+                final ContentEntry entry = featureSource.getEntry();
+                final ContentState state = entry.getState(this.tx);
+                state.fireFeatureAdded(featureSource, cur);
```
aaime: I know that uDig pays attention to events, @jodygarnett , is having a delayed notification of the insert (e.g., after 100 write calls) going to cause issues?

modules/library/jdbc/src/main/java/org/geotools/jdbc/KeysFetcher.java
```diff
@@ -0,0 +1,553 @@
...
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * In charge for setting/getting the primary key values for inserts.
+ */
+public abstract class KeysFetcher {
```
aaime: Try to avoid making new public API if we are not forced to, once it's public, it's hard to change

modules/library/jdbc/src/main/java/org/geotools/jdbc/InsertionClassifier.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.TreeMap;
+
+/**
+ * PreparedStatement inserts must be split in function of those criteria:
+ * - useExisting
+ * - type of the geometries
+ * <p/>
+ * This class allows to do the splitting.
+ */
+public class InsertionClassifier {
```
aaime: Try to avoid making new public API if we are not forced to, once it's public, it's hard to change

modules/library/jdbc/src/main/java/org/geotools/jdbc/InsertionClassifier.java
```diff
@@ -0,0 +1,101 @@
...
+    public final boolean useExisting;
+    public final Map<String, Class<? extends Geometry>> geometryTypes;
+
+    public static Map<InsertionClassifier, Collection<SimpleFeature>> classify(
+            SimpleFeatureType featureType, Collection<? extends SimpleFeature> features) {
+        //TODO: would be nicer with guava's MultiMap, but guava is not available and I don't have
```
aaime: Yep, we don't want GeoTools to depend on Guava yet. Please remove the comment.

modules/library/jdbc/src/main/java/org/geotools/jdbc/InsertionClassifier.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.TreeMap;
+
+/**
+ * PreparedStatement inserts must be split in function of those criteria:
+ * - useExisting
+ * - type of the geometries
```
aaime: I don't think we really need to classify geometries based on their type, but only if they are null, or not. In most (all?) databases it's fine to have generic geometry columns that host everything, and the mechanisms we use to write them are not type specific to my knowledge, classifying by geometry type seems only a way to make batch inserts less effective.. unless I'm missing something?

modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
aaime: I guess it depends on the reviewer.. I honestly don't care to see the steps you followed to get to a result, it's a waste of time to me, I always review the final product, not the single commits.

modules/library/jdbc/src/test/java/org/geotools/jdbc/JDBCDataStoreOnlineTest.java
```diff
@@ -139,16 +139,19 @@ public void testCreateSchemaWithConstraints() throws Exception {
...
+        FeatureWriter<SimpleFeatureType, SimpleFeature> w =
+                dataStore.getFeatureWriter( tname("ft2"),Transaction.AUTO_COMMIT);
         w.hasNext();
 
-        SimpleFeature f = (SimpleFeature) w.next();
+        SimpleFeature f = w.next();
         f.setAttribute( 1, new Integer(0));
         f.setAttribute( 2, "hello");
         w.write();
+        w.close();
```
aaime: Ok, this is the type of backwards compatibilty change that scares me. Due to batching, the writers have to be closed before anything else in the same transaction can see the modifications, which for some applications will be a serious problem (since they might need to write a little, read a little, write some more), especially since it will be hard to locate for those upgrading (and people notably do not read upgrade notes until they figure they are having problems).
This may require a discussion in a larger group, but we might want to have batching activated by some flag, and off by default.

aaime:
@mcrmcr can you have a look at this one too, and maybe run the DB2 tests?
For everybody else, can you chime in on the batch insert backwards compatibility issues?
There is another one I did not mention: sloppy code that does not call finalize, and that's used to get the insert right away, will only get (some of) them only after finalization runs, which runs with unpredictable timings.

modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
pvalsecc: Because it's more a function than a method and it's more testable like that. By the way, I'v forgot to add the test. Fixing that.

modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
pvalsecc: This is again more a function than a method. It doesn't need anything from JDBCDataStore instances.

modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCInsertFeatureWriter.java
pvalsecc: I'll wait for you guys to take a decision regarding your "sloppy code support" comment before doing anything on that.
Regarding the performance, I did tests with different DBs and 100 is the sweet spot. IMHO, there is no need to add configuration for that. Unless you want to set it to 1 to disable buffering.

modules/library/jdbc/src/main/java/org/geotools/jdbc/InsertionClassifier.java
pvalsecc: Sadly, we do need to classify the geometries based on their types. At least for DB2. Please have a look at [DB2Util#prepareGeometryValue](https://github.com/geotools/geotools/blob/6b7fb3e34ccd12cb446c47ae472cbeb1b25dec55/modules/plugin/jdbc/jdbc-db2/src/main/java/org/geotools/data/db2/DB2Util.java#L157-157).

modules/library/jdbc/src/test/java/org/geotools/jdbc/JDBCDataStoreOnlineTest.java
pvalsecc: How do we initiate this discussion?

aaime: It's best to start a thread on the gt-devel mailing list

pvalsecc:
OK, I've pushed a squashed commit with fixes from Andrea's comments.

pvalsecc:
I've followed Andrea's recommendation and I've made the batch size to be configurable with a default of 1. That way, we don't break backward compatibility. I've also reverted most UTs I had to fix and just made that they are run with the default batch size of 1. The rest of the UTs are run with a batch size of 100.

modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCInsertFeatureWriter.java
aaime: @jodygarnett are you there, can you tell us if this is a problem? @fgdrf is also a udig developer, he might know too

modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
```diff
@@ -418,6 +424,14 @@ public void setFetchSize(int fetchSize) {
         this.fetchSize = fetchSize;
     }
 
+    public int getBatchInsertSize() {
```
aaime: Public API without a javadoc (same for the setter)

modules/plugin/jdbc/jdbc-oracle/src/main/java/org/geotools/data/oracle/OracleDialect.java
```diff
@@ -1248,20 +1248,20 @@ private String getCompatibleAxisName(CoordinateSystemAxis axis, int dimensionIdx
...
     public String getSequenceForColumn(String schemaName, String tableName,
             String columnName, Connection cx) throws SQLException {
-        String sequenceName = (tableName + "_" + columnName + "_SEQUENCE").toUpperCase();
+        String sequenceName = (tableName + "_" + columnName + "_%").toUpperCase();
         PreparedStatement st = null;
         String sql;
         
         try {
-            sql = "SELECT * FROM USER_SEQUENCES WHERE SEQUENCE_NAME = ?";
+            sql = "SELECT SEQUENCE_NAME FROM USER_SEQUENCES WHERE SEQUENCE_NAME like ?";
```
aaime: This is a change in behavior that probably needs its own tests

aaime:
I've added a couple more comments, but it's minor stuff. Now, can core developers please help reviewing this on?  The change is really large, and I don't pretend to understand it all in detail (it simply does not fit the time I can dedicate during weekends, made of bits of one hour long).
Anyways, if no other core developers show up, I'll be merging this one next week, no point in delaying it further if no one else is going to check it out.

modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
pvalsecc: I refuse to add stupid comments like: "get the batch insert size" and the attribute itself has enough documentation.

aaime: The javadoc will have to be much more than that, and explain the consequences of batching to those that are not aware of it, like, not being able to immediately see the effects of an insert because it's stuck in the batch queue, and still hasn't hit the database.

aaime:
One month in, all feedback I provided addressed, no one else caring to comment... merging, past this point I assume any further issue with the pull request will be addressed directly by those that did not care to participate during the review.

PR #1012, opened 2015-10-22
GEOT-5260 Map missing core java.sql data types to spatialite

oscarfonts:
See https://osgeo-org.atlassian.net/browse/GEOT-5260


oscarfonts:
Just realized Geopackage has the same problem.
Will generate a new Pull Request fixing both plugins.

PR #1013, opened 2015-10-22

ridethepenguin:
This is the backport to the 14.x branch.


bencaradocdavies:
Thanks @ridethepenguin that looks good to me.

PR #1015, opened 2015-10-24
updated refs to svn to git and version

ianturton:
removed references to SVN and added refs to git

### DIFF
--- a/docs/user/tutorial/map/style.rst
+++ b/docs/user/tutorial/map/style.rst
@@ -45,10 +45,10 @@ Example
 -------
 
 The example code is available
- * Directly from svn: StyleLab.java_
+ * Directly from git: StyleLab.java_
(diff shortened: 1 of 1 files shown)


docs/user/tutorial/map/style.rst
```diff
@@ -135,7 +135,7 @@ Dependencies
 Please ensure your pom.xml includes the following::
 
   <properties>
-      <geotools.version>2.7-SNAPSHOT</geotools.version>
+      <geotools.version>14.0</geotools.version>
```
fgdrf: @ianturton I'm wondering about the Release Version on master-branch. Is it because demo-code is tested against this version? Is it worth to think about  @RELEASE@ token or the Shinx replacemant |version| that could be replaced with the final release version during release-procedure (rename.xml) or during Sphinx-doc build?

PR #1020, opened 2015-10-26
GEOT-5260 Map missing core java.sql data types to spatialite & geopkg

oscarfonts:
See https://osgeo-org.atlassian.net/browse/GEOT-5260


modules/unsupported/geopkg/src/test/java/org/geotools/geopkg/GeoPkgDataStoreAPITest.java
```diff
@@ -0,0 +1,88 @@
+package org.geotools.geopkg;
```
jodygarnett: header :)

modules/unsupported/geopkg/src/test/java/org/geotools/geopkg/GeoPkgDataStoreAPITest.java
```diff
@@ -0,0 +1,88 @@
...
+        map.put(GeoPkgDataStoreFactory.DBTYPE.key, "geopkg");
+        map.put(GeoPkgDataStoreFactory.DATABASE.key, "foo.gpkg");
+
+        GeoPkgDataStoreFactory factory = new GeoPkgDataStoreFactory();
+        factory.setBaseDirectory(tmp.getRoot());
+        dataStore = factory.createDataStore(map);
+    }
+
+    @Test
+    public void testCreateDataTypes() throws Exception {
```
jodygarnett: Is there no other geopkg test we can add this to?

modules/plugin/jdbc/jdbc-spatialite/src/test/java/org/geotools/data/spatialite/SpatiaLiteDataStoreAPITestSetup.java
```diff
@@ -28,7 +30,16 @@
     protected SpatiaLiteDataStoreAPITestSetup() {
         super(new SpatiaLiteTestSetup());
     }
-    
+
+    protected void setUpData() throws Exception {
+        //kill all the data
+        super.setUpData();
+        try {
+            dropDataTypesTable();
```
jodygarnett: Perhaps I am a bit slow, I thought these tests do clean up activities (like dropDataTypesTable) in the teardown method, and in the setup method prior to setting up the data (incase there was a previous failed run).

jodygarnett:
general feedback on headers, a question about the geopkg test case.

modules/plugin/jdbc/jdbc-spatialite/src/test/java/org/geotools/data/spatialite/SpatiaLiteDataStoreAPITestSetup.java
oscarfonts: Agreed, conceptually.

But just followed what I found on its superclass (and other related classes):
https://github.com/geotools/geotools/blob/master/modules/library/jdbc/src/test/java/org/geotools/jdbc/JDBCDataStoreAPITestSetup.java#L38

This setUpData method is different from setUp method, both called by "connect" method in JDBCTestSupport, which in turn is called in the setUp method from OnlineTestCase. This last one is the true junit setup method. There are equivalent tearDown/disconnect methods all the way down, but rarely implemented. Just guessing, but given the nontrivial construct, there may be cases (uncaught exceptions during setup?) where the teardown code is not reached.

I can try to change it, but I'd rather be conservative and stick to what is already there.

modules/unsupported/geopkg/src/test/java/org/geotools/geopkg/GeoPkgDataStoreAPITest.java
oscarfonts: Sure, there are some "online" tests and the first intent was to add it there. But, ehm, I didn't manage to run the existing tests! I should have asked in the list...

But in fact, I then preferred to deliver a simpler test which actually can be always run, not depending on an external condition which is difficult to meet and in most cases implies test will be skipped.

As far as I understood, the "online" tests were made optional and needed a configuration file because they depend on external services (namely a running database instance). Geopkg doesn't need such an external service, so their tests don't need to be "online" and can be always run.

bradh:
Does this need to be rebased?

bradh:
Something has gone horribly wrong with git (or github) in https://github.com/oscarfonts/geotools/commit/ed7029398d4d29cd5b96322928f584ebd9f6126e - it has hundreds of changed files, including some you almost certainly didn't want to modify.

oscarfonts:
@bradh I just pulled the master branch from geotools:master to oscarfonts:GEOT-5260, bringing all the changes since october.

Going to https://github.com/geotools/geotools/pull/1020/files you can see the actual diff, which affects only 5 files.

Excuse me if I missed something and there was a better way to do this.

modules/unsupported/geopkg/src/test/java/org/geotools/geopkg/GeoPkgDataStoreAPITest.java
jodygarnett: Fair enough, thanks for explaining.

jodygarnett:
merged to master, thanks for the hard work @oscarfonts (and @bradh)

PR #1026, opened 2015-10-30
[GEOT-5276] Added fix for new features added outside existing feature…

robward-scisys:
Added fix so that when using GeoServer if a new feature is added outside the extent of the existing feature class the new feature is correctly returned.

Added separate test so existing tests not impacted.


aaime:
Seems ok to me, but I have no way to verify. Anyone around that can check SDE changes?

PR #1027, opened 2015-10-30
[GEOT-5277] ImageWorker equates setting background values to setting nodata

aaime:


simboss:
Are we going to backport to 2.8, in time, right?

aaime:
Well... it's a fix, and affects only code paths used when jai-ext is enabled, so, I don't see why not. +1 on backporting

PR #1030, opened 2015-11-03
Implements GetStyles request on WMS 1.1.1 [GEOT-5309]

Vampouille:
See: https://osgeo-org.atlassian.net/browse/GEOT-5309


modules/extension/wms/src/test/java/org/geotools/data/wms/test/GetStyleOnlineTest.java
```diff
@@ -0,0 +1,136 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ *
+ *    (C) 2004-2008, Open Source Geospatial Foundation (OSGeo)
+ *    (C) 2005, David Zwiers
```
jodygarnett: This is a new file created this year? So the info should be (C) 2015 Open Source Geospatial Foundation (OSGeo)

modules/extension/wms/src/test/java/org/geotools/data/wms/test/GetStyleOnlineTest.java
```diff
@@ -0,0 +1,136 @@
...
+ */
+public class GetStyleOnlineTest extends OnlineTestCase {
+
+    @Override
+    protected String getFixtureId() {
+        return "WMS";
+    }
+
+    @Test
+    public void testGetStyle() throws Exception {
```
jodygarnett: Since this is a single test you may be able to add it to the existing GeoServerOnlineTest?

The test itself look fine thank you.

Vampouille:
Yes, but I don't known where to put this new test. I need fixture so I need OnlineTestCase. Currently, all online tests are related to postgis : 

ImageMosaicPostgisIndexOnlineTest
JDBCTestSupport
PostgisNGCreateDatabaseOnlineTest
PostGisIndexTest

modules/extension/wms/src/test/java/org/geotools/data/wms/test/GetStyleOnlineTest.java
Vampouille: This is fixed

jodygarnett:
Sorry, was getting confused with GeoTools which as a GeoServerOnlineTest (fixture can be configured to test against a local default GeoServer)

Vampouille:
I moved tests to LocalGeoServerOnlineTest class

jodygarnett:
Thank you.

@bencaradocdavies we are good to apply this to master and back port. I will do so now during the GeoServer meeting.

@Vampouille I have two things to check with you:
- Confirmation you have a contribution agreement on file (there was a link when you made the pull request)
- Do you have a JIRA issue for this?

modules/extension/wms/src/main/java/org/geotools/data/wms/request/AbstractGetStylesRequest.java
```diff
@@ -34,11 +35,10 @@
...
      * @param onlineResource
-     * @param layers
+     * @param properties
      */
-    public AbstractGetStylesRequest( URL onlineResource, Layer[] layers) {
-        super(onlineResource, null);
-        this.layers = layers;
+    public AbstractGetStylesRequest( URL onlineResource, Properties properties ) {
+        super(onlineResource, properties);
     }
```
jodygarnett: I am having some problems with this public API change, you have orphaned the private Layer[] layers field (and the getLayers() method that refers to it).

This is making my review go slower :(

modules/extension/wms/src/main/java/org/geotools/data/wms/WMS1_1_0.java
```diff
@@ -212,47 +203,12 @@ public InternalGetLegendGraphicRequest( URL onlineResource, WMS1_1_0 creator) {
         protected void initVersion() {
             setProperty(VERSION, "1.1.0");
         }
-        
-        public Response createResponse(HTTPResponse httpResponse) throws ServiceException, IOException {
-			return new GetLegendGraphicResponse(httpResponse);
-		}	    
-	}
-	
-	public static class InternalGetStylesRequest extends AbstractGetStylesRequest {
```
jodygarnett: I am not sure I understand did you remove GetStyles support from WMS 1.1.0 ... in order to add it to WMS 1.1.1?

jodygarnett:
Trying to merge this and got a bit stuck on public API changes (not too concerned since these classes are only used within this module).

What is more interesting is the appearance of GetStyles being added to WMS 1.1.1 (as expected) by removing it from WMS 1.1.0 (not expected). Am I reading the code wrong ...

Vampouille:
About public API change, I change signature according to describeLayer method which is already implemented.
GetStyles requests are only defined in WMS 1.1.1 (not in 1.1.0 and not in 1.3.0) according to OGC specification (http://www.opengeospatial.org/standards/wms). So I limit implementation from 1.1.1 version

aaime:
I was about to say that the method is defined in the SLD extension for WMS, but indeed the SLD 1.1 specification (based on WMS 1.3) has no trace GetStyles and PutStyles, it just addes the support for remote SLDs in GetMap, DescribeLayer and GetLegendGraphics, while GetStyles, PutStyles are gone.

Btw, I've checked, we haven't received a contribution agreement yet, without it, the pull request cannot be merged (it is clearly stated that you need to send one when more thanone file gets modified, https://github.com/geotools/geotools/blob/master/CONTRIBUTING.md )

jodygarnett:
Since it is an optional operation I do not think we need to be limited by WMS version (we can add the operation to all of them).

aaime:
18 days in, all feedback I provided addressed, no one else caring to comment... merging, past this point I assume any further issue with the pull request will be addressed directly by those that did not care to participate during the review.

aaime:
Crap, I merged the wrong pull request, I though I was merging the batch insert one. Reverting

aaime:
I apologize, can anyone re-open this pull request?

PR #1037, opened 2015-11-10
[GEOT-5287] Render fails to ignore empty geometries in some very specific cases

fergonco:


aaime:
Manually merged to squash the two commits in one. Backported to 14.x and 13.x

PR #1039, opened 2015-11-12
[GEOT-5289] Use java.util.ServiceLoader instead of javax.imageio.spi.ServiceRegistry

don-vip:
PR for https://osgeo-org.atlassian.net/browse/GEOT-5289

The goal is to make GeoTools run with Java 9, now that [JDK-8068749](https://bugs.openjdk.java.net/browse/JDK-8068749) has been delivered.
As a consequence FactoryRegistry do not extends javax.imageio.spi.ServiceRegistry anymore but provides almost the same public API as before to minimize the impact as much as possible.

### DIFF
--- a/modules/library/main/src/main/java/org/geotools/filter/expression/PropertyAccessors.java
+++ b/modules/library/main/src/main/java/org/geotools/filter/expression/PropertyAccessors.java
@@ -2,7 +2,7 @@
  *    GeoTools - The Open Source Java GIS Toolkit
  *    http://geotools.org
  * 
- *    (C) 2002-2008, Open Source Geospatial Foundation (OSGeo)
+ *    (C) 2002-2015, Open Source Geospatial Foundation (OSGeo)
  *
  *    This library is free software; you can redistribute it and/or
  *    modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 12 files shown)


aaime:
Hi don-vip. This is great work, but I've found a licensing issue, some of the classes you added seem a slightly refactored version of OpenJDK own code, which is under a different licence, e.g.: 

http://hg.openjdk.java.net/jdk8u/jdk8u60/jdk/file/935758609767/src/share/classes/javax/imageio/spi/PartiallyOrderedSet.java

Not sure how to solve that, even if LGPL and GPL2+classpath are similar... they are not the same. @jodygarnett do you know what's the best course of action?

A clean rewrite of them would be probably best.

don-vip:
You're right, there is a licensing problem :( I don't know GeoTools source code very well, is there similar classes that could offer the same functionalities than PartiallyOrderedSet and DigraphNode?

aaime:
I'm not aware of similar classes... does not necessarily mean they are not there, GeoTools is too large to fit in one person's head :-p

don-vip:
Here's another possibility: use reflection: https://github.com/don-vip/geotools/commit/695db328bf647ee27c190634b01722120beecd90

If it's ok with this approach, should I create a new clean PR and close this one?

aaime:
@jodygarnett do you have an opinion on this one? Anyone else?

PR #1041, opened 2015-11-12
GEOT-5296: Apply WFS timeout to read timeout

awaterme:
WFSConfig.timeoutMillis is applied to read timeout of HTTPClient, too.
See [GEOT-5296](https://osgeo-org.atlassian.net/browse/GEOT-5296)

### DIFF
--- a/modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/WFSDataStoreFactory.java
+++ b/modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/WFSDataStoreFactory.java
@@ -107,6 +107,7 @@ public class WFSDataStoreFactory extends WFSDataAccessFactory implements DataSto
         http.setPassword(config.getPassword());
         int timeoutMillis = config.getTimeoutMillis();
         http.setConnectTimeout(timeoutMillis / 1000);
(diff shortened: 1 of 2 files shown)


modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/WFSDataStoreFactory.java
```diff
@@ -107,6 +107,7 @@ public WFSDataStore createDataStore(final Map<String, Serializable> params)
         http.setPassword(config.getPassword());
         int timeoutMillis = config.getTimeoutMillis();
         http.setConnectTimeout(timeoutMillis / 1000);
+        http.setReadTimeout(timeoutMillis / 1000);
```
aaime: Believe the very same change should be applied to WFSDataAccessFactory

awaterme:
Adjusted WFSDataAccessFactory, too.

aaime:
I see we have to update the copyright header year, but besides that, it looks good to me.
@NielsCharlier is the author of this module, what do you think Niels?

NielsCharlier:
Yeah that is fine (not the initial creator of the module, but indeed did a lot of work to make it usable)

PR #1044, opened 2015-11-13
[GEOT-5298] fixed read access while content changed

fgdrf:
It fixes broken Readers while internal storage has changed. Having this in mind the reader is temporary inconsintent until another Reader is created but only if quantity of features for this specific FeatureType in DataStore changed in between

### DIFF
--- a/modules/library/data/src/main/java/org/geotools/data/memory/MemoryFeatureReader.java
+++ b/modules/library/data/src/main/java/org/geotools/data/memory/MemoryFeatureReader.java
@@ -17,7 +17,11 @@
 package org.geotools.data.memory;
 
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
 import java.util.NoSuchElementException;
 
 import org.geotools.data.DataSourceException;
@@ -30,20 +34,27 @@ import org.opengis.feature.simple.SimpleFeature;
 import org.opengis.feature.simple.SimpleFeatureType;
 
 public class MemoryFeatureReader implements FeatureReader<SimpleFeatureType, SimpleFeature>{
-    ContentState state;
+
     SimpleFeatureType featureType;
     Iterator<SimpleFeature> iterator;
 
     public MemoryFeatureReader(ContentState state, Query query) throws IOException {
-        this.state = state;
         featureType = state.getFeatureType();
-        String typeName = getFeatureType().getTypeName();
-        MemoryDataStore store = (MemoryDataStore) state.getEntry().getDataStore();
-        iterator = store.features(typeName).values().iterator();
+
+        final MemoryDataStore store = (MemoryDataStore) state.getEntry().getDataStore();
+        final Map<String, SimpleFeature> features = store.features(featureType.getTypeName());
+        synchronized (features) {
+            final Collection<SimpleFeature> featureCollection = features.values();
+            final List<SimpleFeature> internalCollection = new ArrayList<SimpleFeature>();
+            if (features != null) {
+                internalCollection.addAll(featureCollection);
+            }
+            iterator = internalCollection.iterator();
+        }
     }
 
     public SimpleFeatureType getFeatureType() {
-        return state.getFeatureType();
+        return featureType;
     }
 
     public SimpleFeature next()
--- a/modules/library/data/src/test/java/org/geotools/data/memory/MemoryFeatureReaderTest.java
+++ b/modules/library/data/src/test/java/org/geotools/data/memory/MemoryFeatureReaderTest.java
@@ -1,0 +1,118 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ * 
+ *    (C) 2015, Open Source Geospatial Foundation (OSGeo)
+ *    
+ *    This library is free software; you can redistribute it and/or
+ *    modify it under the terms of the GNU Lesser General Public
+ *    License as published by the Free Software Foundation;
+ *    version 2.1 of the License.
+ *
+ *    This library is distributed in the hope that it will be useful,
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ *    Lesser General Public License for more details.
+ */
+package org.geotools.data.memory;
+
+import java.io.IOException;
+import java.util.NoSuchElementException;
+
+import org.geotools.data.DataTestCase;
+import org.geotools.data.DefaultTransaction;
+import org.geotools.data.FeatureReader;
+import org.geotools.data.FeatureWriter;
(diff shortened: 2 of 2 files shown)


modules/library/data/src/main/java/org/geotools/data/memory/MemoryFeatureReader.java
```diff
@@ -30,20 +34,27 @@
 import org.opengis.feature.simple.SimpleFeatureType;
 
 public class MemoryFeatureReader implements FeatureReader<SimpleFeatureType, SimpleFeature>{
-    ContentState state;
+
     SimpleFeatureType featureType;
     Iterator<SimpleFeature> iterator;
 
     public MemoryFeatureReader(ContentState state, Query query) throws IOException {
```
aaime: Not your fault but... why is the query even passed around?

modules/library/data/src/main/java/org/geotools/data/memory/MemoryFeatureReader.java
```diff
@@ -30,20 +34,27 @@
...
-        this.state = state;
         featureType = state.getFeatureType();
-        String typeName = getFeatureType().getTypeName();
-        MemoryDataStore store = (MemoryDataStore) state.getEntry().getDataStore();
-        iterator = store.features(typeName).values().iterator();
+
+        final MemoryDataStore store = (MemoryDataStore) state.getEntry().getDataStore();
+        final Map<String, SimpleFeature> features = store.features(featureType.getTypeName());
+        synchronized (features) {
```
aaime: I don't see corresponding synchronizations in the writer. It would seem this change makes the exception less likely, but not impossible?
What about using a Concurrent map also to store this map instead? Is the memory overhead too high?

modules/library/data/src/test/java/org/geotools/data/memory/MemoryFeatureReaderTest.java
```diff
@@ -0,0 +1,102 @@
+package org.geotools.data.memory;
```
aaime: Lacks copyright header

modules/library/data/src/test/java/org/geotools/data/memory/MemoryFeatureReaderTest.java
```diff
@@ -0,0 +1,102 @@
...
+        // start iterating through content
+        if (featureReader.hasNext()) {
+            featureReader.next();
+            currentFeatureCount++;
+        }
+        SimpleFeature newFeature = SimpleFeatureBuilder.template(roadType, null);
+    
+        memoryDataStore.addFeature(newFeature);
+    
+        while (featureReader.hasNext()) {
```
aaime: So this would fail without your changes?

modules/library/data/src/main/java/org/geotools/data/memory/MemoryFeatureReader.java
fgdrf: Ha, good question. Thats the next issue I'm going to file. Just want to
separate commits and bugs
Am 14.11.2015 14:53 schrieb "Andrea Aime" <notifications@github.com>:

aaime: Yep, thanks :-)

modules/library/data/src/main/java/org/geotools/data/memory/MemoryFeatureReader.java
fgdrf: Agree. I'm still evaluating how to refactor MemoryFeatureWriter because:
* Iterators are not synchronized if using Collections.synchronizedMap(), while access like put and get are --> Writer iterates to the end to add a new Feature ..
* implementing Iteration behavior and iterate to the feature + 1 on the next request could be a performance issue for writes  
* lack of knowledge of the workflow for Writers:
  * if iterator.hasNext() is false and next produces a new one
  * have the user to call write() before creating an other new one?
  * current Implementation looks like it would'nt work but I have to verifiy with a test
  * Ah! ContentFeatureStore.addFeatures(SimpleFeature, FeatureWriter) verifies my guessing, the user have to call write() after adding a feature

Where can I find some details about rendering behavior and event notifications on changes within a datastore. Normally if a User interacts with a map using UI its synchronized using UI-Thread. If the objects comming from a stream from another than the UI thread, how and when gets the map and its resources notified and re-rendered.

However I'm unsure about the current design for different reasons:
* the behavior is not really stateless (Iterator-Thing) 
* in addition it depends on a Query (what happens if the new features would match criterias for different Writers?)  

All in all I have a big Question Mark and looking forward for any feedback.

modules/library/data/src/test/java/org/geotools/data/memory/MemoryFeatureReaderTest.java
fgdrf: :+1: done

modules/library/data/src/main/java/org/geotools/data/memory/MemoryFeatureReader.java
aaime: Hi Frank, yes, one has to call write() to actually apply the changes.
About the UI notifications, I have no clue, GeoServer has no use of listeners so I'm not up to speed with their implementation and usage. I believe the users for events are using uDig and the gt-swing module, which is presently unmaintained (or, and possibly gt-swt too). Maybe ask on the developer list.

aaime:
@fgdrf, am I right in saying that we need @jodygarnett attention on events and synchronization comments above before merging this one?

modules/library/data/src/main/java/org/geotools/data/memory/MemoryFeatureReader.java
```diff
@@ -30,20 +34,27 @@
...
 
     public MemoryFeatureReader(ContentState state, Query query) throws IOException {
-        this.state = state;
         featureType = state.getFeatureType();
-        String typeName = getFeatureType().getTypeName();
-        MemoryDataStore store = (MemoryDataStore) state.getEntry().getDataStore();
-        iterator = store.features(typeName).values().iterator();
+
+        final MemoryDataStore store = (MemoryDataStore) state.getEntry().getDataStore();
+        final Map<String, SimpleFeature> features = store.features(featureType.getTypeName());
```
jodygarnett: This is driving me bonkers, the point is to have the Map<String,SimpleFeature> on each Entry as the canonical truth. And a diff map on each entry that could be used by readers/writers.

I looks like this transition to contentdatastore was done with a minimal amount of code change, introducing all kinds of opportunity for error.

jodygarnett:
@fgdrf I am content with your change, would like to consider moving to a better design in the future.

aaime:
@jodygarnett any reason why this has not been merged then? :)

aaime:
@jodygarnett what do we do with this one?

jodygarnett:
It is a pull request against 14.x and not master.

jodygarnett:
Applied to master as https://github.com/geotools/geotools/commit/eeb504df3a836b447077037bfe75cd21ec5e4b4b

PR #1048, opened 2015-11-16
[GEOT-5302] remove support for EnvFunction lookup tables

fgdrf:
see https://osgeo-org.atlassian.net/browse/GEOT-5302 for details and comments.

Implemented option 1 and added two functions that provides remove support. In addition I updated the javadoc comment for setGlobalValue

### DIFF
--- a/modules/library/main/src/main/java/org/geotools/filter/function/EnvFunction.java
+++ b/modules/library/main/src/main/java/org/geotools/filter/function/EnvFunction.java
@@ -20,12 +20,12 @@ package org.geotools.filter.function;
 import static org.geotools.filter.capability.FunctionNameImpl.parameter;
 
 import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentMap;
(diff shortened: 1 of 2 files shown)


modules/library/main/src/main/java/org/geotools/filter/function/EnvFunction.java
```diff
@@ -90,6 +90,7 @@
  *
```
jodygarnett: update header

jodygarnett:
Do we have any documentation on how EnvFunctions work that should be updated as part of this pull request?

fgdrf:
Update javadoc and added serveral SLD & code java examples how to use isNull and isNil. Added serveral test cases to verify behavior.
EnvFunction.setGlobalValue() now supports null values, changed internal Map from ConcurrentHashMap to a synchronized HashMap to be thread save for gets and puts.

modules/library/main/src/main/java/org/geotools/filter/function/EnvFunction.java
```diff
@@ -20,12 +20,12 @@
 import static org.geotools.filter.capability.FunctionNameImpl.parameter;
```
jodygarnett: Please update header

jodygarnett:
Other then updating headers this one is good to go.

bencaradocdavies:
@fgdrf I think this was manually merged by @jodygarnett on 14.x and master.

PR #1049, opened 2015-11-17
[GEOT-5307] Fixes to Footprint behavior

tbarsballe:
Two minor, related imageMosaic fixes:
* [GEOT-5307] Retain source images when extractin alpha channel
* Catch exceptions thrown by SidecarFootprintProvider implementations. I am not sure if this one actually causes problems, but one of the implementations did throw an exception when I was stepping through it in the debugger, so it is at least possible.


aaime:
I believe you forgot to commit the tests?

tbarsballe:
A test case for the first issue exists here: https://github.com/tbarsballe/geotools/commit/b4c54b5f3997aaf03b945d5a36a4206cbffa5b9f
However, it is waiting on another similar bug before it will actually pass; see https://osgeo-org.atlassian.net/browse/GEOT-5306

tbarsballe:
I don't have a test for the second one, as I am not sure it can cause problems during normal excecution, at least not in a way that would be visible to a test case (any exceptions thrown get caught by the outer catch block; this could cause a provider to be skipped, but so far has only occured on the last provider in the list)

aaime:
We cannot merge without a test case. Can you create one that only fails with the specific fix done here?

aaime:
And oh, the commit messages should be in the form "[GEOT-XYWZ] title of the ticket"

tbarsballe:
Sure, I will create one. It is a bit awkward that this will be a subset of the GEOT-5306 test case, but I guess it is worth it so we can merge now, in case GEOT-5306 doesn't get fixed anytime soon.

tbarsballe:
I can rebase to update the commit messages as well; at time of commit the GEOT ticket didn't exist yet.

tbarsballe:
Updated with a test case; squashed, rebased, and renamed commit.

tbarsballe:
Updated test case; looks like image byte values behave differently on different systems :P

aaime:
Looks good to me, again, best to also have @dromagnoli opinion

dromagnoli:
Looks good to me too.

PR #1051, opened 2015-11-18
[GEOT-5306] Handle mosaic with alpha and ROI

tbarsballe:
The JAI mosiac operator ignores the ROI when computing source weights if an alpha channel is also provided.
This fix adds a step to premultiply the ROI by the alpha channel before doing a mosiac operation (assuming both are provided).
As part of his fix, the JAI "Multiply" operation has been added to org.geotools.image.ImageWorker


tbarsballe:
Note: This change depends on #1049 to pass tests; Travis CI will not succed until that PR has been merged.

aaime:
Changes look sensible to me, but I'd prefer to have @dromagnoli opinion's as well

dromagnoli:
Looks good to me. I have only 2 notes to add:
1) roi.getAsImage may load all the data in memory in case roi is a ROIShape instance. However, there are already other pieces of code which call getAsImage so I'm not suggesting to improve that in this pull request. We should rework these calls (as an instance by converting ROIShape to ROIGeometry which uses deferred loading for getAsImage) into a different JIRA, so I'm fine with your change.
2) I have also seen other pieces of code doing alpha Masking on top of ROI, using Mosaic op instead of Multiply (as an instance on gt-coverage's Mosaic). We can consider setting up some utility method to do alpha masking based on ROI and re-use that where needed. Again, I'm not saying that we need to do that in this pull request but we should think about that soon or later.

tbarsballe:
@dromagnoli Thanks for the review. I will merge this in after #1049 gets a review + merge, as it is a prerequesite for this fix to pass tests.

For (1) - good to know, I saw that call being used in a bunch of places.
For (2) - from what I saw, imagemosaic masks the final mosaic with the mosaic'd ROIs using a mosiacROI method in ImageWorker.

PR #1052, opened 2015-11-20
[GEOT-4506] Fix a race condition when initialising JDBCDataStore type mappings

dpovey:
Note: this fix is the smallest change that fixes the observed ConcurrentModificationException, however there is probably more work to be done to make this class properly threadsafe.  The added test was proved to fail reliably before the fix was made.


aaime:
Hi @dpovey , looks good, thanks for the test. The test class needs copyright headers. Since it's new it's best if you add them yourself.

dpovey:
Done,  Also had to remerge the master due to someone else adding a test
with the same name.  Thanks for the review.

aaime:
Hi @dpovey , I've manually merged the changes (keep you as an author) and backported them to 14.x and 13.x
Thanks again for the contribution

PR #1054, opened 2015-11-21
[GEOT-5311] ShapefileDumper, utility class to dump generic feature collection to shapefile(s)

aaime:


modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/RemappingFeatureCollection.java
```diff
@@ -0,0 +1,227 @@
...
+import org.opengis.feature.type.AttributeDescriptor;
+import org.opengis.feature.type.GeometryDescriptor;
+import org.opengis.filter.Filter;
+
+/**
+ * FeatureCollection that remaps attribute names using a given map.
+ * 
+ * @author Mauro Bartolomeoli, mbarto@infosia.it
+ */
+class RemappingFeatureCollection extends DecoratingSimpleFeatureCollection {
```
aaime: Backported from GeoServer. Could have used the transform feature store instead but that would have added a dependency, and this code has been in use for some time without issues.

PR #1062, opened 2015-12-02
remove dead code in PropertyDescriptorImpl

hgryoo:
When type is null, the code that throws NullPointerException is repeated.


aaime:
Thank you.

PR #1063, opened 2015-12-02
[GEOT-5318] GridCoverageRenderer fails on indexed image with null background color

dvntucker:
Added a null check in the GridCoverageRender when performing color model expansion. Shouldn't need to expand the color model if no background color was provided.

Includes unit test.
No doc changes needed.


tbarsballe:
Looks good to me.

PR #1064, opened 2015-12-05
[GEOT-5289] Use java.util.ServiceLoader instead of javax.imageio.spi.ServiceRegistry (Reflection approach)

don-vip:
PR for https://osgeo-org.atlassian.net/browse/GEOT-5289 taking into account feedbacks from #1039 

The goal is to make GeoTools run with Java 9, now that [JDK-8068749](https://bugs.openjdk.java.net/browse/JDK-8068749) has been delivered.
As a consequence FactoryRegistry do not extends javax.imageio.spi.ServiceRegistry anymore but provides almost the same public API as before to minimize the impact as much as possible.


jodygarnett:
As per email I think we may want to make our own service registry replacement.

don-vip:
Is there an e-mail discussion somewhere? I can't find it on https://sourceforge.net/p/geotools/mailman/geotools-devel I'm not sure to look at the right place...

aaime:
Hi, I've create a PartiallySortedSet similar to the javax one (but written from scratch, although based on the same graph concepts) here: https://github.com/geotools/geotools/blob/master/modules/library/xml/src/main/java/org/geotools/PartiallyOrderedSet.java

We need to make sure also the other classes we need are clean IP wise.

jodygarnett:
Putting this on hold until after code freeze. I would like to make this a priority for the next release cycle (to line up with the scheduled java 9 release)

bencaradocdavies:
Code freeze is over. @jodygarnett?

karussell:
We are interested in this change as well. Can we support it somehow (time, money, review)?

jodygarnett:
Code freeze is over (x2). Java 9 has been delayed until 2017.

Would love to knock this one out if there are willing parties to work/fund this.

jodygarnett:
I would like to double check that the FactoryRegistery is all new work (and not copied from javax).

kamenik:
Fresh news - JDK 9 Developer Preview is now available. Could you please look at this pull request again?

aaime:
Last time I checked this pull request could not be even considered because it seems to contain copy/pasted code from the JDK. Has anything changed regarding that?

don-vip:
Nothing has changed as I understood you were going to rewrite the PR in your own way?

kamenik:
I thought it is ok with your version of PartiallyOrderedSet. What is missing yet?

aaime:
@kamenik going by memory, did not check, there should still be a large chunk of copy/pasted code in FactoryRegistry, but I might be wrong (you'll have to click open the diff, it's too big and github does not show it by default)

aaime:
@don-vip it does not have to be "my way", it just needs to be written from scratch instead of copying code off the JDK.

don-vip:
@aaime I understand your concern, but after having updated my PR to latest GeoTools codebase that's not something I'll do. It is a major rework that should be done by GeoTools developers. For now I'm using my fork, which works fine. Feel free to reuse the parts of my PR that are OK for GeoTools and rewrite the parts that are not: https://github.com/don-vip/geotools/tree/JDK-8068749-reflection-v4 (I don't create a new PR as I know it won't be accepted for the code added in FactoryRegistry). As a reminder Java 9 is released in two months!

ansell:
The only changes necessary to be compile compatible with Java-9 for this particular case is to change all references to ``javax.imageio.spi.ServiceRegistry.lookupProviders`` to ``java.util.ServiceLoader.load`` and add ``.iterator()``. ``java.util.ServiceLoader`` was added in Java-6, so it isn't necessary to actually move to Java-9 to use it.

This patch seems to add extra complexity on top of the basic changes necessary. I will write up a simpler PR and propose it as a replacement for this, as this is such a small issue, it shouldn't stop people from further testing with Java-9-ea/rc.

ansell:
Sorry for my confusion earlier, I had looked at the simple modules requiring just the lookupProviders API to be migrated to ServiceLoader.load and missed the inheritance of ServiceRegistry in the FactoryRegistry class. 

Is it acceptable to either rewrite FactoryRegistry or fully replace the current FactoryRegistry without keeping backwards compatibility, given GeoTools is regularly bumping major versions? 

Alternatively, can the public ServiceRegistry API (from Javadoc) be used to replace any required missing methods to FactoryRegistry using the ServiceLoader API, without legal concerns about copying from the GPL OpenJDK to this LGPL project. I haven't looked into the complexity of that yet as I don't want to prejudice any coding I do, having only directly used ServiceLoader.load and ServiceRegistry.lookupProviders in my own code in the past.

aaime:
@ansell we try very hard to avoid breaking backwards compatibility. Unfortunately I did no research in this topic yet, so I don't have an answer to your questions, sorry! :-(

ansell:
Thanks for getting back to me. The complexity is due to the extension of ServiceRegistry by FactoryRegistry, making the entire ServiceRegistry API part of the GeoTools API. At minimum we need to keep the functionality that the other GeoTools modules are using, but I would also like to keep backwards compatibility if possible. 

I got stuck on my first two attempts because I reworked the generics during the process, but will try again at some stage soon to see if it can be done cleanly without touching the generics parameters (and without looking at OpenJDK source code). This is blocking my progress with checking that my projects still work with Java-9 so it will still be a high priority for me.

nicolaiparlog:
In #1670, I extended the `FactoryRegistry` with the `ServiceRegistry` methods GeoTools code is calling without copying any JDK code. I used GeoTools' `PartiallyORderedSet`.

don-vip:
Closing this PR in favor of Nicolai's :)

PR #1067, opened 2015-12-14
GEOT-5322 - better clipping of invalid polygons

DBlasby:
Moved clip() from StreamingRenderer to GeometryClipper.

clipFailResistant could be improved, but is better than what's there at the moment.

Added testcase.  Removed some commented out code.


modules/library/main/src/main/java/org/geotools/geometry/jts/GeometryClipper.java
```diff
@@ -76,6 +79,49 @@ public GeometryClipper(Envelope bounds) {
...
+     * If this fails, the original geometry is returned.
+     * 
+     * see {@link #clip(Geometry, boolean) clip}
+     * 
+     * @param g
+     * @param ensureValid
+     * @param scale 0=double precision Precision Model
+     * @return
+     */
+    public Geometry clipFailResistant(Geometry g, boolean ensureValid, double scale) {
```
aaime: robustClip maybe?

modules/library/main/src/test/java/org/geotools/geometry/jts/GeometryClipperTest.java
```diff
@@ -1,3 +1,20 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ * 
+ *    (C) 2015, Open Source Geospatial Foundation (OSGeo)
```
aaime: Thanks for adding the copyright headers. The year should be 2010 - 2015, as shown by the history of file at:

````
svn log http://svn.osgeo.org/geotools/trunk/modules/library/main/src/test/java/org/geotools/geometry/jts/GeometryClipperTest.java
````

File: modules/library/main/src/main/java/org/geotools/geometry/jts/GeometryClipper.java [hunk 3]
```diff
@@ -76,6 +79,49 @@ public GeometryClipper(Envelope bounds) {
...
+        try {
+            return clip(g, ensureValid); 
+        } catch(TopologyException e) {
+            // move the points to a precision grid
+            GeometryPrecisionReducer reducer;
+            if (scale == 0)
```
aaime: Our code conventions demand braces around single line if/else blocks.

modules/library/main/src/main/java/org/geotools/geometry/jts/GeometryClipper.java
```diff
@@ -76,6 +79,49 @@ public GeometryClipper(Envelope bounds) {
...
+                reducer = new GeometryPrecisionReducer(new PrecisionModel());//double precision model
+            else
+                reducer = new GeometryPrecisionReducer(new PrecisionModel(scale));
+            Geometry reduced = reducer.reduce(g); //this will try to fix problems with the geometry.
+            
+            //if the reducer can not construct a result, then return the unclipped geometry
+            // alternatively, we could try the clip with ensureValid=false
+            if (reduced.isEmpty())
```
aaime: Same here

modules/library/main/src/main/java/org/geotools/geometry/jts/GeometryClipper.java
aaime: Hum... hum... not robust either... it's more like "clip if you can, don't if you don't". It made sense in the context of streaming renderer, where not clipping is not always fatal (it is only if they line is dashed and very long compared to the viewing area, an instant OOM generator due to a JDK bug), not sure it does make sense as a public method though. @jodygarnett , what do you think?

DBlasby: If you use scale=0, then it will attempt to fix invalid polygons - inside the precisionmodel code  It does a polygon.buffer(0), which will often "fix" an invalid polygon (but not always).  I can add that explicitly - its probably better to attempt the buffer(0) first, then attempt to re-compute the intersection, and then move the points around (precisionmodel).  That's likely a bit better (but more time-consuming) than that what the original code did (buffer(0) might be failing because its being applied after the precisionmodel instead of trying it both before and after).  The original code wasn't noticing the precisionmodel failures - which is what this code does.  I expect the original author thought that if the precisionmodel failed it would throw an exception.  

I could keep the original code and just add;
if (reduced.isEmpty()) {
       throw TopologyException(...);
   }

 I didn't do much work making it more robust than it was  - I didn't see any references to the clipper outside the renderer.  I'd need more test data and use cases to improve it.  The main problem is GIGO - invalid polygons aren't easy to robustly work with.  I added the function here so it could grow, instead of everyone "doing a little" in their own code (i.e. the clip() method in streamingrenderer).

I can move this function back into streamingrenderer, but seems a bit weird to write computation geometry code and test cases in the streamingrenderer instead of in the GeometryClipper (or some type subclass/wrapper).  

So, I can;
a) Move code back to StreamingRenderer, just do the .isEmpty()/throw
b) keep code where it is, but be more aggressive in trying to convert the invalid polygon to a valid polygon (i.e. add an explicit buffer(0))
c) add GeometryClipper Interface, and then a pixel-based renderer wrapper 

I'd recommend (b) or (c) - (a) is easier but I don't really think the CG code and tests should be in the rendrerer and the code is likely a useful starting place for others.

To get actual robust clipping with invalid polygons would be a very very big task (and we havent even talked about what to do when JTS cannot compute the intersection even with valid polygons).  Everyone might have a different opinion on what to do - and likely require some type of hints into the function (i.e. perturbPoints=allowed, moveClipRectange=allowed, etc...).  Like I said - need lots of use cases and test cases.  For example, it might make sense to, if the polygon is invalid, use the simpler non-robust clipping method since the input is already invalid and likely ok to return something invalid.

aaime: Maybe just leave it there and call it "tryClip" or whatever makes it evident that a clip is attempted, but might not succeed, and that will not make it throw an exception.
Or call it robustClip, and make it throw an exception in case of failure, that will be caught by the StreamingRenderer and just move on using the original geometry.

aaime:
Hum... I have closed this one by mistake (don't remember doing it), reopening

modules/library/render/src/main/java/org/geotools/renderer/lite/StreamingRenderer.java
```diff
@@ -2557,7 +2557,7 @@ private void processSymbolizers(final Graphics2D graphics,
                     // we need to preserve the topology if we end up applying buffer for perp. offset
                     boolean preserveTopology = style instanceof LineStyle2D && ((LineStyle2D) style).getPerpendicularOffset() != 0 &&
                             (source instanceof Polygon || source instanceof MultiPolygon);
-                    Geometry g = clip(shape.getGeometry(), clipper, preserveTopology);
```
DBlasby: I have a question about this.  I originally just thought it was weird terminology, but I think its a bug.
If you look you'll see the "boolean preserveTopology".  However, if you look at the original code (clip() function in streaming renderer, removed in this pull request) you'll see that it actually flips this boolean (line 2611).

I expect the flip is incorrect, but not sure.

NOTE: if you don't flip it, then it uses the non-robust clip() which doesnt fail.

If flipping is correct, then I'll change it to;
 Geometry g = clipper.clipFailResistant(shape.getGeometry(),  !preserveTopology,  1);

Or, better;
 boolean dontPreserveTopology = ...
Geometry g = clipper.clipFailResistant(shape.getGeometry(),  !dontPreserveTopology,  1);

modules/library/render/src/main/java/org/geotools/renderer/lite/StreamingRenderer.java
```diff
@@ -2605,22 +2605,6 @@ private void processSymbolizers(final Graphics2D graphics,
         }
     }
 
-    private Geometry clip(Geometry geometry, final GeometryClipper clipper,
-            boolean preserveTopology) {
-        try {
-            return clipper.clip(geometry, !preserveTopology);
```
DBlasby: preserveTopology is flipped here.

GeometryClipper : 
 public Geometry clip(Geometry g, boolean ensureValid)

modules/library/main/src/main/java/org/geotools/geometry/jts/GeometryClipper.java
DBlasby: I did a quick test.  For my polygon, doing a buffer(0) makes it valid.  I'll add this in.

modules/library/main/src/main/java/org/geotools/geometry/jts/GeometryClipper.java
```diff
@@ -76,6 +79,58 @@ public GeometryClipper(Envelope bounds) {
...
+     * Otherwise, it will return the original (unclipped) geometry. 
+     * 
+     * see {@link #clip(Geometry, boolean) clip}
+     * 
+     * @param g
+     * @param ensureValid
+     * @param scale 0=double precision Precision Model
+     * @return
+     */
+    public Geometry tryClip(Geometry g, boolean ensureValid, double scale) {
```
jodygarnett: I will think about this one a bit since it is a public method ...

modules/library/render/src/main/java/org/geotools/renderer/lite/StreamingRenderer.java
aaime: Yeah, by a cursory look it really looks like a mistake. Funny... I made several hand checks for it, must have been a last minute change.

DBlasby: As a side effect, I think the map is drawing quicker!  I don't have great metrics, but I think it is.

jodygarnett:
Squashed and merged as 33a125eea0bda1cebe6e6a187210c7cb387539cd 

This is suitable for back port to 14.x

aaime:
Guys, you merged code that's eating exceptions... most static analysis tools will flag this as a violation.
Could you please log the exceptions at FINE or FINER level instead, guarding it to avoid the overhead when that level is not enabled, like this:

https://github.com/geotools/geotools/blob/master/modules/library/render/src/main/java/org/geotools/renderer/crs/ProjectionHandler.java#L532

PR #1070, opened 2015-12-17
[GEOT-5330] Transformer.transformQuery() does not preserve sortBy filter

ridethepenguin:
See [related JIRA issue](https://osgeo-org.atlassian.net/browse/GEOT-5330) for details.


aaime:
Feel free to backport as needed

PR #1077, opened 2016-01-02
Add removeSchema support to geobuf DataStore.  Fixes issue GEOT-5332

jericks:
Fixes https://osgeo-org.atlassian.net/browse/GEOT-5332.

Thanks!
Jared


jericks:
For what its worth, the failure is not Geobuf related, but in the CSS module:

[INFO] Geobuf DataStore ................................... SUCCESS [  7.072 s]
[INFO] Cartographic CSS parser ............................ FAILURE [  7.851 s]

Jared

aaime:
Compile issue in CSS fixed on master. Strange, does not happen with all java 8 versions.

PR #1078, opened 2016-01-03
Make Property DataStore support removeSchema.  Solves issue GEOT-5333.

jericks:
https://osgeo-org.atlassian.net/browse/GEOT-5333

Thanks!
Jared

### DIFF
--- a/modules/plugin/property/src/main/java/org/geotools/data/property/PropertyDataStore.java
+++ b/modules/plugin/property/src/main/java/org/geotools/data/property/PropertyDataStore.java
@@ -2,7 +2,7 @@
  *    GeoTools - The Open Source Java GIS Toolkit
  *    http://geotools.org
  *
- *    (C) 2002-2014, Open Source Geospatial Foundation (OSGeo)
+ *    (C) 2002-2016, Open Source Geospatial Foundation (OSGeo)
  *
  *    This library is free software; you can redistribute it and/or
  *    modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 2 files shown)


aaime:
Looking good to me. @jodygarnett you are marked as the module maintainer in the pom.xml, can you confirm?

aaime:
Ah, @tbarsballe is the author of the current store version, so we could also get feedback from him.

modules/plugin/property/src/main/java/org/geotools/data/property/PropertyDataStore.java
```diff
@@ -147,4 +147,21 @@ protected ContentFeatureSource createFeatureSource(ContentEntry entry) throws IO
             return new PropertyFeatureSource( entry, Query.ALL );
         }
     }
+
+    @Override
+    public void removeSchema(Name typeName) throws IOException {
+        super.removeSchema(typeName.getLocalPart());
```
tbarsballe: This will always throw an UnsupportedOperationException. Should be:
```
removeSchema(typeName.getLocalPart());
```

In order to call the other implementation.

Maybe also add a test of this method (using a `Name`) to your test case?

tbarsballe:
Other than the one issue I pointed out, this looks good.

I do have one other minor comment - the removeSchema implementations use an explicit `@Override` annotation. It would be nice if `createSchema` also did this, if only for consitencies sake (see [here](https://github.com/jericks/geotools/blob/GEOT-5333/modules/plugin/property/src/main/java/org/geotools/data/property/PropertyDataStore.java#L82)); if you could add that while fixing the one removeSchema method, that would be appreciated.

jericks:
Thanks @tbarsballe!  I fixed the removeSchema(Name) method and added the Override where appropriate.

PR #1079, opened 2016-01-06
GEOT-5334, fixing overflow in ClassificationFunction.round.

jdeolive:
Fixed by removing intermediate conversion to int. Removed uneccessary use of
Long wrapper class as well.


aaime:
Looks good to me.

jdeolive:
Thanks for the review Andrea!

PR #1084, opened 2016-01-18
[14.x] XML serialialization bug for GeometryCollection (Generic)

rishg:
There is a bug in xml serialization module when using generic geometrycollection, that is, a WKT : GEOMETRYCOLLECTION ( LINESTRING (...), POLYGON (...), POINT (...) ). However, this is something that is supported in WMS.

### DIFF
--- a/modules/extension/xsd/xsd-gml3/src/main/java/org/geotools/gml3/simple/GML32FeatureCollectionEncoderDelegate.java
+++ b/modules/extension/xsd/xsd-gml3/src/main/java/org/geotools/gml3/simple/GML32FeatureCollectionEncoderDelegate.java
@@ -16,6 +16,7 @@
  */
 package org.geotools.gml3.simple;
 
+import com.vividsolutions.jts.geom.GeometryCollection;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


aaime:
Check the contribution guide: https://github.com/geotools/geotools/blob/master/CONTRIBUTING.md
For a change like this one we need a contribution agreement singed and sent, a test to go with the code, and the pull request must be done against master (single commit) and then we can backport it to 14.x

rishg:
@aaime Thank you for pointing it out. Added some unit test but build failed. So, kindly ignore the PR #1085 . Will fix and contribute after the agreement.

rishg:
@aaime Do you think I can push https://github.com/rishg/geotools/commit/735abef17023b8eeca7d952c665ff7cbaf04cd03 as bug fix to master.  We are doing the following: 
```
Add shortcut when querying for global min/max in Solr plugin.

- Uses Solr's the sort feature to only query for one result instead of the entire Solr core.
- Also supports a NearestVisitor for dates.
- Fixes slowness in geoserver's GetCapabilities call on a large Solr core.
```

PR #1085, opened 2016-01-18
[14.x] Bug fix for XML serialization of GeometryCollection (Generic)

rishg:
Proposing a fix for xml serialization of GeometryCollection (example GEOMETRYCOLLECTION (POINT (19 19), LINESTRING (x y, y z) ). Generic geometry collections gets visualized in WMS but are not in WFS.


rishg:
Need to fix unit test. Got an email from moderator for contribution agreement.

PR #1086, opened 2016-01-20
Add config level for log4jlogger [GEOT-5625]

jodygarnett:
This will allow config to be used, it ends up one step below INFO level.

* https://osgeo-org.atlassian.net/browse/GEOT-5625

### DIFF
--- a/modules/library/metadata/src/main/java/org/geotools/util/logging/Log4JLogger.java
+++ b/modules/library/metadata/src/main/java/org/geotools/util/logging/Log4JLogger.java
@@ -18,7 +18,6 @@ package org.geotools.util.logging;
 
 import java.util.logging.Level;
 
-
 /**
  * An adapter that redirect all Java logging events to the Apache's
  * <A HREF="http://logging.apache.org/log4j">Log4J</A> framework.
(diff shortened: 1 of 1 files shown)


modules/library/metadata/src/main/java/org/geotools/util/logging/Log4JLogger.java
```diff
@@ -33,6 +32,10 @@
  * @see Logging
  */
 final class Log4JLogger extends LoggerAdapter {
+    
+    private static final int CONFIG_INT = org.apache.log4j.Level.INFO_INT-1;
+    private static final org.apache.log4j.Level CONFIG = org.apache.log4j.Level.toLevel( CONFIG_INT, org.apache.log4j.Level.TRACE );
```
jodygarnett: default level should be DEBUG here (sigh!)

aaime: Needs some test and docs for config log4j

jodygarnett:
Putting this on hold until after code freeze.

bencaradocdavies:
Code freeze is over.

jodygarnett:
Does anyone care to persue this? Would love to have less verbose startup.

jodygarnett:
Attaching to bug ticket and closing

PR #1091, opened 2016-01-28
ignore transitive dependency from picocontainer to xml-apis

jodygarnett:
We do not use any XML configuration with piccontainer (the XSD module configures dependency injection by hand).


aaime:
No objection, but wondering what's the driver for this change? :-) Just a curiosity.

PR #1092, opened 2016-01-28
fixes for default types as String & improved logging

ianturton:


modules/unsupported/wps/src/main/java/org/geotools/data/wps/WPSProcess.java
```diff
@@ -124,20 +126,28 @@ protected WPSProcess(ProcessFactory factory) {
...
 			return null;
 		} catch (IOException e) {
+			LOGGER.info(e.getMessage());
 			return null;
 		}
 		
 		// if there is an exception in the response, return null
 		// TODO:  properly handle the exception?
-		if (response.getExceptionResponse() != null || response.getExecuteResponse() == null) {
+		if (response.getExceptionResponse() != null) {
```
aaime: Wouldn't it be better to throw an exception instead?

aaime:
Copyright headers not updated, no ticket reference, no test.

aaime:
Cough cough :-)

modules/unsupported/wps/src/main/java/org/geotools/data/wps/WPSProcess.java
jodygarnett: Throw exception, or at least change the logger level to "fine"

modules/unsupported/wps/src/main/java/org/geotools/data/wps/WPSProcess.java
```diff
@@ -53,7 +55,7 @@
  * @source $URL$
  */
 public class WPSProcess extends AbstractProcess {
-
+	private static final Logger LOGGER = Logging.getLogger(WPSProcess.class.getName());
```
jodygarnett: why use of getName() here?

jodygarnett:
This change is too late for the 15-beta I am happy to see you working on this module.

jodygarnett:
@ianturton ping?

PR #1096, opened 2016-02-02
Fix javadoc to be consistent with the code

mprins:
Several different names were used in the doc+comments for the primary key metadata table.

fixes [GEOT-4886](https://osgeo-org.atlassian.net/browse/GEOT-4886)


jodygarnett:
So no actually change of functionality here then ...

mprins:
@jodygarnett that is correct

PR #1098, opened 2016-02-03
[GEOT-5357] Add shortcut when querying for global min/max in Solr plugin.

rishg:
- Uses Solr's  sort feature to only query for one result instead of the entire Solr core.
- Also supports a NearestVisitor for dates.
- Fixes slowness in geoserver's GetCapabilities call on a large Solr core.

### DIFF
--- a/modules/unsupported/solr/src/main/java/org/geotools/data/solr/SolrFeatureSource.java
+++ b/modules/unsupported/solr/src/main/java/org/geotools/data/solr/SolrFeatureSource.java
@@ -2,7 +2,7 @@
  *    GeoTools - The Open Source Java GIS Toolkit
  *    http://geotools.org
  * 
- *    (C) 2014, Open Source Geospatial Foundation (OSGeo)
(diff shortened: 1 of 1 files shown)


aaime:
No objection on the change, looks like an improvement, yet it's not following the rules stated at https://github.com/geotools/geotools/blob/master/CONTRIBUTING.md (that github asked you to review when creating the pull request).

In particular:
* Copyright headers have not been updated
* There are no tests
* The commit message does not refer to an existing ticket

aaime:
You are right about the tests, as a performance optimization no visible behavior changes. It's ok not to have tests in this case. 
The commit message has not been fixed yet, you can rebase your commits locally, squash them into one, reword it, and then push --force to the branch that created this pull request, the pull request will be updated.

aaime:
Hi @jdeolive do you think it's good to go?

jdeolive:
Yup, the change looks good me, nice optimization.

PR #1101, opened 2016-02-05
Correcting SimpleDateFormat String

emerkle826:
The Format string has the MONTH (MM) and MINUTES (mm) keys swapped.
  This PR fixes that.


emerkle826:
Need to write tests, closing.

PR #1102, opened 2016-02-05
GEOT-5355 Correcting SimpleDateFormat String

emerkle826:
The Format string has the MONTH (MM) and MINUTES (mm) keys swapped.
Correcting the format string and adding assertions to tests.

### DIFF
--- a/modules/unsupported/geopkg/src/test/java/org/geotools/geopkg/GeoPackageTest.java
+++ b/modules/unsupported/geopkg/src/test/java/org/geotools/geopkg/GeoPackageTest.java
@@ -2,7 +2,7 @@
  *    GeoTools - The Open Source Java GIS Toolkit
  *    http://geotools.org
  *
- *    (C) 2002-2015, Open Source Geospatial Foundation (OSGeo)
+ *    (C) 2002-2016, Open Source Geospatial Foundation (OSGeo)
  *
  *    This library is free software; you can redistribute it and/or
  *    modify it under the terms of the GNU Lesser General Public
@@ -29,9 +29,13 @@ import java.sql.Connection;
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.Statement;
+import java.text.SimpleDateFormat;
 import java.util.ArrayList;
+import java.util.Calendar;
+import java.util.Date;
 import java.util.List;
 import java.util.Set;
+import java.util.TimeZone;
 import java.util.logging.Level;
 
 import org.apache.commons.io.FileUtils;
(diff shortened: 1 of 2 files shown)


modules/unsupported/geopkg/src/test/java/org/geotools/geopkg/GeoPackageTest.java
```diff
@@ -164,6 +168,45 @@ void assertTableExists(String table) throws Exception {
...
+                Date parsedDate = sdf.parse(dateString);
+                // assert the month value is the same for NOW and the parsed date.
+                assertEquals("Month value does not match", new SimpleDateFormat("MMMM").format(now),
+                    new SimpleDateFormat("MMMM").format(parsedDate));
+                // assert the minute value is the same for NOW and the parsed date.
+                assertEquals("Minutes value does not match", new SimpleDateFormat("mm").format(now),
+                    new SimpleDateFormat("mm").format(parsedDate));
+            }
+        }
+        finally {
```
smithkm: try with resources would be clearer and more concise

smithkm:
The copyright headers need to be updated to 2016.

You should report this as a bug in JIRA and mention the ticket here (Note that the other PRs mention a GEOT-#### ticket).  The JIRA tickets are used to generate the release notes so if you don't do this, no one will know about your fix.

The OSGeo JIRA is here: https://osgeo-org.atlassian.net

emerkle826:
@smithkm Thanks for the feedback. I created the JIRA ticket (https://osgeo-org.atlassian.net/browse/GEOT-5355), updated the PR title and made the recommended code changes.

smithkm:
OK, looks good.  Should be an easy backport to the stable and maintenance branches.

modules/unsupported/geopkg/src/test/java/org/geotools/geopkg/GeoPackageTest.java
```diff
@@ -164,6 +168,35 @@ void assertTableExists(String table) throws Exception {
...
+            if (rs.next()) {
+                final String dateString = rs.getString(1);
+                // parse the date with the expected format string
+                Date parsedDate = sdf.parse(dateString);
+                // assert the month value is the same for NOW and the parsed date.
+                assertEquals("Month value does not match", new SimpleDateFormat("MMMM").format(now),
+                    new SimpleDateFormat("MMMM").format(parsedDate));
+                // assert the minute value is the same for NOW and the parsed date.
+                assertEquals("Minutes value does not match", new SimpleDateFormat("mm").format(now),
+                    new SimpleDateFormat("mm").format(parsedDate));
```
bencaradocdavies: @emerkle826 this is a race condition. What makes you think that ```now``` is in the same minute at the fixture set up time?

modules/unsupported/geopkg/src/test/java/org/geotools/geopkg/GeoPackageTest.java
```diff
@@ -164,6 +168,35 @@ void assertTableExists(String table) throws Exception {
...
+            Statement st = cx.createStatement();
+            ResultSet rs = st.executeQuery(String.format("SELECT last_change FROM %s;",
+                GeoPackage.GEOPACKAGE_CONTENTS))){
+            
+            if (rs.next()) {
+                final String dateString = rs.getString(1);
+                // parse the date with the expected format string
+                Date parsedDate = sdf.parse(dateString);
+                // assert the month value is the same for NOW and the parsed date.
+                assertEquals("Month value does not match", new SimpleDateFormat("MMMM").format(now),
```
bencaradocdavies: This test could also fail if run close to the end of a month.

bencaradocdavies:
@emerkle826 perhaps you could (1) save nowBeforeSetup and nowAfterSetup to ensure you get the full time range of expected ```last_change```, (2) use a {{Pattern}} to extract the month and minute from ```dateString```, (3) assert that each of month and minute matches either nowBeforeSetup or nowAfterSetup. If the minute of nowBeforeSetup and nowAfterSetup differ by anything other than 59, 0, or 1, do not test the minute as your test has likely been interrupted by e.g. host load on a virtual machine.

emerkle826:
@bencaradocdavies I created a new PR to fix the test. I think it should be sufficient to just ensure that the `last_change` time is between the `start_time` and `end_time` that I now capture before and after entry setup in the test. This doesn't guarantee the SDF string doesn't get messed up again or reverted, as there are cases where the MONTH and MINUTES may not be the same, but don't cause the incorrectly parsed date to be outside the `start_time` and `end_time` range. But it would be another race for that to pass. Please see https://github.com/geotools/geotools/pull/1103 and let me know if I should change it still.

bencaradocdavies:
Fix merged in https://github.com/geotools/geotools/pull/1103

tbarsballe:
Backported to 14.x

PR #1103, opened 2016-02-06
GEOT-5355 Fixing bad test

emerkle826:
Changed the test to capture dates before and after the entry setup
so the "last_change" parsed date can be compared.


bencaradocdavies:
@emerkle826 thanks very much. That looks like a complete fix. Your new assertions are simpler and stronger than those I suggested!

tbarsballe:
Backported to 14.x

PR #1106, opened 2016-02-10
[GEOT-5360] Issue serializing xml (WFS) for Geometry collections

rishg:
WMS seems to work just fine with GeometryCollection, however, WFS does not.
The classes added here encodes the geometries within the geometrycollection.
Update GML2FeatureCollectionEncoderDelegate.java


aaime:
From a cursory look it seems ok, maybe add a test covering GML2 collection encoding too?

PR #1107, opened 2016-02-11
NetCDF/GRIB: Minor improvements on the Indexer Creator testing tool

dromagnoli:
adding support for elevation attribute as double


bradh:
Is there a ticket that explains what this is trying to fix?

dromagnoli:
Nope, since the tool is still a test executable, something like a "standalone" piece for quicker debug and test. I can create it if needed.

PR #1108, opened 2016-02-11
[GEOT-4317] url encoding of WMS layer name

HendrikPeilke:
Since wms layer names with spaces are possible with ArcGIS Server, here is an update on how urls on the wms requests are built (see https://osgeo-org.atlassian.net/browse/GEOT-4317)

Signed-off-by: HendrikPeilke <Hendrik.Peilke@ibykus.de>

### DIFF
--- a/modules/extension/wms/src/main/java/org/geotools/data/wms/request/AbstractGetFeatureInfoRequest.java
+++ b/modules/extension/wms/src/main/java/org/geotools/data/wms/request/AbstractGetFeatureInfoRequest.java
@@ -2,7 +2,7 @@
  *    GeoTools - The Open Source Java GIS Toolkit
  *    http://geotools.org
  *
- *    (C) 2004-2008, Open Source Geospatial Foundation (OSGeo)
+ *    (C) 2004-2016, Open Source Geospatial Foundation (OSGeo)
  *    
  *    This library is free software; you can redistribute it and/or
  *    modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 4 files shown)


aaime:
The solution seems correct, but we need a test for it, and copyright header updates.
See all the minimum requirements in the contribution guidelines here:

aaime:
Actually hum... wait a second, the WMS specification

aaime:
Ops, reopening

HendrikPeilke:
Sorry, missed the headers and tests. There are tests, which explicitely ask for a '+'-conversion... But with ArcGIS Server this won't work. And I guess changing it to '%20' will also work with other servers, so changing the tests and the code for styles also seems best (e.g.testGetFinalURL inside https://github.com/geotools/geotools/blob/master/modules/extension/wms/src/test/java/org/geotools/data/wms/test/AbstractGetMapRequestTest.java) or what do you think? If yes, should I also update the upgrade info docs?

jodygarnett:
strategy seems fine, focus on the test rather than docs.

We have a code freeze tomorrow, but since this is a bug fix it won't effect you.

aaime:
Merging, apparently %20 is always valid (from what I can gather on internet searches at least)

aaime:
I have backported on 14.x. 13.x has the last release just now, since I could not find any authoritative statement on "+" or "%20" I've decided to use caution and not backport there (as we might not have resource to make another 13.x release)

PR #1111, opened 2016-02-12
[GEOT-5363]:Better management of absolute file path on Hints config

dromagnoli:


aaime:
Copyright headers not updated

dromagnoli:
Copyrights updated. Merging

PR #1112, opened 2016-02-12
[GEOT-5363]:Better management of absolute file path on Hints config

dromagnoli:


aaime:
Looks like there are compile failures in Travis

dromagnoli:
Fixed and merged

PR #1120, opened 2016-02-17
[GEOT-5367] Add a group by visitor

nmco:
The current aggregation visitors (AverageVisitor, CountVisitor, MaxVisitor, MedianVisitor, MinVisitor, StandardDeviationVisitor and SumVisitor) don’t support a “group by” behavior. For example, we are able to compute the average electrical consumption per building. But we are not able to ask the average electrical consumption per building type.

The idea is to add new "group by" visitor that will be able to group features by type and apply a visitor on them. The JDBCDataStore will be extend to natively support the "group by" behavior. The existing Aggregator process will also be extended to support the "group by" behavior.

All of this will be done preserving backwards compatibility.

Mailing list discussion about this:
http://osgeo-org.1560.x6.nabble.com/Adding-quot-group-by-quot-support-to-the-current-aggregation-visitors-td5249393.html

Issue: https://osgeo-org.atlassian.net/browse/GEOT-5367

### DIFF
--- a/modules/library/main/src/main/java/org/geotools/feature/visitor/GroupByVisitor.java
+++ b/modules/library/main/src/main/java/org/geotools/feature/visitor/GroupByVisitor.java
@@ -1,0 +1,340 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ *
+ *    (C) 2016, Open Source Geospatial Foundation (OSGeo)
+ *
+ *    This library is free software; you can redistribute it and/or
+ *    modify it under the terms of the GNU Lesser General Public
+ *    License as published by the Free Software Foundation;
+ *    version 2.1 of the License.
+ *
+ *    This library is distributed in the hope that it will be useful,
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ *    Lesser General Public License for more details.
+ */
+package org.geotools.feature.visitor;
+
+import com.vividsolutions.jts.geom.Envelope;
+import com.vividsolutions.jts.geom.Geometry;
+import com.vividsolutions.jts.geom.Point;
+import org.geotools.feature.DefaultFeatureCollection;
+import org.opengis.feature.Feature;
+import org.opengis.feature.FeatureVisitor;
+import org.opengis.feature.simple.SimpleFeature;
+import org.opengis.filter.expression.Expression;
+import org.opengis.util.ProgressListener;
+
+import java.util.*;
+import java.util.stream.Collectors;
+
+/**
+ * Group features by one or several attributes and applies an aggregator visitor to each group.
+ */
+public class GroupByVisitor implements FeatureCalc {
+
+    private final Aggregate aggregateVisitor;
+    private final Expression aggregateAttribute;
+    private final FeatureCalc aggregateVisitorProtoType;
+    private final List<Expression> groupByAttributes;
+    private final ProgressListener progressListener;
+
+    private final InMemoryGroupBy inMemoryGroupBy = new InMemoryGroupBy();
+
+    private CalcResult optimizationResult = CalcResult.NULL_RESULT;
+
+    public GroupByVisitor(Aggregate aggregateVisitor, Expression aggregateAttribute,
+                          List<Expression> groupByAttributes, ProgressListener progressListener) {
+        this.aggregateVisitor = aggregateVisitor;
+        this.aggregateAttribute = aggregateAttribute;
+        this.groupByAttributes = groupByAttributes;
+        this.progressListener = progressListener;
+        aggregateVisitorProtoType = aggregateVisitor.create(aggregateAttribute);
+    }
+
+    public boolean wasOptimized() {
+        return optimizationResult != null;
+    }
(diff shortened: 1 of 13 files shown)


jodygarnett:
ah - wow nice work. Also the evening before a code freeze - I will do my best!

modules/unsupported/process-feature/src/main/java/org/geotools/process/vector/AggregateProcess.java
```diff
@@ -67,18 +56,34 @@
...
+        return execute(features, aggAttribute, functions, singlePass, null, progressListener);
+    }
+
     @DescribeResult(name = "result", description = "Aggregation results (one value for each function computed)")
     public Results execute(
             @DescribeParameter(name = "features", description = "Input feature collection") SimpleFeatureCollection features,
             @DescribeParameter(name = "aggregationAttribute", min = 0, description = "Attribute on which to perform aggregation") String aggAttribute,
             @DescribeParameter(name = "function", description = "An aggregate function to compute. Functions include Count, Average, Max, Median, Min, StdDev, and Sum.", collectionType = AggregationFunction.class) Set<AggregationFunction> functions,
             @DescribeParameter(name = "singlePass", description = "If True computes all aggregation values in a single pass (this will defeat DBMS-specific optimizations)", defaultValue = "false") boolean singlePass,
+            @DescribeParameter(name = "groupByAttributes", description = "List of group by attributes", collectionType = String.class) List<String> groupByAttributes,
```
jodygarnett: Is it worth making this one optional - to preserve backwards compatibility? Probably not ...

modules/library/main/src/main/java/org/geotools/feature/visitor/GroupByVisitor.java
```diff
@@ -0,0 +1,454 @@
...
+ * Group features by one or several attributes and applies an aggregator visitor to each group.
+ */
+public class GroupByVisitor implements FeatureCalc {
+
+    /**
+     * This Enum contains all the visitors that can be used with the group by visitor.
+     */
+    public enum AggregateVisitor {
+
+        Average {
```
jodygarnett: Violates enum naming conventions, consider AggregateVisitor.ENUM.

These things are not really visitors, you re using them to create a visitor are you not. I do kind of see what you are doing here providing a consistent create and wrap access to associate both the feature collection visitor and the calculation result.

Is it worth factoring this out into a standalone class?

nmco:
Hi Jody, 

Thank you for the review, I have updated the pull request with all your suggestions.

Regarding the documentation there is no documentation related with aggregation visitors where can I add documentation for this ?

nmco:
The build fail is related with xmlcodegen module (so I think is unrelated with this pull request). Long story short, version 3.2 of maven-plugin-plugin seems to be buggy to solve this o my local build I forced the 3.4 version to something like this (still I have no idea why this is happening now): 
```
<build>
    <plugins>
      <plugin>
        <groupId>org.apache.maven.plugins</groupId>
        <artifactId>maven-plugin-plugin</artifactId>
        <version>3.4</version>
      </plugin>
    </plugins>
  </build>
```

nmco:
Ooops wrong button.

jodygarnett:
Docs for aggregate functions are here:

* FeatureCollection: http://docs.geotools.org/latest/userguide/library/main/collection.html#aggregate-functions

modules/library/main/src/main/java/org/geotools/feature/visitor/GroupByInternalVisitor.java
```diff
@@ -0,0 +1,166 @@
...
+import org.geotools.util.UnsupportedImplementationException;
+import org.opengis.filter.expression.Expression;
+
+/**
+ * This Enum contains all the visitors that can be used with the group by visitor.
+ * Each visitor supported by the group by visitor has to provide a consistent way to
+ * create a new instance and if the visitor result can be optimized a way to wrap
+ * the raw result.
+ */
+public enum GroupByInternalVisitor {
```
jodygarnett: Thanks for breaking this out into seperate class, I will refactor this to:

````
public enum Aggregate {
   AVERAGE { ... },
   COUNT {...},
   ...
}
````

The enum is not acting as a visitor (since feature collection does not "accept" it).

nmco:
GroupByInternalVisitor renamed to Aggregate. I add some documentation for this visitor after  the aggregation functions section (if this is not the correct place to put it or if the documentation I wrote is not clear enough please let me know).

jodygarnett:
I have pushed up your first two commits, and renamed to Aggregate.

I will try and rescue your docs shortly.

bencaradocdavies:
@nmco pom changes required to build should be included your pull request. See:
https://osgeo-org.atlassian.net/browse/GEOT-5370

I will push the pom change as soon as I have tested that it does not break Maven 3.0.4. Plugin pinning occurs in the ```pluginManagement``` section of the top-level pom.

bencaradocdavies:
GeoTools master pinned to maven-plugin-plugin:3.4:
https://github.com/geotools/geotools/commit/889011b2e9257ca533132b630ecfa0a2cfe34d37

This fixes the build for Maven 3.3.9 and does not break it for Maven 3.0.4.

nmco:
Thank you @bencaradocdavies !

jodygarnett:
docs merged to master

PR #1125, opened 2016-03-08
GEOT-4909, fixing issues when saxon used as encoder transformer.

jdeolive:
- Ensuring attributes returns empty string rather than null for no uri
- Using “xml” rather than “XML” for output method key value (mandated
but the spec)

### DIFF
--- a/modules/extension/xsd/pom.xml
+++ b/modules/extension/xsd/pom.xml
@@ -81,6 +81,23 @@
     </plugins>
   </build-->
 
+  <profiles>
+    <profile>
+      <id>saxon</id>
+      <dependencies>
+         <dependency>
+           <groupId>net.sf.saxon</groupId>
+           <artifactId>Saxon-HE</artifactId>
+           <version>9.6.0-6</version>
+           <scope>test</scope>
+         </dependency>
+      </dependencies>
+      <properties>
+        <test.args>-Djavax.xml.transform.TransformerFactory=net.sf.saxon.jaxp.SaxonTransformerFactory</test.args>
+      </properties>
+    </profile>
+  </profiles>
+
   <modules>
     <module>xsd-core</module>
     <module>xsd-gml2</module>
--- a/modules/extension/xsd/xsd-core/src/main/java/org/geotools/xml/Encoder.java
+++ b/modules/extension/xsd/xsd-core/src/main/java/org/geotools/xml/Encoder.java
@@ -2,7 +2,7 @@
  *    GeoTools - The Open Source Java GIS Toolkit
  *    http://geotools.org
  *
- *    (C) 2002-2015, Open Source Geospatial Foundation (OSGeo)
+ *    (C) 2002-2016, Open Source Geospatial Foundation (OSGeo)
  *
  *    This library is free software; you can redistribute it and/or
  *    modify it under the terms of the GNU Lesser General Public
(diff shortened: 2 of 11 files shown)


modules/unsupported/wfs-ng/pom.xml
```diff
@@ -206,6 +206,15 @@
       <version>2.0.0-2</version>
       <scope>test</scope>
 	</dependency>
+  <!-- enable this dependency to test saxon -->
```
aaime: Maybe make it a profile, so that one can test it just adding an extra to the command line?

jdeolive:
Yeah, I went down that road but then realized running actually requires adding a system property which then requires updating the test config... so ditched it. But maybe i'll try again because being able to switch by simply engaging the profile is nice.

jdeolive:
Ok, went ahead and added the profile that adds the dependency and sets the system property. Required a minor change to the root pom to pass through a new property called "test.args". It also now engages saxon for all of the xsd tests rather than for just the wfs module. Turns out this was a worth while effort since lots of other cases didn't work with saxon. But happy to report all xsd tests pass with and without saxon now.

aaime:
Nicely done Justin!

The Travis build is failing in a weird way in gt-jdbc, I don't think it has anything to do with your changes:

````
Tests run: 2, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.085 sec <<< FAILURE! - in org.geotools.jdbc.JDBCTransactionStateTest
testSetTransactionNullWithExternalConnection(org.geotools.jdbc.JDBCTransactionStateTest)  Time elapsed: 0.003 sec  <<< ERROR!
org.mockito.exceptions.misusing.UnfinishedStubbingException: 
Unfinished stubbing detected here:
-> at org.geotools.jdbc.JDBCTransactionStateTest.setupLogHandler(JDBCTransactionStateTest.java:64)

E.g. thenReturn() may be missing.
Examples of correct stubbing:
    when(mock.isOk()).thenReturn(true);
    when(mock.isOk()).thenThrow(exception);
    doThrow(exception).when(mock).someVoidMethod();
Hints:
 1. missing thenReturn()
 2. although stubbed methods may return mocks, you cannot inline mock creation (mock()) call inside a thenReturn method (see issue 53)

	at org.geotools.jdbc.JDBCTransactionStateTest.setupLogHandler(JDBCTransactionStateTest.java:73)
````

Oddly enough this does not happen on ares...

jdeolive:
Hmmm... yeah, looks unrelated, just built locally and didn't see the failure. I bumped the PR to kick off another travis build so we'll see if maybe this was just a one off.

jdeolive:
Looks like Travis is happy. If I don't hear anything else I'll go ahead and merge this one.

aaime:
Hi @jdeolive, it may be that this change broke a few wfs tests, see here:
http://ares.boundlessgeo.com/jenkins/view/geoserver/job/geoserver-master/2304/consoleText

jdeolive:
Rats, knew I should have run the geoserver tests :) I'm on the road right now but I'll take a look as soon as I can. If anyone is blocked on this feel free to roll back this pull request and I'll re-merge when I get a chance to sort it out.

jdeolive:
Looked into it and this change is not playing nice with the recent changes to speed up gml writing. Basically the gml writer is always using the gml 2/3 namespace which then leads to problems when encoding in the gml 3.2 context which requires a different namespace. I'm continuing to investigate.

PR #1126, opened 2016-03-10
Allow customization to the CREATE TABLE statement

sayonkumarsaha:
For developing a datastore for SAP HANA, the `CREATE TABLE` statement needs to be changed to `CREATE COLUMN TABLE`.


aaime:
Looks ok to me, merging

PR #1127, opened 2016-03-12
Ensuring that gml 3.2 namespace used when encoding gml 3.2.

jdeolive:
- Updated QualifiedName to accept a namespace uri
- Grabbing the namespace uri from the namespace context

### DIFF
--- a/modules/extension/xsd/xsd-gml3/src/main/java/org/geotools/gml3/simple/GML32FeatureCollectionEncoderDelegate.java
+++ b/modules/extension/xsd/xsd-gml3/src/main/java/org/geotools/gml3/simple/GML32FeatureCollectionEncoderDelegate.java
@@ -2,7 +2,7 @@
  *    GeoTools - The Open Source Java GIS Toolkit
  *    http://geotools.org
  *
- *    (C) 2015-2016, Open Source Geospatial Foundation (OSGeo)
+ *    (C) 2016-2016, Open Source Geospatial Foundation (OSGeo)
  *
  *    This library is free software; you can redistribute it and/or
  *    modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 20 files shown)


jdeolive:
This should fixe the issues that the recent #1125 introduced to GeoServer.

modules/extension/xsd/xsd-gml3/src/main/java/org/geotools/gml3/simple/GML32FeatureCollectionEncoderDelegate.java
```diff
@@ -2,7 +2,7 @@
  *    GeoTools - The Open Source Java GIS Toolkit
  *    http://geotools.org
  *
- *    (C) 2015-2016, Open Source Geospatial Foundation (OSGeo)
+ *    (C) 2016-2016, Open Source Geospatial Foundation (OSGeo)
```
aaime: It was correct before (2015 - 2016) :-p

aaime:
There are several issues with copyright headers updates... seems that a mass replace "2015 -> 2016" was used, while 2016 should have been used as the second year if necessary.

aaime:
Merging to get the builds back to normal, I'll fix the copyright headers in a separate commit.

jdeolive:
Sorry about the headers, indeed I used a mass update to update the headers and I guess wasn't careful enough about reviewing the final diff.

Is there a better way to update the copyright headers? I have to admit the more recent strictness around copyright headers is quite cumbersome. Especially for changes that touch a lot of files. Is it not possible to use a copyright header that stays relevant for everything moving forward?

aaime:
Sigh, painful topic :-(

So, you should really ask @jodygarnett , for the record I've tried because I'm a bit tired of having to enforce the copyright header updates on each and every pull request I review (Jody's approach would be to perform a manual merge and do the updates in the process,  but I'm not too thrilled about that process, too much overhead).

He made me notice that we are already using a somehow simplified approach, the more complex and more correct one being that one has to list each and every year in which the file has been modified.

For GeoServer there are scripts lurking around that you can apply to correctly update the header of every file modified in the last update, check the geoserver-devel list archives for references, for Geotools I still bhaven't seen an equivalent one that we can use.

jdeolive:
I'll wait to hear from @jodygarnett but my opinion is that it is excessive based on what I have seen on other OS projects do. Even those that are part of the eclipse foundation which from what I have seen has the most strict rules when it comes to legal stuff. I have a haunting suspicion that we're adding unnecessary work for ourselves here.

PR #1129, opened 2016-03-14
[GEOT-5360] Unit test GML2

rishg:
Unit tests for GML2 - follow up on the previous commit.


aaime:
Merged and backported to 14.x. Thank you!

PR #1130, opened 2016-03-17
[GEOT-5079] PostGISDialect Error with PostGIS 2.0

srstclair:
Fixes [GEOT-5079](https://osgeo-org.atlassian.net/browse/GEOT-5079).

When PostGIS geometry/geography dimensionaly can't be determined from the `geometry_columns` or `geography_columns` tables (due to missing records or permissions issue), `PostGISDialect` falls back to sampling the first geometry with a dimension function. The `DIMENSION` function doesn't exist in PostGIS 2.x; [`ST_DIMENSION`](http://postgis.net/docs/ST_Dimension.html) should be used instead.

Commit includes test cases for both geometry and geography types.

I can backport to 14.x and 13.x once this is merged to master.

### DIFF
--- a/modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/PostGISDialect.java
+++ b/modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/PostGISDialect.java
@@ -2,7 +2,7 @@
  *    GeoTools - The Open Source Java GIS Toolkit
  *    http://geotools.org
  *
- *    (C) 2002-2015, Open Source Geospatial Foundation (OSGeo)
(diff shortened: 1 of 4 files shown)


srstclair:
I tested this against PostGIS 1.5 and 2.x also.

jdeolive:
Patch looks really good to me, comes with tests and even has header updated (which I can never remember to do). Nice work.

shane-axiom:
I cleaned up the way the tests get a connection and dialect, there were some unnecessary operations in the commit before.

srstclair:
I cleaned up the way the tests get a connection and dialect, there were some unnecessary operations in the commit before.

aaime:
Manually merged on master to make a minor fix (the test setup lacked copyright updates) and backported to 14.x and 13.x.
Thanks a lot!

PR #1133, opened 2016-03-18
[GEOT-5378] Add check for functions deprecated in PostGIS 2.1.0

srstclair:
Fixes [GEOT-5378 Add check for functions deprecated in PostGIS 2.1.0](https://osgeo-org.atlassian.net/browse/GEOT-5378)

PostGISDialect was using deprecated functions ST_Force_2D and ST_Estimated_Extent, which were deprecated in 2.1.0 in favor of ST_Force2D and ST_EstimatedExtent, respectively. This commit checks the PostGIS version and returns the new versions of the function names when PostGIS 2.0+ is detected.

Please review the new `PostGISDialect.getVersion()` and `PostGISDialect.getVersionSafe()` methods, which were needed to detect the version without having access to an existing database connection.

I didn't add a new test case here because all of the changed methods are covered by existing tests. I tested against PostGIS 1.5 and 2.2.

Will backport to 14.x and 13.x when merged to master.

Also, this Bash script was used to detect PostGIS functions used by GeoTools and could be useful when checking future deprecations.
https://gist.github.com/srstclair/a58ed186ef2794f60b2a

### DIFF
--- a/modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/PostGISDialect.java
+++ b/modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/PostGISDialect.java
@@ -153,6 +153,8 @@ public class PostGISDialect extends BasicSQLDialect {
 
     static final Version V_2_0_0 = new Version("2.0.0");
 
+    static final Version V_2_1_0 = new Version("2.1.0");
+
     static final Version V_2_2_0 = new Version("2.2.0");
 
     static final Version PGSQL_V_9_0 = new Version("9.0");
@@ -229,6 +231,7 @@ public class PostGISDialect extends BasicSQLDialect {
     public void initializeConnection(Connection cx) throws SQLException {
         super.initializeConnection(cx);
         getPostgreSQLVersion(cx);
+        getVersion(cx);
     }
 
     @Override
@@ -306,7 +309,7 @@ public class PostGISDialect extends BasicSQLDialect {
                 Boolean.TRUE.equals(hints.get(Hints.FEATURE_2D));
 
             if (force2D) {
-                sql.append("encode(ST_AsBinary(ST_Force_2D(");
+                sql.append("encode(ST_AsBinary(" + getForce2DFunction() + "(");
                 encodeColumnName(prefix, gatt.getLocalName(), sql);
                 sql.append(")),'base64')");
             } else {
@@ -336,7 +339,7 @@ public class PostGISDialect extends BasicSQLDialect {
                 sql.append("),'base64')");
             } else {
                 if (NON_CURVED_GEOMETRY_CLASSES.contains(gatt.getType().getBinding())) {
-                    sql.append("encode(ST_AsBinary(ST_Simplify(ST_Force_2D(");
+                    sql.append("encode(ST_AsBinary(ST_Simplify(" + getForce2DFunction() + "(");
                     encodeColumnName(prefix, gatt.getLocalName(), sql);
                     sql.append("), " + distance + preserveCollapsed + ")),'base64')");
                 } else {
@@ -347,7 +350,7 @@ public class PostGISDialect extends BasicSQLDialect {
                     sql.append(") THEN ");
                     encodeColumnName(prefix, gatt.getLocalName(), sql);
                     sql.append(" ELSE ");
-                    sql.append("ST_Simplify(ST_Force_2D(");
+                    sql.append("ST_Simplify(" + getForce2DFunction() + "(");
                     encodeColumnName(prefix, gatt.getLocalName(), sql);
                     sql.append("), " + distance + preserveCollapsed + ") END),'base64')");
                 }
@@ -359,7 +362,7 @@ public class PostGISDialect extends BasicSQLDialect {
     @Override
     public void encodeGeometryEnvelope(String tableName, String geometryColumn,
             StringBuffer sql) {
-        sql.append("ST_AsText(ST_Force_2D(ST_Envelope(");
+        sql.append("ST_AsText(" + getForce2DFunction() + "(ST_Envelope(");
         sql.append("ST_Extent(\"" + geometryColumn + "\"::geometry))))");
     }
     
@@ -389,7 +392,8 @@ public class PostGISDialect extends BasicSQLDialect {
                 if (att instanceof GeometryDescriptor) {
                     // use estimated extent (optimizer statistics)
                     StringBuffer sql = new StringBuffer();
-                    sql.append("select ST_AsText(ST_force_2d(ST_Envelope(ST_Estimated_Extent('");
+                    sql.append("select ST_AsText(" + getForce2DFunction() + "(ST_Envelope("
+                            + getEstimatedExtentFunction() + "('");
                     if(schema != null) {
                         sql.append(schema);
                         sql.append("', '");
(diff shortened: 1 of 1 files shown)


wrass:
@srstclair Thanks very much for adding this PR, looks good.  Just note though that these functions were changed in PostGIS 2.1.0, so this PR could do with the addition of `static final Version V_2_1_0 = new Version("2.1.0");` somewhere around line 155 along with a minor amendment on lines 1220 and 1230.

srstclair:
@wrass Thanks for looking it over.

I thought about setting the version check to 2.1.0, but I checked the docs for both functions:

http://postgis.net/docs/ST_Force2D.html
http://postgis.net/docs/ST_EstimatedExtent.html

They both say that the updated functions were added in 2.0.x:

> Changed: 2.1.0. Up to 2.0.x this was called ST_Force_2D.
> Changed: 2.1.0. Up to 2.0.x this was called ST_Estimated_Extent.

I took that to mean that the updated functions were added in 2.0.x, and the old versions were officially deprecated in 2.1.x. I don't mind changing to 2.1.x if you think that's better.

wrass:
@srstclair To be fair, I don't think the docs are all that clear on this, but the Release Notes for 2.1.0 should help to clarify: http://postgis.net/docs/release_notes.html#idp107324528

Ii looks like the functions were renamed in this commit for 2.1.0beta1 (this was over a year after 2.0.0 was released): https://github.com/postgis/postgis/commit/c17e2878f7706254d4f562cceaa61896c3b03457

Hope this helps.

shane-axiom:
@wrass Thanks for tracking that down. I'll update to look for >= 2.1.0 and update the PR.

srstclair:
@wrass Updated.

aaime:
Hi Shane, looks good to me, including the new functions to check the version... one thing that's giving me pause is that this makes 3 functions to get the postgis version... it would probably make more sense to move getting the version in the initConnection method and then just use the fields. @jdeolive do you see issues with this plan?

jdeolive:
Agreed, getting and storing the version in initConnection makes sense to me.

srstclair:
@aaime @jdeolive Changed PostGISDialect to set PostGIS version during [`initializeConnection`](https://github.com/geotools/geotools/pull/1133/commits/d8bc89d05c237fc969381704dc472ef5bdc31af5#diff-6683308a31e67377243891bf9cd85aa0R234). All good?

aaime:
Yep, all good. Merging

bencaradocdavies:
This change breaks the geoserver-master-app-schema-online build on ares Jenkins and locally against postgis:
http://ares.boundlessgeo.com/jenkins/job/geoserver-master-app-schema-online/365/consoleText

Locally on Debian unstable:
```
postgresql-9.5 9.5.1-1 amd64
postgresql-9.5-postgis-2.2 2.2.2+dfsg-1 amd64
```

From the ares failure:
```
testResultHitsWithFilter(org.geoserver.test.onlineTest.DataReferenceWfsPostgisTest)  Time elapsed: 6665 sec  <<< ERROR!
org.postgresql.util.PSQLException: ERROR: function postgis_lib_version() does not exist
```

bencaradocdavies:
However this manual test returns the version as expected:
```
$ psql -h localhost -d refdata -U refdata -W
Password for user refdata: 
psql (9.5.1)
SSL connection (protocol: TLSv1.2, cipher: ECDHE-RSA-AES256-GCM-SHA384, bits: 256, compression: off)
Type "help" for help.

refdata=> select postgis_lib_version();
 postgis_lib_version 
---------------------
 2.2.2
(1 row)
```

shane-axiom:
Blargh, thanks for reporting. Checking into it now...

bencaradocdavies:
@shane-axiom I have a fix.

bencaradocdavies:
@shane-axiom the problem is that the fixture setup SQL ```Postgis_Data_ref_set.sql``` contains:
```
SET search_path = classificationmanager, pg_catalog;
```
leaving the public schema off the search path. The solution is to schema-qualify the function call:
```
rs = st.executeQuery("select public.PostGIS_Lib_Version()");
```
I am testing this now.

srstclair:
Hmm, what happens if a user creates the PostGIS extension in a non-`public` schema, e.g.

```
CREATE SCHEMA postgis;
CREATE EXTENSION postgis SCHEMA postgis;
```

and then adds the non-public schema to the desired users' search paths using

```
ALTER USER username SET search_path = postgis,public;
```

I don't see any other examples of schema qualified function calls in [PostGISDialect](https://github.com/geotools/geotools/blob/master/modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/PostGISDialect.java), which seems to indicate
that it's expected the database user will have the PostGIS schema included in the search path...

bencaradocdavies:
@shane-axiom I have pushed a working fix:
https://github.com/geotools/geotools/commit/1d2fbf159fef9bbc816cb317989a6d0065a31b23

@srstclair I had not considered the possibility that someone might create the extension in a different schema. I have never seen anyone do this. The pattern I usually see is that the creation of the extension in the public schema. The fixture setup SQL ```Postgis_Data_ref_set.sql``` makes explicit reference to the public schema in some parts. In other parts ```search_path``` is reset to include ```public```. I am happy to reconsider this change; I am just very keen to fix this build as others are working on Oracle support and I do not want this failure to get in their way.

srstclair:
@bencaradocdavies Understood about not wanting to hold up other work. I'm not familiar with the App-Schema Online Tests (I did just review [this page](http://docs.geoserver.org/latest/en/developer/programming-guide/app-schema/index.html)), and I'm not finding `Postgis_Data_ref_set.sql` anywhere. Is that a build server specific configuration?

I'm guessing that the tests must rely on the `public` schema being excluded from the `search_path`? Maybe those tests could be changed? I think it's important to not assume PostGIS is in `public` if possible.

Thanks for looking into this.

bencaradocdavies:
@srstclair I think you are right and that it would be better for me to revert https://github.com/geotools/geotools/commit/1d2fbf159fef9bbc816cb317989a6d0065a31b23 and instead have the gs-app-schema-postgis-test fixture clean up its mess. The underlying problem is that the setup script both assumes that postgis is installed in the public schema but then leaves ```search_path``` set to a value that does not include it. I think that it is better to have local fixes for gs-app-schema-postgis-test than impose a bad workaround on other users.

The SQL is ```/RefDataSet/Postgis_Data_ref_set.sql``` in http://download.osgeo.org/webdav/geotools/org/geotools/data/refdataset-1.0/1.0.7/refdataset-1.0-1.0.7.jar

The SQL is a generated database dump and is not under version control. Yes, the jar file is the canonical curated source. The horror. This has been discussed on the lists. I hope you see why I am loathe to modify this file.

Apologies for acting in haste, and thanks for your swift and helpful response.

shane-axiom:
@bencaradocdavies Great, thank you. The non-public PostGIS schema is admittedly rare but I have seen it, and it's nice to keep things clean sometimes. Let me know if there's anything I can do.

srstclair:
@bencaradocdavies Great, thank you. The non-public PostGIS schema is admittedly rare but I have seen it, and it's nice to keep things clean sometimes. Let me know if there's anything I can do.

bencaradocdavies:
Better fix applied in GeoServer: https://github.com/geoserver/geoserver/commit/a92b206bd1f25bbfbfb6b33bfe6a046949d8dfb0

PR #1136, opened 2016-03-22
GEOT-5383 imagemosaic resolution value in mosaic property file picks first file found

NielsCharlier:


bencaradocdavies:
Copyright header dates?

NielsCharlier:
Oh bugger, it suddenly dawned on me that I had forgotten again and I thought I'd quickly add it before anyone notices and perhaps gets annoyed again, but you beat me to it, @bencaradocdavies

simboss:
I would split this into 2 PRs.

I am happy with the fix for 5832, less happy with the other one which I believe would need a separate discusion which I am going to start on the ticket.

NielsCharlier:
Simone, the reason I grouped them together is because they use the same test data, so the second commit is going to be dependant on the first no matter what. But if the first commit is already approved, I can make a new pull request with only that one and merge it, and this one will automatically adjust to becoming only the second commit.

NielsCharlier:
Strange, the two build failure are related to heterogeneous granules, suggesting it has something to do with my changes, but they pass on my computer and I can't reproduce them.
Telling from the stack trace. Seems to be related to the amount of levels, so I will wait for the discussion. I don't actually mind what we pick, as long as it is not unpredictable.

NielsCharlier:
First commit now has its own PR: https://github.com/geotools/geotools/pull/1137

simboss:
Thx for splitting.

I am ok with this new approach, I will talk a little to daniele and andrea tomorrow about the impact of this change and then report here or on the ticket.

simboss:
Ciao Niels,
sorry for the delay.

I have not had time to discuss this thoroughly with Andrea and Daniele, will do this on Tue and get back to you.

NielsCharlier:
Thanks. Still haven't been able to reproduce the failures, if you guys have a clue what could be causing this...

dromagnoli:
Hi Niels,
I think that I can take a look on this tomorrow. Sorry for the delay

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/ImageMosaicConfigHandler.java
```diff
@@ -894,40 +894,32 @@ public void updateConfiguration(GridCoverage2DReader coverageReader,
...
             }
+
+            // configuration need to be updated
+            if (needUpdate) {
+                catalogConfigurationBean.setHeterogeneous(true);                
+                mosaicConfiguration.setLevels(resolutionLevels);
+                mosaicConfiguration.setLevelsNum(numberOfLevels);
+                getConfigurations().put(mosaicConfiguration.getName(), mosaicConfiguration);
+            }
+
```
dromagnoli: I think that with this approach, heterogeneous won't be set to true in this case:
GranuleA_xResolutions: 0.1,  0.2,  0.4  --> (3 levels)
GranuleB_xResolutions: 0.1,  0.2          --> (2 levels)
MosaicConfiguration is already set with GranuleA properties.

Analyzing GranuleB:
  - Utils.homogeneousCheck return true since it uses Math.min
  - numberLevels = 2 < originalNumberOfLevels = 3
Therefore needUpdate is never set to true so no heterogeneous check is set to true.

If granules would have been analyzed in the opposite order (GranuleB, then GranuleA), 
numberOfLevels = 3 > originalNumberOfLevels = 2 will have triggered needUpdate = true.

Wondering if the failing tests on this pull request are due to this.
I have checkout your branch on my local machine and it's failing here too.

dromagnoli:
Hi @NielsCharlier 
please, see my comment in the code.

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/ImageMosaicConfigHandler.java
NielsCharlier: You are right! I knew there still had to be an order dependency considering the non-consistent results of the build, but I just couldn't see it. Thanks!

NielsCharlier:
New patch was added, let's see what the build says...

NielsCharlier:
Okay, success! Are you happy with merging now?

dromagnoli:
Hi @NielsCharlier. Sorry for the delay.
I think it won't introduce any regression (it just takes the best resolution instead of random one) so I would say +1 to merge.

PR #1139, opened 2016-03-23
[GEOT-5387]Solr version update from 4.9.0 to 5.2.1

rishg:
update to a newer version of Solr.

### DIFF
--- a/modules/unsupported/solr/pom.xml
+++ b/modules/unsupported/solr/pom.xml
@@ -70,7 +70,7 @@
     <dependency>
       <groupId>org.apache.solr</groupId>
       <artifactId>solr-solrj</artifactId>
-      <version>4.9.0</version>
+      <version>5.2.1</version>
       <exclusions>
         <exclusion>
           <groupId>org.slf4j</groupId>
--- a/modules/unsupported/solr/src/main/java/org/geotools/data/solr/FieldLayerMapper.java
+++ b/modules/unsupported/solr/src/main/java/org/geotools/data/solr/FieldLayerMapper.java
@@ -2,7 +2,7 @@
  *    GeoTools - The Open Source Java GIS Toolkit
  *    http://geotools.org
  *
- *    (C) 2015, Open Source Geospatial Foundation (OSGeo)
+ *    (C) 2015-2016, Open Source Geospatial Foundation (OSGeo)
  *
  *    This library is free software; you can redistribute it and/or
  *    modify it under the terms of the GNU Lesser General Public
@@ -17,7 +17,7 @@
 package org.geotools.data.solr;
 
 import org.apache.solr.client.solrj.SolrQuery;
-import org.apache.solr.client.solrj.impl.HttpSolrServer;
+import org.apache.solr.client.solrj.impl.HttpSolrClient;
 import org.apache.solr.client.solrj.response.FacetField.Count;
 import org.apache.solr.client.solrj.response.QueryResponse;
 import org.apache.solr.common.params.FacetParams;
@@ -47,7 +47,7 @@ public class FieldLayerMapper implements SolrLayerMapper {
     }
 
     @Override
-    public List<String> createTypeNames(HttpSolrServer solrServer) throws Exception {
+    public List<String> createTypeNames(HttpSolrClient solrServer) throws Exception {
         List<String> names = new ArrayList<>();
 
         SolrQuery query = new SolrQuery();
--- a/modules/unsupported/solr/src/main/java/org/geotools/data/solr/FilterToSolr.java
+++ b/modules/unsupported/solr/src/main/java/org/geotools/data/solr/FilterToSolr.java
@@ -2,7 +2,7 @@
  *    GeoTools - The Open Source Java GIS Toolkit
  *    http://geotools.org
  * 
- *    (C) 2014, Open Source Geospatial Foundation (OSGeo)
+ *    (C) 2014-2016, Open Source Geospatial Foundation (OSGeo)
  *
  *    This library is free software; you can redistribute it and/or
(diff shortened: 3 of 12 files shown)


aaime:
The change seems good (you forgot to update the copyright headers though, please fix that).
That said, we need to better understand the consequences of the library upgrade. Will the store still work against older Solr versions? Are the online tests in the module working with this change?

rishg:
Tested gt-solr jar (with the changes) against a solr 4.9 server and there were no issues.

aaime:
@jdeolive do you have an opinion on this one?

aaime:
Btw @rishg you did not (directly) answer the second question " Are the online tests in the module working with this change?"
Did you try them out?

jdeolive:
+1. Solr is about to release 6.0 so I think moving to 5.x makes sense. Wondering though if we should move to the latest 5.x, which I think is 5.5.

rishg:
@aaime I just finished trying out the online tests. It works fine except for the cases testing field **_geo3_**. After digging around a bit, it seems like Solr has removed this particular (minX minY maxX maxY) format from indexing.   
See comments https://cwiki.apache.org/confluence/display/solr/Spatial+Search

rishg:
At present, I prefer to keep it at 5.2.1 but I do not see a big difference between 5.2.1 and 5.5.

rishg:
also, please let me know how you would like to proceed on this PR. Would you like to have that one field deleted from online tests, or some other form of update?

jdeolive:
The geo3 field is the bbox field type... which looks like it's trivial to port to a new syntax:

    "minX minY maxX maxY" -> "ENVELOPE(minX, maxX, maxY, minY)"

I just issued a pull request against your branch to make the syntax changes... but I haven't tested them yet. Since you have the test environment handy can you do a quick test to see if that fixes it?

Also, for future reference it's a little nicer (and safer) if you issue pull requests from a feature branch rather than your copy of master.

rishg:
@jdeolive Ran the tests, a couple things that would have to be changed: 
1. BboxStrategy's decode method subclass of SolrSpatialStrategy to the following: 
` @Override public Geometry decode(String str) throws ParseException {
            String[] bbox = str.split(",");
            if (bbox.length != 4) {
                throw new ParseException("Illegal bounding box: " + str);
            }
            bbox[0] = bbox[0].substring(bbox[0].indexOf("(")+1, bbox[0].length());
            bbox[3] = bbox[3].substring(0, bbox[3].indexOf(")"));
            Envelope env = new Envelope(parseDouble(bbox[0]), parseDouble(bbox[2]),
                parseDouble(bbox[3]), parseDouble(bbox[1]));
            return JTS.toGeometry(env);
        }`
2. SolrFeatureSourceTest: testBboxFilterWithBboxType queries Solr with the old (mixX minY ...) format and it fails.
3. isDisjointTo has been deprecated in lucene, so the testDisjointFilter throws an IOException. The proposed solution is in this thread :  https://issues.apache.org/jira/browse/LUCENE-5692

jdeolive:
Thanks Gaurish, appreciate the work on this one.

For (1) I would say that change makes sense.
For (2) seems like it should be trivial to switch to the new syntax?
For (3) I would say rather than try to implement the workaround described which is a bit clunky, let's just not try to natively encode the disjoint filter. Simply removing this [line](https://github.com/geotools/geotools/blob/master/modules/unsupported/solr/src/main/java/org/geotools/data/solr/FilterToSolr.java#L192) should do it. Unless @aaime feels strongly about this one?

aaime:
Nope, no strong feelings about removing native disjoint support, does not seem like a widely used spatial operator anyways.
Some hesitation about the bbox sytanx, is the code going to work across versions? Like, could it read both new and old syntax? Or am I missing the point, and this is just about data loading in the tests?

jdeolive:
Right, I think we would want to handle both syntaxes ideally, for backwards compatibility reasons in order to be able to work against both solr 4.x and 5.x+

rishg:
Thanks for your quick response, Justin and Andrea.  
For (1), I I will add a condition to ensure that there is a backword compatibilty. 
I am not sure if 2 is that trivial. Could you point me in the direction of where do you think the fix would go? https://github.com/geotools/geotools/blob/master/modules/unsupported/solr/src/test/java/org/geotools/data/solr/SolrFeatureSourceTest.java#L386-L388
Thanks again!

rishg:
@jdeolive I have updated the files and tests pass at this point. Please take a look at the https://github.com/rishg/geotools/blob/master/modules/unsupported/solr/src/main/java/org/geotools/data/solr/SolrSpatialStrategy.java

jdeolive:
Thanks @rishg. The useCql flag in the BBoxStrategy isn't quite right. I think it's a bad idea to set static state that used in the encode method, but set in the decode method. I would suggest removing that flag and just change encode to use the new syntax. Afaik I was the driving force behind even supporting bbox types and I know in the flavour or solr we use both types are supported. Plus we're on solr 5. If anyone complains about it I'll add some logic in that will look at the solr version, and use that to determine which way to encode. That work for you @aaime?

rishg:
Thanks for responding @jdeolive. 
Pushed with your comments.

jdeolive:
Thanks @rishg, this one is ready to go once it has merged with the latest master. There is a minor conflict in one of the copyright headers. Should be easy to fix. Once that is done I would say we merge this one.

rishg:
@jdeolive - Thanks Justin for the comments. I have made the changes asked. Should be good to go now, I hope :)

jdeolive:
Yup, looks great to me. As soon as Travis is happy I'll merge this one in. Thanks for handling all the feedback so quickly @rishg .

rishg:
Awesome! Thanks again.

PR #1140, opened 2016-03-24
[GEOT-5349] Use encodeSchemaName(...) while looking up GT_PK_METADATA

mprins:
This provides a fix for [GEOT-5349](https://osgeo-org.atlassian.net/browse/GEOT-5349) by encoding the schema name.


aaime:
Looking good. Writing a test would require a schema that fails when not passed to the encodeSchemaName method, but that is out of the control of the test suite. So I guess this one is ok not having a test. Merging

PR #1145, opened 2016-03-27
[GEOT-5389] Fix JTS version reference in documentation.

bradh:
Also includes a bonus typo fix from the JTS docs.

### DIFF
--- a/docs/developer/procedures/add.rst
+++ b/docs/developer/procedures/add.rst
@@ -176,8 +176,8 @@ Examples of Updating JTS Jar
 
 2. Here is an example of how to deploy the JTS binary jar::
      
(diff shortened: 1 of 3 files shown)


mprins:
I think the JTS example here is  unfortunate, as it's been available in maven central for some years: http://search.maven.org/#artifactdetails%7Ccom.vividsolutions%7Cjts%7C1.13%7Cjar

bradh:
@mprins Feel free to make more extensive edits if you believe they are needed.

tbarsballe:
@bradh - This looks good, do you want to merge it so that it gets into the 15-beta release later this week?

PR #1148, opened 2016-03-30
Porting over same build speedup changes as with GeoServer

aaime:


aaime:
Discarding it, the benefit is so small that it probably does not warrant the extra complication

PR #1151, opened 2016-03-31
Removed accessor method that nobody uses in GeometryAttributeImpl

hgryoo:
I removed unused, unnecessary getter/setter methods.
the interface returns Object not Geometry and nobody uses this "impl" class directly.


aaime:
I don't think we should accept this change as is, the class is public after all, some users of the library could have used them method.
In my opinion it should be deprecated first in 15.x, migration notes should be updated, and then we can remove it in 16.x

hgryoo:
I marked methods as deprecated.
we need the change proposal? or is it trivial?

PR #1153, opened 2016-04-01

ridethepenguin:
Current code in `XmlAttributeNodePointer` assumes an XML attribute called `id` must be the feature identifier and tries to retrieve it calling `feature.getIdentifier()`; however, if the attribute is not mapped using `<idExpression>` but as a regular `<ClientProperty>`, it will not be used as the feature identifier, in which case `feature.getIdentifier()` will return null (the original code throws a NPE).

This PR solves the issue in the simplest way, falling back to normal attribute extraction when `feature.getIdentifier()` returns nothing.


bencaradocdavies:
@ridethepenguin that looks great, nice fix. Full build including GeoServer with app-schema online tests passes.

PR #1154, opened 2016-04-01
[GEOT-5390] Fixed ColorModelFactory returning cached ColorModel objects with the wrong data type.

sikeoka:
[GEOT-5390] Fixed ColorModelFactory returning cached ColorModel objects with the wrong data type.


aaime:
The current Travis failure seems to have nothing to do with the change (hard crash in the JVM while running over some wfs client tests), I've kicked the build again

simboss:
It looks good to me.

Will you take care of the backport?

jodygarnett:
Thanks for the fix, can I double check if OSGeo has a contributor agreement on file for yourself or mousebird consulting? Thanks for the chat at eclipsecon :D

PR #1156, opened 2016-04-04
Adds ysld unsupported module.

jdeolive:
YSLD is a Yaml based representation of YSLD with some syntactic nicities added to the mix. The module was developed by Boundless and is being contributed to the community. Thanks Boundless!


modules/unsupported/ysld/pom.xml
```diff
@@ -0,0 +1,100 @@
...
+  
+  <parent>
+    <groupId>org.geotools</groupId>
+    <artifactId>unsupported</artifactId>
+    <version>15-SNAPSHOT</version>
+  </parent>
+  
+  <groupId>org.geotools</groupId>
+  <artifactId>gt-ysld</artifactId>
+  <name>Ysld Parser/Encoder</name>
```
jodygarnett: Ylsd or YSLD :)

modules/unsupported/ysld/pom.xml
```diff
@@ -0,0 +1,100 @@
...
+    </dependency>
+    <dependency>
+      <groupId>org.geotools</groupId>
+      <artifactId>gt-render</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.yaml</groupId>
+      <artifactId>snakeyaml</artifactId>
+      <version>1.13</version>
```
jodygarnett: Use of dependency management for version numbers, especially for things like hamcrest and findbugs that may conflict.

modules/unsupported/ysld/src/main/java/org/geotools/ysld/Band.java
```diff
@@ -0,0 +1,92 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ *
+ *    (C) 2014-2016, Open Source Geospatial Foundation (OSGeo)
+ *
```
jodygarnett: This is one for @smithkm - should we have a (C) 2015-2016 Boundless to reflect this code has shipped in another product?

jdeolive:
Thanks @jodygarnett. Changed Ysld to YSLD (assuming that is the decided upon convention), added kevin and myself and module maintainers and moved dependency versions into dependency management section.

jodygarnett:
Added an updated readme, and changed the headers to show file history.

modules/unsupported/ysld/src/main/java/org/geotools/ysld/Band.java
smithkm: @jodygarnett I guess that comes down to that argument you had with Arne Kepp about the GWC headers.  I have no idea which is the correct way to do this to make the lawyers happy.

PR #1157, opened 2016-04-04

ridethepenguin:
This is the backport of #1153 to the 14.x branch.


bencaradocdavies:
Thanks @ridethepenguin that looks good.

PR #1158, opened 2016-04-04

ridethepenguin:
This is the backport of #1153 to the 13.x branch.


bencaradocdavies:
@ridethepenguin this looks quite different to the changes for master and 14.x. It looks to me that you have created an entirely new mapping file and properties file for the test. Is there a technical reason for this, rather than backporting the SWE fixture? Too many entangled dependencies?

If possible, I like to reduce divergence between the branches to make future maintenance easier. I know you usually do this as well, so I would like to know why you chose to not do so in this case.

ridethepenguin:
@bencaradocdavies right, I should have added a comment to the PR explaining this. I had to refactor the code because `SweValuesTest` and related mapping / data files were altogether missing on the 13.x branch, since [the bug-fix they originated from](https://osgeo-org.atlassian.net/browse/GEOT-5251) was never ported to this branch. I thought it'd be quicker to refactor the code for the present fix rather than backport all the missing pieces.

Let me know how you think it's best to proceed: I don't mind leaving the backport to 13.x aside (14.x is what I really cared about), or spending some more time trying to port over all the dependencies (but this is going to result in a different set of changes anyway, maybe I should create two different PRs then).

bencaradocdavies:
@ridethepenguin I assumed that this was the case but wanted to check. Thanks for the extra effort required for this backport to 13.x.

ridethepenguin:
@bencaradocdavies awesome, thanks for a careful merging (as usual)!

PR #1163, opened 2016-04-05
small fix to feature factory lookup

jodygarnett:


aaime:
@jodygarnett could you please check the contribution rules, seems you've missed a few steps: https://github.com/geotools/geotools/blob/master/CONTRIBUTING.md

mprins:
see also: https://osgeo-org.atlassian.net/browse/GEOT-5400
see also: https://sourceforge.net/p/geotools/mailman/message/34992659/

aaime:
Just read your mail, quoting "That sounds like a pretty serious glitch, I wonder how the test cases are passing?".
A bit ironic there is no test in the pull requests, tests that are not there cannot fail :-p

bencaradocdavies:
@jodygarnett do you have a test case for this? And some copyright header updates?  :-P

modules/library/main/src/main/java/org/geotools/feature/AttributeTypeBuilder.java
```diff
@@ -197,9 +198,7 @@
 	 *
 	 */
     public AttributeTypeBuilder() {
-		this( new FeatureTypeFactoryImpl() );
-		init();
```
aaime: Btw, removing this init means the userData map won't be initialized to a empty map, calling addUserData right after the constructor seems like it will generate a NPE (unless I'm missing something, just had another quick look)

jnh5y:
All, any ideas of how to create a unit test which would fail here?  If so, I'd be interested in assisting.

PR #1165, opened 2016-04-06
[GEOT-5401] FastBBOX should not cast to SimpleFeature

wulfsberg:
A pretty minor change in the code, simply letting the feature object drop through to the standard expression evaluation, rather than hitting a class cast exception.
The test contains a fair amount of boilerplate to set up the PropertyAccessor to show this.

### DIFF
--- a/modules/library/render/src/test/java/org/geotools/renderer/lite/FastBBOXTest.java
+++ b/modules/library/render/src/test/java/org/geotools/renderer/lite/FastBBOXTest.java
@@ -1,0 +1,100 @@
+package org.geotools.renderer.lite;
+
+import com.vividsolutions.jts.geom.Envelope;
+import com.vividsolutions.jts.geom.Geometry;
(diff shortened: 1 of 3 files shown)


modules/library/render/src/test/java/org/geotools/renderer/lite/FastBBOXTest.java
```diff
@@ -0,0 +1,100 @@
+package org.geotools.renderer.lite;
```
aaime: Missing copyright header. Since this is your code, we cannot add it for you.

wulfsberg:
Added the standard copyright text to the test class. (Sorry for missing that in the first place; I was in that programming space where test files have a special status; not in the legal space where they are just files :)

aaime:
Manually merged to squash in a single commit, backported to 14.x and 13.x

PR #1166, opened 2016-04-07
jar lookup of manifest and version details

jodygarnett:
Used to verify geotools execution environment.

### DIFF
--- a/modules/library/metadata/src/main/java/org/geotools/factory/GeoTools.java
+++ b/modules/library/metadata/src/main/java/org/geotools/factory/GeoTools.java
@@ -409,23 +409,18 @@ public final class GeoTools {
      * @return Version (or null if unavailable)
      */
     public static Version getVersion(Class<?> type) {
-        final URL classLocation = type.getResource(type.getSimpleName() + ".class");
+        final URL classLocation = classLocation(type);
         String path = classLocation.toString();
 
         // try and extract from maven jar naming convention
         if (classLocation.getProtocol().equalsIgnoreCase("jar")) {
-            String location = path.substring(0, path.lastIndexOf("!") + 1);
-            int dash = location.lastIndexOf("-");
-            int dot = location.lastIndexOf(".jar");
-
-            if (dash != -1 && dot != -1) {
-                String version = location.substring(dash + 1, dot);
-                return new Version(version);
+            String jarVersion = jarVersion(path);
+            if( jarVersion != null ){
+                return new Version(jarVersion);
             }
             // try manifest
             try {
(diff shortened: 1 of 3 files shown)


jodygarnett:
@tbarsballe could you review please, adding this functionality to GeoTools next to other jar and environment maintenance methods.

tbarsballe:
Looks good overall.
One thing I am wondering is how this would behave if running from an unexploded war. It looks like it should work fine, but that seems like something you might want to test manually.

modules/library/metadata/src/main/java/org/geotools/factory/GeoTools.java
```diff
@@ -376,13 +381,103 @@ public static String getBuildTimestamp() {
...
+     * <li>build.timestamp=21-Mar-2016 21:30</li>
+     * </ul>
      */
     public static Properties getBuildProperties() {
         Properties props = new Properties();
         props.putAll(PROPS);
         return props;
     }
 
+    public static Version getVersion(Class<?> type) {
```
tbarsballe: Missing javadoc

modules/library/metadata/src/test/java/org/geotools/factory/GeoToolsTest.java
```diff
@@ -20,8 +20,19 @@
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
```
tbarsballe: ^ Headers say 2007-2008 ?

jodygarnett:
I expect this will fail from an unexploded war - since  the protocol would either be custom (ie not jar) or zip. The protocol will fail in a stable way, returning null for "unknown" similar to how it behaves in an IDE.

tbarsballe:
Yes, looks like this will not work in JBOSS - their paths look like `vfs:/var/jboss/workspace/BuildSvr_FNMOC/jboss/geoserver/deployments/geoserver.war/WEB-INF/lib/gt-xsd-wcs-13.2.jar/org/geotools/wcs/getCoverage.xsd`

(Custom protocol + no `!`)

See https://osgeo-org.atlassian.net/browse/GEOS-7139

jodygarnett:
I could try search for ".war/" to catch the jboss case, another option is to "count backwords the correct number of directories based on the package name. I will have to factor out the method to make it testable, but yes let me cover that case and we can merge.

jodygarnett:
@tbarsballe Please see recent commit which supports your jboss use-case - factored out non public utility methods to make the logic testable.

tbarsballe:
Looks good, I am quite happy with this now.

PR #1168, opened 2016-04-11
[GEOT-5396] Support Rotated Pole projection

mfilocha:
Add new transformation used often by Numerical Weather Forecasting community to support "Rotated Pole" coordinates. That transformation is also available in proj.4 as "+proj=ob_tran".

For initial idea of this pull request, see:
http://osgeo-org.1560.x6.nabble.com/Rotated-pole-coordinate-system-td5204800.html

This transformation is intended to replace a "GeneralOblique" one for two reasons:
- better self-descriptive name
- _latitudeOfOrigin_ parameter is consistent with its name (latitude of a new origin, not of a transformed Pole, as in proj.4).

For details, see:
http://osgeo-org.1560.x6.nabble.com/Problem-with-General-Oblique-transform-latitude-td5257089.html
And:
[GEOT-5396] Support Rotated Pole projection
https://osgeo-org.atlassian.net/browse/GEOT-5396

### DIFF
--- a/modules/library/referencing/src/test/resources/org/geotools/referencing/test-data/scripts/RotatedPole.txt
+++ b/modules/library/referencing/src/test/resources/org/geotools/referencing/test-data/scripts/RotatedPole.txt
@@ -1,0 +1,79 @@
+// Test points for the RotatedPole projection
+//
+// (C) 2016, Open Source Geospatial Foundation (OSGeo)
+//
+// The following test points were calculated with eqtoll and proj.4 tools.
+// "eqtoll" is a part of the MetOffice Unified Model suite 
+// 
+//
+// Maciej Filocha (ICM) - May 2015, April 2016
+//
+// --------------------------------------------------------------------------
+// How to run this script:
+//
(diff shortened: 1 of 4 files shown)


modules/library/referencing/src/test/resources/org/geotools/referencing/test-data/scripts/RotatedPole.txt
```diff
@@ -0,0 +1,79 @@
...
+
+//  
+// First part test points are taken from output of the "eqtoll" tool
+// 
+source crs = _UM_
+target crs = _WGS84_
+//print crs
+
+source pt = (0, 0)
+target pt = (19.30000,52.500000)
```
bencaradocdavies: Good, this is as I expected.

bencaradocdavies:
@aaime @jodygarnett this is a new feature. Is it blocked by the code freeze? Build passes, and it is a small extension, but it is a new feature, and this is gt-referencing.

simboss:
I would not stop this as this something hidden behind factories which kicks in only if explicitly requested.

Strictly speaking there is new API (as there is a new public class for the projection) but you will (almost)  never interact with it.

PR #1169, opened 2016-04-12
fix change style dialog method only return null bug

beihaifeiwu:


tbarsballe:
This fix does not follow the Contributer guidelines: https://github.com/geotools/geotools/blob/master/CONTRIBUTING.md
Notably, it is missing a test case, a JIRA bug report, and the copyright headers require updating.

bencaradocdavies:
@beihaifeiwu thanks for the fix. We discussed in our committee meeting. As this module lacks tests (and is unsupported), your fix is not making things worse.

PR #1170, opened 2016-04-14
Include YSLD in the default build

jericks:
I would like to include YSLD in the default build so it gets published to the Maven repository so projects using GeoTools (like GeoScript) can use YSLD.

https://osgeo-org.atlassian.net/browse/GEOT-5402

Thanks!
Jared


jdeolive:
Thanks Jared!

PR #1171, opened 2016-04-14
[GEOT-5403] fix issue with next/hasNext

ianturton:
see [GEOT-5403](https://osgeo-org.atlassian.net/projects/GEOT/issues/GEOT-5403) for full discussion

### DIFF
--- a/modules/library/main/src/main/java/org/geotools/data/store/FilteringFeatureIterator.java
+++ b/modules/library/main/src/main/java/org/geotools/data/store/FilteringFeatureIterator.java
@@ -67,7 +67,7 @@ public class FilteringFeatureIterator<F extends Feature> implements FeatureItera
     }
 
     public F next() throws NoSuchElementException {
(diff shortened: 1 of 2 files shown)


modules/library/main/src/main/java/org/geotools/data/store/FilteringFeatureIterator.java
```diff
@@ -67,6 +67,9 @@ public boolean hasNext() {
     }
 
     public F next() throws NoSuchElementException {
+        if(next==null&&this.hasNext()==false) {
+            throw new NoSuchElementException();
```
dromagnoli: minor suggestions:
please, check the formatting (need some whitespaces). 
I would also personally use !this.hasNext() instead of explicit check with false.

if (next == null && !this.hasNext()) {

aaime:
Looks good to me, I'd add the copyright header to the test class too while you're at it

PR #1173, opened 2016-04-15
GEOT-5258, Support for storing dates as bigint with postgis.

jdeolive:
Allows for supporting a wider range of dates than the postgresql timestamp type. This is rework of what was originally pull request #997.

### DIFF
--- a/modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/BigDate.java
+++ b/modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/BigDate.java
@@ -14,22 +14,21 @@
  *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
  *    Lesser General Public License for more details.
  */
-package org.geotools.data.postgis.ps;
+package org.geotools.data.postgis;
 
-import org.geotools.data.postgis.PostgisUDTTestSetup;
-import org.geotools.jdbc.JDBCUDTOnlineTest;
-import org.geotools.jdbc.JDBCUDTTestSetup;
+import java.util.Date;
 
 /**
- * 
- *
- * @source $URL$
+ * Used internally by PostGIS dialect to signal the use of the "bigdate" domain type.
+ * <p>
+ * This type is used to store a date as a raw long value in order to provide a greather range of 
+ * date values than the built in postgresql timestamp types. Using a raw long allows for representing
+ * a date range of approximately +- 290 million.
+ * </p>
  */
-public class PostgisUDTOnlineTest extends org.geotools.data.postgis.PostgisUDTOnlineTest {
+class BigDate extends Date {
 
-    @Override
-    protected JDBCUDTTestSetup createTestSetup() {
-        return new PostgisUDTTestSetup(new PostGISPSTestSetup());
+    private BigDate(long milliseconds) {
+        super(milliseconds);
     }
-
 }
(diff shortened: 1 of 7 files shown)


modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/BigDate.java
```diff
@@ -0,0 +1,34 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ *
+ *    (C) 2002-2016, Open Source Geospatial Foundation (OSGeo)
```
jodygarnett: New file should be 2016 :)

jodygarnett:
Thanks @jdeolive - this is much cleaner, I like how the type conversion was detected and handled during the visitor.

I would be happy to fix up the headers and merge this in for the beta2 (and lean on the geotools-devel list for testing).

modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/BigDate.java
```diff
@@ -0,0 +1,34 @@
...
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ *    Lesser General Public License for more details.
+ */
+package org.geotools.data.postgis;
+
+import java.util.Date;
+
+/**
+ * Used internally by PostGIS dialect to signal the use of the "bigdate" domain type.
```
aaime: Wondering, if used internally, does it have to be public, or could it be made package private? I did not actually check if there are impediments, just wondering.

modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/PostgisFilterToSQL.java
```diff
@@ -133,7 +134,18 @@ public Object visit(Function function, Object extraData) throws RuntimeException
             throw new RuntimeException(e);
         }
     }
-
```
aaime: Header not updated

modules/plugin/jdbc/jdbc-postgis/src/test/java/org/geotools/data/postgis/PostgisUDTOnlineTest.java
```diff
@@ -16,16 +16,30 @@
  */
 package org.geotools.data.postgis;
```
aaime: Header not updated

modules/plugin/jdbc/jdbc-postgis/src/test/java/org/geotools/data/postgis/PostgisUDTTestSetup.java
```diff
@@ -18,6 +18,9 @@
 
 import org.geotools.jdbc.JDBCUDTTestSetup;
 
+import static java.lang.String.format;
```
aaime: Header not updated

modules/plugin/jdbc/jdbc-postgis/src/test/java/org/geotools/data/postgis/ps/PostgisUDTOnlineTest.java
```diff
@@ -25,7 +25,7 @@
  *
  * @source $URL$
  */
-public class PostgisUDTOnlineTest extends JDBCUDTOnlineTest {
+public class PostgisUDTOnlineTest extends org.geotools.data.postgis.PostgisUDTOnlineTest {
```
aaime: Header not updated

aaime:
Looks good, found some minor header issues

jdeolive:
Thanks for the review guys. I'll update the headers. We should also be able to make BigDate package protected.

PR #1174, opened 2016-04-16
Store MemoryDataStore features in each MemoryEntry

jodygarnett:
The goal with ContentDataStore was to use objects, rather than Maps for organization. This refactor changes MemoryDataStore from Map<String,Map<String,Feature>> to a MemoryEntry managed by the ContentDataStore superclass.

Although this refactor is functionally identical changing to objects makes the logic easier to follow.

To enable this refactor ContentEntry is no longer a final class.


jodygarnett:
GeoServer builds correctly after this change (not that it uses MemoryDataStore much).

jodygarnett:
While this can wait until after RC1 until being merged - I would like it applied to master before next geotools training course.

PR #1176, opened 2016-04-22
[GEOT-5406] Ignore NetCDF grid_mapping_name that is present but unsup…

bencaradocdavies:
…ported

The current behaviour of ```NetCDFProjection``` when encountering an unsupported NetCDF ```grid_mapping_name``` is to throw a ```NullPointerException```. This fix changes its behaviour to instead log a message and return null, consistent with the behaviour earlier in the method for a missing ```grid_mapping_name```.

https://osgeo-org.atlassian.net/browse/GEOT-5406


simboss:
Obvious fix, I would add a test ;)

bencaradocdavies:
@simboss test added; always worthwhile. :-)

bencaradocdavies:
Added even more testing goodness.

simboss:
I would go ahead and merge this.

Would you be so kind to backport this to 14.x?

bencaradocdavies:
Sure. Cherry-picked onto 14.x in https://github.com/geotools/geotools/commit/b87246b11bd53e02abeb48722860a4cd9b55e508

PR #1177, opened 2016-04-27
[GEOT-5409] GroupByVisitor does not allow collection wrappers to optimize visit

aaime:


modules/library/main/src/test/java/org/geotools/feature/visitor/GroupByVisitorTest.java
```diff
@@ -282,6 +291,21 @@ public void testMergingGroupByResults() throws Exception {
                 new Object[]{"THEATER", 200.0}
         });
     }
+    
+    @Test
+    public void testFeatureAttributeVisitor() {
+        GroupByVisitor visitor = buildVisitor("energy_consumption", "Average", new String[]{"building_type"});
+        List<Expression> expressions = visitor.getExpressions();
+        Set<String> names = new HashSet<String>();
```
nmco: new HashSet<String>() could be new HashSet<>()

modules/library/main/src/test/java/org/geotools/feature/visitor/GroupByVisitorTest.java
```diff
@@ -305,14 +329,21 @@ private void testVisitor(String aggregateAttribute, String aggregateVisitor,
...
+            String[] groupByAttributes) {
         GroupByVisitorBuilder visitorBuilder = new GroupByVisitorBuilder()
                 .withAggregateAttribute(aggregateAttribute, buildingType)
                 .withAggregateVisitor(aggregateVisitor);
         for (String groupByAttribute : groupByAttributes) {
             visitorBuilder.withGroupByAttribute(groupByAttribute, buildingType);
         }
         GroupByVisitor visitor = visitorBuilder.build();
-        featureCollection.accepts(visitor, new NullProgressListener());
         return visitor;
```
nmco: The visitor variable is not need could simply be: `return visitorBuilder.build();`

modules/library/main/src/main/java/org/geotools/feature/visitor/GroupByVisitor.java
```diff
@@ -337,4 +343,11 @@ public Map toMap() {
             return result;
         }
     }
+
+    @Override
+    public List<Expression> getExpressions() {
+        List<Expression> result = new ArrayList<Expression>(groupByAttributes);
```
nmco: Type argument Expression is not need

modules/library/main/src/test/java/org/geotools/feature/visitor/GroupByVisitorTest.java
aaime: We often do this on purpose to simplify debugging (gives a place where the variable can be examined before exiting the method), but sure, in this case it adds no value

nmco:
Tried it on a real system and the visitor execution is correctly delegated. Sorry for the missed interface :)

aaime:
Applied Nuno's feedback, merging

PR #1179, opened 2016-05-03
Guard log of value in ResultSetFeature.setAttribute

groldan:
org.geotools.jdbc.JDBCFeatureReader$ResultSetFeature was
converting all attribute values in setAttribute() to String
just for a no-op log statement. This patch guards the log
call with a Logger.isLoggable(Level) to avoid building a
large String when not needed, which is especially a performance
hit for geometry values.


jodygarnett:
Looks good, headers? And merge when Travis says go ...

PR #1182, opened 2016-05-03
Improve GeoTools Release Guide

tbarsballe:
After doing the GeoTools 15-RC1 release, I found that the release guide was rather out-of-date. This intends to remedy the problems I found:

**Add manual JIRA release instructions**

The current release instructions rely on the `geotools-release-jira` job on the ares jenkins server. This job has not been working for about a yer. Until someone finds the time to fix the JIRA release scripts, the Release Guide should be updated with manual release steps.

**Indicate `geotools-release-publish` is also responsible for deploying artifacts to maven.**

This is not made clear anywhere in the existing release guide, and caused some confusion when trying to start the GeoWebCache release after the artifacts had been built by the `geotools-release` job.
 
**Remove deprecated news sites**

A number of the sites in the "Announce the release" section have not been used in 2+ years. These have been removed.

### DIFF
--- a/docs/developer/procedures/release.rst
+++ b/docs/developer/procedures/release.rst
@@ -47,25 +47,11 @@ Since we don't make any release from master, ensure you select the right CITE te
 Release in JIRA
 ---------------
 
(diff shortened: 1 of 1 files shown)


jodygarnett:
Thanks - only merging to master (since it captures our procedures)

PR #1184, opened 2016-05-04
[GEOT-5397] fixed org.geotools.styling.FeatureTypeStyleImpl.clone()

bestrauss:
New test: org.geotools.styling.FeatureTypeStyleImplTest


aaime:
Thank you. Backported also to the other two active branches, 15.x and 14.x

PR #1185, opened 2016-05-04
[GEOT-5398] ECQL: format property name ID as "ID"

bestrauss:
The ECQL parser still accepts the deprecated keyword ID. Thus property
names ID, iD, Id, id must be delimited, and hence will be delimited.


aaime:
Looks good to me, thanks for the fix

aaime:
We'll probably need a backport to 15.x ad 14.x, right now I'm travelling and it's a bit hard to do the backports myself and check stuff builds. Any chance you can make further pull requests for those branches?

PR #1186, opened 2016-05-06
isolate provider to prevent SPI failure if no JAI [GEOT-561]

jodygarnett:
See: https://osgeo-org.atlassian.net/browse/GEOT-561

### DIFF
--- a/modules/library/referencing/src/main/java/org/geotools/referencing/operation/transform/WarpTransform2D.java
+++ b/modules/library/referencing/src/main/java/org/geotools/referencing/operation/transform/WarpTransform2D.java
@@ -2,7 +2,7 @@
  *    GeoTools - The Open Source Java GIS Toolkit
  *    http://geotools.org
  *
- *    (C) 2005-2008, Open Source Geospatial Foundation (OSGeo)
+ *    (C) 2005-2016, Open Source Geospatial Foundation (OSGeo)
  *
  *    This library is free software; you can redistribute it and/or
  *    modify it under the terms of the GNU Lesser General Public
@@ -19,6 +19,7 @@ package org.geotools.referencing.operation.transform;
 import java.awt.geom.Point2D;
 import java.awt.geom.Rectangle2D;
 import java.io.Serializable;
+import java.lang.reflect.Array;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
 import java.util.Calendar;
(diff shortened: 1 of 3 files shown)


modules/library/referencing/src/main/java/org/geotools/referencing/operation/transform/WarpTransform2D.java
```diff
@@ -411,14 +404,14 @@ public ParameterValueGroup getParameterValues() {
...
+            p[c++] = new Parameter(WarpTransform2DProvider.Y_COEFFS, poly.getYCoeffs());
             float s;
-            if ((s=poly.getPreScaleX ()) != 1) p[c++] = new Parameter(Provider. PRE_SCALE_X, s);
-            if ((s=poly.getPreScaleY ()) != 1) p[c++] = new Parameter(Provider. PRE_SCALE_Y, s);
-            if ((s=poly.getPostScaleX()) != 1) p[c++] = new Parameter(Provider.POST_SCALE_X, s);
-            if ((s=poly.getPostScaleY()) != 1) p[c++] = new Parameter(Provider.POST_SCALE_Y, s);
+            if ((s=poly.getPreScaleX ()) != 1) p[c++] = new Parameter(WarpTransform2DProvider. PRE_SCALE_X, s);
+            if ((s=poly.getPreScaleY ()) != 1) p[c++] = new Parameter(WarpTransform2DProvider. PRE_SCALE_Y, s);
+            if ((s=poly.getPostScaleX()) != 1) p[c++] = new Parameter(WarpTransform2DProvider.POST_SCALE_X, s);
+            if ((s=poly.getPostScaleY()) != 1) p[c++] = new Parameter(WarpTransform2DProvider.POST_SCALE_Y, s);
```
jodygarnett: I do wonder if I should make a useless static Provider class so that these public parameters do not have to change?
{code}
/** Placeholder interface used for parameter definition */
static interface Provider {
        public static final ParameterDescriptor X_COEFFS = new DefaultParameterDescriptor("xCoeffs", float[].class, null, null);
        ...
}
{code}

The actual provider had to move to a separate class (and use reflection) to avoid crashing SPI factory lookup.

cuttlefish:
This has been causing us issues in GeoGig. The changes look reasonable to me. I'll try it out in GeoGig to see if it resolves our issue.

cuttlefish:
I tried merging this into 14.1
It resolved the error we were getting in GeoGig. Thanks @jodygarnett!

jodygarnett:
I would like to backport this for the GeoMesa project.

PR #1187, opened 2016-05-09
Improving thread safety of SimpleFeatureTypeImpl.getTypes

elahrvivaz:
* Avoids the chance of a second thread seeing a partially filled type list

I didn't add a unit test as thread safety is hard to test reliably, but it should be evident how the old code could result in a second thread seeing a non-null but empty or partially filled type list.

Signed-off-by: Emilio Lahr-Vivaz <elahrvivaz@ccri.com>

### DIFF
--- a/modules/library/main/src/main/java/org/geotools/feature/simple/SimpleFeatureTypeImpl.java
+++ b/modules/library/main/src/main/java/org/geotools/feature/simple/SimpleFeatureTypeImpl.java
@@ -76,10 +76,11 @@ public class SimpleFeatureTypeImpl extends FeatureTypeImpl implements
         if (types == null) {
             synchronized (this) {
                 if (types == null) {
-                    types = new ArrayList<AttributeType>();
+                    ArrayList<AttributeType> temp = new ArrayList<AttributeType>();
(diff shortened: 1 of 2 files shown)


aaime:
Copyright headers were not updated. Adding a MT test should be quick, we have a few of them in the code base, e.g.:
https://github.com/geotools/geotools/blob/master/modules/library/main/src/test/java/org/geotools/filter/function/FilterFunctionPropertyTest.java
https://github.com/geotools/geotools/blob/master/modules/library/render/src/test/java/org/geotools/map/MapContentConcurrencyTest.java
https://github.com/geotools/geotools/blob/master/modules/unsupported/geojson/src/test/java/org/geotools/geojson/GEOT5036RegressionTest.java

They don't need to fail every single time, a failure every few runs is normally enough to make devs aware of a possible regression

elahrvivaz:
Updated headers and added a test that fails approximately 1 in 5 times before the patch (run through intellij).

aaime:
Thanks! Merging on master, right now we are very close to the release, so it's probably safer not to backport, but you might want to open pull requests for 15.x and 14.x once the 15.0 release is out.

PR #1188, opened 2016-05-11
[GEOT-5416] Fix so SortOrder.DESCENDING works

ianturton:
[GEOT-5416](https://osgeo-org.atlassian.net/browse/GEOT-5416) - Currently SortOrder.DESCENDING is handled in the same way as SortOrder.ASCENDING which is clearly wrong. This PR Fixes the issue.

### DIFF
--- a/modules/library/main/src/main/java/org/geotools/data/DataUtilities.java
+++ b/modules/library/main/src/main/java/org/geotools/data/DataUtilities.java
@@ -2,7 +2,7 @@
  *    GeoTools - The Open Source Java GIS Toolkit
  *    http://geotools.org
  * 
- *    (C) 2003-2015, Open Source Geospatial Foundation (OSGeo)
+ *    (C) 2003-2016, Open Source Geospatial Foundation (OSGeo)
  *    
  *    This library is free software; you can redistribute it and/or
  *    modify it under the terms of the GNU Lesser General Public
--- a/modules/library/main/src/main/java/org/geotools/data/DataUtilities.java
+++ b/modules/library/main/src/main/java/org/geotools/data/DataUtilities.java
@@ -116,6 +116,7 @@ import org.opengis.filter.FilterFactory2;
 import org.opengis.filter.expression.Expression;
 import org.opengis.filter.expression.PropertyName;
 import org.opengis.filter.sort.SortBy;
+import org.opengis.filter.sort.SortOrder;
 import org.opengis.geometry.BoundingBox;
 import org.opengis.metadata.citation.Citation;
 import org.opengis.referencing.FactoryException;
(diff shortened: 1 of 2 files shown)


modules/library/main/src/main/java/org/geotools/data/DataUtilities.java
```diff
@@ -2395,6 +2396,7 @@ public int compare(SimpleFeature f1, SimpleFeature f2) {
             };
         } else {
             final PropertyName PROPERTY = sortBy.getPropertyName();
+            final SortOrder ORDER = sortBy.getSortOrder();
```
bencaradocdavies: @ianturton neither of these local final variables should be uppercase; although they are final, their value varies between invocations, so they are not lexically constant. I know the first was there already and you are only being consistent, but this practice violates the code conventions. Perhaps change these to ```propertyName``` and ```sortOrder``` respectively?

modules/library/main/src/main/java/org/geotools/data/DataUtilities.java
```diff
@@ -2404,7 +2406,11 @@ public int compare(SimpleFeature f1, SimpleFeature f2) {
                         return 0; // cannot perform comparison
                     }
                     if (value1 instanceof Comparable && value1.getClass().isInstance(value2)) {
-                        return ((Comparable<Object>) value1).compareTo(value2);
+                        if(ORDER == SortOrder.ASCENDING) {
```
bencaradocdavies: Missing space between ```if``` and ```(```.

modules/library/main/src/main/java/org/geotools/data/DataUtilities.java
```diff
@@ -2404,7 +2406,11 @@ public int compare(SimpleFeature f1, SimpleFeature f2) {
...
+                            return ((Comparable<Object>) value1).compareTo(value2);
+                        }else {
```
bencaradocdavies: Missing space between ```}``` and ```else```.

modules/library/main/src/test/java/org/geotools/data/collection/SortedFeatureCollectionTest.java
```diff
@@ -50,6 +50,12 @@ public void testSortAttribute() throws Exception {
                 new SortBy[] { sort });
         checkSorted(sorted, DataUtilities.sortComparator(sort));
     }
+    public void testSortAttributeDescending() throws Exception {
```
bencaradocdavies: Should be a blank line before this method.

bencaradocdavies:
@ianturton this fix makes sense to me and has a unit test. Needs a few style tweaks and then one Jenkins CI passes I think it should be merged.

ianturton:
Fixed minor style issues

bencaradocdavies:
That looks great @ianturton you are a champ. Merge as soon as this PR has been anointed with Jenkins CI holy water.

bencaradocdavies:
Thanks @ianturton.

PR #1192, opened 2016-05-13
protect against process factories that are unavailable [GEOT-5418]

jodygarnett:
Also protect against parameters that are not available on the classpath.

 https://osgeo-org.atlassian.net/browse/GEOT-5418

### DIFF
--- a/modules/unsupported/process/src/main/java/org/geotools/process/function/ProcessFunctionFactory.java
+++ b/modules/unsupported/process/src/main/java/org/geotools/process/function/ProcessFunctionFactory.java
@@ -2,7 +2,7 @@
  *    GeoTools - The Open Source Java GIS Toolkit
  *    http://geotools.org
  *
- *    (C) 2008, Open Source Geospatial Foundation (OSGeo)
+ *    (C) 2008-2016, Open Source Geospatial Foundation (OSGeo)
  *
  *    This library is free software; you can redistribute it and/or
  *    modify it under the terms of the GNU Lesser General Public
@@ -22,6 +22,7 @@ import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.logging.Logger;
 
 import org.geotools.data.Parameter;
 import org.geotools.feature.NameImpl;
@@ -41,9 +42,6 @@ import org.opengis.filter.expression.Literal;
  * value can be seen as a filter function
  * 
  * @author Andrea Aime - GeoSolutions
- * 
- *
- *
  *
  * @source $URL$
  */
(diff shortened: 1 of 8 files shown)


jodygarnett:
@tbarsballe and @jnh5y will this solve your geomesa deployment issue

tbarsballe:
Have a bug report: https://osgeo-org.atlassian.net/browse/GEOT-5418

tbarsballe:
This looks to have fixed the issue; GeoMesa builds after applying this change to GeoTools.
I was not able to determine what process was actually causing the error - it looks like the GeoMesa build might be suppressing some logging.

jnh5y:
@jodygarnett @tbarsballe thanks guys!  Can we get this ported to the 15.x series as well?

tbarsballe:
That's the plan. Ideally this will be included in the 15.0 release.
@jodygarnett - I think this is good to merge; If you want to merge to master, I can backport to 15.x.

jnh5y:
As a note, I very, very briefly tested this change as a backport to 15-RC1, and it worked.

simboss:
What about adding a test?

aaime:
Agreed about the test. Should be easy and clean, the Processors class has registerProcessFactory/removeProcessFactory methods that allow the test to temporarily add a mock factory claiming not to be available.

jodygarnett:
I have had a go at the following:
````
        ProcessFactory good = new TestFactory(new NameImpl("test", "good"), true);
        ProcessFactory evil = new TestFactory(new NameImpl("test", "evil"), false);
        ProcessFunctionFactory factory = new ProcessFunctionFactory();
        try {
            Processors.addProcessFactory(good);
            Processors.addProcessFactory(evil);
            
            List<FunctionName> list = factory.getFunctionNames();
            
            assertTrue("available", list.contains(new NameImpl("test", "good")));
            assertFalse("unavailable", list.contains(new NameImpl("test", "evil")));
        } finally {
            Processors.removeProcessFactory(good);
            Processors.removeProcessFactory(evil);
        }
````

Unfortunately the add/remove process factory methods are not being respected, the factories are not returned by `Processors.getProcessFactories()` which returns a single `BufferFeatureCollectionFactory` factory.

**Comment by aaime** [comment]
Thanks for adding the test. Not sure about the programmatic registration issue, GeoServer is using it without problems as far as I know.

File: modules/unsupported/process/src/main/java/org/geotools/process/function/ProcessFunctionFactory.java [hunk 1]
```diff
@@ -137,32 +135,37 @@ private synchronized void init() {
             processToFunction = new HashMap<Name,FunctionName>();
             functionNames = new ArrayList<FunctionName>();
             for (ProcessFactory factory : factories) {
+                if( !factory.isAvailable() ) continue;
```
aaime: Not a stopped, you might just ignore the comment, but this does not align with our coding conventions, it should have been:

````
if(!factory.isAvailable()) {
  continue;
}

File: modules/unsupported/process/src/main/java/org/geotools/process/function/ProcessFunctionFactory.java [hunk 2]
```diff
@@ -137,32 +135,37 @@ private synchronized void init() {
...
+                            FunctionName functionName = new FunctionNameImpl(processName, result, args);
+                            functionNames.add(functionName);
+                            processToFunction.put(processName, functionName);
                         }
-
-                        FunctionName functionName = new FunctionNameImpl(processName, result, args);
-                        functionNames.add(functionName);
-                        processToFunction.put(processName, functionName);
+                    }
+                    catch (Throwable unavailable){
```
aaime: Similarly here, there should have been a space between ) and {

modules/unsupported/process/src/test/java/org/geotools/process/factory/VectorIdentityRTProcess.java
```diff
@@ -1,3 +1,19 @@
+/*
```
aaime: Thanks for adding the missing headers

modules/unsupported/process/src/main/java/org/geotools/process/function/ProcessFunctionFactory.java
bencaradocdavies: And ```catch``` should be on the same line as the preceding closing brace:
```
} catch (Throwable unavailable) {

jodygarnett:
Thanks for feedback, squashed and merged.

PR #1193, opened 2016-05-16
[GEOT-5420] Performance regression creating image mosaic reader against databases with thousands of tables

aaime:
Copyright headers are all already updated to 2016


aaime:
Closing, as written it breaks netcdf multicoverage support

PR #1197, opened 2016-05-17
[GEOT-5420] Performance regression creating image mosaic reader agaist databases with thousands of tables

aaime:
Does not require a new test in that it's just a different solution to the GEOT-5168 (pure performance optimization)


dromagnoli:
Looks good to me. Waiting for Travis check for merge

PR #1200, opened 2016-05-27
Changed references to JDK 7 to be JDK 8

dvntucker:
- Updated a few references to JDK 7 in some documentation
- Removed warning about Maven and Apt-Get. Ubuntu repositories contain Maven 3 since at least 14. Added a reference to official maven install guide if user's would like the most up-to-date version.

### DIFF
--- a/docs/user/tutorial/quickstart/artifacts/pom.xml
+++ b/docs/user/tutorial/quickstart/artifacts/pom.xml
@@ -11,7 +11,7 @@
     <url>http://maven.apache.org</url>
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <geotools.version>13.2</geotools.version>
+        <geotools.version>14.3</geotools.version>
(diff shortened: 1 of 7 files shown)


jodygarnett:
Do we need to change any of the pom.xml file compiler settings to make use of source / target `1.8` ?

dvntucker:
Updated the quickstart documentation to contain the Maven compiler settings.

docs/user/tutorial/quickstart/artifacts/pom.xml
```diff
@@ -11,7 +11,7 @@
     <url>http://maven.apache.org</url>
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <geotools.version>13.2</geotools.version>
+        <geotools.version>14.3</geotools.version>
```
mprins: this should be 15.0 since that's the java 8 version right?

PR #1201, opened 2016-05-31
Return null instead of throwing exception when parsing empty dates [GEOT-5435]

fxprunayre:
For all types (eg. Double) a check is made on ```""``` and return null in this case. Apply similar approach to empty date/dateTime/time/duration types instead of throwing SAXException.

Let me know if this is relevant or not.

Related discussion http://osgeo-org.1560.x6.nabble.com/WFS-feature-with-empty-date-td5268563.html


aaime:
Hi Francois, the changes looks good to me. Before merging the following should happen though:

- A ticket is created in jira to track this change (for the release notes)
- The commit message should refer the ticket
- The pull request should first be made against master, then backported. No need to close this one though, just make sure to have one for master as well.

aaime:
I've merged the master version, let's see if the GeoServer build and the subsequent CITE test pass, then we can backport.

PR #1203, opened 2016-06-02
GEOT-5430 Check for empty or null geometry before doing offset

dvntucker:
[GEOT-5430]

Unfortunately I haven't been able create a test case for this due to the nature of the issue. I'm all ears for suggestions. The issue boils down to the following:

- StreamingRenderer clips a MultiPolygon that's actually out of the envelope bounds. This produces an empty polygon.
- Offset code then attempts to offset the vector, but fails since the geometry is empty

Except in most cases the code shouldn't even make it this far, since features outside the bounds should be filtered out anyway. I'm not sure why it doesn't happen in the case of the data in the ticket, it could have something to do with the projection?

In any case, this is a pretty innocuous null + empty check anyway, and test coverage is already pretty good there anyway, so maybe we could merge this without a specific test case?

### DIFF
--- a/modules/library/render/src/main/java/org/geotools/renderer/lite/StreamingRenderer.java
+++ b/modules/library/render/src/main/java/org/geotools/renderer/lite/StreamingRenderer.java
@@ -2,7 +2,7 @@
  *    GeoTools - The Open Source Java GIS Toolkit
  *    http://geotools.org
  * 
- *    (C) 2004-2015, Open Source Geospatial Foundation (OSGeo)
+ *    (C) 2004-2016, Open Source Geospatial Foundation (OSGeo)
  *
  *    This library is free software; you can redistribute it and/or
  *    modify it under the terms of the GNU Lesser General Public
@@ -2561,7 +2561,8 @@ public class StreamingRenderer implements GTRenderer {
                     Geometry g = clipper.clipSafe(shape.getGeometry(), preserveTopology, 1);
                     
                     // handle perpendincular offset as needed
-                    if(style instanceof LineStyle2D && ((LineStyle2D) style).getPerpendicularOffset() != 0) {
+                    if(style instanceof LineStyle2D && ((LineStyle2D) style).getPerpendicularOffset() != 0
+                            && g != null && !g.isEmpty()) {
                         LineStyle2D ls = (LineStyle2D) style;
                         double offset = ls.getPerpendicularOffset();
                         // people applying an offset on a polygon really expect a buffer instead,
--- a/modules/library/main/src/main/java/org/geotools/geometry/jts/GeometryClipper.java
+++ b/modules/library/main/src/main/java/org/geotools/geometry/jts/GeometryClipper.java
@@ -94,7 +94,7 @@ public class GeometryClipper {
      * @param g
      * @param ensureValid
      * @param scale Scale used to snap geometry to precision model, 0 to disable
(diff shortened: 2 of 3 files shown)


modules/library/render/src/main/java/org/geotools/renderer/lite/StreamingRenderer.java
```diff
@@ -2561,7 +2561,8 @@ private void processSymbolizers(final Graphics2D graphics,
                     Geometry g = clipper.clipSafe(shape.getGeometry(), preserveTopology, 1);
                     
                     // handle perpendincular offset as needed
-                    if(style instanceof LineStyle2D && ((LineStyle2D) style).getPerpendicularOffset() != 0) {
+                    if(style instanceof LineStyle2D && ((LineStyle2D) style).getPerpendicularOffset() != 0
+                            && g != null && !g.isEmpty()) {
```
jodygarnett: No problem with the additional safety checks.

Is this a case of ``clipSafe(shape.getGeometry(), preserveTopology, 1)`` api not being respected? If so please update the javadocs on the method with a clarification.

dvntucker: I have updated the JavaDoc to point out that it can return null or an empty geometry.

aaime:
Here is a test that triggers the error you described, committed to master already, just remove the @Ignore and verify your fix does the trick:
https://github.com/geotools/geotools/commit/0d8578819346c48c0265be23674dcb2495a8d515

As for always having tests, I'm surprised @jodygarnett did not explain the rationale, which is included in the geoserver pull request review guidelines, that he helped to author:
https://github.com/geoserver/geoserver/wiki/Pull-request-review-guide
Look for "Presence of tests". It is not related to code coverage per se, but to the dynamics in which the code base evolves, and warding against regressions.

dvntucker:
Thanks so much for the test case Andrea.

Looking through the debugger at your test case, if I'm understanding it correctly:

- the BBox filter is buffered by a small amount
- which causes our multipolygon to be part of the FeatureCollection, despite being technically out of the request bounds
- eventually the multipolygon is clipped because it doesn't actually fall within the rendering bounds, which are now transformed to screen space

Does that sound right? Just trying to understand the problem. That explains why my attempts at a test case weren't working, my multipolygon was too far outside the bounds to be captured in that manner.

dvntucker:
I have un-@Ignore'd aaime's test case.

aaime:
The problem is that the renderer uses bbox vs bbox intersection tests for performance reasons (real intersections are way slower than anything else the rendering might be doind), so anything "surrounding" the rendering area gets captured :-)

jodygarnett:
Oh don't worry I had explained. I was however daunted by the test data size @dvntucker provided (300k to force testing a null check). Attempts to simplify ended up getting filtered etc...

aaime:
@jodygarnett agreed a 300kb shapefile is no good for a test. For next time and similar issues:
* Setup the shapefile for failing
* Conditional breakpoint in debugger that only stops if the result is empty or null
* Get the source geometry, get the WKT of it in the debugger (in Eclipse it would be the "Display" tab, that allows to play with the current variables on the stack and call methods on them)
* Create a property file with just one feature with that geometry (hopefully not too big)
* Create a rendering test that uses it mimicking the many others

jodygarnett:
thanks @dvntucker and @aaime

PR #1205, opened 2016-06-04
[GEOT-5428] Support NetCDF rotated pole projection

bencaradocdavies:
See: https://osgeo-org.atlassian.net/browse/GEOT-5428

CF Conventions section for rotated pole projection:
http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/build/ch05s06.html

@simboss or @dromagnoli please review. I will make some line comments to explain a few of the details.


modules/plugin/coverage-multidim/netcdf/src/main/java/org/geotools/coverage/io/netcdf/crs/NetCDFProjection.java
```diff
@@ -111,6 +112,29 @@ public String getName() {
...
     }
+
+    /**
+     * Subclasses override this if they wish to adjust OGC parameters after they are read from NetCDF. This is the inverse of
+     * {@link #getNetcdfParameters(ParameterValueGroup)}.
+     * 
+     * @param netcdfParameters parameter values read from NetCDF
+     * @return parameter values used for OGC projection
+     */
+    public ParameterValueGroup getOgcParameters(ParameterValueGroup netcdfParameters) {
```
bencaradocdavies: The problem with the current implementation is that it assumes the mapping between NetCDF and OGC parameters is a simple renaming. This is not true for the rotated pole projection. These two new methods, whose default implementations do nothing, provide a lightweight mechanism for projection subclasses to change parameter values that is better than adding more ```if``` clauses to ```handleParam```. If we see another case that needs these methods, a more extensive refactoring might be justified.

dromagnoli:
@bencaradocdavies: Looks good to me.
+1 on merging it

bencaradocdavies:
Thanks @dromagnoli. Merged and, after the builds passed, cherry-picked onto 15.x (build also passed).

PR #1206, opened 2016-06-04
Return null instead of throwing exception when parsing empty dates [GEOT-5435]

fxprunayre:
Same as #1201 for master.


aaime:
Looks good to me, merging

PR #1214, opened 2016-06-15
Added processes for convolve and normalizing a coverage

jdeolive:
- Convolve process simply exposes the Convolve raster operation
- The normalize process takes a raster, computes its max value, and then involves the divide operation by the max to produce a raster with values between [0,1], "normalizing" the values within it

### DIFF
--- a/modules/unsupported/process-raster/src/main/java/org/geotools/process/raster/ConvolveCoverageProcess.java
+++ b/modules/unsupported/process-raster/src/main/java/org/geotools/process/raster/ConvolveCoverageProcess.java
@@ -1,0 +1,71 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ *
+ *    (C) 2016, Open Source Geospatial Foundation (OSGeo)
+ *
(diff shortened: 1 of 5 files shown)


modules/unsupported/process-raster/src/main/java/org/geotools/process/raster/ConvolveCoverageProcess.java
```diff
@@ -0,0 +1,71 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ *
+ *    (C) 2011-2016, Open Source Geospatial Foundation (OSGeo)
```
aaime: These are new files, shouldn't the be just "(C) 2016, Open Source Geospatial Foundation (OSGeo)"
(also considering the last GSIP on the topic?)

aaime:
One minor question on the copyright headers, besides that, looks good to me.

jdeolive:
Thanks for reviewing Andrea. My bad on the headers, will fix those. Once travis is happy I'll go ahead and merge. Thanks!

PR #1215, opened 2016-06-16
[GEOT-5421] NullPointerException when GridCoverage goes offscreen

bencaradocdavies:
This 14.x backport includes ```final``` qualifier on ```error``` variable so compiles with ```<source>1.7</source>```. Full build with OpenJDK 8 passes.

See: https://osgeo-org.atlassian.net/browse/GEOT-5421


aaime:
On 14.x it has to be built with Java 7 though?

aaime:
Anyways, yes, I don't see any potential side effect, so if it builds with java 7 we're good.

aaime:
Right... do you have java 7 to start with? Otherwise I can do a build myself

bencaradocdavies:
@aaime Travis CI has Java 7 for 14.x; this was my main reason for the PR.  :-)

bencaradocdavies:
I see an Oracle JDK 7 and OpenJDK 7 build running on Travis CI.

I agree: relying on ```<source>1.7</source>``` is not enough because new API in libraries could also break the build on Java 7. The only way to be sure is to build with Java 7.

If the Travic CI builds on Java 7 pass, you should be good to go.  :-)

Unfortunately openjdk-7-jdk was removed from my platform (debian/unstable) because of timezone package dependency conflicts, so I am relying on Travis CI.

bencaradocdavies:
Thanks @aaime. I hope that the release goes smoothly.

PR #1216, opened 2016-06-16
GEOT-5441: Fixed gmlparsing of nested members

mtoonen:


aaime:
Nice, patch with tests, thank you.

Looking at the code, the coding conventions are not respected (spacing mostly). It would be nice if you could fix that (without reformatting the whole file, just the parts that you have added).

modules/extension/xsd/xsd-gml3/src/main/java/org/geotools/gml3/bindings/PolygonTypeBinding.java
```diff
@@ -98,15 +99,38 @@ public Class getType() {
     public Object parse(ElementInstance instance, Node node, Object value)
         throws Exception {
         //TODO: schema allows no exterior ring, but what the heck is that all about ?
-        LinearRing exterior = (LinearRing) node.getChildValue("exterior");
+        Object o = node.getChildValue("exterior");
+
+        LinearRing exterior = null;
+        if(o instanceof String){
```
aaime: I'm confused, when does it happen that one gets back a String instead a geometry? By your tests it seems the ring case, but that looks to me more of an issue in another binding not working as it should.. like, a RingBinding should be returning a LinearRing, the bug imho should be searched in there and/or in the CurvedGeometryFactory.

mtoonen: Yes, I figured it could be something like that, but I was a bit lost on where that should be fixed. Will try monday to figure it out where the more correct place would be. 
Thanks!

PR #1217, opened 2016-06-16
GEOT-5442: fix same separator in gml:coordinates

mtoonen:


modules/extension/xsd/xsd-gml2/src/main/java/org/geotools/gml2/bindings/GMLCoordinatesTypeBinding.java
```diff
@@ -131,13 +131,23 @@ public Object parse(ElementInstance instance, Node node, Object value)
         CoordinateSequence seq = null;
         int i = 0;
         int ncoords = tuples.countTokens(); //number of coordinates
+        if(cs.equals(ts)) {
+            ncoords = ncoords/2;
```
aaime: Doesn't this assume 2d coordinates? What if the GML has ts=cs but it's representing 3d coordinates?
Hmm... I guess we cannot recognize anymore that case by just looking at the string, like it's done today, one would have to read the srs and make assumptions (it's not uncommon to declare a 2d SRS but provide 3d ordinates).
Oh well, it's not perfect but GML 2 is old and this patch seems to make the parser work better than before for this particular case.

aaime:
Merging, let's see how the CITE tests in GeoServer like this change :-)

modules/extension/xsd/xsd-gml2/src/main/java/org/geotools/gml2/bindings/GMLCoordinatesTypeBinding.java
mtoonen: Yes, you're correct, it does assume 2D, but I figured: it was better than before:-).
However, I might be wrong, but in my tests I think I encountered GML3 geometries, which were parsed with this binding.

mtoonen:
Thanks for reviewing/merging!

PR #1218, opened 2016-06-20
[GEOT-5445] Add alias handling to ScalarValidator

tbarsballe:
PR for https://osgeo-org.atlassian.net/browse/GEOT-5445

While this is fix is an improvement,, there should really be some sort of validation for Aliases. However, this does not seem to be implemented anywhere in the YSLD validators currently and implementation appears to be nontrivial.

### DIFF
--- a/modules/unsupported/ysld/src/test/java/org/geotools/ysld/validate/YsldValidateTest.java
+++ b/modules/unsupported/ysld/src/test/java/org/geotools/ysld/validate/YsldValidateTest.java
@@ -780,4 +780,50 @@ public class YsldValidateTest {
         List<MarkedYAMLException> errors = validate(builder.toString());
         assertThat(errors, empty());
     }
+    
+    @SuppressWarnings("unchecked")
+    @Test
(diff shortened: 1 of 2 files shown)


modules/unsupported/ysld/src/test/java/org/geotools/ysld/validate/YsldValidateTest.java
```diff
@@ -780,4 +780,48 @@ public void testColourMapAsVariable() throws Exception {
         List<MarkedYAMLException> errors = validate(builder.toString());
         assertThat(errors, empty());
     }
+    
+    @SuppressWarnings("unchecked")
+    @Test
+    public void testColourAsVariable() throws Exception {
```
smithkm: A comment to explain  what the sub  tests are doing would help.

tbarsballe:
Added  a comment to the test

smithkm:
Thanks, merging.

jodygarnett:
was this backported?

tbarsballe:
Not to my knowledge

PR #1219, opened 2016-06-21
Property data store fails to write geometries with 3d ordinates

aaime:


aaime:
Merging to get some feedback from the GeoServer build and CITE tests

PR #1221, opened 2016-06-23
increased oracle nvarchar2 max size to 4000 [GEOT-5451]

ianturton:
fix for https://osgeo-org.atlassian.net/browse/GEOT-5451 
I can't see any way to add a unit test for this.

### DIFF
--- a/modules/plugin/jdbc/jdbc-oracle/src/test/java/org/geotools/data/oracle/OracleDataStoreOnlineTest.java
+++ b/modules/plugin/jdbc/jdbc-oracle/src/test/java/org/geotools/data/oracle/OracleDataStoreOnlineTest.java
@@ -99,6 +99,8 @@ public class OracleDataStoreOnlineTest extends JDBCDataStoreOnlineTest {
         	vBuffer.append("x");
         }
         fBuilder.add(vBuffer.toString());
+     // used to fail here
+        dataStore.createSchema(featureType);
     }
 
 }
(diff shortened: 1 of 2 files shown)


aaime:
Sounds like a good candidate for a new createSchema test in OracleDataStoreOnlineTest? There are a couple extras already.

modules/plugin/jdbc/jdbc-oracle/src/test/java/org/geotools/data/oracle/OracleDataStoreOnlineTest.java
```diff
@@ -77,5 +78,27 @@ public void testCreateSpatialIndexNameTooLong() throws Exception {
...
+        
+        dataStore.createSchema(featureType);
+        SimpleFeatureBuilder fBuilder = new SimpleFeatureBuilder(featureType);
+        fBuilder.add(null);
+        StringBuffer vBuffer = new StringBuffer(4000);
+        //to be honest I can't tell from the oracle docs if 4000 or 3999 is the actual limit
+        for(int i=0;i<3999;i++) {
+        	vBuffer.append("x");
+        }
+        fBuilder.add(vBuffer.toString());
```
aaime: Unsure what is happening here... is the feature builder going to throw an exception if we go above the limits?
Btw, there is also FeatureTypes.getFieldLength that could come in handy for a test

ianturton: I didn't know about getFieldLength - but since this is an Oracle error on write I don't think it helps.

aaime: If it happens on write... I don't see the write in the test. Sorry to be a pest Ian, just trying to understand how this test would fail if the fix was not in place

PR #1228, opened 2016-06-29
Oracle DataStore is unable to add empty geometries [GEOT-742]

ianturton:
See [GEOT-742](https://osgeo-org.atlassian.net/browse/GEOT-724) for full discussion.

This fixes two problems:

  1. Null Pointer thrown when there is no coordinate sequence
  1. Inserts a null into the table as Oracle is too dumb to handle empty geometries (see [discussion](https://gis.stackexchange.com/questions/200270/is-it-possible-to-store-empty-geometries-in-an-oracle-database)).


aaime:
Whoa, closing an almost 11 years old ticket? Kudos :-)

PR #1231, opened 2016-06-30
Change test data for granule collector to not need axis order

dvntucker:
Changed the test data for GranuleCollectorSPITest to no longer need the axis order change. This should fix the failure on OpenJDK 8.


simboss:
Let's see if this fixes the build

bencaradocdavies:
Thanks @simboss, this fixed the build for me with openjdk-8-jdk 8u91-b14-3 amd64 on debian/unstable.

bencaradocdavies:
But ouch, did I just see it fail on Oracle JDK 8 on ares Jenkins?

bencaradocdavies:
@simboss I could not reproduce the Oracle JDK 8 failure locally in Maven or Eclipse. This change breaks geotools-master on ares Jenkins.

dvntucker:
Hmm, I have an idea what's going on. Most likely the files are being indexed in a different order depending on the OS, causing the bounding box used in the test to be invalid. Ideally the test should use the bounding box of the mosaic CRS rather than a fixed one. Should be able to fix sometime this weekend.

PR #1232, opened 2016-07-05
[GEOT-5458] WFS 1.0.0 / Tolerate empty OnlineResource.

fxprunayre:
See https://osgeo-org.atlassian.net/browse/GEOT-5458


aaime:
Hi @NielsCharlier, here @fxprunayre made a fix for the wfs store, I was wondering if wfs-ng might be affected by the same issue. Quickly glancing over the code, it seems that a null online resource is considered, so maybe not affected.

PR #1241, opened 2016-07-18
WFS Legacy Naming (GEOT-4883)

NielsCharlier:
https://osgeo-org.atlassian.net/browse/GEOT-4883

### DIFF
--- a/modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/impl/WFSDataAccessFactory.java
+++ b/modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/impl/WFSDataAccessFactory.java
@@ -119,7 +119,7 @@ public class WFSDataAccessFactory implements DataAccessFactory {
     }
 
     /** Access with {@link WFSDataStoreFactory#getParametersInfo()  */
-    private static final WFSFactoryParam<?>[] parametersInfo = new WFSFactoryParam[18];
+    private static final WFSFactoryParam<?>[] parametersInfo = new WFSFactoryParam[19];
 
     private static final int GMLComplianceLevel = 2;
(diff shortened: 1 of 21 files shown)


modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/impl/WFSDataAccessFactory.java
```diff
@@ -400,6 +400,18 @@ public T lookUp(final Map params) throws IOException {
...
+    /**
+     * Optional {@code Integer} OCG GML compliance level. i.e. (simple feature) 0, 1 or 2
+     */
+    public static final WFSFactoryParam<Boolean> GML_COMPATIBLE_TYPENAMES;
+    static {
+        String name = "WFSDataStoreFactory:GML_COMPATIBLE_TYPENAMES";
+        String title = "GmlCompatibleTypeNames";
+        String description = "Use Gml Compatible TypeNames (replace : by _).";
+        parametersInfo[18] = GML_COMPATIBLE_TYPENAMES = new WFSFactoryParam<Boolean>(name,
+                Boolean.class, title, description, false);
```
jodygarnett: Is there any reason why this should be anything other than true?

I guess backwards compatibility for wfs-ng users, but since GML output would not of worked in this case previously...

aaime: Jody, a data store must care of data access, the idea that it is tasked of policing the name is just... wrong. Whatever is tasked of producing GML and/or WFS in general should care of making names valid. WFS is not the only reason why one would use a store and names should be reported as they are.

PR #1242, opened 2016-07-20
Add pgraster 32 bit float tests

gevaerts:
The new testcase reproduces GEOT-5463.

The only changes to resources.zip are three new sql files, */pgraster32bit.sql.

Note that due to the error occuring in a worker thread the tests don't actually fail.


bencaradocdavies:
Thanks for the PR. Closing as you indicate that this test should fail. PRs should include fixes that make the PR pass. I note that Travis CI passes for this, presumably because the test is disabled on Travis CI.

gevaerts:
I do indeed assume that Travis CI won't run the test (it's in the set that's normally ignored due to being "online"), but due to the threaded worker architecture the exception doesn't get passed back to the original caller anyway so apart from the stack trace appearing in stderr there's not much trace of it. That does of course make the test a bit useless in general, but I submitted the PR anyway mainly because it does actually seem to reproduce the bug in question reliably so people can now actually look at it.

Ideas on how to make the test actually fail as expected would be most welcome.

PR #1243, opened 2016-07-21
WFS-NG mapserver encoding issue (GEOS-7607)

NielsCharlier:
https://osgeo-org.atlassian.net/browse/GEOS-7607


NielsCharlier:
this builds locally

jodygarnett:
Runs for me, checking GEOS-7607 it looks like @aaime may of located some other WFS servers we could hit. See issue for details

jodygarnett:
Merged to master, thanks Niels

PR #1250, opened 2016-07-22
Added a bit of additional logging to the image mosaic reader.

dvntucker:
Should ameliorate some of the issues with no errors when creating an image mosaic store.

### DIFF
--- a/modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/ImageMosaicReader.java
+++ b/modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/ImageMosaicReader.java
@@ -470,6 +470,8 @@ public class ImageMosaicReader extends AbstractGridCoverage2DReader implements S
                 } 
                 initReaderFromURL(source, localHints);
             } catch (Exception e) {
+                LOGGER.log(Level.FINE, "Exception generated while trying to initialize ImageMosaicReader: "
+                        + e.getMessage(), e);
                 throw new DataSourceException(e);
             }
         }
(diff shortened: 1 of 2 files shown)


modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/Utils.java
```diff
@@ -1102,8 +1102,8 @@ static URL checkSource(Object source, Hints hints) {
                     sourceURL = new URL(tempSource);
                     source = DataUtilities.urlToFile(sourceURL);
                 } catch (MalformedURLException e) {
-                    sourceURL = null;
-                    source = null;
+                    LOGGER.log(Level.FINE, "Malformed URL given to ImageMosaicReader.", e);
+                    throw new IllegalArgumentException(e);
```
jodygarnett: So this is an actual change of behaviour

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/ImageMosaicReader.java
```diff
@@ -470,6 +470,8 @@ public ImageMosaicReader(Object source, Hints uHints) throws IOException {
                 } 
                 initReaderFromURL(source, localHints);
             } catch (Exception e) {
+                LOGGER.log(Level.FINE, "Exception generated while trying to initialize ImageMosaicReader: "
+                        + e.getMessage(), e);
                 throw new DataSourceException(e);
```
jodygarnett: Just wrapping for an IOException? This is a chance to provide a message and caused by - so that the caller can get more details.

aaime:
Agree with @jodygarnett review

dvntucker:
Going to close this for now, hopefully will address it later.

PR #1253, opened 2016-07-27
[GEOT-5392] Mosaicking refactor

dvntucker:
Lots of changes here. Refactored a lot of the Mosaicker into its own package. 

- GranuleCollector becomes SubmosaicProducer. SubmosaicProducer is now responsible for mosaicking its own contents.
- ReprojectingSubmosaicProducer implementation for OTF reprojection.
- All backwards compatible with imagemosaic
- Test cases for the new mosaic implementation

Mainly putting this PR together for ease of review/commenting


aaime:
It seems there are conflicts

dvntucker:
Yep, I'll take a look this morning. Hopefully nothing too crazy.

aaime:
The first two commits create a lot of new public API (visible to anyone using Geotools, not just inside the mosaic) apparently only for the sake of moving 3 classes in their own package. I recommend moving those classes back and return the visibility to package private.

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/granulecollector/SubmosaicProducer.java
```diff
@@ -0,0 +1,43 @@
...
+
+import org.geotools.gce.imagemosaic.GranuleDescriptor;
+import org.geotools.gce.imagemosaic.MosaicElement;
+import org.opengis.referencing.crs.CoordinateReferenceSystem;
+
+/**
+ * Responsible for creating subsets of the whole imagemosaic. Since there are frequently parts of
+ * a mosaic that need to be handled differently, a submosaic producer is responsible for handling
+ * those parts separately before they're added to the whole mosaic. For example a single dimension
+ * can be handled on its own before being
```
aaime: "before being..." incomplete javadoc

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/granulecollector/SubmosaicProducer.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+public interface SubmosaicProducer {
+    boolean accept(GranuleDescriptor granuleDescriptor);
+
+    List<MosaicElement> createMosaic() throws IOException;
+
+    boolean isDoInputTransparency();
+
+    boolean isHasAlpha();
```
aaime: hasAlpha or isAlphaEnabled

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/granulecollector/SubmosaicProducer.java
```diff
@@ -0,0 +1,43 @@
...
+import org.geotools.gce.imagemosaic.MosaicElement;
+import org.opengis.referencing.crs.CoordinateReferenceSystem;
+
+/**
+ * Responsible for creating subsets of the whole imagemosaic. Since there are frequently parts of
+ * a mosaic that need to be handled differently, a submosaic producer is responsible for handling
+ * those parts separately before they're added to the whole mosaic. For example a single dimension
+ * can be handled on its own before being
+ */
+public interface SubmosaicProducer {
```
aaime: New public interface, no javadocs on its methods

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/granulecollector/SubmosaicProducer.java
```diff
@@ -0,0 +1,43 @@
...
+    boolean accept(GranuleDescriptor granuleDescriptor);
+
+    List<MosaicElement> createMosaic() throws IOException;
+
+    boolean isDoInputTransparency();
```
aaime: isInputTransparencyEnabled maybe?

dvntucker:
This is now rebased off of the formatting changes.

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/GranuleDescriptor.java
```diff
@@ -137,14 +140,25 @@
 
     OverviewsController overviewsController;
 
+    private CoordinateReferenceSystem granuleCRS;
+
+    public GeneralEnvelope coverageEnvelope;
```
aaime: The granule CRS is added and exposed via a getter, but this one is public?
The name is also confusing, what you really have is the granule/reader envelope, not the "coverage" one, indeed the descriptor does not deal with coverages at all.

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/GranuleDescriptor.java
```diff
@@ -137,14 +140,25 @@
 
     OverviewsController overviewsController;
 
+    private CoordinateReferenceSystem granuleCRS;
```
aaime: Looks like this one is a dead store, not initialized, nor used anywhere. The CRS information is also most likely available in the envelope (barring bugs, the envelope should always have a CRS, unless the reader does not have one at all)

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/GranuleDescriptor.java
```diff
@@ -97,31 +98,33 @@
...
+    /**
+     * Logger.
+     */
     private final static Logger LOGGER = org.geotools.util.logging.Logging
             .getLogger(GranuleDescriptor.class);
 
     private static final String AUXFILE_EXT = ".aux.xml";
 
-    /** Hints to use for avoiding to search for the ImageMosaic format */
+    /**
```
aaime: Looks like there still are lots of javadoc related pure reformats in this pull request. Please do not use automatic formatting, it makes bug fixes backports harder down the road :-)

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/GranuleDescriptor.java
```diff
@@ -137,14 +140,25 @@
...
+    }
+
+    public void setGranuleCRS(CoordinateReferenceSystem granuleCRS) {
+        this.granuleCRS = granuleCRS;
+    }
+
     /**
      * This class represent an overview level in a single granuleDescriptor.
-     * 
+     * <p>
```
aaime: Extra empty paragraph here

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/granulecollector/DefaultSubmosaicProducerFactory.java
```diff
@@ -0,0 +1,136 @@
...
+import org.geotools.resources.coverage.FeatureUtilities;
+import org.opengis.filter.Filter;
+
+/**
+ * Create granule collectors based on whether the request was for a stacked mosaic with additional domains requested
+ */
+public class DefaultSubmosaicProducerFactory implements SubmosaicProducerFactory {
+
+    @Override
+    public List<SubmosaicProducer> createCollectors(RasterLayerRequest request,
```
aaime: Inconsistent naming, the javadoc and method name talks about collectors, but returns submosaic producers

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/granulecollector/ReprojectingSubmosaicProducer.java
```diff
@@ -0,0 +1,178 @@
...
+        try {
+            granuleCRS = CRS.decode(granuleCRSCode);
+        } catch (FactoryException e) {
+            throw new IllegalStateException(e);
+        }
+
+        CRSBoundMosaicProducer submosaicProducer = this.perCrsMosaicProducers.get(granuleCRSCode);
+        if (submosaicProducer == null) {
+            submosaicProducer = new CRSBoundMosaicProducer(response, dryRun, granuleCRS);
+            perCrsMosaicProducers.put(granuleCRSCode, submosaicProducer);
```
aaime: As discussed during the last voice synch up, this breaks sorting, the data structure should just be a list of crs bound producers, and every time a crs changes, a new bound producer should be added to the list

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/granulecollector/ReprojectingSubmosaicProducerFactory.java
```diff
@@ -0,0 +1,37 @@
...
+import org.geotools.gce.imagemosaic.RasterLayerRequest;
+import org.geotools.gce.imagemosaic.RasterLayerResponse;
+import org.geotools.gce.imagemosaic.RasterManager;
+
+/**
+ * Creates a SubmosaicProducer that can handle reprojecting granules into the mosaic's target CRS
+ */
+public class ReprojectingSubmosaicProducerFactory implements SubmosaicProducerFactory {
+    @Override
+    public List<SubmosaicProducer> createCollectors(RasterLayerRequest request,
```
aaime: Makes it look like heterogeneous crs mosaic are not going to support stacking merge behavior on ND mosaics correctly (since the default  submosaic producer has special handling for it and will return multiple submosaic producers for it, but not this one)

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/properties/CRSExtractor.java
```diff
@@ -0,0 +1,66 @@
...
+    public static final String DEFAULT_ATTRIBUTE_NAME = "crs";
+
+    private String crsCode;
+
+    public CRSExtractor() {
+        this(new CRSExtractorSPI(), Collections.singletonList(DEFAULT_ATTRIBUTE_NAME));
+    }
+
+    public CRSExtractor(PropertiesCollectorSPI spi, List<String> propertyNames) {
+        super(spi, propertyNames);
```
aaime: The list of property names is provided because other extractors are setup with regexes and can extract multiple values (groups) in one shot, but for this one, I'd actually throw an exception if there are multiple properties, what purpose does it serve to collect the same crs value multiple times in different attributes? The same goes for DateExtractor, FSDateExtractor, they could use a "SingleValuePropertiesCollector" base class, or something like that.

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/catalog/GranuleCatalog.java
```diff
@@ -147,4 +151,12 @@ protected MultiLevelROI getGranuleFootprint(SimpleFeature sf) {
      * @throws IOException in case something bad happens
      */
     public abstract void drop() throws IOException;
+
+    public SubmosaicProducerFactory getCollectorsFactory() {
```
aaime: It is not the job of the granule catalog to know about the collectors/mosaic producers, it is just to know about the granule originators and descriptors, but not about how to build the results. The raster manager should pick it from the MosaicConfigurationBean on initialization instead, and probably make it a field to avoid re-instantiating it over and over.

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/granulecollector/SubmosaicProducer.java
```diff
@@ -0,0 +1,42 @@
...
+import org.geotools.gce.imagemosaic.GranuleDescriptor;
+import org.geotools.gce.imagemosaic.MosaicElement;
+import org.opengis.referencing.crs.CoordinateReferenceSystem;
+
+/**
+ * Responsible for creating subsets of the whole imagemosaic. Since there are frequently parts of a mosaic that need to be handled differently, a
+ * submosaic producer is responsible for handling those parts separately before they're added to the whole mosaic. For example a single dimension can
+ * be handled on its own before being
+ */
+public interface SubmosaicProducer {
```
aaime: So this new interface takes over the responsibiliites of both GranuleCollector and MosaicProducer, mixing them in a single class?

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/RasterLayerResponse.java
```diff
@@ -259,410 +252,11 @@ public ColorInterpretation getColorInterpretation() {
...
-
-    /**
      * A class doing the mosaic operation on top of a List of {@link MosaicElement}s.
-     * 
-     * @author Simone Giannecchini, GeoSolutions SAS
      *
+     * @author Simone Giannecchini, GeoSolutions SAS
      */
-    private class Mosaicker {
+    public static class Mosaicker {
```
aaime: If it's public/static it could well go in its own file now

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/RasterLayerResponse.java
```diff
@@ -875,150 +511,74 @@ private MosaicElement createMosaic() throws IOException {
 
     /**
      * This class is responsible for putting together the granules for the final mosaic.
-     * 
+     *
      * @author Simone Giannecchini, GeoSolutions SAS
-     * 
      */
     private class MosaicProducer implements GranuleCatalogVisitor {
```
aaime: Looks like this class has no use anymore? I don't see it being used/instantiated anywhere.

aaime: Never mind, found a usage in RasterLayerResponse.prepareResponse, the search output in Eclipse looked it preparedResponse was a method of MosaicProdurcer instead.

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/granulecollector/ReprojectingSubmosaicProducerFactory.java
aaime: Btw, when this is done, make sure to also add a last sortby on the crs field to avoid jumps across different CRSs in granules having the same values for the main sort fields (e.g., if sorting on time to have most recent on top, make sure features with the same date are not in mixed crs order, but packed instead)

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/granulecollector/DefaultSubmosaicProducerFactory.java
```diff
@@ -0,0 +1,136 @@
...
+            RasterManager rasterManager, RasterLayerResponse response, boolean dryRun) {
+        // get merge behavior as per request
+        List<SubmosaicProducer> defaultSubmosaicProducers = new ArrayList<>();
+        MergeBehavior mergeBehavior = request.getMergeBehavior();
+
+        // prepare dimensions management if needed, that is in case we use stacking
+        if (mergeBehavior.equals(MergeBehavior.STACK)) {
```
aaime: This code is orthogonal to the normal collection strategies and reprojection itself, it's designed to allow supporting WMS-EO, in which a mosaic has N bands split as separate granules, that need to be selected, collected and mosaicked separately, and then band-merged to return the custom band composition requested in the WMS-EO extensions.

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/granulecollector/ReprojectingSubmosaicProducer.java
dvntucker: I have a fix for the SortBy stuff mostly ready, I just need to do some more testing and create a test case.

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/GranuleDescriptor.java
dvntucker: I'll change the name to be more accurate and take a look at the visibility to make sure it's correct.

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/GranuleDescriptor.java
dvntucker: Ok, I'll take a look at getting the CRS from the Envelope instead. Probably a better place to get it.

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/GranuleDescriptor.java
dvntucker: I'll take a look at fixing some of these. Just an unfortunate side effect of the differences between IntelliJ/Eclipse formatting. I reformatted my entire branch with Eclipse; however some of the multiline comments had been expanded by IntelliJ already and for some reason the Eclipse formatter won't change the format of existing comments.

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/granulecollector/DefaultSubmosaicProducerFactory.java
dvntucker: I'll fix up the JavaDoc. Some of these were just automatically refactored from existing classes.

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/granulecollector/ReprojectingSubmosaicProducerFactory.java
dvntucker: I'm not sure we're going the resources to support a stacked mosaic in this. I'm going to take a look at what it would take, maybe I can do it fairly simply. We may have additional time/money to tackle this in September though. Hopefully it's not a deal-breaker, since stacking merge is still available in the default mosaic config.

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/properties/CRSExtractor.java
dvntucker: I can make a change for that.

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/catalog/GranuleCatalog.java
dvntucker: I'll see if I can move this somewhere where it makes sense.

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/granulecollector/SubmosaicProducer.java
dvntucker: That's basically right yes. So that the new class can collect information needed to handle granule groups that may need special handling (such as reprojecting).

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/RasterLayerResponse.java
dvntucker: I can refactor out. I think there are some other inner classes that can be moved to the outer level now.

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/granulecollector/DefaultSubmosaicProducerFactory.java
dvntucker: Maybe I should move that code into another specific subclass that's only used if the STACK behavior is used?

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/granulecollector/ReprojectingSubmosaicProducer.java
dvntucker: SortBy is respected now. I do this by grouping by every CRS + unique sort value. This could be fairly inefficient if the sort is by a large continuous range of values, but I think in many cases people will be sorting by a smaller number of discrete values (say, resolution or dates)

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/properties/CRSExtractor.java
dvntucker: I'm not sure if we're going to have time to make this change in this PR. It's slightly more involved because PropertyCollector itself makes assumptions about having a list of properties. I think ideally PropertyCollector would be an interface, or PropertyCollector itself would be even more generic and have different Multi/Single value implementations.

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/GranuleDescriptor.java
dvntucker: This has been renamed and been made private.

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/catalog/GranuleCatalog.java
dvntucker: Moved all of this to RasterManager, removed all references in the Catalog code.

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/properties/CRSExtractor.java
aaime: No problem, was not a big deal

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/granulecollector/ReprojectingSubmosaicProducer.java
```diff
@@ -158,13 +206,52 @@ public CRSBoundMosaicProducer(RasterLayerResponse rasterLayerResponse, boolean d
...
+                    + "Cannot proceed with mosaicking.", e);
+            }
+
+            //we need to see if the incoming granule matches the "template" feature this producer
+            //was initialized with.
+            if (sortBy != null) {
+                SimpleFeature granuleFeature = granuleDescriptor.getOriginator();
+                SimpleFeature templateFeature = templateDescriptor.getOriginator();
+                for (SortBy sortKey : sortBy) {
+                    Object incomingAttribute = granuleFeature
```
aaime: I'm not sure I understand why this is necessary. I though ensuring uniform CRS in the producer was enough. E.g., if I have all features in the same CRS but I'm sorting on date, do I really need to roll each feature in its own sub-mosaicker?

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/RasterLayerResponse.java
```diff
@@ -1174,10 +1176,11 @@ private void handleSortByClause(final Query query) {
...
 
                 // assign to query if sorting is supported!
-                final SortBy[] sb = clauses.toArray(new SortBy[] {});
+
+                this.sortBy = clauses.toArray(new SortBy[] {});
                 if (rasterManager.getGranuleCatalog()
-                        .getQueryCapabilities(rasterManager.getTypeName()).supportsSorting(sb)) {
-                    query.setSortBy(sb);
+                        .getQueryCapabilities(rasterManager.getTypeName()).supportsSorting(sortBy)) {
+                    query.setSortBy(sortBy);
```
aaime: In case of heterogeneous CRS I'd add the crs property as a last sort-by property, so that we don't end up creating sub-mosaics for nothing (mosaicking is an expensive operation, we should try to do as few as possible within the respect of the user request)

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/granulecollector/DefaultSubmosaicProducerFactory.java
aaime: Given that we plan to refactor the mosaic in the future to simplify its structure further, I'd say to let this one go for the moment. We are also considering moving the stacking approach up in the food chain and have it implemented purely in GeoServer later (as it's only used by WMS-EO at the moment).

aaime:
Just added a few more minor comments, everything looks otherwise pretty good. Looks like there are merge conflicts though.

dvntucker:
I'll squash and rebase. I also want to rename one of the packages (granule collector) before merging.

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/RasterLayerResponse.java
dvntucker: I think we're going to leave this for the user to control with the SortBy parameter for now. In the future I'd like to update maybe SubmosaicProducerFactory to have an optional SortBy parameter so that the RasterLayerRequest can query.

dvntucker:
Simplified the Sorting along the lines we discussed last week.

bmmpxf:
@aaime Is this ready to be merged now?

PR #1254, opened 2016-07-28
Add URL encode filter function

newmanw:
- Added a filter function to URL encode a string

JIRA Issue: https://osgeo-org.atlassian.net/browse/GEOT-5472

### DIFF
--- a/modules/library/main/src/main/java/org/geotools/filter/function/string/URLEncodeFunction.java
+++ b/modules/library/main/src/main/java/org/geotools/filter/function/string/URLEncodeFunction.java
@@ -1,0 +1,100 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ *
+ *    (C) 2016, Open Source Geospatial Foundation (OSGeo)
+ *
+ *    This library is free software; you can redistribute it and/or
+ *    modify it under the terms of the GNU Lesser General Public
+ *    License as published by the Free Software Foundation;
+ *    version 2.1 of the License.
+ *
+ *    This library is distributed in the hope that it will be useful,
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ *    Lesser General Public License for more details.
+ */
+
+package org.geotools.filter.function.string;
+
+import static org.geotools.filter.capability.FunctionNameImpl.parameter;
+
+import java.io.UnsupportedEncodingException;
+import java.net.URLEncoder;
+
+import org.geotools.filter.FunctionExpressionImpl;
+import org.geotools.filter.capability.FunctionNameImpl;
+import org.opengis.filter.capability.FunctionName;
+
+/**
+ * URL encodes a string.
+ *
+ * This function expects:
+ * <ol>
+ * <li>Literal: String to be URL encoded
+ * <li>Literal: (Optional) Boolean indicating if string should be form URL encoded (defaults to false)
+ *
+ * @see URLEncoder#encode(String)
+ *
+ * @author Billy Newman (BIT Systems)
+ */
+public class URLEncodeFunction extends FunctionExpressionImpl {
+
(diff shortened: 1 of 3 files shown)


modules/library/main/src/main/java/org/geotools/filter/function/FilterFunction_strURLEncode.java
```diff
@@ -0,0 +1,62 @@
...
+    /**
+     * URL encode the string.
+     *
+     * @param feature
+     * @return The URL encoded string
+     * @throws IllegalArgumentException if the argument is not a string
+     */
+    @Override
+    public Object evaluate(Object feature) {
+        String arg0;
```
dvntucker: Can this variable be renamed? Even if it's just something as simple as "stringToBeEncoded"? Arg0 isn't very informative :)

modules/library/main/src/main/java/org/geotools/filter/function/FilterFunction_strURLEncode.java
```diff
@@ -0,0 +1,62 @@
+package org.geotools.filter.function;
```
jodygarnett: header required for new file.

The "FilterFunction_strURLEncode was just a naming convention used by a code generator, please consider renaming the class to follow java contentions - ``URLEncodeFunction``.

modules/library/main/src/test/java/org/geotools/filter/function/string/FilterFunction_strURLEncodeTest.java
```diff
@@ -0,0 +1,56 @@
+package org.geotools.filter.function.string;
```
jodygarnett: header needed for new files

modules/library/main/src/main/java/org/geotools/filter/function/FilterFunction_strURLEncode.java
```diff
@@ -0,0 +1,62 @@
...
+            arg0 = getExpression(0).evaluate(feature, String.class);
+        } catch (Exception e) // probably a type error
+        {
+            throw new IllegalArgumentException(
+                    "Filter Function problem for function strURLEncode argument #0 - expected type String");
+        }
+
+        try {
+            // Using URLEncoder, spaces are converted to plus signs, but must be %20 for URL calls
+            return URLEncoder.encode(arg0, "utf-8").replaceAll("\\+", "%20");
```
dvntucker: I'm a little uncomfortable with this, because the next user could just as easily think + is the correct behavior instead of %20 (since technically both are correct depending on which part of the URL is being encoded). Would it be possible to add a boolean flag on the filter function that toggles whether this operation is performed or not? I'm not totally familiar with the filter functions but it seems like this should be possible.

modules/library/main/src/main/java/org/geotools/filter/function/FilterFunction_strURLEncode.java
newmanw: Agreed, and done.  Thanks!

modules/library/main/src/main/java/org/geotools/filter/function/FilterFunction_strURLEncode.java
newmanw: Added header.

Renamed class, and moved to org.geotools.filter.function.string package from org.geotools.filter.function

modules/library/main/src/test/java/org/geotools/filter/function/string/FilterFunction_strURLEncodeTest.java
newmanw: Added header

modules/library/main/src/main/java/org/geotools/filter/function/FilterFunction_strURLEncode.java
newmanw: Yeah no problem, I was wondering about this and glad you brought it up.  I added an optional    2nd parameter 'formUrlEncode' that defaults to false if not provided.

Thoughts?

PR #1262, opened 2016-08-01
[GEOT-5478] Fix ExtractBoundsFilterVisitor AND'd DURING filters

jahhulbert-ccri:
Signed-off-by: Andrew Hulbert <andrew.hulbert@ccri.com>

### DIFF
--- a/modules/library/main/src/main/java/org/geotools/filter/visitor/ExtractBoundsFilterVisitor.java
+++ b/modules/library/main/src/main/java/org/geotools/filter/visitor/ExtractBoundsFilterVisitor.java
@@ -33,6 +33,7 @@ import org.opengis.filter.PropertyIsGreaterThanOrEqualTo;
 import org.opengis.filter.PropertyIsLessThan;
 import org.opengis.filter.PropertyIsLessThanOrEqualTo;
 import org.opengis.filter.PropertyIsLike;
+import org.opengis.filter.PropertyIsNil;
 import org.opengis.filter.PropertyIsNotEqualTo;
 import org.opengis.filter.PropertyIsNull;
 import org.opengis.filter.expression.Add;
@@ -54,6 +55,20 @@ import org.opengis.filter.spatial.Intersects;
 import org.opengis.filter.spatial.Overlaps;
 import org.opengis.filter.spatial.Touches;
 import org.opengis.filter.spatial.Within;
+import org.opengis.filter.temporal.After;
+import org.opengis.filter.temporal.AnyInteracts;
+import org.opengis.filter.temporal.Before;
+import org.opengis.filter.temporal.Begins;
+import org.opengis.filter.temporal.BegunBy;
+import org.opengis.filter.temporal.During;
+import org.opengis.filter.temporal.EndedBy;
+import org.opengis.filter.temporal.Ends;
+import org.opengis.filter.temporal.Meets;
+import org.opengis.filter.temporal.MetBy;
+import org.opengis.filter.temporal.OverlappedBy;
+import org.opengis.filter.temporal.TContains;
+import org.opengis.filter.temporal.TEquals;
+import org.opengis.filter.temporal.TOverlaps;
 import org.opengis.referencing.crs.CoordinateReferenceSystem;
 
 import com.vividsolutions.jts.geom.Envelope;
@@ -157,7 +172,7 @@ public class ExtractBoundsFilterVisitor extends NullFilterVisitor {
     }
     /**
      * Please note we are only visiting literals involved in spatial operations.
-     * @param literal, hopefully a Geometry or Envelope
+     * @param expression, a literal expression, hopefully a Geometry or Envelope
      * @param data Incoming BoundingBox (or Envelope or CRS)
      * 
      * @return ReferencedEnvelope updated to reflect literal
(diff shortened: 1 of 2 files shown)


aaime:
Wondering, was that the only temporal filter missing?

jahhulbert-ccri:
i'll double check some of the others

jahhulbert-ccri:
@aaime looks like the others here https://github.com/geotools/geotools/tree/master/modules/library/main/src/main/java/org/geotools/filter/temporal aren't really implemented either...I can try to look into them for this PR or do a follow-on bug fix. lemme know. (I usually just use DURING and BETWEEN) but I can checkout out the other CQL statements.

jahhulbert-ccri:
Also, I was wondering if maybe instead of defaulting to null it could have a base class that defaulted to infinity() for the extent rather than null but that was a larger change

jnh5y:
Are there any remaining functions which are not overridden?  E.g., will we lose with an spatial + attribute filter?

jahhulbert-ccri:
only remaining one according to my IDE is 
```
    @Override
    public Object visit(PropertyIsNil filter, Object data) {
        return super.visit(filter, data);
    }
```

jnh5y:
Yeah, I don't know if isNil is important or not.  To test isNil and isNull, I added tests like this:

+    @Test
+    public void testAndIsNull() throws Exception {
+        final Period p = period("2016-01-01T00:00:00.000-0500", "2106-01-02T00:00:00.000-0500");
+        final Filter f = ff.and(ff.bbox("geom", -10, -10, 10, 10, null),
+                ff.isNull(ff.literal("someDate")));
+        final Envelope env = (Envelope) f.accept(visitor, null);
+        assertEquals(new Envelope(-10, 10, -10, 10), env);
+    }    
+
+    @Test
+    public void testAndIsNil() throws Exception {
+        final Period p = period("2016-01-01T00:00:00.000-0500", "2106-01-02T00:00:00.000-0500");
+        final Filter f = ff.and(ff.bbox("geom", -10, -10, 10, 10, null),
+                ff.isNil(ff.literal("someDate"), null));
+        final Envelope env = (Envelope) f.accept(visitor, null);
+        assertEquals(new Envelope(-10, 10, -10, 10), env);
+    }


The second with isNil fails...

jahhulbert-ccri:
@jnh5y added those fixes in for Nil and null!

modules/library/main/src/main/java/org/geotools/filter/visitor/ExtractBoundsFilterVisitor.java
```diff
@@ -19,22 +19,7 @@
...
-import org.opengis.filter.PropertyIsBetween;
-import org.opengis.filter.PropertyIsEqualTo;
-import org.opengis.filter.PropertyIsGreaterThan;
-import org.opengis.filter.PropertyIsGreaterThanOrEqualTo;
-import org.opengis.filter.PropertyIsLessThan;
-import org.opengis.filter.PropertyIsLessThanOrEqualTo;
-import org.opengis.filter.PropertyIsLike;
-import org.opengis.filter.PropertyIsNotEqualTo;
-import org.opengis.filter.PropertyIsNull;
+import org.opengis.filter.*;
```
jodygarnett: I tend to leave this alone, since developers will hit Control-Shift-O on occasion to organize imports.

jahhulbert-ccri: yeah I believe that intellij did that for me automatically...sorry about that.

modules/library/main/src/main/java/org/geotools/filter/visitor/ExtractBoundsFilterVisitor.java
```diff
@@ -397,7 +383,79 @@ public Object visit(Subtract expression, Object data) {
...
+    }
+
+    @Override
+    public Object visit(TOverlaps contains, Object data) {
+        return infinity();
+    }
+
+    @Override
+    public Object visit(PropertyIsNil filter, Object data) {
+        return infinity();
```
jodygarnett: So if I understand correctly you are overriding all the other "new filters" that were added for Filter 2.0 to behave in a similar fashion.

jodygarnett:
Thanks for the additional fixes, waiting on travis to double check.

I see you have seperate backports in place, is it worth squashing and merging (and backporting a single commit) or are you going to copy the commits around?

modules/library/main/src/main/java/org/geotools/filter/visitor/ExtractBoundsFilterVisitor.java
jahhulbert-ccri: yeah...i originally just did the DURING query so that it would behave like a BETWEEN query for temporal predicates but then did it the same for the others like toverlaps, tequals, etc.

jahhulbert-ccri:
Jody, I can carry the commit around. I'm just keeping them up to date with a cherry-pick each time off the current branch since the file wasn't really different between the branches that works OK it seems.

jahhulbert-ccri:
@jodygarnett fixed up the imports...lemme know if that's the style that the eclipse editor will be happy with...i'll squash once you think its all ready and then push the backport branches.

jahhulbert-ccri:
squashed it

PR #1273, opened 2016-08-09
GEOT-5484 - FootprintBehavior not used in single granule path

dvntucker:
FootprintBehavior is not used when the single granule optimized path is taken. This causes areas outside the bounds of granule not to be properly cut from the resulting image.


modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/RasterLayerResponse.java
```diff
@@ -804,6 +804,9 @@ private MosaicElement createMosaic() throws IOException {
                                     // Adding globalRoi to the output
                                     RenderedOp rop = (RenderedOp) mosaic;
                                     rop.setProperty("ROI", in.roi);
+
+                                    mosaic = footprintBehavior.postProcessMosaic(mosaic, in.roi,
```
aaime: I'm not sure, but what if image bounds contains raster bounds instead? Will we miss post processing?

aaime:
Lacks a test

dvntucker:
Yep, sorry. Was so excited about fixing the issue I forgot it. Should be able to test pretty easily.

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/RasterLayerResponse.java
dvntucker: Hmm, maybe, yeah. I guess in a case where your raster fills the entire image bounds and your footprint has, say, a hole right in the middle then it should need to do the footprint handling as well right? I'll try a test case and see what happens.

dvntucker: Ok, I just checked and I don't think this is an issue. In this case presumably it should only matter if we have a footprint that has holes in it right? In that case it never enters this code path because the bounds aren't rectangular (see line 759 where it converts the ROI to a rectangle). Maybe there is a case I'm not thinking of though.

jodygarnett:
This now has a test, are we good to go @dvntucker ?

PR #1277, opened 2016-08-11
[GEOT-5485]: Supporting ImageMosaic's inner overviews on ImagePyramid

dromagnoli:


modules/plugin/imagepyramid/src/main/java/org/geotools/gce/imagepyramid/ImageLevelsMapper.java
```diff
@@ -0,0 +1,222 @@
...
+import java.util.logging.Logger;
+
+import org.geotools.data.DataUtilities;
+import org.geotools.factory.Hints;
+import org.geotools.gce.imagemosaic.ImageMosaicReader;
+
+/**
+ * Parse imagePyramid property files and setup the mapping to provide the proper 
+ * {@link ImageMosaicReader} for the required image choice. 
+ * Starting with 16.x we are supporting ImagePyramids of ImageMosaic with internal
```
aaime: I thought this was supposed to be backported to 15.x too? No need to cite the version in the javadoc anyways

modules/plugin/imagepyramid/src/main/java/org/geotools/gce/imagepyramid/ImageLevelsMapper.java
```diff
@@ -0,0 +1,222 @@
...
+    }
+
+    /**
+     * Cache of {@link ImageMosaicReader} objects for the different levels.
+     */
+    ConcurrentHashMap<Integer, ImageMosaicReader> readers = new ConcurrentHashMap<Integer, ImageMosaicReader>();
+
+    public void dispose() {
+        for (Entry<Integer, ImageMosaicReader> element : readers.entrySet()) {
+            element.getValue().dispose();
```
aaime: I'd wrap this one is try/catch to make sure we dispose as many readers as possible even in case of failure

modules/plugin/imagepyramid/src/main/java/org/geotools/gce/imagepyramid/ImagePyramidFormat.java
```diff
@@ -206,20 +206,15 @@ public boolean accepts(Object source, Hints hints) {
             }
```
aaime: Shouldn't this one advertise support for the overview policy now? Generally speaking, I'd change the code so that it advertises support for all the mosaic params (calling onto the mosaic format to grab them first)

modules/plugin/imagepyramid/src/main/java/org/geotools/gce/imagepyramid/ImageLevelsMapper.java
dromagnoli: Right. I have simply moved the original code in the specific class.
I will add the try/catch wrapping (note that ImageMosaicReader dispose already simply logs the exceptions without rethrowing them)

modules/plugin/imagepyramid/src/main/java/org/geotools/gce/imagepyramid/ImageLevelsMapper.java
dromagnoli: Going to fix it

modules/plugin/imagepyramid/src/main/java/org/geotools/gce/imagepyramid/ImagePyramidFormat.java
dromagnoli: I think that other params management (beside overviewPolicy) is out of scope right now (there is no code handling/parsing them). We could consider opening a specific JIRA to add this improvement too.

aaime: I though we were just passing everything down as-is? (besides the grid geometry and overview policy, which are used locally to figure out which mosaic reader to use). Can you make an example of a parameter that we need to "manage" or "parse"?

dromagnoli: You are right, that piece of code doing custom parsing trick me. The params are directly passed down as-is. I have updated the PR.

PR #1280, opened 2016-08-15
Fix list formatting in junit.rst

leif81:
"Lists must always start a new paragraph -- that is, they must appear after a blank line."

http://docutils.sourceforge.net/docs/user/rst/quickstart.html#lists


leif81:
I should have mentioned that prior to this fix the list items are all squished onto a single line as seen here:

http://docs.geotools.org/latest/developer/conventions/test/junit.html

bradh:
Looks like a valid fix.

PR #1281, opened 2016-08-16
Fix link to FeatureCollection in swing faq

leif81:
Crossing my fingers this syntax for linking to other docs works (?). I borrowed it from another geotools rst example.

The link was sending the user here:
http://docs.geotools.org/library/main/collection

Instead of here:
http://docs.geotools.org/latest/userguide/library/main/collection.html


tbarsballe:
Looks good, thanks for the fix.

PR #1282, opened 2016-08-16
Wfs-ng: encode spatial filters in wfs 2.0 (GEOT-5132)

NielsCharlier:
https://osgeo-org.atlassian.net/browse/GEOT-5132

### DIFF
--- a/modules/extension/xsd/xsd-fes/src/main/java/org/geotools/filter/v2_0/bindings/BinarySpatialOpTypeBinding.java
+++ b/modules/extension/xsd/xsd-fes/src/main/java/org/geotools/filter/v2_0/bindings/BinarySpatialOpTypeBinding.java
@@ -2,7 +2,7 @@
  *    GeoTools - The Open Source Java GIS Toolkit
  *    http://geotools.org
  *
- *    (C) 2002-2011, Open Source Geospatial Foundation (OSGeo)
+ *    (C) 2002-2016, Open Source Geospatial Foundation (OSGeo)
  *
  *    This library is free software; you can redistribute it and/or
  *    modify it under the terms of the GNU Lesser General Public
@@ -20,7 +20,14 @@ import javax.xml.namespace.QName;
 
 import org.geotools.filter.v2_0.FES;
 import org.geotools.xml.AbstractComplexBinding;
+import org.geotools.xml.Encoder;
+import org.geotools.xml.EncoderDelegate;
+import org.opengis.filter.expression.Expression;
+import org.opengis.filter.expression.Literal;
 import org.opengis.filter.spatial.BinarySpatialOperator;
+import org.xml.sax.ContentHandler;
+
+import com.vividsolutions.jts.geom.Geometry;
 
 /**
  * <pre>
@@ -55,6 +62,18 @@ public class BinarySpatialOpTypeBinding extends AbstractComplexBinding {
     
     @Override
     public Object getProperty(Object object, QName name) throws Exception {
-        return FESParseEncodeUtil.getProperty((BinarySpatialOperator) object, name);
+        Expression e = FESParseEncodeUtil.getProperty((BinarySpatialOperator) object, name);
+        if (e instanceof Literal && ((Literal) e).getValue() instanceof Geometry) {
+            return new EncoderDelegate() {
+                @Override
+                public void encode(ContentHandler output) throws Exception {                    
+                    Encoder encoder = new Encoder(new org.geotools.gml3.v3_2.GMLConfiguration());
+                    encoder.setInline(true);
+                    encoder.encode(((Literal) e).getValue(), org.geotools.gml3.v3_2.GML.AbstractGeometry, output);
+                }
+            };
+        } else {
+            return e;
+        }
     }
 }
--- a/modules/extension/xsd/xsd-fes/src/test/java/org/geotools/filter/v2_0/bindings/FilterTypeBindingTest.java
+++ b/modules/extension/xsd/xsd-fes/src/test/java/org/geotools/filter/v2_0/bindings/FilterTypeBindingTest.java
@@ -205,7 +205,6 @@ public class FilterTypeBindingTest extends FESTestSupport {
         
         Element e = getElementByQName(doc, FES.Intersects);
         assertNotNull(e);
-        print(e);
 
         assertNotNull(getElementByQName(e, FES.ValueReference));
         assertNotNull(getElementByQName(e, GML.Point));
--- a/modules/extension/xsd/xsd-fes/src/test/java/org/geotools/filter/v2_0/bindings/FilterTypeBindingTest.java
+++ b/modules/extension/xsd/xsd-fes/src/test/java/org/geotools/filter/v2_0/bindings/FilterTypeBindingTest.java
@@ -2,7 +2,7 @@
  *    GeoTools - The Open Source Java GIS Toolkit
  *    http://geotools.org
  *
- *    (C) 2002-2015, Open Source Geospatial Foundation (OSGeo)
+ *    (C) 2002-2016, Open Source Geospatial Foundation (OSGeo)
  *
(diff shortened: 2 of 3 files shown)


modules/extension/xsd/xsd-fes/src/main/java/org/geotools/filter/v2_0/bindings/BinarySpatialOpTypeBinding.java
```diff
@@ -55,6 +62,18 @@ public Class getType() {
     
     @Override
     public Object getProperty(Object object, QName name) throws Exception {
-        return FESParseEncodeUtil.getProperty((BinarySpatialOperator) object, name);
+        Expression e = FESParseEncodeUtil.getProperty((BinarySpatialOperator) object, name);
+        if (e instanceof Literal && ((Literal) e).getValue() instanceof Geometry) {
+            return new EncoderDelegate() {
+                @Override
+                public void encode(ContentHandler output) throws Exception {                    
+                    Encoder encoder = new Encoder(new org.geotools.gml3.v3_2.GMLConfiguration());
```
bencaradocdavies: FES 2.0 has no explicit dependency on GML 3.2.1, or any GML version for that matter. Why should GML 3.2.1 be used here? FES 2.0 can be used to encode filters with any GML version.

NielsCharlier: okay, what do you suggest? how should a gml version be chosen then?

NielsCharlier: We're going to have to pick one. So perhaps the question back at you is, why should GML 3.2.1 not be used here?

bencaradocdavies: @NielsCharlier I think it is perfectly fine for wfs-ng to decide to make all its WFS 2.0 / FES 2.0 requests using GML 3.2.1, which is guaranteed to be supported. The problem is that this module may be used for other purposes. The API is lacking. There should probably be some sort of factory or configuration being injected into this binding. Looks at ```GML3EncodingUntils``` for example. I agree that GML 3.2.1 is likely a good choice, but it just seems wrong to hard-code this choice here.

NielsCharlier: But surely when it comes to encoding, there are sometimes multiple valid ways to encode the same data, and then the encoder makes a choice for us.

bencaradocdavies: Yes, I agree that the encoder will have to pick one, but should this choice be hardcoded here? ``FESConfiguration`` declares a dependency on GML 3.2.1 ```GMLConfiguration``. Do you have access to an ```FESConfiguration``` instance in this scope? Could this functionality be centralised in ```FESParseEncodeUtil```?

NielsCharlier: I can successfully change this line:
Encoder encoder = new Encoder(new org.geotools.gml3.v3_2.GMLConfiguration());
to this line:
Encoder encoder = new Encoder(new FESConfiguration());

Because,as you said, FESConfiguration depends on gml3.v3_2.GMLConfiguration.

But, that still leaves us with this line:
encoder.encode(((Literal) e).getValue(), org.geotools.gml3.v3_2.GML.AbstractGeometry, output);

And I must specify AbstractGeometry, because that is the whole reason of the EncoderDelegate, to explicitly specify that the object needs to be encoded as a GML geometry.

But considering that, as you pointed out, FESConfiguration is already depending on GML3.2 Configuration in geotools. That means that we don't actually have a choice here, in our geotools implementation FES does depend on GML 3.2

NielsCharlier: In other words, the explicit mention of "org.geotools.gml3.v3_2.GML.AbstractGeometry" cannot be avoided, because it is the ntire reason of this block of code. 
This is exactly necessary, because the Encoder has no way to just "know" from the scheme that a geometry must be encoded here (unlike what was case with the old filter scheme) and that why it was previously failing to do so.

So in summary, the lack of dependency of a particular GML version that makes you criticise this block of code, is the exact reason that this block of code is necessary in the first place... 

So, rather you must question the whole api and implementation of the geotools Encoder as it is... which is a whole other can of worms.

bencaradocdavies: @NielsCharlier I agree that not much more can be done without a major API refactor to allow specification of GML version. My remarks were not a criticism, just an observation that the GML version was being hardcoded. Under the circumstances, I do not see a better solution, and am comfortable with your original code.

modules/extension/xsd/xsd-fes/src/test/java/org/geotools/filter/v2_0/bindings/FilterTypeBindingTest.java
```diff
@@ -2,7 +2,7 @@
  *    GeoTools - The Open Source Java GIS Toolkit
  *    http://geotools.org
  *
- *    (C) 2002-2015, Open Source Geospatial Foundation (OSGeo)
+ *    (C) 2002-2016, Open Source Geospatial Foundation (OSGeo)
```
bencaradocdavies: And our new policy is that we no longer require copyright date updates. Just leave existing OSGeo headers as they are.  :-)

modules/extension/xsd/xsd-fes/src/test/java/org/geotools/filter/v2_0/bindings/FilterTypeBindingTest.java
```diff
@@ -205,9 +205,10 @@ public void testEncodeSpatial() throws Exception {
         
         Element e = getElementByQName(doc, FES.Intersects);
         assertNotNull(e);
+        print(e);
```
bencaradocdavies: Please do not print to stdout/stderr. Please log instead so output can be silenced by the build infrastructure.

modules/unsupported/wfs-ng/src/test/java/org/geotools/data/wfs/online/v2_0/GeoServerOnlineTest.java
```diff
@@ -74,8 +74,8 @@ public GeoServerOnlineTest() {
     
     public static Filter createSpatialFilter() {
         GeometryFactory gf = new GeometryFactory(); 
-        Coordinate[] coordinates = { new Coordinate(39, -107), new Coordinate(38, -107),
-                new Coordinate(38, -104), new Coordinate(39, -104), new Coordinate(39, -107) };
+        Coordinate[] coordinates = { new Coordinate(-107, 39), new Coordinate(-107, 38),
+                new Coordinate(-104, 38), new Coordinate(-104, 39), new Coordinate(-107, 39) };
```
bencaradocdavies: Good. This agrees with the axis order setting above.

modules/extension/xsd/xsd-fes/src/test/java/org/geotools/filter/v2_0/bindings/FilterTypeBindingTest.java
NielsCharlier: oops

bencaradocdavies:
Thanks @NielsCharlier.

PR #1283, opened 2016-08-18
Upgrade to Mongo module to fix GEOT-5492

tingold:
Fixes GEOT-5492 by:
-upgrade version of mongo-java-driver to 3.3.0 in pom.xml
-replacing usages of “ensureIndex” (deprecated) with “createIndex” in various classes
-UnknownHostException not explicitly thrown

Unit tests remain valid and functional tests pass in GeoServer

### DIFF
--- a/modules/unsupported/mongodb/pom.xml
+++ b/modules/unsupported/mongodb/pom.xml
@@ -1,28 +1,28 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- =======================================================================    
-        Maven Project Configuration File                                        
-                                                                                
-        The Geotools Project                                                    
-            http://www.geotools.org/                                            
-                                                                                
+<!-- =======================================================================
+        Maven Project Configuration File
+
+        The Geotools Project
+            http://www.geotools.org/
+
         Version: $Id: pom.xml 2014-03-21 13:32:49Z tkunicki $
      ======================================================================= -->
 <project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0"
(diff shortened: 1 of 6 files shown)


tbarsballe:
Note: MongoDB 3.3 driver is not explicitly backards compatible with MongoDB < 3.0.
Given that the current version of MongoDB is not compatible with the current version of GeoServer without this change, I feel like this change is fine. 

However, this is probably not something we want to backport to older versions of GeoTools.

aaime:
I just checked what's in Ubuntu 16.04 LTS, the official package is version 2.6.10, meaning this version will still be popular for the next two years. Is there any chance to make the module work for both versions, 2.x and 3.x?

tbarsballe:
I took a little bit of a deeper look into this, and it turns out my information is wrong. The 3.3.0 driver is compatible with MongoDB 2.4 and newer.
The old 2.11 driver is only advertised as compatible with MongoDB 2.4 (and older)

Details here: https://docs.mongodb.com/ecosystem/drivers/driver-compatibility-reference/#reference-compatibility-mongodb-java

So this should be backwards compatible. More importantly, the current driver is not compatible with MongoDB 2.6.

aaime:
Oh ok, then the upgrade seems like good news :-)

tbarsballe:
Just to be sure, I tested the new driver against both MongoDB 2.4.9 and 3.2.9 on Ubuntu 14.04, and it worked fine. So we appear to be safely backwards compatible.

dvntucker:
Should we back port this? I'm not sure but it should be compatible with current stable.

tbarsballe:
I would hesitate to backport it, if only because driver changes are the sort of thing that seems likely to break existing configurations.

However, Given that the old (2.11) driver is only advertised as compatible with mongodb 2.4 and older, it may be safe to backport this to 2.9.x.

PR #1285, opened 2016-08-21
[GEOT-5489] - NPE when using an interpolation in reprojecting mosaic

dvntucker:
Fixed a reference to the local variable that should have been a local variable. Also made the method static since it's really meant to be used statically anyway.


tbarsballe:
Looks good, but can you include JIRA commit number in the commit message

dvntucker:
Thanks @tbarsballe, merging.

PR #1286, opened 2016-08-27
Fix visibility of JFontChooser.showDialog [GEOT-5501]

jodygarnett:
For details see https://osgeo-org.atlassian.net/browse/GEOT-5501


jodygarnett:
Thanks, I will back port to stable.

PR #1287, opened 2016-08-27
[GEOT-5502] First implementation of Pole of inaccessibility.

chau-intl:
Vladimir Agafonkin came up with an algorithm for calculating poles of inaccessibility to give good labelling points for odd shaped polygons. 

See https://www.mapbox.com/blog/polygon-center/ for a discussion of algorithm

This implements this functionality and makes it available as a function so it could be added to GeoServer SLD files.


modules/unsupported/polylabel/src/main/java/org/geotools/polylabel/PolyLabeller.java
```diff
@@ -0,0 +1,132 @@
...
+/**
+ * Based on Vladimir Agafonkin's Algorithm
+ * https://www.mapbox.com/blog/polygon-center/
+ * 
+ * @author Ian Turton
+ * @author Casper Børgesen
+ *
+ */
+public class PolyLabeller {
+	private static final Logger LOGGER = Logging.getLogger(PolyLabeller.class.getName());
```
smithkm: Using tabs, should be spaces.

modules/unsupported/polylabel/src/main/java/org/geotools/polylabel/PolygonLabelProcess.java
```diff
@@ -0,0 +1,48 @@
...
+ * @author Casper Børgesen
+ *
+ */
+
+public class PolygonLabelProcess extends StaticMethodsProcessFactory<PolygonLabelProcess>{
+	public PolygonLabelProcess() {
+		super(Text.text("PolygonLabelProcess"), "polygonlabelprocess", PolygonLabelProcess.class);
+	}
+	
+	@DescribeProcess(title = "Polygon label process", description = "Calculate the the Pole of accessibility, the most distant interior point i a polygon.")
```
smithkm: Typo, 'i a polygon'  Should be 'in'?

modules/unsupported/polylabel/src/main/java/org/geotools/polylabel/Cell.java
```diff
@@ -0,0 +1,109 @@
...
+
+  public Point getPoint() {
+	  return gb.point(x, y);
+  }
+  
+  // signed distance from point to polygon outline (negative if point is
+  // outside)
+  private double pointToPolygonDist(Point point, MultiPolygon polygon) {
+    boolean inside = polygon.contains(point);
+/*    double minDistSq = Double.POSITIVE_INFINITY;
```
smithkm: Best to remove commented out code before submitting.  If it's really important that it be retained, use a comment to explain why.

modules/unsupported/polylabel/src/main/java/org/geotools/polylabel/Cell.java
```diff
@@ -0,0 +1,109 @@
...
+public class Cell implements Comparable<Cell> {
+  static final private GeometryBuilder gb = new GeometryBuilder();
+  private static final double SQRT2 = 1.4142135623730951;
+  private double x;
+  private double y;
+  private double h;
+  private double d;
+  private double max;
+
+  Cell(double x, double y, double h, MultiPolygon polygon) {
```
smithkm: Comment explaining the class would be nice.  Even if it's only meant for implementation, it will help with maintaining.

ianturton:
Casper will sign CLA on Monday when he has checked his contract

aaime:
Thinking out loud and without having checked the pull request.... is this worth turning into a text symbolizer vendor option too? If it's fast enough it could become the default poly center implementation, with the vendor option still allowing it to turn it off and get the old behavior, but most of the people would get the benefit without needing to alter all their SLDs.

ianturton:
Once we've tried this out on a some realistic polygons we could consider moving it into the main styling package and making it the default.

smithkm:
OK nitpicks addressed.  Looks good.

dr-jts:
A bit late to the party here.... is this algorithm in use at all?

There's a few places where it can be sped up dramatically:

- Use `IndexedPointInArea `rather than `contains`
- Use `IndexedFacetDistance `rather than plain `Distance `
- With these in use there is no need to check for polygon validity

I have implemented a version like this and the performance goes from 153 s to 680 ms on a world polygon dataset. Will be posting it soon.

aaime:
We tried it out hoping it would lead to better label points, the results have not been very satisfactory to our cartographers (and we verified implementation making QGIS compute the same using its pole of inaccesibility implementation, which returned the same points). Looks like MapBox is using something else now, as their maps have the labels in better positions than the pole. Still, good to have it improved, in some cases it's better than the built-in one we have.

dr-jts:
Thanks for pointing out the updated version.  That should be a lot better (not requiring validity, and should be faster).

Interesting that this isn't considered ideal labelling.   I did notice that sometimes the point is far from the centre of the polygon, when there's still enough room there for a label.  Perhaps a combination of trying the centre and the furthest point would be better.

dr-jts:
Testing indicates the latest version of PolyLabeller is indeed a lot faster.  A version using indexed Point-in-Polygon and Distance is still faster, but only about 3x on a "typical" dataset.  So probably not worth the (much) greater complexity (in terms of dependencies).

dr-jts:
@aaime etc:  Following up on this...   It looks like Mapnik is now using a modified PolyLabeller algorithm, which provides some weighting from the polygon centroid to stop the compute point being too far away.  See this:
https://github.com/mapnik/mapnik/pull/3811

giohappy:
Nice the centroid fix approach. It's something we also were thinking about to improve the PolyLabel efficacy for labeling. Indeed, from our experience, the pole of inaccessibility often doesn't give great visual results (and most of map services don't employ it straight).
We could consider porting it into GT one day...

dr-jts:
I found a bug in the master [version of this code](https://github.com/geotools/geotools/blob/d2a5c6f5475a254b81428005b2094c5e88b20702/modules/unsupported/process-geometry/src/main/java/org/geotools/process/geometry/Cell.java#L80).  The code uses getCoordinates() on polygon geometries.  This is incorrect if the polygon contains holes, since it merges all coordinates from all rings into a single let.  (It's also slow for polygons with many holes).

The fix is to add a 3rd level to the loop to handle polygon shell and hole rings separately.  

Even better, use [IndexdFacetDistance](https://github.com/locationtech/jts/blob/master/modules/core/src/main/java/org/locationtech/jts/operation/distance/IndexedFacetDistance.java) to make it much faster.

aaime:
Thanks for the report @dr-jts 
Right now nobody is "owning" that code, it's in an unsupported module. Maybe @ianturton wants to have a look? Otherwise PRs welcomed :-) (if nobody takes it, maybe best to open a ticket to avoid losing the intel).

dr-jts:
Sure, if no-one else picks up on this I will open a new issue.

For the record, I have an implementation using indexing [here](https://github.com/dr-jts/jts-ports/tree/master/src/main/java/org/geotools/polylabelfast).

PR #1293, opened 2016-09-01
[GEOT-5503]: Avoid error through encoding of null geoms, fixed NPE

awaterme:
Proposed bugfix...

### DIFF
--- a/modules/library/xml/src/main/java/org/geotools/xml/gml/GMLComplexTypes.java
+++ b/modules/library/xml/src/main/java/org/geotools/xml/gml/GMLComplexTypes.java
@@ -2,7 +2,7 @@
  *    GeoTools - The Open Source Java GIS Toolkit
  *    http://geotools.org
  * 
- *    (C) 2004-2008, Open Source Geospatial Foundation (OSGeo)
+ *    (C) 2004-2016, Open Source Geospatial Foundation (OSGeo)
  *    
  *    This library is free software; you can redistribute it and/or
  *    modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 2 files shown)


dvntucker:
Hi @awaterme, the code looks good. Can you clarify why non-SimpleType's would be skipped? Not saying it's wrong, just want to understand it better. In general I'd expect that sort of decision to be made by downstream code. Thanks!

awaterme:
Hi @dvntucker, thanks for review this!

These were the reasons for my decision:

1. This patch affects the printing of AbstractFeatureType, more precisely the printing of attributes of SimpleFeatures. SimpleFeatures may have simple types or geometries as attributes. In either case it make no sense to attempt to print these attributes, when null. So it should be valid to skip nulls.
2. A lot of geometry types (MultiPoint, MultiLine, MultiPolygon,...) explicitly check their parameters on encode() and throw an exception when called with null (UnsupportedOperation etc). As it was clearly the intention of the author to prohibit calls with null I considered it harmful to change this behaviour. Especially as geometries might be used in contexts other than SimpleFeatures.
3. Andrea Aime pointed out on the mailing list, that he would expect upper layers to handle nulls.

Hope this is valid for you.

Best regards,
Andreas

PS: Cant resist to use the opportunity to ask a similar question: Many simple types (XSBooleanBinding, XSDateBinding, XSDateTimeBinding, ...) currently refuse parsing empty strings (""). 
This is quite a problem for WFS-T clients submitting null values as empty elements, because this ends in exceptions. Even the GeoTools own WFS-T client (gt-wfs, FeatureStore-API) sends null attributes as empty elements making it incompatible with GeoServer. 
Is this by intention? What do you think about making the bindings more flexible and treat "" as null?

dvntucker:
Thanks for the explanation. 

As for your question, I'm not sure, but unless I'm mistaken empty strings aren't valid for those types, so I would guess that the clients are wrong for sending along empty values for them. I'm not an expert in that area, but maybe someone on the devel list could better answer your questions. Cheers.

PR #1294, opened 2016-09-01
[GEOT-5505][GEOT-5504] Adjusting mbtiles module for PBF vector tiles

smithkm:
Fixes GEOT-5505 by using `baselayer` type instead of `base_layer`.  `base_layer` still accepted but logs a warning.

Implements GEOT-5504 by adding the `pbf` format.


modules/unsupported/mbtiles/src/main/java/org/geotools/mbtiles/MBTilesMetadata.java
```diff
@@ -1,17 +1,48 @@
...
 public class MBTilesMetadata {
     
+    private static Logger LOGGER = Logging.getLogger(MBTilesMetadata.class); 
+    
     protected static Pattern patternEnvelope = Pattern.compile(" *(\\-?[0-9\\.]*) *, *(\\-?[0-9\\.]*) *, *(\\-?[0-9\\.]*) *, *(\\-?[0-9\\.]*) *");
     
-    public enum t_type {OVERLAY, BASE_LAYER} ;
+    public enum t_type {
+        OVERLAY("overlay"), 
+        BASE_LAYER("baselayer");
```
tbarsballe: Underscore in the enum, but not in the identifier? I assume this is inteded, but just want to verify.

modules/unsupported/mbtiles/src/main/java/org/geotools/mbtiles/MBTilesMetadata.java
```diff
@@ -1,17 +1,48 @@
...
+            this.identifier = identifier;
+        }
+        
+        public final String identifier;
+        
+        public static t_type lookUp(final String s) {
+            return Arrays.stream(t_type.values())
+                .filter((t)->t.identifier.equalsIgnoreCase(s))
+                .findAny()
+                .orElseThrow(()->new IllegalArgumentException(String.format("Unknown mbtiles type '%s'",s)));
```
tbarsballe: From an API standpoint, shouldn't the lookup method have `throws IllegalArgumentException`?
I am not quite sure how this works with the Java 8 constructs you are using here?

modules/unsupported/mbtiles/src/main/java/org/geotools/mbtiles/MBTilesMetadata.java
```diff
@@ -155,7 +191,11 @@ public void setBoundsStr(String boundsStr) {
             double miny = Double.parseDouble(matcherEnvelope.group(2));
             double maxx = Double.parseDouble(matcherEnvelope.group(3));
             double maxy = Double.parseDouble(matcherEnvelope.group(4));
-            setBounds(new ReferencedEnvelope(minx, maxx, miny, maxy, null));
```
tbarsballe: Now that this is surrounded by a try/catch, should we be throwing an exception up-stack to notify calling code that there was a failure?

modules/unsupported/mbtiles/src/main/java/org/geotools/mbtiles/MBTilesMetadata.java
smithkm: That was the problem I was fixing.  The original code used an underscore for both which violated the spec.  I changed the string used in the file to not have an underscore (complying with the spec) but left the enum identifier as it was to avoid breaking anything.  During parsing I make it accept either string so that files written with older versions of GeoTools that did this wrong will still work.

modules/unsupported/mbtiles/src/main/java/org/geotools/mbtiles/MBTilesMetadata.java
smithkm: IllegalArgumentException is a RunTimeException so throws is optional.  I'm happy to add it though.

modules/unsupported/mbtiles/src/main/java/org/geotools/mbtiles/MBTilesMetadata.java
tbarsballe: Makes sense (I should really actually read the description of the PR).

modules/unsupported/mbtiles/src/main/java/org/geotools/mbtiles/MBTilesMetadata.java
tbarsballe: Probably better to have it than not to have it. Not a major issue though.

modules/unsupported/mbtiles/src/main/java/org/geotools/mbtiles/MBTilesMetadata.java
smithkm: The exception happens if "EPSG:4326" can't be parsed to a CRS.  I think that should never actually happen.

tbarsballe: Fair enough.

modules/unsupported/mbtiles/src/test/java/org/geotools/mbtiles/MBTilesFileVectorTileTest.java
```diff
@@ -0,0 +1,68 @@
...
+import org.junit.Before;
+import org.junit.Test;
+
+public class MBTilesFileVectorTileTest {
+    
+    File dbfile;
+    
+    @Before
+    public void setUp() {
+        dbfile = new File(MBTilesFileVectorTileTest.class.getResource("planet.mbtiles").getFile());
```
bencaradocdavies: @smithkm this bad URL->File conversion breaks the build in a path with spaces. The failure causes attempts to access filesystem paths containing ```%20``` that do not exist. Building in a path with spaces is a proxy for testing that code supports i18n-safe URL<->File conversions.

Fixed on master; please include this fix in any backport:
https://github.com/geotools/geotools/commit/d3d878b9a8cf9161b0a91d1b61b47ac8b06b68fb

Please use DataUtilities for all URL<->File conversions:
http://docs.geotools.org/latest/developer/conventions/code/url.html

PR #1295, opened 2016-09-02
[GEOT-5506]: Specific encoding for XSISimpleTypes.Date and DateTime

awaterme:
Fix for [GEOT-5506](https://osgeo-org.atlassian.net/browse/GEOT-5506)


aaime:
Looks good to me, merging. Mind, the wfs plugin is about to be phased out in favor of wfs-ng, which might be using a different gml encoder for transactios (I'm assuming it is, but I have not checked the code)

PR #1300, opened 2016-09-08
[GEOT-5513] Changes for supporting AbstractGridFormat.BANDS parameter 

mantesat:
[GEOT-5513](https://osgeo-org.atlassian.net/browse/GEOT-5513)

Adding a BANDS parameter in AbstractGridFormat. Also, changes include adjustments on GridCoverageRenderer  for supporting readers using this parameter and  "resetting" of selection channels on raster symbolizers (when necessary) so as to use bands defined by the BANDS parameter.

### DIFF
--- a/modules/library/render/src/main/java/org/geotools/renderer/lite/gridcoverage2d/ChannelSelectionUpdateStyleVisitor.java
+++ b/modules/library/render/src/main/java/org/geotools/renderer/lite/gridcoverage2d/ChannelSelectionUpdateStyleVisitor.java
@@ -1,0 +1,108 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ *
+ *    (C) 2016, Open Source Geospatial Foundation (OSGeo)
+ *
+ *    This library is free software; you can redistribute it and/or
+ *    modify it under the terms of the GNU Lesser General Public
+ *    License as published by the Free Software Foundation;
+ *    version 2.1 of the License.
+ *
+ *    This library is distributed in the hope that it will be useful,
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ *    Lesser General Public License for more details.
+ */
+
+package org.geotools.renderer.lite.gridcoverage2d;
+
+import org.geotools.coverage.grid.io.AbstractGridFormat;
+import org.geotools.styling.ChannelSelection;
+import org.geotools.styling.ColorMap;
+import org.geotools.styling.ContrastEnhancement;
+import org.geotools.styling.RasterSymbolizer;
+import org.geotools.styling.SelectedChannelType;
+import org.geotools.styling.ShadedRelief;
+import org.geotools.styling.Symbolizer;
+import org.geotools.styling.visitor.DuplicatingStyleVisitor;
+import org.opengis.filter.expression.Expression;
+
+/**
+ * ChannelSelectionUpdateStyleVisitor is a {@link DuplicatingStyleVisitor} that is used 
+ * to "reset" style symbolizer's selection channel order when the GridCoverageReader 
+ * used to read a coverage supports band selection. If a reader supports band selection
+ * then channel ordering is done by band selection from the reader, so symbolizer does
+ * not need to re-apply selection channel order.
+ * 
+ * Also, see {@link AbstractGridFormat#BANDS} for the reader band selection parameter 
+ * description.
+ * 
+ * @source $URL$
+ * @version $Id$
+ *
+ */
+
+public class ChannelSelectionUpdateStyleVisitor extends DuplicatingStyleVisitor{
+
+    private SelectedChannelType[] channels;
+    
+    public ChannelSelectionUpdateStyleVisitor(SelectedChannelType[] channels){
+        super();
+        this.channels = channels;
+    }
+    
+    @Override
+    public void visit(RasterSymbolizer raster) {
+
+        ChannelSelection channelSelection = createChannelSelection();
+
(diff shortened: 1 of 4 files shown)


modules/library/render/src/main/java/org/geotools/renderer/lite/gridcoverage2d/ChannelSelectionUpdateStyleVisitor.java
```diff
@@ -0,0 +1,85 @@
...
+	
+	private SelectedChannelType[] channels;
+	
+	public ChannelSelectionUpdateStyleVisitor(SelectedChannelType[] channels){
+		super();
+		this.channels = channels;
+	}
+	
+    @Override
+    public void visit(RasterSymbolizer raster) {
```
aaime: Thinking out loud, could have this been done by just overriding the parent's "protected ChannelSelection copy(ChannelSelection channelSelection)" ? Not a big deal anyways.

modules/library/render/src/main/java/org/geotools/renderer/lite/gridcoverage2d/GridCoverageRenderer.java
```diff
@@ -970,6 +976,38 @@ private RenderedImage renderImage(final GridCoverage2DReader reader,
...
         
+        // Check if reader supports band selection, and rearrange raster channels order in
+        // symbolizer. Reader should have taken care o proper channel order, based on initial
+        // symbolizer channel definition
+        if (reader.getFormat().getReadParameters().getDescriptor().descriptors().
+        		contains(AbstractGridFormat.BANDS) && readParams!=null){
+	        // if bands are selected, alter the symbolizer to use bands in order 0,1,2,... since the
+	        // channel order defined by it previously is taken care of the reader
+            int[] bandIndices = null;
+        	for (int i = 0; i < readParams.length; i++) {
```
aaime: Formatting issues here, there are tabs (rule, make sure new code abides to the formatting rules, but don't reformat pre-existing code). Most IDEs have facilities to only format selected portions of code, or reformat on save only what has been modified.

modules/library/render/src/main/java/org/geotools/renderer/lite/gridcoverage2d/GridCoverageRenderer.java
```diff
@@ -970,6 +976,38 @@ private RenderedImage renderImage(final GridCoverage2DReader reader,
...
+	            if (channels!=null){
+		            int i=0;
+		            for (SelectedChannelType channel:channels){
+		                //Remember, channel indices start from 1
+		                channel.setChannelName(Integer.toString(i+1));
+		                i++;
+		            }
+
+		            ChannelSelectionUpdateStyleVisitor channelsUpdateVisitor = new ChannelSelectionUpdateStyleVisitor(channels);
+	                symbolizer.accept(channelsUpdateVisitor);
```
aaime: By itself this does nothing. The proper way to use a duplicating visitor is to make it duplicate the entire symbolizer, and replace the old symbolizer by the one returned by the visit. The above code modifying in place the old symbolizer can cause trouble as styles can be cached by the streaming renderer caller.

modules/library/render/src/main/java/org/geotools/renderer/lite/gridcoverage2d/GridCoverageRenderer.java
```diff
@@ -1164,6 +1202,35 @@ public void paint(final Graphics2D graphics, final GridCoverage2DReader gridCove
 
         setupInterpolationHints(interpolation);
         
+        //
+        // Band selection
+        //
+        int[] bandIndicesTmp = null;
+        ChannelSelection channelSelection = symbolizer.getChannelSelection();
```
aaime: Does it make sense to do the band selection and check for the bands param support just once? Or at least externalize that in a single method? You are two copies grabbing the indices from the channel selection, here and above.

aaime:
I have checked the build failure locally, it appears to be legit, I also get:

````
Failed tests: 
  StreamingRendererTest.testDrawIntepolation:222 expected:<null> but was:<javax.media.jai.InterpolationBicubic@38de5daf>
````

**Comment by aaime** [comment]
Btw, I see there are no tests but there is not a reader supporting BANDS here either. app-schema routinely makes changes in GT and adds tests to GeoServer instead, I guess it might be good for this case too.

**Comment by mantesat** [comment]
Comments by aaime have been addresed on commit  498ab65

File: modules/library/render/src/main/java/org/geotools/renderer/lite/gridcoverage2d/GridCoverageRenderer.java [hunk 5]
```diff
@@ -1164,8 +1178,14 @@ public void paint(final Graphics2D graphics, final GridCoverage2DReader gridCove
 
         setupInterpolationHints(interpolation);
         
+        //
+        // Band selection
+        //
+        applyBandsSelectionParameterIfNeeded(gridCoverageReader, readParams, symbolizer);
+        RasterSymbolizer finalSymbolizer = setupSymbolizerIfBandsSelectionIsUsed(symbolizer);
```
aaime: Aren't you modifying the symbolizer also if the bands parameters are not supported in the reader this way? Please add a test that renders a coverage from a reader with channel selection (e..g, select the second band out of a RGB sources) and make sure this code is not interfering (e.g, you are getting back the green one, not the red one, suggestion, create a compressed geotiff that's fully green to simplify the test checks)

mantesat: Fixed in [54bdacb](https://github.com/mantesat/geotools/commit/54bdacb58e7edb75639ec16aac993c132b7f7c90)

aaime:
Manually merged to add some more test coverage and a couple of fixes

PR #1301, opened 2016-09-08
[GEOT-5516] NetCDF add_offset and scale_factor attributes not respected

dvntucker:
- Removed the calls to change the default enhance mode. The default enhance mode already includes Coordinate System, and this call turns off scale/offset/missing enhancement.
- Updated getRawDataType to just get the data type from the variable, rather than the original (since the original doesn't include datatype change from scale/offset.


bencaradocdavies:
Test fails on Travis CI with:
```
testScaleAndOffset(org.geotools.coverage.io.netcdf.NetCDFReaderTest)  Time elapsed: 0.029 sec  <<< ERROR!
java.lang.ClassCastException: org.geotools.gce.imagemosaic.ImageMosaicReader cannot be cast to org.geotools.coverage.io.netcdf.NetCDFReader
	at org.geotools.coverage.io.netcdf.NetCDFReaderTest.testScaleAndOffset(NetCDFReaderTest.java:204)
```

dvntucker:
Failure is fixed. Just changed the test to use NetCDFReader directly, since using the format finder was returning ImageMosaic.

dromagnoli:
Hi Devon.
I'm wondering if this behavior should be somehow made configurable.

In case some GeoServer users already have several NetCDF data configured and some kind of style configured on top of them, dealing with quantities in a colorMap, or some other rendering transformation doing math on the values. I think that with the new jar, things will stop working (or behave differently) due to the different values being returned so that the styles need to be updated.
So we should take this into account by making it configurable or just explicitly say this somewhere in the documentation for the version migration instructions.

What do you think about it?

jodygarnett:
It would be good to sort this out before the code freeze. We can add this to the release notes if that helps @dromagnoli?

dromagnoli:
@jodygarnett @dvntucker 
Wondering if adding a system property too would also make sense (in case it won't require too much effort) together with the release notes for migration to latest version.
The new scale/offset management code added by Devon would be the default and that property can be used to explicitly disable it for backward compatibility in case anyone needs it.

What do you think?

dvntucker:
A system property would probably be easiest at this point. I'll try to get it in this morning.

simboss:
I would go for the system var too. Ideally it should have been the other way around at least for a release the new behavior should NOT be enabled by default but rather explicitly.

Anyway, life isn't perfect, I am happy to have a backword compatibility flag. I'd open a JIRA report to remove that thing at a certain point while documenting the change to be done on installation to make things work.

modules/plugin/coverage-multidim/netcdf/src/main/java/org/geotools/coverage/io/netcdf/NetCDFReader.java
```diff
@@ -641,6 +648,11 @@ private CoverageSource getGridCoverageSource(final String coverageName) throws I
             // create, cache and return
             CoverageSource source = access.access(new NameImpl(coverageName), null,
                     AccessType.READ_ONLY, null, null);
+
+            if (source instanceof NetCDFSource) {
+
+            }
+
```
dromagnoli: What is this block?

modules/plugin/coverage-multidim/netcdf/src/main/java/org/geotools/coverage/io/netcdf/NetCDFReader.java
```diff
@@ -142,6 +144,12 @@
 
     String defaultName = null;
 
+    static {
+        if (Boolean.getBoolean("org.geotools.coverage.io.netcdf.coordSysEnhanceOnly")) {
+            NetcdfDataset.setDefaultEnhanceMode(EnumSet.of(NetcdfDataset.Enhance.CoordSystems));
+        }
+    }
+
```
dromagnoli: Is this the right place where to put the systemProperty check?
I have seen other EnhancingMode in SPI and Utilities.

modules/plugin/coverage-multidim/netcdf/src/main/java/org/geotools/imageio/netcdf/utilities/NetCDFUtilities.java
```diff
@@ -453,7 +452,7 @@ public static int getDimensionLength(Variable var, final int dimensionIndex) {
      */
     public static int getRawDataType(final VariableIF variable) {
         VariableDS ds = (VariableDS) variable;
-        final DataType type = ds.getOriginalDataType();
+        final DataType type = ds.getDataType();
```
dromagnoli: Does the system property check influences the dataType too?

modules/plugin/coverage-multidim/netcdf/src/main/java/org/geotools/coverage/io/netcdf/NetCDFReader.java
dvntucker: Hmm, looks like a mistaken commit, I'll remove it.

modules/plugin/coverage-multidim/netcdf/src/main/java/org/geotools/imageio/netcdf/utilities/NetCDFUtilities.java
dvntucker: It doesn't now, but it probably should. If you're not enhancing the variables then you probably want the original data type. I'll change this to check for the ScaleMissing system property.

modules/plugin/coverage-multidim/netcdf/src/main/java/org/geotools/coverage/io/netcdf/NetCDFReader.java
dvntucker: I can put it in the SPI, that would probably make more sense.

dvntucker:
Ok, made a few changes to this:

* Old behavior is now default.
* All the Enhance modes can now be enabled/disabled via System property.
* Adding documentation to the multidim docs for these properties.
* Getting the data type checks for Scale/Offset setting and returns based on that.

Missing is: a real test for all this. I have the original test there set to @Ignore right now (since it will fail without a System property). Happy to write legit tests for these, but AFAICT it's pretty difficult to test params controlled by System params like that (threading issues with other tests and so on). Not sure if anyone has good ideas for that.

Ideally it would probably be a read parameter anyway, which would be more testable. However I'm sort of out of time to work on this, and it seemed quite difficult to thread a new read param all the way through the various layers to the Netcdf-Java code, especially since the actual NetcdfDataset is wrapped in the ImageIO API and read via that. Certainly open to suggestions on that front as well.

bencaradocdavies:
@dvntucker multithreaded Maven builds modules in parallel, but each module test phase gets it own JVM instance, which runs tests sequentially (in the default configuration, which we use). This means that you should be fine setting System properties in your test fixture, as long as you restore them afterwards. A bit ugly, but much better to have test coverage of functionality enabled by system properties.

More evidence of this practice:
```
$ find . -name "*Test.java" -exec grep -H System.setProperty {} \; | grep -v '//' | wc -l
55
$ find modules/plugin/coverage-multidim/netcdf/ -name "*Test.java" -exec grep -H System.setProperty {} \;
modules/plugin/coverage-multidim/netcdf/src/test/java/org/geotools/coverage/io/netcdf/CoordinateVariableTest.java:        System.setProperty("user.timezone", "GMT");
modules/plugin/coverage-multidim/netcdf/src/test/java/org/geotools/coverage/io/netcdf/CoordinateVariableTest.java:        System.setProperty("netcdf.coordinates.enablePlugins", "true");
modules/plugin/coverage-multidim/netcdf/src/test/java/org/geotools/coverage/io/netcdf/NetCDFCRSTest.java:        System.setProperty(NetCDFCRSAuthorityFactory.SYSTEM_DEFAULT_USER_PROJ_FILE, netcdfPropertiesPath);
modules/plugin/coverage-multidim/netcdf/src/test/java/org/geotools/coverage/io/netcdf/NetCDFReaderTest.java:        System.setProperty("netcdf.coordinates.enablePlugins", "true");
modules/plugin/coverage-multidim/netcdf/src/test/java/org/geotools/coverage/io/netcdf/PostGisIndexTest.java:        System.setProperty(NetCDFCRSAuthorityFactory.SYSTEM_DEFAULT_USER_PROJ_FILE,
modules/plugin/coverage-multidim/netcdf/src/test/java/org/geotools/coverage/io/netcdf/NetCDFMosaicReaderTest.java:        System.setProperty("org.geotools.referencing.forceXY", "true");
modules/plugin/coverage-multidim/netcdf/src/test/java/org/geotools/coverage/io/netcdf/NetCDFMosaicReaderTest.java:        System.setProperty("user.timezone", "GMT");
modules/plugin/coverage-multidim/netcdf/src/test/java/org/geotools/coverage/io/netcdf/NetCDFMosaicReaderTest.java:        System.setProperty("org.geotools.shapefile.datetime", "true");
```

dvntucker:
Ok, didn't know that about the threaded tests. I've run into issues before using the forceXY property in tests, but I guess I misunderstood the root issue. I'll fix my tests then.

dvntucker:
Actually, thinking about it a bit more, changing the system property at runtime isn't going to do anything since the defaults are set in a static block right now. We could either:

- Test the scale/offset behavior by changing the NetcdfDataset.defaultEnhanceMode at runtime in the test (and changing back after the test). This would test the scale/offset behavior, but leave the System property configuration itself untested.

- Put in the time to actually make the enhance mode a read parameter of the NetcdfReader. I took a stab at this, but it ends up being quite messy, and needs to use a method that is explicitly documented as only being accidentally public. This is what this looks like: https://github.com/dvntucker/geotools/pull/5/files

So... yeah, kind of messy either way. A simple solution that's not very testable or a messy solution that is testable. Unless there's something I'm missing, which is quite possible.

dromagnoli:
Hi @dvntucker.
At this point I would say let's move with option 1.
If needed, we can work on per request parameter (option 2) in the future.

dvntucker:
Ok, updated the test to run after changing the defaults in code, then setting the defaults back after it's run. This PR is ready to go from my perspective.

modules/plugin/coverage-multidim/netcdf/src/main/java/org/geotools/imageio/netcdf/utilities/NetCDFUtilities.java
```diff
@@ -453,7 +489,12 @@ public static int getDimensionLength(Variable var, final int dimensionIndex) {
      */
     public static int getRawDataType(final VariableIF variable) {
         VariableDS ds = (VariableDS) variable;
-        final DataType type = ds.getOriginalDataType();
+        final DataType type;
+        if (Boolean.getBoolean(ENHANCE_SCALE_MISSING)) {
+            type = ds.getDataType();
```
dromagnoli: why not checking the static ENHANCE_SCALE_OFFSET at this point?

dvntucker: That's what I originally had, but unfortunately it's then not programmatically testable. I suppose I could make the variable not final? Neither is my favorite solution.

jodygarnett:
merge conflicts and code freeze holding this up

bencaradocdavies:
https://osgeo-org.atlassian.net/browse/GEOT-5516

PR #1302, opened 2016-09-09
GEOT-5514: Disabled DTD's in gt-xml to protect against XEE attacks

aaronwaddell:
See: https://osgeo-org.atlassian.net/browse/GEOT-5514

Also added PowerMockito in order to add tests for DocumentFactory.java and protect against regression.

### DIFF
--- a/modules/library/xml/pom.xml
+++ b/modules/library/xml/pom.xml
@@ -135,6 +135,18 @@
         <version>1.9.5</version>
         <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.powermock</groupId>
+      <artifactId>powermock-module-junit4</artifactId>
+      <version>1.5.6</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.powermock</groupId>
+      <artifactId>powermock-api-mockito</artifactId>
+      <version>1.5.6</version>
+      <scope>test</scope>
+    </dependency>
 
   </dependencies>
 
--- a/modules/library/xml/src/main/java/org/geotools/xml/DocumentFactory.java
+++ b/modules/library/xml/src/main/java/org/geotools/xml/DocumentFactory.java
@@ -141,7 +141,7 @@ public class DocumentFactory {
     }
 
     /*
-     * convinience method to create an instance of a SAXParser if it is null.
+     * Convenience method to create an instance of a SAXParser if it is null.
      */
     private static SAXParser getParser() throws SAXException {
         SAXParserFactory spf = SAXParserFactory.newInstance();
@@ -149,8 +149,10 @@ public class DocumentFactory {
         spf.setValidating(false);
 
         try {
-            // spf.setFeature("http://xml.org/sax/features/external-general-entities", false);
-            // spf.setFeature("http://xml.org/sax/features/external-parameter-entities", false);
+            // The following configuration prevents XML External Entity Injection (XXE) attacks
+            // See for more information: https://www.owasp.org/index.php/XML_External_Entity_(XXE)_Processing
+            spf.setFeature("http://xml.org/sax/features/external-general-entities", false);
+            spf.setFeature("http://xml.org/sax/features/external-parameter-entities", false);
 
             SAXParser sp = spf.newSAXParser();
             return sp;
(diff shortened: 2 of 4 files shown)


aaronwaddell:
In order to demonstrate the fixes effectiveness try running /aaronwaddell/geotools-xml-entity-injection using Geotools built from master and then Geotools built from this branch. You should expect that although XML External Entity Injection is possible on master it would not be possible on this branch.

mprins:
Why add powermock dependencies in two places? And different versions at that. Ideally the version and possibly scope is specified in the dependencyManagement section of the parent pom and those modules using the dependency should not specify a version

aaronwaddell:
Powermock provides specific versions for compatible versions of mockito, see here: https://github.com/jayway/powermock/wiki/MockitoUsage. Attempting to use the incorrect version of Powermock for the version of Mockito you are using will usually result in a strange exception and a failed test. This is certainly true in this case.

Powermock has been added to the the parent pom in the dependencyManagement section with a version specified to match the version of Mockito (1.8.5) also specified at the parent pom. By doing so we allow for submodules to make use of Powermockito correctly and not have to worry about version compatibilities, whilst also adhering to Maven dependency management best practices. However, you will note that the version of Mockito in modules/library/xml/pom.xml is specified higher than that of the parent pom (1.9.5 vs 1.8.5). The (higher) version of Powermock is specified again in this pom file in order to account for version compatibilities with Mockito.

aaime:
GeoTools already depends on mockito and easymock, could you please use one of those two instead of adding a third mocking library in the mix.
Also XEE is already handled in the library by allowing users to set their own entity resolver, while I have not looked into this in detail it would be best if the WMS cascading module did the same for consistency, see: https://github.com/geotools/geotools/commit/1f949bc7a7e6735b9230fb15234df8b885b0f104
https://github.com/geoserver/geoserver/commit/71633cb133b4e32778d2f2fde7e617aa6692ec16

aaronwaddell:
Introducing Powermock to a project already making use of Mockito and Easymock may seem undesirable, however neither Easymock or Mockito alone is capable of mocking static methods and therefore testing DocumentFactory. Powermock itself provides extension APIs for both Mockito and Easymock and arguably shouldn't be considered an additional mocking library, but more of an extension of what we already have. I would perhaps suggest that the real issue here is that Geotools uses both Mockito **and** Easymock? Of course there is always the option to not test DocumentFactory, but it is highly likely that the bug crept in unnoticed for this reason.

It appears that Geoserver is using XMLReader as opposed to the higher level XMLParser used in Geotools' DocumentFactory. XMLReader supports setting an EntityResolver directly whereas XMLParser does not, hence the use of features to control DTD parsing. It appears that Geotools has two approaches to XMLParsing, the DocumentFactory and Parser classes. Parser makes use of XMLParser however also uses ParserHandler as a means of setting an EntityResolver. Currently gt-wms uses DocumentFactory and not Parser. It looks like some early work has taken place for gt-wms to use Parser in the form of WMSConfiguration, but when compared to WFSConfiguration it looks like a relatively large amount of work is still required in order to achieve this.

A good point raised by the Geoserver implementation is regarding configurability; should users be able to toggle external entity parsing on and off? I believe the safer answer would be "yes". In light of this I'll add another commit which would allow a user to choose to parse external entities if they wish, but still default to ignoring them.

jodygarnett:
Always a challenge to track discussion across email list, jira and pull requests.

* Please see [GEOT-4404]() / https://github.com/geotools/geotools/pull/141 for an example of introducing an EntityResolver. This represents the alternative approach @aaime refers to allowing entity resolution to be controllable (rather than just disabled).

In particular we use this approach to limit XSD resolution to schemas that have been included in the geotools library (and thus available to downstream applications in a jar file). This allows validation, protects against XEE attacks, without simply disabling.

It would of course be preferable to change the wms client code to the GTXML parser - if you have any leads on volunteers and/or funding it would be a great improvement.

jodygarnett:
Giving it a go to introduce an EntityResolver ... and discovered that XMLSAXHandler has the following:

````
    public InputSource resolveEntity( String pubId, String sysId ) throws SAXException {
        // avoid dtd files
        if(sysId != null && sysId.endsWith("dtd")){
	  return new InputSource(new StringReader(""));
	}
       ...
```

jodygarnett:
Okay have a look at https://github.com/jodygarnett/geotools/commits/GEOT-5514

* disabled doctypes at the sax parser level (primary defence again DTD XXE)
* EntityResolver introduced, available as a Hint, a good idea to be consistent
* note: DTD was already disabled at the entity resolver level
* note: your disable switch is available as a hint (incase you still want it after all of the above)

aaime:
Superceded by another pull request

aaronwaddell:
I was just about to close this :)

At a glance @jodygarnett PR looks ideal. I'll aim to take a more detailed look later on today.

Thanks to all those involved.

PR #1305, opened 2016-09-13
[GEOT-5519] Fix graphic size on custom DPIs

awaterme:
Suggested fix for [GEOT-5519](https://osgeo-org.atlassian.net/browse/GEOT-5519), including test cases.

### DIFF
--- a/modules/library/main/src/main/java/org/geotools/styling/visitor/Measure.java
+++ b/modules/library/main/src/main/java/org/geotools/styling/visitor/Measure.java
@@ -2,7 +2,7 @@
  *    GeoTools - The Open Source Java GIS Toolkit
  *    http://geotools.org
  *
- *    (C) 2013, Open Source Geospatial Foundation (OSGeo)
+ *    (C) 2013-2016, Open Source Geospatial Foundation (OSGeo)
  *
  *    This library is free software; you can redistribute it and/or
  *    modify it under the terms of the GNU Lesser General Public
@@ -142,6 +142,13 @@ class Measure {
     boolean isPixelInPixelDefault() {
         return (uom == null || uom == defaultUnit)
                 && (defaultUnit == null || defaultUnit == NonSI.PIXEL);
+    }
+
+    /**
+     * @return true, if the uom is a real world unit within a symbolizer whose default unit is pixel.
+     */
+    public boolean isRealWorldUnitInPixelDefault() {
+        return isRealWorldUnit() && (defaultUnit == null || defaultUnit == NonSI.PIXEL);
     }
 
 }
--- a/modules/library/main/src/main/java/org/geotools/styling/visitor/RescalingMode.java
+++ b/modules/library/main/src/main/java/org/geotools/styling/visitor/RescalingMode.java
@@ -2,7 +2,7 @@
  *    GeoTools - The Open Source Java GIS Toolkit
  *    http://geotools.org
  *
- *    (C) 2013, Open Source Geospatial Foundation (OSGeo)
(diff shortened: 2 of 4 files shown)


modules/library/main/src/main/java/org/geotools/styling/visitor/Measure.java
```diff
@@ -2,7 +2,7 @@
  *    GeoTools - The Open Source Java GIS Toolkit
  *    http://geotools.org
  *
- *    (C) 2013, Open Source Geospatial Foundation (OSGeo)
+ *    (C) 2013-2016, Open Source Geospatial Foundation (OSGeo)
```
jodygarnett: Thanks! We have just recently relaxed our stance on updating headers - no harm in changing though

modules/library/main/src/main/java/org/geotools/styling/visitor/RescalingMode.java
```diff
@@ -150,5 +141,17 @@ public Expression rescaleToExpression(Expression scaleFactor, Measure measure) {
...
+     * Translates between units and their shortcuts (we can only get the full name from the unit
+     * object
+     */
+    final Map<Unit, String> UNIT_SYMBOLS = new HashMap<Unit, String>() {
+        {
+            put(NonSI.PIXEL, "px");
+            put(NonSI.FOOT, "ft");
+            put(SI.METER, "m");
+        }
+    };
```
jodygarnett: did this code just change position, or tabs to spaces? I do not see a difference...

modules/library/main/src/main/java/org/geotools/styling/visitor/RescalingMode.java
```diff
@@ -76,7 +63,11 @@ public String rescaleToStringInternal(double scaleFactor, Measure measure) {
         @Override
         public String rescaleToStringInternal(double scaleFactor, Measure measure) {
             if (measure.isRealWorldUnit()) {
-                return String.valueOf(measure.value);
+                if(measure.isRealWorldUnitInPixelDefault()){                    
+                    return String.valueOf(measure.value) + UNIT_SYMBOLS.get(measure.uom);
```
jodygarnett: This change looks sensible, continuing to preserve the unit when provided.

modules/library/main/src/main/java/org/geotools/styling/visitor/RescalingMode.java
```diff
@@ -76,7 +63,11 @@ public String rescaleToStringInternal(double scaleFactor, Measure measure) {
...
+                }else{
+                    return String.valueOf(measure.value);
```
jodygarnett: should we explicitly provide a unit of "px" here - since we are making a judgement call

jodygarnett:
This seems a straightforward fix, suitable for backport.

modules/library/main/src/main/java/org/geotools/styling/visitor/RescalingMode.java
awaterme: Hi Jody, I would prefer the current solution because it is consistent with the results of its sister method rescaleToExpression(...). 
rescaleToExpression() has unit tests expecting a result without "px" unit. 
Agree?

modules/library/main/src/main/java/org/geotools/styling/visitor/RescalingMode.java
awaterme: It's not obvious: `UNIT_SYMBOLS` was previously a field of the `KeepUnits` enum constant only. Now it is a field of the `RescalingMode` enum itself, making it accessible for both, `KeepUnits` and `Pixels`.

PR #1310, opened 2016-09-15
[GEOT-5521] Support Azimuthal Equidistant projection

bencaradocdavies:
https://osgeo-org.atlassian.net/browse/GEOT-5521

I would also like to backport this to 15.x for inclusion in 15.2.


bencaradocdavies:
Backported to 15.x.

PR #1313, opened 2016-09-16
[GEOT-724] backport for 14.x

mprins:
fixes "Oracle DataStore is unable to add empty geometries"

Applies patches provided in #1228 to 14.x branch


mprins:
not sure if there is interest for this as 14.x is basically eol, but the work was trivial.

PR #1314, opened 2016-09-18
GEOT-5514: Disabled DTD's in gt-xml to protect against XEE attacks

jodygarnett:
See: https://osgeo-org.atlassian.net/browse/GEOT-5514

This is a followup/replacement to #1302 by @aaronwaddell.

### DIFF
--- a/modules/library/xml/pom.xml
+++ b/modules/library/xml/pom.xml
@@ -135,19 +135,6 @@
         <version>1.9.5</version>
         <scope>test</scope>
     </dependency>
-    <dependency>
-      <groupId>org.powermock</groupId>
-      <artifactId>powermock-module-junit4</artifactId>
-      <version>1.5.6</version>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
-      <groupId>org.powermock</groupId>
-      <artifactId>powermock-api-mockito</artifactId>
-      <version>1.5.6</version>
-      <scope>test</scope>
-    </dependency>
-
   </dependencies>
 
 
--- a/modules/library/xml/pom.xml
+++ b/modules/library/xml/pom.xml
@@ -135,6 +135,18 @@
         <version>1.9.5</version>
         <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.powermock</groupId>
+      <artifactId>powermock-module-junit4</artifactId>
+      <version>1.5.6</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.powermock</groupId>
+      <artifactId>powermock-api-mockito</artifactId>
+      <version>1.5.6</version>
+      <scope>test</scope>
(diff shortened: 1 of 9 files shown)


aaime:
Hi @jodygarnett, I have prepared a quick branch showing that getting rid of the powermock dependency is actually simple, just create and pass an extra hint, see this commit:  https://github.com/aaime/geotools/commit/a7cb5a05446ea1e475d743a25c87ec47979b0a55
(it's added as a public hint, but could as well left protected and available just for the sake of testing).

However, wondering if it's redundant to have both a customizable entity resolver and the feature flags disabling XXE at the same time? The existing test certainly seems not to check at all entity resolution.

aaime:
Hum... tests are failing btw, looks like there is cross-test contamination.

jodygarnett:
The EntityResolver is useful on its own (used for schema resolution).

The disabling is fine, and actually we could turn it on by default as we do not use DTDs when working with WMS. The extra hint I found, to turn off DOCTYPE parsing is the most brutal / effective way to avoid *just* DTDs.

jodygarnett:
Thanks @aaime / @aaryno this now looks ready to merge

aaime:
Nope, not ready, it's still failing tests (the cross-test contamination I was talking about):
````
Tests in error: 
  LayerInheritanceTest.testInheritCapabilities:52 Â» SAXParse DOCTYPE is disallow...
  DimensionParsingTest.testDimensionExtent111:82 Â» IllegalArgument Don't know ho...
  Geot553Test.testGeot553:77 Â» IllegalArgument Don't know how to handle a get re...
  WMSCoverageReaderTest.test4326wms11:210->getReader4326wms11:252 Â» IllegalArgument
````

aaime:
I am going to cut the beta in the meantime, it's not meant for production so we can have this fixed in time for the RC.

aaime:
Duh, confused myself, me and @randomorder are actually in charge of 1.9.x 
@jodygarnett any chance you can fix the tests so that we can merge and backport to the stable series? I guess we'll have to cut the 15.x release tomorrow at this point.

jodygarnett:
Back now - let me have a look!

jodygarnett:
Oh wow "inheritCap.xml" has:

````
<!DOCTYPE WMT_MS_Capabilities SYSTEM "http://terraservice.net/OgcCapabilities_1_1_1.dtd">
````

So it looks like DTD is actually a thing - http://cite.opengeospatial.org/OGCTestData/wms/1.1.1/spec/wms1.1.1.html#xml_dtds.wms_capabilities_dtd

I will restore access!

aaime:
I haven't checked, but if that's not the case yet, make sure that any system property set has a clearproperty removing it at the end of the test

jodygarnett:
It was just a "default" parser setting I had filled in on the assumption that DTD was not used. Doing a bull build now..

PR #1317, opened 2016-09-21
GEOT-5514, make the xml configuration actually configurable (15.x branch)

aaime:
This makes entity resolution actually configurable and adds a coupe of tests proving it actually works

### DIFF
--- a/modules/extension/wms/src/main/java/org/geotools/data/wms/WMS1_0_0.java
+++ b/modules/extension/wms/src/main/java/org/geotools/data/wms/WMS1_0_0.java
@@ -232,7 +232,7 @@ public class WMS1_0_0 extends WMSSpecification {
         
         public Response createResponse(HTTPResponse httpResponse) throws ServiceException,
                 IOException {
-            return new WMSGetCapabilitiesResponse(httpResponse);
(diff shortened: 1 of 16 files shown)


tbarsballe:
Bit of a best-practice question - this makes API changes to the sole constructors for a couple of public classes:
* WMSGetCapabilitiesResponse(httpResponse, hints);
* DescribeLayerResponse(httpResponse, hints);

While these do seem to be core classes that should not be constructed by external libraries, shouldn't we also be including a non-hints version for backwards compatibility?

e.g:

```
public WMSGetCapabilitiesResponse(HTTPResponse response) throws ServiceException, IOException {
    return new WMSGetCapabilitiesResponse(response, new HashMap<String, Object>());
}
```

aaime:
@tbarsballe I did not bother because as you say those are internal classes not particularly useful when instantiated on their own outside of the lib (unfortunately people had and keep on adding public classes carelessly when default visibility would do). 
Adding the extra constructors would have made the debugging harder too, at least during the development.
But yes, adding afterwards the constructors is a good idea, a nice extra for sure.

aaime:
Added a commit to ensure backwards compatibility. Merging.

PR #1324, opened 2016-09-25
[GEOT-5527] PostgreSQL 9.6 breaks ASC_OR_DESC for index

bencaradocdavies:
https://osgeo-org.atlassian.net/browse/GEOT-5527

This pull request targets master.


bencaradocdavies:
With this change, full builds of GeoTools and GeoServer master both pass, including ```gt-jdbc-postgis``` online tests and GeoServer app-schema online tests, both against PostgreSQL 9.6 / PostGIS 2.3.

Note that GeoServer app-schema online tests never failed; only tests that query index information are affected by the change in PostgreSQL 9.6.

aaime:
+1

PR #1325, opened 2016-09-25
[GEOT-5527] PostgreSQL 9.6 breaks ASC_OR_DESC for index

bencaradocdavies:
https://osgeo-org.atlassian.net/browse/GEOT-5527

This pull request targets 15.x.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -104,7 +104,7 @@
     <fork.javac>true</fork.javac>
     <javac.maxHeapSize>256M</javac.maxHeapSize>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    <postgresql.jdbc.version>9.4-1201-jdbc41</postgresql.jdbc.version>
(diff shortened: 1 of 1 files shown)


aaime:
Unsure about this one, it's just a bugfix release, but out of our control... is there a risk to catch regressions in the driver with a last minute change?

mprins:
Considering that PostgreSQL 9.6 is to be released yet and early adopters can just upgrade the driver themselves I'd go for 15.3 (even though downgrading the driver in case of a regression is just as easy)

bencaradocdavies:
I am comfortable waiting for 15.3. It is prudent to do so as 15.x will soon be the maintenance branch and the manual workaround is easy. I have merged https://github.com/geotools/geotools/pull/1324 for master which lets development integration testing continue even on very up-to-date platforms. This PR can wait until after 15.2 is released

jodygarnett:
This is held up on the code freeze

PR #1328, opened 2016-09-30
added test for GEOT-5176 - white space in column names

ianturton:
I believe that this bug was fixed by [PR1183](https://github.com/geotools/geotools/pull/1183), this PR adds a test, just needs a working Oracle instance for testing.

### DIFF
--- a/modules/plugin/jdbc/jdbc-oracle/src/test/java/org/geotools/data/oracle/ColumnEncodingTest.java
+++ b/modules/plugin/jdbc/jdbc-oracle/src/test/java/org/geotools/data/oracle/ColumnEncodingTest.java
@@ -35,5 +35,16 @@ public class ColumnEncodingTest {
         dialect.encodeColumnName("name",buffer);
         assertEquals("NAME", buffer.toString());
     }
+    
+    /**
+     * test for GEOT-5176 Oracle can't handle spaces in column names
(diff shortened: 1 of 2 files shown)


modules/plugin/jdbc/jdbc-oracle/src/test/java/org/geotools/data/oracle/ColumnEncodingTest.java
```diff
@@ -35,5 +35,16 @@ public void test() {
...
+    
+    /**
+     * test for GEOT-5176 Oracle can't handle spaces in column names
+     */
+    @Test
+    public void testspaces() {
+        StringBuffer buffer = new StringBuffer();
+        OracleDialect dialect = new OracleDialect(null);
+        dialect.encodeColumnName("name with space",buffer);
+        assertEquals("NAME WITH SPACE", buffer.toString());
```
Gnafu: Shouldn't you check for "\"NAME WITH SPACE\"" ?

ianturton: good point

aaime:
Well, let's merge and see how the build server likes it...

PR #1329, opened 2016-10-01
[GEOT-5534] Make streaming renderer map the default geometry attribute to the real geometry attribute name

nmco:
Associated issue:
https://osgeo-org.atlassian.net/browse/GEOT-5534

This pull request also adds a test case for this.


aaime:
Yep, looks sane. Please backport

PR #1333, opened 2016-10-03
[GEOS-7576] Make rescale visitors rescale polygons graphic-margin parameter

nmco:
Associated issue:
https://osgeo-org.atlassian.net/browse/GEOS-7576

### DIFF
--- a/modules/library/api/src/main/java/org/geotools/styling/PolygonSymbolizer.java
+++ b/modules/library/api/src/main/java/org/geotools/styling/PolygonSymbolizer.java
@@ -2,7 +2,7 @@
  *    GeoTools - The Open Source Java GIS Toolkit
  *    http://geotools.org
  * 
- *    (C) 2002-2008, Open Source Geospatial Foundation (OSGeo)
+ *    (C) 2002-2016, Open Source Geospatial Foundation (OSGeo)
  *
  *    This library is free software; you can redistribute it and/or
  *    modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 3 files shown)


modules/library/api/src/main/java/org/geotools/styling/PolygonSymbolizer.java
```diff
@@ -2,7 +2,7 @@
  *    GeoTools - The Open Source Java GIS Toolkit
  *    http://geotools.org
  * 
- *    (C) 2002-2008, Open Source Geospatial Foundation (OSGeo)
+ *    (C) 2002-2016, Open Source Geospatial Foundation (OSGeo)
```
aaime: No need to update the copyright dates anymore, just add a standard header with the current year if the file is new or does not have a header

modules/library/main/src/main/java/org/geotools/styling/visitor/UomRescaleStyleVisitor.java
```diff
@@ -197,6 +197,7 @@ public void visit(PolygonSymbolizer poly) {
         Unit<Length> uom = copy.getUnitOfMeasure();
         rescaleStroke(copy.getStroke(), uom);
         rescaleFill(copy.getFill(), uom);
+        scaleIntArrayOption(copy.getOptions(), TextSymbolizer.GRAPHIC_MARGIN_KEY, uom);
```
aaime: While it's the same value, it's confusing to see TextSymbolizer referenced here, PolygonSymbolizer gets the right constant in this same pull request.

modules/library/api/src/main/java/org/geotools/styling/PolygonSymbolizer.java
aaime: Mind, this means you don't need to change the commit either, leave them be, just remember that you don't need to alter them anymore for the future changes.

modules/library/main/src/main/java/org/geotools/styling/visitor/UomRescaleStyleVisitor.java
nmco: Ouch ... I added the PolygonSymbolizer constant to make it clear and forgot to updated it (facepalm).

jodygarnett:
Judging from the above this is not yet ready to merge...

nmco:
Updated.

PR #1335, opened 2016-10-03
[GEOS-7576] Make rescale visitors rescale polygons graphic-margin parameter (backport 14.x)

nmco:
Associated issue:
https://osgeo-org.atlassian.net/browse/GEOS-7576


nmco:
The build failed only in OSX.

aaime:
The OSX failure is not a problem per se, it just run out of time and was starting to generate the docs "The job exceeded the maximum time limit for jobs, and has been terminated."

PR #1338, opened 2016-10-05
Copy provided hints to safely add required DEFAULT_NAMESPACE_HINT_KEY and VALIDATION_HINT [GEOT-5514]

jodygarnett:
For discussion see: https://osgeo-org.atlassian.net/browse/GEOT-5514

The change is needed to respect the ENTITY_RESOLVER hint provided when this code is used by GeoWebCache.


DBlasby:
The only comment I have for this is perhaps to first check if the key is already in the hints before adding it (just incase someone wanted to over-ride it)...

DBlasby:
Looking at history - In the past, hints were not passed in.  So, this isn't a problem.

smithkm:
Yep, although it is a good point.

jodygarnett:
I do not mind respecting a validation hint if passed in, but we need to lock down the wms schema passed in (or we won't access the right precompiled schema entities - each of which have parsing/encoding methods).

smithkm:
OK, I'm happy with it as is, but if you want to add the capacity for overriding default hints other than the namespace one, that would be icing on the pull request.

jodygarnett:
Revised based on feedback, also cleaned up test story.

I am comfortable squashing and merging this for ease of backporting.

DBlasby:
Looks good, jody!

PR #1339, opened 2016-10-05
[GEOT-5538] Extra space before schemalocation in getfeature response breaks parsing

mtoonen:
https://osgeo-org.atlassian.net/browse/GEOT-5538


aaime:
Manually merged to avoid the spurios merge commit and fix the commit message (it should refer to the ticket, basically, have the same text as the pull request title).
Backported to 16.x and 15.x while I was at it.

PR #1340, opened 2016-10-06
[GEOT-5514] Use NoExternalEntityResolver when parsing SVG Icons

tbarsballe:


aaime:
Yep, that will protect. What's giving me pause is that the rest of the system is configurable and defaults to allow entity expansion, this one is not configurable and defaults not to.
From ideal to more realistic I'd have:
- The system configurable but defaulting not to allow entity expansion
- The system configurable and default to allow entity expansion
- An inconsistent situation doing different things in different spots, but that still allows to lock down the system fully

jodygarnett:
This is a factory, so we should be able to use geotools factory hints - if not I could try and look up a global geotools hint for the entity resolver...

modules/library/metadata/src/main/java/org/geotools/factory/GeoTools.java
```diff
@@ -196,6 +196,22 @@ private static Properties loadProperites(String resource) {
     /**
      * The {@linkplain System#getProperty(String) system property} key for the default
      * value to be assigned to the {@link Hints#
+     * ENTITY_RESOLVER} hint.
+     *
+     * This setting specifies the XML Entity resolver to be used when configuring a SAXParser,
+     * by default it NoLocalExternalEntityResolver.
```
tbarsballe: a) Shouldn't this be PreventLocalEntityResolver (instead of NoLocalExternalEntityResolver)
b) We are in a javadoc, so can we @link to the class?

modules/plugin/svg/src/test/java/org/geotools/renderer/style/SVGGraphicFactoryTest.java
```diff
@@ -32,20 +37,21 @@
  */
 public class SVGGraphicFactoryTest extends TestCase {
 
-    private SVGGraphicFactory svg;
```
tbarsballe: Any particular reason we switched to re-instantiating a new factory each time? I don't see anywhere we modify it, and in the case where we need to instantiate one with hints we can do a special case.
This change just seems unnecessarily verbose to me.

modules/library/metadata/src/main/java/org/geotools/factory/GeoTools.java
```diff
@@ -196,6 +196,22 @@ private static Properties loadProperites(String resource) {
...
+     *
+     * @see Hints#ENTITY_RESOLVER
+     * @see #getDefaultHints
+     */
+    public static final String ENTITY_RESOLVER = "org.xml.sax.EntityResolver";
```
tbarsballe: Where are we setting PreventLocalEntityResolver to be the default, becuase it doesn't look like we are doing so here, and getDefaultHints doesn't seem to have been modified.

modules/plugin/svg/src/test/java/org/geotools/renderer/style/SVGGraphicFactoryTest.java
jodygarnett: getting ready for when we have a save vs non safe default

modules/library/metadata/src/main/java/org/geotools/factory/GeoTools.java
jodygarnett: We are not doing that yet, we are currently unsafe - just to confirm we can supply PreventLocalEntityResolver via a hint.

tbarsballe: Okay, the javadoc above does not match the code behavior then

modules/library/metadata/src/main/java/org/geotools/factory/GeoTools.java
jodygarnett: The down side of not using {{@link ClassName}} in javadocs is the refactor won't catch the rename.

modules/library/render/src/main/java/org/geotools/renderer/style/SLDStyleFactory.java
```diff
@@ -1366,7 +1329,7 @@ private Icon getIcon(ExternalGraphic eg, Object feature, double size) {
 
 		// scan the external graphic factories and see which one can be used
 		Iterator<ExternalGraphicFactory> it = DynamicSymbolFactoryFinder
-				.getExternalGraphicFactories();
+				.getExternalGraphicFactories(new Hints(renderingHints));
```
DBlasby: I'm not sure about this change...

modules/plugin/svg/src/main/java/org/geotools/renderer/style/SVGGraphicFactory.java
```diff
@@ -85,6 +99,56 @@
...
+            if( this.resolver == null ){ // use null instance rather than check each time
+                this.resolver = NullEntityResolver.INSTANCE;
+            }
+        }
+        else {
+            this.resolver = defaultResolver();
+        }
+    }
+    private static final Logger LOGGER = org.geotools.util.logging.Logging.getLogger("org.geotools.renderer.style.SVGGraphicFactory");
+    protected static EntityResolver defaultResolver() {
```
DBlasby: Not sure if this should be here - look like generic code useful to anyone looking for a resolver.  Perhaps put somewhere else?

tbarsballe:
Compile failure:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.1:compile (default-compile) on project gt-svg: Compilation failure
[ERROR] /home/travis/build/geotools/geotools/modules/plugin/svg/src/main/java/org/geotools/renderer/style/SVGGraphicFactory.java:[71,34] error: package com.sun.xml.internal.stream does not exist
```

tbarsballe:
Manually merged to master

tbarsballe:
Backported to 15.x

PR #1341, opened 2016-10-06
[GEOT-3318] Added support for reading from InputStream with GeoTiffReader.

initjsk:
[Issue: GEOT-3318](https://osgeo-org.atlassian.net/browse/GEOT-3318)

### DIFF
--- a/modules/plugin/geotiff/src/test/java/org/geotools/gce/geotiff/GeoTiffReaderTest.java
+++ b/modules/plugin/geotiff/src/test/java/org/geotools/gce/geotiff/GeoTiffReaderTest.java
@@ -1157,14 +1157,24 @@ public class GeoTiffReaderTest extends org.junit.Assert {
      */
     @Test
     public void testCanReadInputStream() throws IOException {
+        
         File rasterfile = TestData.file(GeoTiffReaderTest.class, "geo.tiff");
-        FileInputStream is = new FileInputStream(rasterfile);
-
-        // Read coverage
-        GeoTiffReader reader = new GeoTiffReader(is);
-        GridCoverage2D gridCoverage = reader.read(null);
-
-        assertTrue(gridCoverage != null && gridCoverage.getNumSampleDimensions() == 1);
+        GeoTiffReader reader = null;
+        
+        try(FileInputStream is = new FileInputStream(rasterfile)) {
+            // Read coverage
+            reader = new GeoTiffReader(is);
+            GridCoverage2D gridCoverage = reader.read(null);
+    
+            assertTrue(gridCoverage != null && gridCoverage.getNumSampleDimensions() == 1);
+            
+            gridCoverage.dispose(true);
+        }
+        finally {
+            if(reader != null) {
+                reader.dispose();
+            }
+        }
     }
 
     /**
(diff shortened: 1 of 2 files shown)


modules/plugin/geotiff/src/test/java/org/geotools/gce/geotiff/GeoTiffReaderTest.java
```diff
@@ -1150,4 +1151,34 @@ public void testFileDoesNotExist() throws Throwable {
...
+    /**
+     * The GeoTiffReader should be able to read from an InputStream
+     */
+    @Test
+    public void testCanReadInputStream() throws IOException {
+        File rasterfile = TestData.file(GeoTiffReaderTest.class, "geo.tiff");
+        FileInputStream is = new FileInputStream(rasterfile);
+
+        // Read coverage
+        GeoTiffReader reader = new GeoTiffReader(is);
```
aaime: Always remember to dispose of the reader like in the other tests, there might be build issues in Windows otherwise (due to file locking)

modules/plugin/geotiff/src/test/java/org/geotools/gce/geotiff/GeoTiffReaderTest.java
```diff
@@ -1150,4 +1151,34 @@ public void testFileDoesNotExist() throws Throwable {
...
         }
     }
+
+    /**
+     * The GeoTiffReader should be able to read from an InputStream
+     */
+    @Test
+    public void testCanReadInputStream() throws IOException {
+        File rasterfile = TestData.file(GeoTiffReaderTest.class, "geo.tiff");
+        FileInputStream is = new FileInputStream(rasterfile);
```
aaime: The stream should be closed at the end of the test (best to use a try/with/resources)

modules/plugin/geotiff/src/test/java/org/geotools/gce/geotiff/GeoTiffReaderTest.java
```diff
@@ -1150,4 +1151,34 @@ public void testFileDoesNotExist() throws Throwable {
...
+        GridCoverage2D gridCoverage = reader.read(null);
+
+        assertTrue(gridCoverage != null && gridCoverage.getNumSampleDimensions() == 1);
+    }
+
+    /**
+     * The GeoTiffReader should be able to read from an ImageInputStream
+     */
+    @Test
+    public void testCanReadImageInputStream() throws IOException {
```
aaime: Same comments as above apply here.

aaime: Not sure either, I've just restarted the build, let's see what it does ;-)

dvntucker:
Looks OK to me. Unsure whether this should be squashed or not. Not really something that would be backported for the most part.

dvntucker:
@dromagnoli @simboss Any issues with merging this? I tested it locally and it works for me. Wanted to check with module maintainers first, which is technically our policy

simboss:
we'll harmonized and merge before EOW

dromagnoli:
Hi @dvntucker. No issues on my side. I'm going to squash-merge that.

PR #1343, opened 2016-10-10
modify numberFormat to take an optional language string (GEOT-5543)

ianturton:
Fix for [GEOT-5543](https://osgeo-org.atlassian.net/browse/GEOT-5543) takes current locale into account when formatting the number. This is more likely to be what the user is expecting.

### DIFF
--- a/modules/library/main/src/main/java/org/geotools/filter/function/FilterFunction_numberFormat.java
+++ b/modules/library/main/src/main/java/org/geotools/filter/function/FilterFunction_numberFormat.java
@@ -16,20 +16,24 @@
  */
 package org.geotools.filter.function;
 
-
-import static org.geotools.filter.capability.FunctionNameImpl.*;
+import static org.geotools.filter.capability.FunctionNameImpl.parameter;
 
 import java.text.DecimalFormat;
 import java.text.DecimalFormatSymbols;
+import java.util.Arrays;
+import java.util.HashSet;
 import java.util.Locale;
+import java.util.logging.Logger;
 
 import org.geotools.filter.FunctionExpressionImpl;
 import org.geotools.filter.capability.FunctionNameImpl;
+import org.geotools.util.logging.Logging;
 import org.opengis.filter.capability.FunctionName;
+import org.opengis.filter.expression.Expression;
 
 /**
- * Formats a number into a string given a certain pattern (specified in the format accepted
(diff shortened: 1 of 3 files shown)


modules/library/main/src/main/java/org/geotools/filter/function/FilterFunction_numberFormat.java
```diff
@@ -43,7 +43,6 @@
             parameter("format", String.class),
             parameter("number", Number.class));
     
-    public static DecimalFormatSymbols decimalFormatSymbols = DecimalFormatSymbols.getInstance(Locale.ENGLISH);
```
aaime: There was a commit in the past dedicated to just set the English locale... it seems that you and @sampov2  have different needs there. If used locally for a test, the local locale seems to make more sense, if used inside a web application (e.g., GeoServer) using english seems to make more sense instead (e.g., GML is not using the default locale, it's using the English one). 
See: https://github.com/geotools/geotools/commit/d4a43a347ea81d70c1cdc5261a73186490997ce8

See also:
https://osgeo-org.atlassian.net/browse/GEOT-4987
https://osgeo-org.atlassian.net/browse/GEOT-4986

Time to add that third parameter to specify the intended locale?

ianturton:
Updated to take an optional third parameter containing a language code to set the locale.

modules/library/main/src/main/java/org/geotools/filter/function/FilterFunction_numberFormat.java
```diff
@@ -52,32 +66,47 @@ public FilterFunction_numberFormat() {
...
+                Expression second = getExpression(2);
+
+                localeString = second.evaluate(feature, String.class);
+            }
+        } catch (Exception e) // probably a type error
+        {
+            throw new IllegalArgumentException(
+                    "Filter Function problem for function NumberFormat argument #2 - expected type String");
+        }
+        if (localeString != null && !localeString.isEmpty() && languages.contains(localeString)) {
```
aaime: So if the locale specified is unknown the code will ignore it right? Are the other functions doing the same, graceful degradation instead of early fail?

ianturton: The convention does seem to be to throw an exception - I've added one.

aaime:
+1

PR #1345, opened 2016-10-12
GEOT-5541: Create a function to return number of complex attributes

riniangreani:
https://osgeo-org.atlassian.net/browse/GEOT-5541

So I made the function in main, but the test resides in gt-complex because it needs some dependencies there. There are extra tests in geoserver with app-schema config, but I'll push that after this one is accepted. Thanks.


bencaradocdavies:
@riniangreani that looks fine to me. I am waiting until GeoServer 2.10.x is branched before merging anything on GeoTools master so we get cross-project continuous integration. Should happen Real Soon Now.

riniangreani:
Thanks Ben!

PR #1350, opened 2016-10-14
[GEOT-5542]: Updating GT Pom to depend on ImageIO-Ext 1.1.16

dromagnoli:


aaime:
I believe there is a communication gap here (to decide if merge right away or not). Is 1.1.16 a bugfix release or does it contain possibly de-stabilizing changes?

dromagnoli:
No destabilizing changes.
I have just prepared the PR during the release creation and I have waited for the release to be finished in order to avoid issues in the middle such as boundless repository artifacts not being cached yet, and JIRA GT version tags mismatches. Going to merge right now

PR #1355, opened 2016-10-16
Cql examples refactor

dnltsk:
cleaned up ECQLExamples and CQLExamples
- not using using named keys instead of magic numbers
- unify space indents


tbarsballe:
Some build failures here, looks like some syntax errors in ECQLExamples.java and CQLExamples.java

e.g. 
```
(default-compile) on project docs: Compilation failure: Compilation failure:
[ERROR] /home/travis/build/geotools/geotools/docs/src/main/java/org/geotools/cql/ECQLExamples.java:[60,25] error: non-static variable USING_EXIT cannot be referenced from a static context
```

PR #1357, opened 2016-10-17
Numbers with decimals in ShapefileDataStore always less than 10

alexandreDavid:
With the current code, shapefile columns can't be bigger than 9.999... In Qgis, columns are truncated and information is modified.
My proposition is to split the chars number to define the same length before and after the decimal point.

### DIFF
--- a/modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/ShapefileDataStore.java
+++ b/modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/ShapefileDataStore.java
@@ -378,7 +378,8 @@ public class ShapefileDataStore extends ContentDataStore implements FileDataStor
                 header.addColumn(colName, 'N', Math.min(fieldLen, 33), 0);
             } else if (Number.class.isAssignableFrom(colType)) {
                 int l = Math.min(fieldLen, 33);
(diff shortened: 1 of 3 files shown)


ianturton:
We could consider going with GDAL/OGR's default of "Real (floating point) fields without an explicit width are treated as width 24 with 15 decimal places of precision."  http://www.gdal.org/drv_shapefile.html but for any serious work users should set the field lengths in the schema for getFieldLength to find.

alexandreDavid:
Fix of precision for Double and Float.

alexandreDavid:
I've done some modifications. Can you tell me if it's okay for you or if I must add something to my pull request?
Regards.

ianturton:
can you add a test case to prove it is doing what we expect

aaime:
I second @ianturton suggestion, overall it looks good but needs a test checking the expected number of decimals is respected. Tests are a requirement for any change (unless it can be proven they are impossible or just too hard to implement), see: https://github.com/geotools/geotools/blob/master/CONTRIBUTING.md

alexandreDavid:
I've added a unit test to check columns formats. More, others unit tests already check its.
Regards

alexandreDavid:
Hi, changed code is tested in totallity.
Regards.

PR #1359, opened 2016-10-22
[GEOS-7768] Filter function "greaterEqualThan" is not properly working

matchami:
gt/gte/lt/lte methods in StaticGeometry modified to delegate to FilterFactory2 as suggested by Andrea.  Have also tidied up code formatting.  Actual changes are lines 44-52 and 700-727

### DIFF
--- a/modules/library/main/src/main/java/org/geotools/filter/function/StaticGeometry.java
+++ b/modules/library/main/src/main/java/org/geotools/filter/function/StaticGeometry.java
@@ -19,20 +19,18 @@ package org.geotools.filter.function;
 
 
 
+import org.geotools.factory.CommonFactoryFinder;
+import org.opengis.filter.FilterFactory2;
+
 import com.vividsolutions.jts.algorithm.MinimumBoundingCircle;
 import com.vividsolutions.jts.algorithm.MinimumDiameter;
 import com.vividsolutions.jts.geom.Geometry;
 import com.vividsolutions.jts.geom.GeometryCollection;
 import com.vividsolutions.jts.geom.LineString;
+import com.vividsolutions.jts.geom.OctagonalEnvelope;
 import com.vividsolutions.jts.geom.Point;
 import com.vividsolutions.jts.geom.Polygon;
-import com.vividsolutions.jts.geom.OctagonalEnvelope;
 import com.vividsolutions.jts.io.WKTReader;
-
-
-
-
-
 
 /**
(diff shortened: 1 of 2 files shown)


aaime:
The diff is not readable bcause the entire file got reformatted. The contribution instructions specifically warn against that:

> Please make sure you're following the coding conventions, and otherwise avoid any reformats to the existing code, as they make it harder to review your changes.

matchami:
Hi Andrea,

Try this:  git diff -w master filterfix -- modules/library/main/src/main/java/org/geotools/filter/function/StaticGeometry.java

with regard to the decision to tidy up the formatting, my apologies I genuinely mis-remembered the instructions on this.  I think I confused the bit saying please DO put in missing copyright statements with the bit saying please DON'T fix formatting.  I mistakenly thought that bringing old code up to the coding standards would be a good thing and I think as I was pulled up on code formatting a previous pull request I was overly conscious that it was there!  I will not make that mistake again.  In this case, do you genuinely want me to put the non-compliant code formatting back in and resubmit the pull request?  It seems like a bit of a backwards step ??? but I can do that if you want?

Thanks
Iain

matchami:
BTW  you should also note that the paragraph you have quoted from the code contribution guidelines actually supports my changes: "and OTHERWISE avoid any reformats to the existing code."  The implication here (which is why I was confused) is that reformats to implement coding standards are desired.  

I do realise that the follow sentence suggests a separate commit for these changes - which clearly I did not get right.

Thanks

aaime:
Reformats are not welcomed, they are tolerated if and only if:

- They are done in a separate commit (can be in the same pull request) to allow reviewing the actual changes directly on GitHub (I don't know how you can think we're not aware that desktop and command line tools can ignore spaces, it just takes more time and the first quick check never goes to those lenghts)
- They are backported to all active branches (to avoid conflicts)

The commit guidelines are often written by non native English speakers, if they are not clear enough feel free to clarify them in a separate pull request.

aaime:
Btw, if you are interested in how the pull request get reviewed, there is a guide here that can help a bit:
https://github.com/geoserver/geoserver/wiki/Pull-request-review-guide
It's for GeoServer, but most of it applies to GeoTools as well.

matchami:
Thanks Andrea , that's really helpful.  Also I will submit a proposed
change to the contribution guide as you suggest.

Cheers
Iain

PR #1319, opened 2016-09-22
[GEOT-5526] Apps-Schema: Native Encoding  of Nested Filters

ridethepenguin:
Related JIRA issue:
https://osgeo-org.atlassian.net/browse/GEOT-5526

Related ML discussion:
http://osgeo-org.1560.x6.nabble.com/App-Schema-Encoding-of-Filters-on-Nested-Attributes-td5285165.html

Unit tests and docs will be added in a separate GeoServer PR.

PR #1362, opened 2016-10-25
[GEOT-5526] App-Schema: Native Encoding of Nested Filters

ridethepenguin:
Straightforward backport of #1319 to Geotools 14.x.


bencaradocdavies:
@ridethepenguin this is an unsupported branch, but I do not see the harm in your extra efforts.

PR #1364, opened 2016-10-25
[GEOT-5556] Add mosaic config to config map when updating mosaic

dvntucker:
It looks like when the mosaic is being updated, the MosaicConfiguration is never put in the "configurations" map. During normal operation this isn't an issue, but with a reprojecting mosaic this is needed because it needs info from that config.


dvntucker:
Woops, pulled the trigger on this too soon. Forgot the test.

PR #1365, opened 2016-10-26
Added a workaround for the reprojecting mosaic test failing

dvntucker:
This will skip the test for now when the expected CRS isn't found. Hopefully will add proper fix in the next release.


aaime:
This will disable the heterogeneous test constantly, when it's passing the crs is 3857. The mosaic for that test is different than the other two.

dvntucker:
Ok, added an "expectedCRS" parameter to the tests to compensate.

jodygarnett:
Thanks, waiting for build to complete

jodygarnett:
Backported to 16.x

PR #1367, opened 2016-10-29
[GEOT-5558][GEOS-7524] Ignore no matching properties

nmco:
When creating a sub type using DataUtilies if one of the provided properties doesn't map to an attribute descriptor a NULL attribute descriptor will be added to the new sub type attributes descriptors list.

Since the typical code assumes that the attributes descriptors of a feature type are not NULL this may provoke a NULL pointer exception. A situation like this happens for example when a provided property is valid but the corresponding attribute descriptor is hidden by a security rule (read only attributes for example).

This pull request fix this issue and adds a test case.

Associated issues:
- https://osgeo-org.atlassian.net/browse/GEOT-5558
- https://osgeo-org.atlassian.net/browse/GEOS-7524


aaime:
Looks good to me

PR #1371, opened 2016-10-31
[GEOT-5561] Add an SLD vendor option to underline labels

nmco:
This pull request adds a ``underlineText`` vendor option that will instruct the renderer to underline labels. The underline will work like a typical word processor text underline. The thickness and position of the underline will be defined by the font and color will be the same as the text. Spaces will also be underlined.

Note that CSS and YSLD modules have also been updated to take in account this new vendor option.

To see the expected result we can check the GeoServer documentation pull request:
https://github.com/geoserver/geoserver/pull/1934

Associated issue:
https://osgeo-org.atlassian.net/browse/GEOT-5561


modules/library/render/src/main/java/org/geotools/renderer/label/LabelPainter.java
```diff
@@ -605,27 +661,113 @@ public void paintCurvedLabel(LineStringCursor cursor) {
...
+        // compute the advance based on the first char of the label
+        GlyphVector glyphVector = line.getComponents().get(0).getGlyphVector();
+        double advance = glyphVector.getGlyphMetrics(0).getAdvance() * 0.5f;
+        // extract from the linestring the portion associated with the layer
+        LineString labelLineString = cursor.getSubLineString(startOrdinate - advance, endOrdinate - advance);
+        // compute the underline linestring
+        LiteShape underlineLineString = computeCurvedUnderline(labelLineString, metrics);
+        if (drawingHalo) {
+            // draw the underline halo
+            graphics.draw(underlineLineString);
```
aaime: What guarantees (here and in the other equivalent method for straight line) that the current stroke is the right color and thickness?

aaime: Btw, I am looking at a local word processor, the underline seem to be different... like, thinner... could just be that java does it differently, but the code here does not seem to be setting the underline thickness in the stroke at all?

image

nmco: Regarding the first question:
Both the methods when drawing the halo are invoked in a situation where configureHalo() has already be invoked, this was on propose to avoid resetting all the options again, but this is not very clear adding a comment about this.

nmco: Regarding the second question:
This is my fault, the original underline looked very thin so I perform some tests doubling the original thickness:
```
graphics.setStroke(new BasicStroke(metrics.getUnderlineThickness() * 2));
```

modules/library/render/src/test/java/org/geotools/renderer/lite/LabelUnderlineTest.java
```diff
@@ -0,0 +1,69 @@
...
+import java.awt.*;
+import java.awt.image.BufferedImage;
+import java.io.File;
+
+import static java.awt.RenderingHints.*;
+
+/**
+ * Tests labels underling.
+ */
+public class LabelUnderlineTest {
```
nmco: Ha! I used fonts I found on other tests, switching to Vera font.

modules/library/render/src/test/resources/org/geotools/renderer/lite/test-data/underlineStyle.sld
```diff
@@ -0,0 +1,146 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
nmco: Done.

nmco:
Hi @aaime thanks for the review ! Update according to your suggestions, answered you questions and updated documentation.

aaime:
Thanks, looks good. Waiting for the Travis build to finish (if you have commit rights, please merge yourself when the Travis build is green)

bencaradocdavies:
@nmco, for future reference, there is also a little extra code in the setup that pulls in the Vera font that is included in the test source code. Some existing test classes have it. OpenJDK ships with no fonts and system replacements are often quite different to the Oracle ones. Using Vera is a robust solution.

nmco:
Hi @bencaradocdavies, yup Andrea told me I already updated the tests to use Vera, thanks for the feedback :)

bencaradocdavies:
Sure @nmco, my point is that future test classes need to register Vera in their setup.

nmco:
Ah! correct ... sorry is end of the day and my brain is tired.

bencaradocdavies:
@nmco beginning of the day here and mine is not yet fully awake.  :-)

nmco:
Test updated to correctly setup the Vera font, let's wait for Travis again :)

nmco:
Travis build is green, so merging.

bencaradocdavies:
@nmco this breaks the build on OpenJDK because the CSS font-family in ```underlineStyle.sld``` should be "Bitstream Vera Sans" not "Vera". This causes some system-dependent font to be used.

PR #1372, opened 2016-11-02
[GEOT-5555] added weightClusterPosition, default value=false (master)

CozC:
Added an option to gs:PointStacker, called weightClusterPostion, default value=false (not to break the original functionality) 
The flag computes the cluster position as a rough average of all the points which were clustered.


modules/unsupported/process-feature/src/main/java/org/geotools/process/vector/PointStackerProcess.java
```diff
@@ -143,11 +144,16 @@ public SimpleFeatureCollection execute(
             normalize = argNormalize;
         }
 
+        boolean weightClusterPosition=false;
+        if (argWeightClusterPosition!=null){
```
aaime: Formatting rules not respected. Can you format the lines you added/changed (and only those) based on the official formatting guidelines? See http://docs.geotools.org/latest/developer/conventions/code/style.html

modules/unsupported/process-feature/src/main/java/org/geotools/process/vector/PointStackerProcess.java
```diff
@@ -267,7 +273,7 @@ private Coordinate getStackedPointLocation(PreserveLocation preserveLocation, St
                     stkPt = new StackedPoint(indexPt, new Coordinate(centreX, centreY));
                     stackedPts.put(stkPt.getKey(), stkPt);
                 }
-                stkPt.add(pout);
+                stkPt.add(pout,weightClusterPosition);
```
aaime: Is there a case where one would want to add one point as weighted and another as nearest, into the same StackedPoint? If not, maybe better to add the flag as a field in StackedPoint, and set it via a setter. The add method interface would not change, only its implementation.

modules/unsupported/process-feature/src/test/java/org/geotools/process/vector/PointStackerProcessTest.java
```diff
@@ -214,6 +219,41 @@ public void testReprojected() throws NoSuchAuthorityCodeException, FactoryExcept
         checkResultPoint(result, new Coordinate(-121.813201, 48.777343), 2, 2, null, null);
     }
     
+    @SuppressWarnings("deprecation")
+	@Test
```
aaime: Formatting issue here (using tab?)

jodygarnett:
@CozC are you in position to update this pull request?

CozC:
Salve All, 

I finished with it, the flag added is in production in my personal service (custom compole over geosercr 2.8.5)

Did I made a mistake, or I was missing something in the updating procedure? 


Have a nice evening. 


...

tbarsballe:
This PR currently has merge conflicts that need to be resolved before it can be merged.

CozC:
Salve,

I will take a look asap.

Have a nice day. 

...

lbesnard:
This option seems like a great feature. It would be great to see it being resolve/merge

CozC:
Salve,

The feature is already on latest geotools version (just checked)
Upon the merge, it was a mixup but the code went through (I do not  
remember the number)

https://osgeo-org.atlassian.net/browse/GEOT-5555

Have a nice day.

PS: The parameter was  a small step towards a better clustering, it has  
some issues with overlapping clusters, due to the fact that the original  
implementation had the idea to create a grid and then, cluster points  
appertaining to that specific grid. If the grid splits an agglomeration of  
points, then the overlapping happens. To fix this is, you can increase the  
grid size or decrease the graphical point used to represent the cluster  
(Eg. a circle with smaller radius and some tweaking on sld rules)

I do have on my board an implementation closest to Openlayers clustering,  
but it is on pending due to time constraints.



On Mon, 19 Feb 2018 08:25:15 +0200, lbesnard <notifications@github.com>  
wrote:

>
> This option seems like a great feature. It would be great to see it  
> being resolve/merge
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub, or mute the thread.



-- 
Cosmin Cioranu

PR #1373, opened 2016-11-03
[GEOT-5563] DefaultRenderingExecutor stops working after a layer fail

matchami:
After a rendering failure on any layer, DefaultRenderingExecutor will not issue any more rendering requests, even if the offending layer is removed from the MapContent.  This is caused because the task counter is not decremented on a failure. 

DefaultRenderingExecutor modified to decrement the tasks latch on both fail and completion, thus clearing it's own internals and allow subsequent submissions to be rendered correctly.


aaime:
Looking at the recent module history maybe @jodygarnett or @ianturton could have a look at this one

ianturton:
looks good to me, is this a small enough change to not need a contributor agreement?

aaime:
I believe it does, there is only one non test file modified.

PR #1374, opened 2016-11-03
Amended contribution guidelines as suggested by Andrea

matchami:
After discussion with Andrea, I suggest the following wording change to the contribution agreement would better reflect the desires of the developers.


aaime:
+1 from me, matches how reformats have been managed so far (in a separate commit, and as an exception). Let's hear if there are objections.

Honestly I would not mind having a policy like qgis one, where a mis-formatted file breaks the build, however, they are using out of IDE tool to both format and check the files, while even using checkstyle, I believe we'd be having spurious reformats due to small IDE inconsistencies (we tried with jalopy in the past, I was the one promoting it, it failed quickly and spectacularly). Indeed when @dvntucker reformatted the image mosaic module recently, he had to resort to someone using Eclipse because IntelliJ would not be an exact match.

The other counter argument would be that we'd have to mass reformat everything, and that would create a wall on git blame usage (yes, one can blame the previous revision, but it's annoying to do). With a large code base in constant maintenance git blame is something that gets used daily (at least by those working on bug fixing and refactors of existing code).

bencaradocdavies:
I like to reformat code blocks that I change, and sometimes even whole methods if I make many changes. Eclipse allows me to select and just reformat the code I touch. This allows me to balance a little housekeeping and avoid adding more malformatted code without adding too much noise for the poor reviewer.

+1 for whole-file reformats being in a separate commit.

aaime:
@bencaradocdavies yep I do the same, reformat whatever blocks I change since that part is already in the diff anyways (and same for methods, if I end up changing them deeply)

PR #1378, opened 2016-11-07
[GEOT-5567] Raster symbol overlap behavior now correctly set

robward-scisys:


aaime:
This pull request is trading a dynamic element, an expression, which can thus use environment variable replacement, with a static value. This sounds like a regression.
However I did not have time to look at all calling points and general usage (I'm on a laptop I normally don't use for development), can you show if I just have the wrong impression?

robward-scisys:
Of the two member variables choosing the Expression is not a problem.

However it should be noted that in RasterSymbolizerImpl the setOverlap and getOverlap methods to access the Expression member data are marked as deprecated.

modules/library/main/src/main/java/org/geotools/styling/RasterSymbolizerImpl.java
```diff
@@ -255,7 +247,12 @@ public Expression getOpacity() {
      */
     @Deprecated
     public Expression getOverlap() {
```
jodygarnett: Why is getOverlap() deprecated; what is the replacement ... checking history this method has been deprecated since creation.

Looking at GeoServer docs this is marked as not implemented - http://docs.geoserver.org/latest/en/user/styling/sld/reference/rastersymbolizer.html#overlapbehavior

The specification matches the javadoc...

modules/library/main/src/main/java/org/geotools/styling/RasterSymbolizerImpl.java
```diff
@@ -64,7 +63,6 @@ public RasterSymbolizerImpl(FilterFactory factory, Description desc, String name
         super(name, desc, (String) null, uom);
         this.filterFactory = factory;
         this.opacity = filterFactory.literal(1.0);
-        this.overlap = filterFactory.literal(OverlapBehavior.RANDOM);
```
jodygarnett: random is listed as the default in the symbology encoding specification - are you leaving this value as null (as not set).

modules/library/main/src/main/java/org/geotools/styling/RasterSymbolizerImpl.java
```diff
@@ -255,7 +247,12 @@ public Expression getOpacity() {
      */
     @Deprecated
     public Expression getOverlap() {
-        return overlap;
+    	OverlapBehavior overlap = getOverlapBehavior();
+    	if(overlap == null)
+    	{
```
jodygarnett: Code formatting, but this answers my question on leaving that value as null.

modules/library/main/src/main/java/org/geotools/styling/RasterSymbolizerImpl.java
```diff
@@ -425,10 +422,12 @@ public void setOpacity(Expression opacity) {
      */
     @Deprecated
     public void setOverlap(Expression overlap) {
-        if (this.overlap == overlap) {
+        if(overlap == null)	{
             return;
         }
-        this.overlap = overlap;
+
+        OverlapBehavior overlapBehavior = OverlapBehavior.valueOf(overlap.toString());
```
jodygarnett: This assumes the expression is a literal, the specification says this must evaluate to one of four literals.

To fix this use overlap.evaulate( null, String.class ) - it will work for literal, env function etc...

tbarsballe: This needs some validation - since Expression can be pretty much anything, and OverlapBehavior only really has 4 valid values, an IllegalArgumentException should be thrown in the case where the Expression is not a valid value

bencaradocdavies:
@robward-scisys are you able to address the concerns raised above?

robward-scisys:
I've made the updates and merged in the latest master.

modules/library/main/src/test/java/org/geotools/styling/visitor/DuplicatingStyleVisitorTest.java
```diff
@@ -304,6 +305,43 @@ public void testRasterSymbolizerWithUOM() throws Exception {
...
+        rasterSymbLiteral.accept(visitor);
+        RasterSymbolizer cloneLiteral = (RasterSymbolizer) visitor.getCopy();
+
+        assertEquals(cloneLiteral.getOverlapBehavior(), rasterSymbLiteral.getOverlapBehavior());
+        assertEquals(cloneLiteral.getOverlap(), rasterSymbLiteral.getOverlap());
+
+        // Try with invalid expression string
+        RasterSymbolizer rasterSymb3 = sf.createRasterSymbolizer();
+        try
+        {
```
tbarsballe: Code formatting - open braces should be on the same line as the try. Same with the catch statements below.

robward-scisys:
Squashed all the commits into 1 commit.

PR #1379, opened 2016-11-08
[GEOT-5568] Avoid deadlocks on schema loading

awaterme:
Pull request for [GEOT-5568](https://osgeo-org.atlassian.net/browse/GEOT-5568)

### DIFF
--- a/modules/extension/xsd/xsd-core/src/main/java/org/geotools/xml/Schemas.java
+++ b/modules/extension/xsd/xsd-core/src/main/java/org/geotools/xml/Schemas.java
@@ -16,16 +16,17 @@
  */
 package org.geotools.xml;
 
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
-import java.io.OutputStream;
 import java.net.MalformedURLException;
 import java.net.URL;
-import java.net.URLDecoder;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedList;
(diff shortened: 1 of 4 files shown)


modules/extension/xsd/xsd-core/src/main/java/org/geotools/xml/Schemas.java
```diff
@@ -341,6 +354,55 @@ public static final XSDSchema parse(String location, ResourceSet resourceSet)
...
+    /**
+     * Fetches the contents of the URI into a local {@link InputStream}.
+     * 
+     * @param uri
+     * @param options
+     * @param response
+     * @return A local resource
+     * @throws IOException
+     */
+    private static InputStream readUriResource(URI uri, ResourceSet resourceSet, Map<?, ?> response)
```
aaime: Could you rework this one so that it uses try-with-resources and returns a byte[], which makes evident it's stored in memory and there are no leftovers to close?

modules/extension/xsd/xsd-core/src/main/java/org/geotools/xml/Schemas.java
```diff
@@ -316,12 +316,25 @@ public static final XSDSchema parse(String location, ResourceSet resourceSet)
...
+        // if remote schema is created on same JVM and synchronizes on Schemas, too.
+        Map<Object, Object> options = resourceSet.getLoadOptions();
+        Map<?, ?> response = getOrCreateResponseFrom(options);
+        InputStream inputStream = readUriResource(uri, resourceSet, response);
+                
         // schema building has effects on referenced schemas, it will alter them -> we need 
         // to synchronize this call so that only one of these operations is active at any time
         synchronized(Schemas.class) {
-            xsdMainResource.load(resourceSet.getLoadOptions());
+            try {
```
aaime: try-with-resources here please

awaterme:
Hi Andrea,

thanks for the review! Regarding:
> I'm a bit worried about it being memory bound, but it can well be that the previous code was memory bound as well (did you check, or do you know?).

The current implementation uses the classic SAX parser with InputSource/InputStream, so the proposed solutions is probably less memory efficient. I also though about this, but I came to the conclusion that the additional memory used for the byte[] is probably negligible compared to the data structures created by the parser. I also considers writing the data to a temp file, but it would be pity to introduce such an overhead. So in order to resolve the problem I did not see a better solution.

> Added a couple of observations on the code.

Adjusted the proposal according to your suggestions.

Also tried to squash the new commit into the prior one to keep the history clean, but I git refuses pushing. 

Best regards,
Andreas

awaterme:
Sorry, I missed to replace one classic try/catch, added one more commit.

PR #1380, opened 2016-11-11
Name of isDirrty method seems like typing mistake

hgryoo:
"boolean isDirty(int index)" method already exists.
it appears that method overloading is intended.
also, fixed spacing.

### DIFF
--- a/modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCFeatureReader.java
+++ b/modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCFeatureReader.java
@@ -524,12 +524,12 @@ public class JDBCFeatureReader implements  FeatureReader<SimpleFeatureType, Simp
             //get the primary key, ensure its not contained in the values
             key = dataStore.getPrimaryKey(featureType);
             int count = md.getColumnCount();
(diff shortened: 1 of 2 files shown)


modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCFeatureReader.java
```diff
@@ -704,7 +704,7 @@ public boolean isDirty(int index) {
             return dirty[index];
         }
 
-        public boolean isDirrty(String name) {
+        public boolean isDirty(String name) {
```
aaime: Given the method is public, it cannot be renamed as it would break clients. Instead, create a new method with the right name and deprecate the old one, making it delegate to the new one.

hgryoo: Ok, I changed.

aaime:
Normally we'd like to have a ticket in Jira and the commit message referring to it, e.g. https://github.com/geotools/geotools/commit/c2b32f91a083834469e81027ba07c5ba0a2d87cd
but this one is a trivial one, I'll merge it directly.

PR #1381, opened 2016-11-11
[GEOT-5571]createCurve method always makes cycle

hgryoo:
https://osgeo-org.atlassian.net/browse/GEOT-5571

### DIFF
--- a/modules/library/referencing/src/main/java/org/geotools/geometry/GeometryBuilder.java
+++ b/modules/library/referencing/src/main/java/org/geotools/geometry/GeometryBuilder.java
@@ -2,7 +2,7 @@
  *    GeoTools - The Open Source Java GIS Toolkit
  *    http://geotools.org
  * 
- *    (C) 2002-2008, Open Source Geospatial Foundation (OSGeo)
+ *    (C) 2002-2017, Open Source Geospatial Foundation (OSGeo)
  *
  *    This library is free software; you can redistribute it and/or
  *    modify it under the terms of the GNU Lesser General Public
@@ -273,23 +273,45 @@ public class GeometryBuilder {
 		// - Start and end params for the CurveSegments will be set
 		return getPrimitiveFactory().createCurve(segments);
 	}
-    public Curve createCurve(PointArray points) throws MismatchedReferenceSystemException, MismatchedDimensionException {
+	
+    @Deprecated
+    public Curve createCurve(PointArray points)
+            throws MismatchedReferenceSystemException, MismatchedDimensionException {
+        return createCurve(points, true);
+    }
+
+    public Curve createCurve(PointArray points, boolean closed)
+            throws MismatchedReferenceSystemException, MismatchedDimensionException {
         if (points == null)
             throw new NullPointerException("Points are required to create a curve");
+        if (points.size() < 2)
+            throw new IllegalArgumentException(
+                    "At least two points are required to create a curve");
 
         // A curve will be created
         // - The curve will be set as parent curves for the Curve segments
         // - Start and end params for the CurveSegments will be set
-        List/*<LineSegment>*/ segmentList = new ArrayList/*<LineSegment>*/();
-        for( int i=0; i<points.size();i++){
+        List/* <LineSegment> */ segmentList = new ArrayList/* <LineSegment> */();
+        for (int i = 0; i < points.size() - 1; i++) {
             int start = i;
-            int end = (i+1)%points.size();
-            DirectPosition point1 = points.getDirectPosition( start, null );
-            DirectPosition point2 = points.getDirectPosition( end, null );
-            LineSegment segment = createLineSegment( point1, point2 );
-            segmentList.add( segment );
+            int end = i + 1;
+            DirectPosition point1 = points.getDirectPosition(start, null);
+            DirectPosition point2 = points.getDirectPosition(end, null);
+            LineSegment segment = createLineSegment(point1, point2);
+            segmentList.add(segment);
         }
-        return getPrimitiveFactory().createCurve( segmentList );
+
+        if (closed) {
+            LineSegment first = (LineSegment) segmentList.get(0);
+            LineSegment last = (LineSegment) segmentList.get(segmentList.size() - 1);
+
+            if (!first.getStartPoint().equals(last.getEndPoint())) {
+                LineSegment segment = createLineSegment(last.getEndPoint(), first.getStartPoint());
+                segmentList.add(segment);
+            }
+        }
+
+        return getPrimitiveFactory().createCurve(segmentList);
     }
 
 	/**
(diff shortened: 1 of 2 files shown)


aaime:
I see a legitimate failure in the build:

{code}
Tests run: 5, Failures: 0, Errors: 2, Skipped: 0, Time elapsed: 0.986 sec <<< FAILURE! - in org.geotools.geometry.iso.primitive.SurfaceTest
testFastSurfaceFactory(org.geotools.geometry.iso.primitive.SurfaceTest)  Time elapsed: 0.013 sec  <<< ERROR!
java.lang.IllegalArgumentException: Start point of first element has to be at the same position as end point of last element
	at org.geotools.geometry.iso.primitive.RingImpl.checkConsistency(RingImpl.java:100)
	at org.geotools.geometry.iso.primitive.RingImpl.<init>(RingImpl.java:84)
	at org.geotools.geometry.iso.primitive.PrimitiveFactoryImpl.createRing(PrimitiveFactoryImpl.java:310)
	at org.geotools.geometry.GeometryBuilder.createRing(GeometryBuilder.java:420)
	at org.geotools.geometry.GeometryBuilder.createSurfaceBoundary(GeometryBuilder.java:450)
	at org.geotools.geometry.GeometryBuilder.createSurfaceBoundary(GeometryBuilder.java:429)
	at org.geotools.geometry.iso.primitive.SurfaceTest.testFastSurfaceFactory(SurfaceTest.java:349)

testFastSurfaceBuilder(org.geotools.geometry.iso.primitive.SurfaceTest)  Time elapsed: 0.012 sec  <<< ERROR!
java.lang.IllegalArgumentException: Start point of first element has to be at the same position as end point of last element
	at org.geotools.geometry.iso.primitive.RingImpl.checkConsistency(RingImpl.java:100)
	at org.geotools.geometry.iso.primitive.RingImpl.<init>(RingImpl.java:84)
	at org.geotools.geometry.iso.primitive.PrimitiveFactoryImpl.createRing(PrimitiveFactoryImpl.java:310)
	at org.geotools.geometry.GeometryBuilder.createRing(GeometryBuilder.java:420)
	at org.geotools.geometry.GeometryBuilder.createSurfaceBoundary(GeometryBuilder.java:450)
	at org.geotools.geometry.GeometryBuilder.createSurfaceBoundary(GeometryBuilder.java:429)
	at org.geotools.geometry.iso.primitive.SurfaceTest.testFastSurfaceBuilder(SurfaceTest.java:324)
{code}

So it seems there is code needing the current behavior. Thinking out loud, how about you deprecate createCurve and replace it with a createOpenCurve and createClosedCurve, or even a new createCurve(..., boolean closed) version that allows to choose if the curve should be open or closed?

hgryoo:
Oh, I'm sorry. I checked the build But it was cached. 
failures are caused in createRing method. how about making "isStrict" boolean member variable in Geometrybuilder class? when the "isStrict" is false value, createRing method repairs coordinates in order to ensure valid ring.

following code is similar to what I thinking.
readCoordinateSequenceRing method in org.geotools.data.postgis.WKBReader class.
```
private CoordinateSequence readCoordinateSequenceRing(int size) throws IOException {
        CoordinateSequence seq = readCoordinateSequence(size);
        if (isStrict)
            return seq;
        if (CoordinateSequences.isRing(seq))
            return seq;
        return CoordinateSequences.ensureValidRing(csFactory, seq);
}
```

hgryoo:
I implemented like what you said. The old createCurve is deprecated and make new createCurve method with isClosed, boolean parameter. If closed value is true and curve is not closed, new LineSegment is added into segment list in order to make curve be closed.

tbarsballe:
The change looks good, but we do require a test case for any changes, could you please add one?

modules/library/referencing/src/main/java/org/geotools/geometry/GeometryBuilder.java
```diff
@@ -424,8 +442,8 @@ public Solid createSolid(SolidBoundary boundary) throws MismatchedReferenceSyste
 		return getPrimitiveFactory().createSolid(boundary);
 	}
 
-    public SurfaceBoundary createSurfaceBoundary(PointArray points ) throws MismatchedReferenceSystemException, MismatchedDimensionException {
-        Curve curve = createCurve( points );
+    public SurfaceBoundary createSurfaceBoundary(PointArray points) throws MismatchedReferenceSystemException, MismatchedDimensionException {
+    	Curve curve = createCurve( points );
```
bencaradocdavies: No tabs in the source code, please. Please see the [style guide](http://docs.geotools.org/latest/developer/conventions/code/style.html). We also provide an Eclipse ```formatter.xml```.

modules/library/referencing/src/main/java/org/geotools/geometry/GeometryBuilder.java
```diff
@@ -273,22 +274,39 @@ public Curve createCurve(List segments) throws MismatchedReferenceSystemExceptio
...
             segmentList.add( segment );
         }
+        
+        if(closed) {
+        	LineSegment first = (LineSegment) segmentList.get(0);
+        	LineSegment last = (LineSegment) segmentList.get(segmentList.size() - 1);
+        	
+        	if(!first.getStartPoint().equals(last.getEndPoint())) {
+        		LineSegment segment = createLineSegment(last.getEndPoint(), first.getStartPoint());
+        		segmentList.add( segment );
```
bencaradocdavies: No spaces inside round brackets.

hgryoo: Thank you for your comment. I found that there are many broken indentation(e.g. tabs are used in processBoundsToRing method) and added space inside round brackets. Do we need to fix it?

hgryoo:
I added unit tests in CurveTest class. But it looks awkward for me. Because GeometryBuilder class is inside of gt-referencing module and it doesn't have own unit test class in the maven module. Although the CurveTest class uses the GeometryBuilder class, actually it doesn't test methods in GeometryBuilder. it just bring factories from the builder.

tbarsballe:
That does seem a bit awkward. You could add a new test class for GeometryBuilder in the gt-referencing module to do direct unit testing of GeometryBuilder methods.
The CurveTest tests are still useful for integration testing.

One other request - could you rebase your commits into a single commit? This makes backporting easier.

modules/library/referencing/src/main/java/org/geotools/geometry/GeometryBuilder.java
bencaradocdavies: I like to reformat code that I change. I use the official GeoTools Eclipse formatter configuration ```build/eclipse/formatter.xml```. I select the block and use Ctrl-Shift-F in Eclipse to limit reformatting to the changed code (or whole method if I change most of it). I do not reformat entire files as this makes the changes too hard to read.

hgryoo:
I reformatted codes that I change and rebased into a single commit. I think it would be better to make new issue for testing of GeometryBuilder methods.
https://osgeo-org.atlassian.net/browse/GEOT-5710

PR #1382, opened 2016-11-14
[GEOT-5570] Scaling up a small image generates a output image with a high number of small tiles

aaime:


dromagnoli:
Looks good to me.

PR #1385, opened 2016-11-16
RescaleStyleVisior visit(Graphic) method now copies the anchor point data

robward-scisys:


modules/library/main/src/main/java/org/geotools/styling/visitor/RescaleStyleVisitor.java
```diff
@@ -162,6 +163,12 @@ public void visit(Graphic gr) {
             displacementCopy = (Displacement) pages.pop();
         }
 
+        AnchorPoint anchorCopy = null;
+        if (gr.getAnchorPoint() != null) {
+            gr.getAnchorPoint().accept(this);
+            anchorCopy = (AnchorPoint) pages.pop();
+        }
```
jodygarnett: good catch, and this rescales the AnchorPoint as well (well the percentages stay the same).

PR #1386, opened 2016-11-17
[GEOT-5575] Relax the XML encoder to be able to encode nested complex types that don't respect the GML object-property model

nmco:
Associated issue:
https://osgeo-org.atlassian.net/browse/GEOT-5575

GeoServer side pull requests that add app-schema test case with complex features:
https://github.com/geoserver/geoserver/pull/1976

### DIFF
--- a/modules/extension/xsd/xsd-core/src/main/java/org/geotools/xml/Encoder.java
+++ b/modules/extension/xsd/xsd-core/src/main/java/org/geotools/xml/Encoder.java
@@ -74,6 +74,8 @@ import org.geotools.xml.impl.GetPropertyExecutor;
 import org.geotools.xml.impl.NamespaceSupportWrapper;
 import org.geotools.xml.impl.SchemaIndexImpl;
 import org.geotools.xs.XS;
+import org.opengis.feature.ComplexAttribute;
+import org.opengis.feature.Property;
 import org.picocontainer.MutablePicoContainer;
 import org.picocontainer.PicoContainer;
 import org.picocontainer.defaults.DefaultPicoContainer;
@@ -195,6 +197,9 @@ public class Encoder {
      * Logger logger;
      */
     private Logger logger;
+
+    /** if true the encoder may encode complex features that map to a complex type that is not GML valid **/
+    private boolean relaxed = Boolean.parseBoolean(System.getProperty("encoder.relaxed", "true"));
 
     /**
      * The configuration used by the encoder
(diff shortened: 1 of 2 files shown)


jodygarnett:
This will need a test case, see GEOS-7860

aaime:
Follows Ben's approach of testing most complex feature related stuff in GeoServer integrated tests.
Test is here: https://github.com/geoserver/geoserver/pull/1976

modules/extension/xsd/xsd-core/src/main/java/org/geotools/xml/Encoder.java
```diff
@@ -618,6 +623,45 @@ public void encode(Object object, QName name, OutputStream out)
         }
     }
 
+    /**
+     * Helper method that checks if the complex feature we want to encode maps
+     * to a complex type that respects the GML object-property model.
+     */
+    private boolean isNonStrippedNestedElement(Object next, XSDElementDeclaration element) {
```
bencaradocdavies: Method name should be ```isNonStripedNestedElement``` (i.e. striped not stripped).

modules/extension/xsd/xsd-core/src/main/java/org/geotools/xml/Encoder.java
```diff
@@ -618,6 +623,45 @@ public void encode(Object object, QName name, OutputStream out)
...
+        if (!(next instanceof ComplexAttribute)) {
+            return false;
+        }
+        ComplexAttribute complex = (ComplexAttribute) next;
+        // let's se if we need to encode this
+        Collection<Property> nestedProperties = complex.getProperties();
+        if (nestedProperties == null || nestedProperties.isEmpty()) {
+            return false;
+        }
+        // let's if all the properties have the same type
```
bencaradocdavies: "let's if" -> "let's see if"

modules/extension/xsd/xsd-core/src/main/java/org/geotools/xml/Encoder.java
```diff
@@ -743,8 +787,15 @@ public void encode(Object object, QName name, ContentHandler handler)
                             }
                         }
                         else {
-                            //add the next object to be encoded to the stack
-                            encoded.push(new EncodingEntry(next, element,entry));                            
+                            if (relaxed && isNonStrippedNestedElement(next, element)) {
```
bencaradocdavies: I think putting an instanceof test first would guard against performance degradation for non-complex cases (consistent with previous ```if``` clause) and is also great documentation. Yes, I saw that you also had a test in the method body, which is fine. Call me paranoid but this method is a real hot spot and avoiding method calls inside it might be important:
```
if (next instanceof ComplexAttribute && relaxed && isNonStripedNestedElement(next, element)) {
```
Note also the change ```isNonStrippedNestedElement``` -> ```isNonStripedNestedElement``` as mentioned above.

You can also fix the format of the previous two lines to be "} else {".  :-)

bencaradocdavies:
@nmco other than the small changes mentioned above, and the test failures noted in https://github.com/geoserver/geoserver/pull/1976, this looks fine and is almost ready to be merged.

modules/extension/xsd/xsd-core/src/main/java/org/geotools/xml/Encoder.java
nmco: Fixed.

modules/extension/xsd/xsd-core/src/main/java/org/geotools/xml/Encoder.java
```diff
@@ -618,6 +623,45 @@ public void encode(Object object, QName name, OutputStream out)
...
+    /**
+     * Helper method that checks if the complex feature we want to encode maps
+     * to a complex type that respects the GML object-property model.
+     */
+    private boolean isNonStrippedNestedElement(Object next, XSDElementDeclaration element) {
+        if (!(next instanceof ComplexAttribute)) {
+            return false;
+        }
+        ComplexAttribute complex = (ComplexAttribute) next;
+        // let's se if we need to encode this
```
nmco: Fixed.

modules/extension/xsd/xsd-core/src/main/java/org/geotools/xml/Encoder.java
nmco: Fixed.

nmco:
Thanks for the review @bencaradocdavies, follow your suggestions and apply a fix for the test (the if in the for loop should have been removed).

bencaradocdavies:
Thanks @nmco. All good.

nmco:
Thanks !

PR #1387, opened 2016-11-17
Mix of image oversample speedups

aaime:


dromagnoli:
+1 to me. Just check that jai-ext 1.0.12 is released and available before merging :)

PR #1388, opened 2016-11-17
[GEOT-5556] Add mosaic config to config map when updating mosaic

dvntucker:
- Added the mosaic configuration to the config map when updating a mosaic, since it wasn't getting added before but it was mean to be (probably didn't matter before, since the config wasn't used).
- Ensure CRS is correctly set for new CRS bound mosaic collector, before it was being set to the previous CRS, not the incoming one.
- Test case for updating mosaic


dromagnoli:
Looks good to me.

aaime:
Will this one be backported to 16.x?

simboss:
I was about to ask the same :)

dvntucker:
I will do a backport PR. It really should be backported.

PR #1389, opened 2016-11-18
Allow DataUtilities.urlToFile and DefaultResourceLocator to respect URL query [GEOT-5579]

jodygarnett:
See:
- https://osgeo-org.atlassian.net/browse/GEOT-5579
- https://osgeo-org.atlassian.net/browse/GEOS-7861

### DIFF
--- a/modules/library/main/src/main/java/org/geotools/styling/DefaultResourceLocator.java
+++ b/modules/library/main/src/main/java/org/geotools/styling/DefaultResourceLocator.java
@@ -44,13 +44,12 @@ public class DefaultResourceLocator implements ResourceLocator {
         URL url = null;
         try {
             url = new URL(uri);
-
-            //url is valid, check if it is relative
+            
             File f = DataUtilities.urlToFile(url);
             if (f != null && !f.isAbsolute()) {
                 //ok, relative url, if the file exists when we are ok
                 if (!f.exists() && sourceUrl != null) {
-                    URL relativeUrl = makeRelativeURL(f.getPath());
+                    URL relativeUrl = makeRelativeURL(f.getPath(), url.getQuery());
                     if (relativeUrl != null) {
                         URL validated = validateRelativeURL(relativeUrl);
                         if (validated != null) {
@@ -62,10 +61,9 @@ public class DefaultResourceLocator implements ResourceLocator {
         } catch (MalformedURLException mfe) {
             LOGGER.fine("Looks like " + uri + " is a relative path..");
             if (sourceUrl != null) {
-                url = makeRelativeURL(uri);
+                url = makeRelativeURL(uri,null);
             }
-            if (url == null)
-            {
+            if (url == null) {
                 url = getClass().getResource(uri);
                 if (url == null)
                         LOGGER.warning("can't parse " + uri + " as a java resource present in the classpath");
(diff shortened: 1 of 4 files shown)


aaime:
Yep, this is what I intended to implement myself... however, believe we'll need a similar change in GeoServer no?

modules/library/main/src/main/java/org/geotools/styling/DefaultResourceLocator.java
```diff
@@ -84,9 +82,14 @@ protected URL validateRelativeURL(URL relativeUrl) {
         return null;
     }
 
-    URL makeRelativeURL(String uri) {
+    URL makeRelativeURL(String uri, String query) {
```
aaime: Mind, the GeoServer version uses the other makeRelative method (which is protected). As far as know, that path will strip away the query string (even if by getting  a URL, it has all the needed information, believe the implementation needs to be changed to preserve the query string)

aaime:
Weird, the build is failing in unrelated code:

{code}
SEVERE: There's code using JDBC based datastore and not disposing them. This may lead to temporary loss of database connections. Please make sure all data access code calls DataStore.dispose() before freeing all references to it
Tests run: 2, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.088 sec <<< FAILURE! - in org.geotools.jdbc.JDBCTransactionStateTest
testSetTransactionNullWithExternalConnection(org.geotools.jdbc.JDBCTransactionStateTest)  Time elapsed: 0.006 sec  <<< ERROR!
org.mockito.exceptions.misusing.UnfinishedStubbingException: 
Unfinished stubbing detected here:
-> at org.geotools.jdbc.JDBCTransactionStateTest.setupLogHandler(JDBCTransactionStateTest.java:64)

E.g. thenReturn() may be missing.
Examples of correct stubbing:
    when(mock.isOk()).thenReturn(true);
    when(mock.isOk()).thenThrow(exception);
    doThrow(exception).when(mock).someVoidMethod();
Hints:
 1. missing thenReturn()
 2. although stubbed methods may return mocks, you cannot inline mock creation (mock()) call inside a thenReturn method (see issue 53)

	at org.geotools.jdbc.JDBCTransactionStateTest.setupLogHandler(JDBCTransactionStateTest.java:73)
{code}

aaime:
Ah, a GeoTools ticket would be needed to track this change in the geotools release notes.

modules/library/main/src/main/java/org/geotools/styling/DefaultResourceLocator.java
jodygarnett: Understood, need this one merged first.

jodygarnett: You are correct, will need to fix that method next. If that implementation used an @overrides annotation it would be broken with a compile error about now ...

modules/library/main/src/test/java/org/geotools/styling/DefaultResourceLocatorTest.java
```diff
@@ -4,11 +4,12 @@
 import java.net.URL;
 
 import org.geotools.data.DataUtilities;
+import org.junit.Test;
 
 import junit.framework.TestCase;
 
 public class DefaultResourceLocatorTest extends TestCase {
-
+    @Test
```
jodygarnett: remove unused annotation

PR #1392, opened 2016-11-19
[GEOT-5580] Use imageio EnhancedImageReadParam instead of ExtendedImageParam

nmco:
This pull request substitute the usage of ```EnhancedImageReadParam``` by by the new imageio```ExtendedImageParam```.

Associated issue:
- https://osgeo-org.atlassian.net/browse/GEOT-5580

### DIFF
--- a/modules/library/coverage/src/main/java/org/geotools/resources/image/ExtendedImageParam.java
+++ b/modules/library/coverage/src/main/java/org/geotools/resources/image/ExtendedImageParam.java
@@ -20,7 +20,9 @@ import javax.imageio.ImageReadParam;
 
 /**
  * Utility class that should be used to pass extra parameters to the NetCdf image reader machinery.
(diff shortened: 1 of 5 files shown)


modules/library/coverage/src/main/java/org/geotools/resources/image/ExtendedImageParam.java
```diff
@@ -1,36 +0,0 @@
...
- *    Lesser General Public License for more details.
- */
-package org.geotools.resources.image;
-
-import javax.imageio.ImageReadParam;
-
-/**
- * Utility class that should be used to pass extra parameters to the NetCdf image reader machinery.
- */
-public class ExtendedImageParam extends ImageReadParam {
```
aaime: Cannot remove a public class all of a sudden, should be deprecated with an indication of what to use instead

nmco:
Correct, I removed it for testing proposes and forgot to add it back :scream_cat:. Thanks for the review Andrea.

PR #1394, opened 2016-11-19
Enable fast lookup for many ESRI WKTs

aaime:


modules/plugin/epsg-hsql/src/test/java/org/geotools/referencing/EsriLookupTest.java
```diff
@@ -0,0 +1,77 @@
+package org.geotools.referencing;
```
jodygarnett: header

modules/library/referencing/src/main/java/org/geotools/referencing/factory/epsg/DirectEpsgFactory.java
```diff
@@ -209,8 +210,9 @@
             case 9103: return NonSI.MINUTE_ANGLE;
             case 9104: return NonSI.SECOND_ANGLE;
             case 9105: return NonSI.GRADE;
-            case 9107: return Units.DEGREE_MINUTE_SECOND;
-            case 9108: return Units.DEGREE_MINUTE_SECOND;
+            // DMS is a way to format information but not a real unit, decode towards DEGREE instead
+            case 9107: return NonSI.DEGREE_ANGLE;// return Units.DEGREE_MINUTE_SECOND;
+            case 9108: return NonSI.DEGREE_ANGLE; // return Units.DEGREE_MINUTE_SECOND;
```
jodygarnett: yeah okay

modules/library/referencing/src/main/java/org/geotools/referencing/factory/IdentifiedObjectFinder.java
```diff
@@ -361,6 +368,10 @@ final IdentifiedObject createFromCodes(final IdentifiedObject object) throws Fac
         }
         return null;
     }
+    
+    protected Set getSpecificCodeCandidates(final IdentifiedObject object) throws FactoryException {
```
jodygarnett: New method, is this a rename of getCodeCandidates? or how do the two differ...

modules/plugin/epsg-hsql/src/test/resources/org/geotools/referencing/esri_tests.properties
```diff
@@ -0,0 +1,36 @@
+2008=PROJCS["NAD_1927_CGQ77_MTM_2_SCoPQ",GEOGCS["GCS_NAD_1927_CGQ77",DATUM["D_NAD_1927_CGQ77",SPHEROID["Clarke_1866",6378206.4,294.9786982]],PRIMEM["Greenwich",0.0],UNIT["Degree",0.0174532925199433]],PROJECTION["Transverse_Mercator"],PARAMETER["False_Easting",304800.0],PARAMETER["False_Northing",0.0],PARAMETER["Central_Meridian",-55.5],PARAMETER["Scale_Factor",0.9999],PARAMETER["Latitude_Of_Origin",0.0],UNIT["Meter",1.0]]
```
jodygarnett: Should we have a:
````
# reference to the ESRI test databse the values come from?
````

jodygarnett (approved):

modules/library/referencing/src/main/java/org/geotools/referencing/factory/IdentifiedObjectFinder.java
aaime: No, it's a new method. The old one end up providing all code candidates (the whole set), so it's suitable only for the full scan, the new one actually makes an effort to returns only the ones relevant to the object being searched.

modules/plugin/epsg-hsql/src/test/resources/org/geotools/referencing/esri_tests.properties
aaime: It comes from a random shapefile I have on disk.

jodygarnett: Gotcha, I am not especially worried since this is test-data and not distributed with the library for use in downstream applications.

I imagine we will gather up additional test-data from ESRI/Oracle/Others; more test data can only improve compatibility with these projects.

jodygarnett:
A lot of warnings in the travis logs around CRS lookup, not sure why it has failed.

aaime:
See the devel list, it's that JDBC transaction test again

jodygarnett:
Grr, kicking travis.

PR #1396, opened 2016-11-23
Fix the number of calculated LabelPositions

mukoki:
The number of labelPositions is lineStringLength/labelDistance
 Current code computes too many positions. It is a regression introduced in 4e29da6f46bacfa3a1cada463eaa72dc0df95a67, committed on 28/08/2015.
 Consequence of this regression is that the painter tries to draw one or more labels at each ends of the line. It is particularly visible with tiles as labels may accumulate on the border of each tile (especially with a nonNull maxDisplacement value).
  To test it use option DISABLE_LETTER_LEVEL_CONFLICT defined in LabelCacheImpl (disabling letter_level_conflict will switch the code to the old function where the number of labels was computed correctly).
  The commit also fix a small encoding problem in letterConflict3.properties.


aaime:
Ah ha, that's interesting. I had noticed the massive performance regression myself as well and put a small band-aid [already](https://github.com/geotools/geotools/commit/f1bf12bb089097971144bd02ee2877b76398cb74) (due to lack of time) with an intention to go back and make a more comprehensive fix (the code is now re-computing twice or more the same positions).

Good catch on this one, I did not notice it. I'll try to review as soon as possible (which due to end of year commitments, might actually take a while)

mukoki:
From commit 8f9535de47b59acc08f09aeee3b92f5f5e28283f label cache uses letters bounding boxes instead of global label bounding box. This may explain part of the performance regression although we did not measure any regression when we submitted the commit.
The present fix is more about visual appearance, but may also improve performance as less labels need to be computed. Here are 3 images to illustrate the problem.
- Image 1 with default behaviour (letter-level conflict) : we can see some useless labels 
- Image 2 simulate old behaviour (with DISABLE_LETTER_LEVEL_CONFLICT = true) : labelling is ok
- Image 3 with default behaviour, we can see label  redundancy at the limit between two tiles 
image1

image2

image3

aaime:
Manually merged along with another regression caused by that fix in label repeat distance, adding extra to the repeat distance that the user did not specify. There are other known regression in place, like the current code ignoring the min group distance...

aaime:
Btw, about the issue of the min distance being added twice as long as the label... that does not make sense  per se, as the user only requested a given distance between the label centers, not the extra.

By adding the length of the label the interpretation would change to "white space between the labels" instead of "space between the label centers", which I can accept to some extent, but adding twice as much 
as the label length does not make sense.

However, I've tried to do that the the test that I've added to cover this issue starts exhibiting wrong behavior again, with some labels being clearly missing. 

Also if the labels are not painted along the line, but horizontal, adding the extra space seems to bear little significant, as the line might be vertical and the width of the text has no effect on label "closeness".

So this behavioral change that was hidden (not advertised) in the refactor needs to be assigned its own ticket and revisited.

PR #1417, opened 2016-12-14
NetCDF variable attributes [GEOS-7901]

braeckel:
This pull request addresses the GeoTools portion of [GEOS-7901](https://osgeo-org.atlassian.net/browse/GEOS-7901), and will be followed by another pull request to the GEOS codebase once this one is finalized.

NetCDF variable attributes are now populated in the Grid2DCoverage properties map so they are available for downstream processing, such as writing out netCDF files.  This includes a new test and a new netCDF file with a good number of netCDF/CF metadata and variable attributes.

There were two assumptions made in this patch based on my (limited) knowledge of the codebase:
- The appropriate way to pass information from GEOT coverageio/netcdf to GEOS netcdf-out (specifically NetCDFOutputManager) is with the Grid2DCoverage properties map.  I was unable to find any other compelling alternatives
- NetCDFResponse is a reasonable place to populate the Grid2DCoverage properties map


bencaradocdavies:
@braeckel the Travis CI failure looks like a bogus jdbc module failure we see from time to time so I have restarted the build for this PR.

We need some feedback from the module maintainers @simboss and @dromagnoli on this one. I would also like their opinion on how this will work together with multidimensional ImageMosaic.

The sample file seemed a little on the large size. 

You add a new file so we need a code contribution agreement if you/UCAR (as appropriate) do not already have one on file with OSGeo. Please also copy @jodygarnett (project officer) in your email:
http://docs.geotools.org/latest/developer/procedures/contribution_license.html

There are a few little formatting issues (just spaces inside round brackets). I know NetBeans is popular at UCAR:
http://docs.geotools.org/latest/developer/conventions/code/style.html

dromagnoli:
Hi,
in relation to Ben's feedbacks, you should do some tests with ImageMosaic of NetCDFs too so you can take a look on how it works. You can start from some tests available here:
https://github.com/geotools/geotools/blob/master/modules/plugin/coverage-multidim/netcdf/src/test/java/org/geotools/coverage/io/netcdf/NetCDFMosaicReaderTest.java

Basically, the ImageMosaic data loading operates on ImageReaders: a low level class which is not geospatially aware. Once a granule to be loaded is retrieved from imageMosaic index, a proper low level ImageReader is setup on it and a read operation is performed to return a RenderedImage. Afterwards, a coverageFactory is used to wrap a GridCoverage on top of the RenderedImage.

https://github.com/geotools/geotools/blob/master/modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/RasterLayerResponse.java#L1202
(line number at time of writing this comment)

Therefore an ImageMosaic of GeoTIFFs will use a TIFFImageReader to create a RenderedImage on top of a granule to be wrapped into GridCoverage (It will not use the GeoTIFFReader for reading operation).
Similarly, an ImageMosaic of NetCDFs will use a NetCDFImageReader to create a RenderedImage (it will not use the NetCDFReader to read).

In addition, I also remember some GridCoverageWrapper usage around in GeoServer.
If my memory serves me right they are used in bands customization and WCS output to deal with properties.  Bands customization occurs once you edit the CoverageBandDetails panel of a configured layer in order, to change, as an instance, the min-max range, the nodata, or the unit of measure which will be reported in the capabilities/describeCoverage documents.

You could setup a GeoServer instance based on this code and check how coverages wrapping is working with these changes.

Does it make sense?

bencaradocdavies:
@dromagnoli are you happy with the use of the ```GridCoverage2D``` properties map to store NetCDF variable attributes? If so, is the key ```additionalVariableAttributes``` satisfactory?

dromagnoli:
I'm ok with that. How testing with ImageMosaic are going on?

braeckel:
I was able to test the ImageMosaic support through the src/web/app Start application.  I went down the rabbit hole and it was quite a lesson on how GeoTools coverages work.  ;)  Unfortunately as alluded to earlier ImageMosaic is different in how it uses Readers and the earlier proposed approach does not work.  There are several potential approaches that I've found, none of which are perfect.

The main challenge with using GridCoverage2D properties is that the main information passed from ImageReaders to ImageMosaic is a RenderedImage.  Since ImageMosaic creates the GridCoverage it doesn't seem like there is a clean way to pass this information unless it is via the image properties.

However, any properties produced by ImageReaders are lost by the several subsequent JAI post-processing and re-assembly steps.  The only properties left are those created by JAI, such as "JAI.ImageReader" and "JAI.ImageReadParam" (see https://docs.oracle.com/cd/E17802_01/products/products/java-media/jai/forDevelopers/jai-imageio-1_0-docs/com/sun/media/jai/operator/ImageReadDescriptor.html#PROPERTY_NAME_IMAGE_READER). 
Another approach would be to create a new interface (GridCoveragePropertyReader, perhaps?) that when found is queried by ImageMosaic for extra properties to put in the GridCoverage2D properties.  This would require adding the interface to a module that is accessible to both ImageMosaic and netCDF extensions, such as org.geotools.coverage.grid.io.imageio. This would also require changes to ImageMosaic to either get the ImageReader instance from the JAI image properties (mentioned earlier) or create a new instance of the ImageReader based upon information in the request and/or RasterManager.  In the latter case this means getting an InputStream, disposing it, and all the rest of the ImageReader lifecycle as currently managed in GranuleDescriptor.

I investigated GridCoverageWrapper (i.e., CoverageDimensionCustomizerReader) and this suffers from the same challenges as ImageMosaic as described above with regards to getting a reference to an appropriate ImageReader.

Another solution is to directly use the image property set by JAI which points to the JAI ImageReader instance.  This would also have to be set by the NetCDFImageReader in cases where JAI is not being used and does not set the property itself.  With this image property the NetCDFOutputManager can get a handle directly to the NetCDFImageReader, which can be queried for the additional information just before data is written.  I have tested this approach and it does work in all cases I was able to run (with/without JAI, with/without ImageMosaic).  However, while it is quite simple and isolates the changes to only NetCDF-related classes it does not feel like a very clean solution due to the dependency on JAI-related properties.

None of the solutions that have come to mind seem ideal, do you have any suggestions or preferences?

bencaradocdavies:
In addition to honouring variable attributes and applying them to NetCDF output, both singly and via ImageMosaic, I also want to implement the inclusion of entire scalar variables together with their attributes, and their mosaicking over time.

For example, suppose I have a collection of GRIB2 files for weather forecasts. Each file is interpreted by NetCDF-Java as having a scalar ```reftime``` variable to represent the time at which the forecast was made. I then use ImageMosaic to join many of these files together into a time series. When I request a coverage via WCS, I want not only the requested gridded data dimensioned by time, but also the additional scalar ```reftime```, now mosaicked and thus dimensioned by time, and renamed ```forecast_reference_time``` to match the CF Conventions standard name.

bencaradocdavies:
I propose [GSIP 158 - NetCDF Output Support for Variable Attributes and Extra Variables](https://github.com/geoserver/geoserver/wiki/GSIP-158) as an alternate approach that will work with both plain NetCDF/GRIB and also ImageMosaic of NetCDF/GRIB files.

bencaradocdavies:
Pull requests https://github.com/geotools/geotools/pull/1564 and https://github.com/geoserver/geoserver/pull/2281 have been submitted for [GSIP 158 - NetCDF output support for variable attributes and extra variables](https://github.com/geoserver/geoserver/wiki/GSIP-158) and this pull request may now be obsolete..

braeckel:
Considering GSIP 158 and your current thread of work I think it is safe to close this pull request.

bencaradocdavies:
Thanks @braeckel, your analysis above was very helpful in guiding the implementation of GSIP 158.

PR #1421, opened 2016-12-18
[GEOT-5594] AzimuthalEquidistant projection missing parameters (master)

bencaradocdavies:
https://osgeo-org.atlassian.net/browse/GEOT-5594

Replaces https://github.com/geotools/geotools/pull/1420 (branch name typo)

PR #1422, opened 2016-12-18
[GEOT-5595] Cannot read/write GeoTIFF with Azimuthal_Equidistant projection (master)

bencaradocdavies:
https://osgeo-org.atlassian.net/browse/GEOT-5595

Note that this PR depends on and incorporates the commit in https://github.com/geotools/geotools/pull/1421

### DIFF
--- a/modules/library/referencing/src/main/java/org/geotools/referencing/operation/projection/AzimuthalEquidistant.java
+++ b/modules/library/referencing/src/main/java/org/geotools/referencing/operation/projection/AzimuthalEquidistant.java
@@ -152,6 +152,20 @@ public class AzimuthalEquidistant {
             return Provider.PARAMETERS;
         }
 
+        /**
+         * Return the values of the parameters that define the projection.
+         * 
+         * @see org.geotools.referencing.operation.projection.MapProjection#getParameterValues()
+         */
+        @Override
+        public ParameterValueGroup getParameterValues() {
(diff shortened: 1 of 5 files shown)


modules/library/referencing/src/test/java/org/geotools/referencing/operation/projection/AzimuthalEquidistantTest.java
```diff
@@ -0,0 +1,56 @@
+/*
```
simboss: Are you sure this change should be here?

Isn't this part of the other PR?

modules/library/referencing/src/main/java/org/geotools/referencing/operation/projection/AzimuthalEquidistant.java
```diff
@@ -152,6 +152,20 @@ public ParameterDescriptorGroup getParameterDescriptors() {
             return Provider.PARAMETERS;
```
simboss: Are you sure this change should be here?

Isn't this part of the other PR?

bencaradocdavies:
@simboss, please see the description; this one relies on https://github.com/geotools/geotools/pull/1421 and the merge will fix all.  :-)

bencaradocdavies:
@simboss that worked as expected: the GEOT-5594 commit does not appear in the merge as viewed in gitk. Apologies for the confusion; just keeping Travis CI happy.

PR #1423, opened 2016-12-19
[GEOT-5602] Reducing length of lock sections

ugomos:
SoftValueHashMap's performance should improve by using the data structures ConcurrentHashMap and ConcurrentLinkedQueue, requiring shorter lock sections.


dromagnoli:
Could you please minimize the reformatting on untouched pieces of code and comments?
That would allows better reviewing of the PR.
(I know... some classes are still improperly formatted but...)

aaime:
Did you make any scalability test? :-)

PR #1426, opened 2016-12-29
[GEOT-5606] Fixed invalid GML3 GeometryCollection encoding.

sikeoka:
Fixed invalid GML3 GeometryCollection encoding and updated the unit test.  Also applying the GEOT-5360 GeometryCollection encoding fix for GML 3.2 to GML 3.1.

https://osgeo-org.atlassian.net/browse/GEOT-5606
https://osgeo-org.atlassian.net/browse/GEOT-5360

This patch can be backported to 15.x and 16.x


aaime:
Yep, looking good, thank you for the fix!

PR #1427, opened 2017-01-11
Use SI.getInstance() to clone si units

shruda:
The si units clone part used NonSI.getInstance() instead of SI.getInstance().


aaime:
The change looks good to me. As per project policy, it requires a test to be merged. See https://github.com/geotools/geotools/blob/master/CONTRIBUTING

ianturton:
replaced by https://github.com/geotools/geotools/pull/1443

PR #1430, opened 2017-01-13
DB2 updates

dwadler:
Fix DB2 SQL generation to use units like METER for DWITHIN and BEYOND. Previously, units were ignored and distance was treated as being in the coordinate system units.

Updated online testcase to correctly test DWITHIN and BEYOND.

### DIFF
--- a/modules/plugin/jdbc/jdbc-db2/src/main/java/org/geotools/data/db2/DB2FilterToSQL.java
+++ b/modules/plugin/jdbc/jdbc-db2/src/main/java/org/geotools/data/db2/DB2FilterToSQL.java
@@ -23,6 +23,7 @@ import java.io.Writer;
 import java.text.SimpleDateFormat;
 import java.util.Date;
 import java.util.HashMap;
+import java.util.Map;
 import java.util.TimeZone;
 import java.util.logging.Logger;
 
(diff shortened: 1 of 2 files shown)


modules/plugin/jdbc/jdbc-db2/src/main/java/org/geotools/data/db2/DB2FilterToSQL.java
```diff
@@ -130,6 +130,29 @@
...
     private static WKTWriter wktWriter = new WKTWriter();
 
+    /**
+     * Conversion factor from common units to meter
+     */
+    private static final Map<String, Double> UNITS_MAP = new HashMap<String, Double>() {
+        {
+            put("kilometers", 1000.0);
+            put("kilometer", 1000.0);
+			put("meters", 1.0);
```
aaime: This line uses tabs, please switch to spaces (yes, there is old code that's not properly formatted, for newer code please follow the coding conventions, and no, do not mass reformat files)

aaime:
Thank you. I don't have experience with DB2 so cannot review the changes or test them, @mcrmcr do you have time to have a look?

dwadler:
Who is @mcrmcr? I did modify the existing online test for DWITHIN to perform a useful test and added a test for BEYOND which ran successfully. There are some other tests that are failing but they were failing before my change - I'm looking into why.

aaime:
@mcrmcr  is the current maintainer of jdbc-db2 (check the module history, https://github.com/geotools/geotools/commits/master/modules/plugin/jdbc/jdbc-db2/src/main )

dwadler:
Does @mcrmcr have a name? I thought Christian Mueller was maintaining the DB2 support. I was the original author in 2005.

aaime:
@mcrmcr is Christian Mueller indeed, the current maintainer. He's the one that has to be comfortable with the changes. Unless he steps down and you become the new maintainer.

mcrmcr:
Merged

PR #1432, opened 2017-01-17
[GEOT-5602] Better concurrent SoftValueHashMap

ugomos:
SoftValueHashMap that uses data structures to scale better in concurrent environments. Unit tests were added to test the behaviour in multithreaded environments. Fixed the issue present in previous (reverted) PR ( https://github.com/geotools/geotools/pull/1431 ): inside the method retainStrongly, it is checked that if the value returned by hash.get(replaceNull(toRemove)) has already been wrapped in a Reference, it must not be wrapped again before insertion in the map.


aaime:
Ugo, can you add comments in the pull request in the sections that got changed?

ugomos:
Yes, I updated my first comment.

aaime:
All right, merging, let's see how the build server likes it

PR #1439, opened 2017-01-24
Purchase Order XML Tutorial update using jaxb

vickdw:
Updated the Purchase Order tutorial bringing the current version of jaxb and filling out the rest of the downloadable classes providing the user with the complete working example.  Updated the associated pom files as shown in the tutorial to show how the POM is filled out adding execution and build plans.  Also updated the tutorial.rst file adding in additional text where needed to outline the steps.  Whitespace was also cleaned up across the project.


docs/src/main/java/org/geotools/po/bindings/CommentBinding.java
```diff
@@ -0,0 +1,44 @@
+package org.geotools.po.bindings;
```
jodygarnett: Thanks!

docs/src/main/java/org/geotools/po/bindings/USAddressBinding.java
```diff
@@ -24,49 +36,58 @@
...
+    ObjectFactory factory;        
+    public USAddressBinding( ObjectFactory factory ) {
+        super();
+        this.factory = factory;
+    }
+
+    /**
+     * @generated
+     */
+    public QName getTarget() {
```
jodygarnett: We are getting a lot of conflicts here, but the code looks the same.  Do we have whitespace conflict I am not seeing? The answer is "yes" and your new code uses the correct formatting :)

PR #1441, opened 2017-01-26
Upgrade EhCache to latest 2.x version

dvntucker:
This one is a little selfish, but we have some conflicts internally with the version of EhCache used in GeoTools and the latest 2.x version. I figure this upgrade is fairly safe in that

- it's only used by ImageMosaic
- there are no code changes required to do the upgrade.

I'd consider upgrading even further, but EhCache 3.x is a significant departure both in API and capabilities.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1046,7 +1046,7 @@
       <dependency>
         <groupId>net.sf.ehcache</groupId>
         <artifactId>ehcache</artifactId>
-        <version>1.6.2</version>
+        <version>2.10.3</version>
       </dependency>
 
       <!-- Tests or legacy -->


aaime:
@dromagnoli I see EhCache is used in the Utils class, but only via a private constructor that nothing is using... weird? Setup as is, does not even seems to be functional (no way to set it, so the portion of the code that can use it does not seem to be reachable)

dvntucker:
Interesting, I didn't notice that. Well removing the dependency entirely would also solve the issue :D

dvntucker:
If no one has any issues then I just removed EhCache instead.

dromagnoli:
Hi @aaime. It is used in a different project depending on GT, through spring injection in application-context.

dvntucker:
Ok, I can change back to the version upgrade.

dvntucker:
Ok, changed this to just upgrade EhCache

PR #1444, opened 2017-01-27
Fixed GEOT-5336

ianturton:
The code was miss calculating the quadrants to compare to when dealing with
diagaonals greater than 180 degrees.


ianturton:
Currently testWorldTwice fails - I'm not sure what the correct answer should be

PR #1445, opened 2017-01-27
Fix for GEOT-5265 Some GeoTiffs don't open

ianturton:
While this fixes the reported problem, it would be good if the GeoTiff reader
could take care of this for itself.


tbarsballe:
Seems like a test case would be valuable here?

Actual fix looks fine.

aaime:
Given that's code supporting documentation I don't think a test is strictly necessary, I'd merge anyways. Just my 2 cents ;-)

ianturton:
I can't think of a way to test this. as Andrea says it's a tutorial

PR #1446, opened 2017-01-28
[GEOT-1710] Urn parser update for new OGC codes

mtCarto:
See https://osgeo-org.atlassian.net/browse/GEOT-1710 for details.

### DIFF
--- a/modules/library/referencing/src/main/java/org/geotools/referencing/factory/URN_Parser.java
+++ b/modules/library/referencing/src/main/java/org/geotools/referencing/factory/URN_Parser.java
@@ -87,11 +87,32 @@ final class URN_Parser extends URI_Parser {
                     if (candidate != null) {
                         final int nameEnd = code.indexOf(URN_SEPARATOR, typeEnd + 1);
                         if (nameEnd >= 0) {
+                            final int versionEnd = code.indexOf(URN_SEPARATOR, nameEnd + 1);
                             final int lastEnd = code.lastIndexOf(URN_SEPARATOR);
-                            Version urnVersion = (lastEnd <= nameEnd) ? null : new Version(
-                                    code.substring(nameEnd + 1, lastEnd));
-                            String urnAuthority = code.substring(typeEnd + 1, nameEnd).trim();
-                            String urnCode = code.substring(lastEnd + 1).trim();
+                            Version urnVersion;
+                            String urnAuthority;
+                            String urnCode;
+                            if (versionEnd != lastEnd && versionEnd != -1) {
+                                urnVersion = (lastEnd <= nameEnd) ?
+                                    null :
+                                    new Version(code.substring(nameEnd + 1, versionEnd));
+                                urnAuthority = code.substring(typeEnd + 1, nameEnd);
+                                urnCode = code.substring(versionEnd + 1).trim();
+                                urnCode = urnCode.replaceAll(String.valueOf(URN_SEPARATOR), ",");
+                            } else {
+                                urnVersion = (lastEnd <= nameEnd) ?
+                                    null :
+                                    new Version(code.substring(nameEnd + 1, lastEnd));
+                                urnAuthority = code.substring(typeEnd + 1, nameEnd).trim();
+                                urnCode = code.substring(lastEnd + 1).trim();
+                            }
+                            if (urnCode.contains("CRS")) {
+                                urnAuthority = "CRS";
+                                urnCode = urnCode.substring(3);
+                            } else if (urnCode.contains("AUTO")) {
+                                urnAuthority = "AUTO";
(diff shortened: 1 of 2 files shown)


modules/library/referencing/src/test/java/org/geotools/referencing/factory/URN_ParserTest.java
```diff
@@ -19,51 +19,46 @@
...
+    @Test public void testParse() throws NoSuchAuthorityCodeException {
         final URN_Parser parser = URN_Parser.buildParser("urn:ogc:def:CRS:EPSG:6.11.2:4326");
-        assertEquals("crs",       parser.type.name);
-        assertEquals("EPSG",      parser.authority);
-        assertEquals("6.11.2",    parser.version.toString());
-        assertEquals("4326",      parser.code);
+        assertEquals("crs", parser.type.name);
+        assertEquals("EPSG", parser.authority);
+        assertEquals("6.11.2", parser.version.toString());
+        assertEquals("4326", parser.code);
```
jodygarnett: this is just white space changes right?

modules/library/referencing/src/main/java/org/geotools/referencing/factory/URN_Parser.java
```diff
@@ -87,11 +87,29 @@ public static URN_Parser buildParser(final String urn) throws NoSuchAuthorityCod
...
+                                urnAuthority = code.substring(typeEnd + 1, nameEnd).trim();
+                                urnCode = code.substring(lastEnd + 1).trim();
+                            }
+                            if (urnCode.contains("CRS")) {
+                                urnAuthority = "CRS";
+                                urnCode = urnCode.substring(3);
+                            } else if (urnCode.contains("AUTO")) {
+                                urnAuthority = "AUTO";
+                                urnCode = urnCode.substring(4);
+                            }
```
jodygarnett: Okay so this parser is still brittle as anything, but not really any worse than before (in terms of having an invalid index exception).

modules/library/referencing/src/test/java/org/geotools/referencing/factory/URN_ParserTest.java
```diff
@@ -19,51 +19,46 @@
...
         assertEquals("EPSG:4326", parser.getAuthorityCode());
     }
 
     /**
      * Parses an invalid URN.
      */
-    @Test
-    public void testInvalidParse() {
+    @Test public void testInvalidParse() {
```
jodygarnett: This is just a whitespace change; do you have a different formatter?

aaime:
Where are these new codes coming from? Do you have a document or spec reference?

jodygarnett:
@aaime from the bug report:

> The OGC Web Services Common specification refers to a new URN format which has to be used in all capabilities documents:
> 
> urn:ogc:def:crs:OGC:1.3:CRS84
> 
> The above document states, at page 57:
> 
> "The OGC URN "urn:ogc:def:crs:OGC:1.3:CRS:84" is here used to refer to the "WGS 84 longitude-latitude" CRS specified in Subclause B.3 of WMS 1.3, previously referenced as "CRS:84".
> 
> The trouble is that once defined it can be used for requests as well. For the moment I can workaround in Geoserver by recognizing the code and replacing it with epsg:4326, but it would
be better to avoid these in the long run.

Let me see if I can find that:

[Definition identifier URNs in OGC namespace](http://portal.opengeospatial.org/files/?artifact_id=30575) - shows one GML example

````
gml:uom="urn:ogc:def:uom:EPSG::9904"
````

mtCarto:
Also referenced OpenGIS_Web_Map_Service_WMS_Implementation_Specification
<http://portal.opengeospatial.org/files/?artifact_id=14416> or the AUTO
code syntax.

Morgan Thompson
Junior Engineer | Boundless
mthompson@boundlessgeo.com

modules/library/referencing/src/test/java/org/geotools/referencing/factory/URN_ParserTest.java
mtCarto: I imported the geotools formatter.xml before starting, maybe it was still using the geogig formatter.

aaime: Sometimes I load by mistake the formatted setup for code examples instead of the main one, may it be the case?

aaime:
Thanks for the reference, when formatting is fixed I believe this can be merged

mtCarto:
Will fixup the formatting today, seems Intellij didn't import the GeoTools formatter correctly.

aaime:
I see the ticket is still open? Can it be closed with "fix for" updated according to merges/backports?

PR #1449, opened 2017-01-31
[GEOT-5626] Initial commit of unsupported mapbox module

mattkrusz:
Issue: https://osgeo-org.atlassian.net/browse/GEOT-5626
See https://github.com/geotools/geotools/wiki/Mapbox-Style for a wiki tracking the progress.


jodygarnett:
Okay we got a build failure:

````
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.15:test (default-test) on project gt-mbstyle: There are test failures.
[ERROR] 
````

1) we can fix the test faiure
2) the profile to build gt-mbstile should not be on by default (ie an unsupported module is not included in the build - it is a play area to work in until ready)

File: modules/unsupported/pom.xml [hunk 1]
```diff
@@ -190,6 +190,12 @@
                 <module>polylabel</module>
             </modules>
         </profile>
+        <profile>               
+            <id>mbstyle</id>
```
jodygarnett: this profile does not look to be on; I wonder why the build failed?

modules/unsupported/pom.xml
```diff
@@ -210,6 +216,7 @@
 				<module>css</module>
 				<module>geojsonstore</module>
 				<module>polylabel</module>
+                <module>mbstyle</module>
```
jodygarnett: ah, it is included in this list of "unsupported" modules - please remove.

This list is built since some of these modules are used by downstream projects; but they have not yet met our QA/doc/maintainer requirements.

mattkrusz:
Thanks Jody -- removed the erroneous entry under "all" in unsupported/pom.xml

jodygarnett:
Thanks for the fix; good luck have fun.

PR #1454, opened 2017-02-02
[GEOT-5627] Fixed create temp file

ugomos:


modules/plugin/imagemosaic/src/test/java/org/geotools/gce/imagemosaic/ImageMosaicReaderTest.java
```diff
@@ -2047,7 +2047,7 @@ public void testRequestInAreaWithNoGranulesBecomesTransparent() throws Exception
         // read and check we actually got a coverage in the requested area
         GridCoverage2D coverage = reader.read(new GeneralParameterValue[] {ggp, transparent});
         assertNotNull(coverage);
-        new GeoTiffWriter(new File("/tmp/test.tiff")).write(coverage, null);
+        new GeoTiffWriter(TestData.temp(this, "test.tiff")).write(coverage, null);
```
simboss: I would add a call to dispose at the end.

simboss:
Do we also need to backport?

Please, check if this applies to older versions.

ugomos:
I updated this PR adding the dispose of the ImageMosaicReader.
Yes, we need also to backport, the same problem is at least in branch 16.x and 15.x.

ugomos:
At last, the line "new GeoTiffWriter(new File("/tmp/test.tiff"))" had no reason to exist, it is now deleted. Added the dispose of the ImageMosaicReader object.

PR #1459, opened 2017-02-06
more dynamic dasharray

wumpz:
Allow to read the dasharray completly from database, instead of getting each numerical part separately from database.


wumpz:
I configured the wrong target path.

wumpz:
Look at PR #1465.

PR #1461, opened 2017-02-07
Readability improvements to InterpolationFunction

jodygarnett:
Javadocs and toString implementation based on StringBuilder. This implementation does not extend any base class, causing it to miss out on a default toString implementation.


jodygarnett:
This is not a behavioural change and is suitable for backport (if anyone cares).

mattkrusz:
This looks correct -- and very useful -- to me. Thanks Jody!

PR #1462, opened 2017-02-07
Create a ColorConverter for HTML / CSS color names (GEOT-5633)

jodygarnett:
This PR adds a Hints.COLOR_NAMES key.

For details see: https://osgeo-org.atlassian.net/browse/GEOT-5633

### DIFF
--- a/modules/library/main/src/main/java/org/geotools/util/ColorConverterFactory.java
+++ b/modules/library/main/src/main/java/org/geotools/util/ColorConverterFactory.java
@@ -42,67 +42,81 @@ import org.geotools.factory.Hints;
  * @source $URL$
  */
 public class ColorConverterFactory implements ConverterFactory {
+    
+    /**
+     * Uses {@link Color#decode(String)} to convert String to Color.
+     */
+    public static Converter CONVERT_STRING = new Converter() {
+        public <T> T convert(Object source, Class<T> target) throws Exception {
+            String rgba = (String) source;
+            try {
+                return target.cast( Color.decode(rgba) );
+            } catch (NumberFormatException badRGB) {
+                // unavailable
+                return null;
+            }
+        }
+    };
+    
+    /**
+     * Converts provided integer to color, taking care to allow rgb and rgba support.
+     */
+    public static Converter CONVERT_NUMBER = new Converter() {
+        public <T> T convert(Object source, Class<T> target) throws Exception {
+            Number number = (Number) source;
+            // is it an integral number, and small enough to be an integer?
+            if (((int) number.doubleValue()) == number.doubleValue() && number.doubleValue() < Integer.MAX_VALUE) {
+                int rgba = number.intValue();
+                int alpha = 0xff000000 & rgba;
+                return target.cast(new Color(rgba, alpha != 0));
+            } else {
+                return null;
+            }
+        }
+    };
+    
+    /**
+     * Converts color to hex representation.
+     */
+    private Converter CONVERT_COLOR_TO_STRING = new Converter() {
+        
+        public <T> T convert(Object source, Class<T> target) throws Exception {
+            Color color = (Color) source;
+            
+            String redCode = Integer.toHexString(color.getRed());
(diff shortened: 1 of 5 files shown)


jodygarnett:
This pull request safely adds the required COLOR_NAMES hint.

In the mbstyle community module a couple of converters are taking shape; when they are ready they can be considered for the gt-main module (providing an oppertunity to simplify css and ysld modules).

jodygarnett:
The color maps collected from css / ysld modules to not quite match up with the information I find online ("gray" should conflict between x11 and web colors for example). As such there are two ignored tests.

aaime:
So if I understand correctly this adds a second "string -> color" conversion, but does not register it in SPI, so it should not cause random conversion issues, and the callers are supposed to use it directly right?

jodygarnett:
It will register it in in SPI, was not going to do so until it passes unit tests.

However it is gated - without the hint it will no do anything.

aaime:
Ah ok, I see

modules/library/main/src/main/java/org/geotools/util/ColorConverterFactory.java
```diff
@@ -42,67 +42,81 @@
...
-                            int alpha = 0xff000000 & rgba;
-                            return new Color(rgba, alpha != 0);
-                        } else {
-                            return null;
-                        }
-                    }
-                };
+                return CONVERT_NUMBER;
             }
         } else if (target.equals(String.class) && source.equals(Color.class)) {
```
mattkrusz: Do we support subclasses of Color? You didn't write this line, but while we're in here, perhaps the Color type equality check should be an assignable check.

modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/parse/NamedColorConverterFactory.java
```diff
@@ -0,0 +1,436 @@
...
+
+import org.geotools.factory.Hints;
+import org.geotools.util.Converter;
+import org.geotools.util.ConverterFactory;
+
+/**
+ * Converter factory ssupporting web or css color names, requires the use of
+ * {@link Hints#COLOR_NAMES} hint with "WEB", "CSS", "X11", or "HTML".
+ * <p>
+ * There are a few subtle differences between WEB, CSS and X11 colors. this and CSS Colors (for the
```
mattkrusz: Not a big deal but could use a slight comment edit for clarity. I understand what you're saying, but I think the formatting broke a sentence here: "this and CSS colors (for the example ...)"

At the same time, might as well remove the duplicate "the the" on the next line also and fix "ssupporting" at the top.

modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/parse/NamedColorConverterFactory.java
```diff
@@ -0,0 +1,436 @@
...
+        colors.put("dodgerblue", new Color(30, 144, 255));
+        colors.put("firebrick", new Color(178, 34, 34));
+        colors.put("floralwhite", new Color(255, 250, 240));
+        colors.put("forestgreen", new Color(34, 139, 34));
+        colors.put("fuchsia", new Color(255, 0, 255));
+        colors.put("gainsboro", new Color(220, 220, 220));
+        colors.put("ghostwhite", new Color(248, 248, 255));
+        colors.put("gold", new Color(255, 215, 0));
+        colors.put("goldenrod", new Color(218, 165, 32));
+        colors.put("gray", new Color(128, 128, 128));
```
mattkrusz: Sorry, a little confused -- so despite the comments, we are leaving this X11 "gray" set to the CSS value (128,128,128) rather than the darker X11 value (190, 190, 190)? I'm not an artist, just making sure I'm on the same page.

jodygarnett:
Closing this pull request, @mattkrusz has a replacement based on email decision to go with just CSS colors - "rebeccapurple" for the win!

PR #1463, opened 2017-02-07
[GEOT-5617] Shapefile attribute (DBF) DbaseFileReader reads numeric empty values as zero

pbackx:
This is a pull request for GEOT-5617. As discussed, I think it is much safer to return a null value than to return a zero value if the reader does not know how to handle an input.

https://osgeo-org.atlassian.net/browse/GEOT-5617


modules/plugin/shapefile/src/test/java/org/geotools/data/shapefile/dbf/DbaseFileReaderTest.java
```diff
@@ -0,0 +1,61 @@
...
+            -----------+---------
+              98245    | <all spaces>
+              98289    | ****************** (= all stars)
+              98538    | 0.000000000
+              98586    | 5.210000000
+         */
+        @Test public void test() throws Exception {
+                InputStream dbf = this.getClass().getResourceAsStream(
+                        "/org/geotools/data/shapefile/test-data/dbase-file-reader/nulls.dbf");
+                DbaseFileReader dbfReader = new DbaseFileReader(Channels.newChannel(dbf), false,
```
aaime: Eeek, tabs! Please use the coding conventions (only reformat this file please)

peterbackx: yes, I was surprised by that. I imported https://github.com/geotools/geotools/blob/master/build/eclipse/formatter.xml in IntelliJ but it looks like it didn't work correctly

pbackx: yes, I was surprised by that. I imported https://github.com/geotools/geotools/blob/master/build/eclipse/formatter.xml in IntelliJ but it looks like it didn't work correctly

pbackx:
The indentation should be fixed now. Upon reviewing the formatter.xml it has the following setting:

`<setting id="org.eclipse.jdt.core.formatter.indentation.size" value="8"/>`

This is why my original commit had so much indentation. I'm not sure if this is interpreted differently by Eclipse, but IntelliJ added 8 spaces for indentation.

PR #1465, opened 2017-02-07
GEOT-5623: dynamic stroke-dasharray improvements

wumpz:
Hopefully now the pull request is correct. Writing this I recognized, that github does not display the small changes I did but the whole files.


wumpz:
I did all the EOL checking local und got using git a quite nice diff file, as you can see looking at GEOT-5623. So strange that happens.

aaime:
Ugh.. yeah, in this state is not reviewable yet. Are you on windows? It might be due to your checkout settings, see https://help.github.com/articles/dealing-with-line-endings/ and http://stackoverflow.com/questions/10418975/how-to-change-line-ending-settings

wumpz:
I check out as is. So no modifications here. At least I hope so.

wumpz:
Due to invalid EOLs this pr is closed.

PR #1466, opened 2017-02-08
[GEOT-5611] Invalid data for ImageMosaic of multi-coverage NetCDF/GRIB files (master)

bencaradocdavies:
https://osgeo-org.atlassian.net/browse/GEOT-5611

This pull request adds unit tests and a fix that uses the ```NetCDFReader.defaultName``` coverage to satisfy requests with unspecified coverage.

@simboss and @dromagnoli, this fix is also needed on 16.x, but I think 15.x is unaffected.


bencaradocdavies:
@simboss and @dromagnoli, a fix for this issue is a must-have for 16.2 and 17-beta next week. Please let me know if you have any objection to this implementation.

dromagnoli:
Hi @bencaradocdavies. No issues on my side.
Did you try your changes in GeoServer too? (creating a store, configuring a layer,...)

bencaradocdavies:
@dromagnoli I tested this PR with full builds of GeoTools (``-Dall``) and GeoServer (``-Prelease``) as well as a manual test of a GeoServer ImageMosaic of four GRIB2 files over time with WCS 2.0 output to NetCDF-4. Without this pull request, all NetCDF-4 ``Temperature_surface`` values were zero. With this pull request, I was able to delete and recreate the ImageMosaic store and publish the ``Temperature_surface`` layer without problems; ``Temperature_surface`` in NetCDF-4 output then contains sensible values.

dromagnoli:
I would say :+1:  +1 then.
(I was asking since I have memories of some code paths in the ResourcePool accessing the reader with no coverageName during initialization of the store or something similar).

bencaradocdavies:
@dromagnoli calls that do not specify ``coverageName`` for a single-coverage NetCDF should work exactly as before. This PR only changes the behaviour for calls that do not specify ``coverageName`` for a multi-coverage NetCDF by using the ``defaultName`` coverage rather than throwing an exception. The ``defaultName`` is typically the first coverage found in the NetCDF file, and it is assumed that it has the same coordinate grid as the other coverages.

Note that it is possible for a NetCDF file to contain multiple different coordinate grids; in practice I have not seen this, and furthermore one organisation for which I worked adopted a business rule that mandated only one grid per file. This seems common. Strict support for multiple different coordinate grids in a a single NetCDF file would require removing all CoverageReader API calls that do not specify a coverage name from gt-imagemosaic. This is major overkill for a situation that I have never seen. Another alternative is to revert all the GEOT-5392 refactoring, again overkill. This PR is in my view the least bad solution.

bencaradocdavies:
Merged and backported to 16.x.

PR #1468, opened 2017-02-08
[GEOT-5635] ImageMosaic loading create one object blocking the finalization thread for a while

aaime:


modules/plugin/imagemosaic/pom.xml
```diff
@@ -207,20 +207,4 @@
   </dependencies>
 
 
-  <build>
```
aaime: I was not sure about this one, just noticed it makes the build slower and harder to debug

aaime: Maybe someone on Windows can try the build and see if removing this causes issues? I've checked the history, but it has been there longer than what we have in git

dromagnoli:
Looks good to me.

modules/plugin/imagemosaic/pom.xml
dromagnoli: Not sure about this, neither.

dromagnoli: I have built the imageMosaic module on Windows (on top of your aaime:replaceSerializableRenderedImage) and I had no issues.

PR #1469, opened 2017-02-08
[GEOT-5636] fixed imageio-ext-gdalgeotiff.jar classpath conflicts

Damianofds:
The imageio-ext-gdalgeotiff.jar can cause classpath loading problems: only classes in imageio-ext-tiff.jar should be loaded.
See related issue on geoserver https://osgeo-org.atlassian.net/browse/GEOS-7981

### DIFF
--- a/modules/plugin/imageio-ext-gdal/pom.xml
+++ b/modules/plugin/imageio-ext-gdal/pom.xml
@@ -150,6 +150,12 @@
       <groupId>it.geosolutions.imageio-ext</groupId>
       <artifactId>imageio-ext-gdalvrt</artifactId>
       <!-- The version number is specified in the parent POM. -->
+      <exclusions>
+        <exclusion>
+          <groupId>it.geosolutions.imageio-ext</groupId>
+          <artifactId>imageio-ext-gdalgeotiff</artifactId>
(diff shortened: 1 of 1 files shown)


modules/plugin/imageio-ext-gdal/pom.xml
```diff
@@ -150,6 +150,12 @@
       <groupId>it.geosolutions.imageio-ext</groupId>
       <artifactId>imageio-ext-gdalvrt</artifactId>
       <!-- The version number is specified in the parent POM. -->
+      <exclusions>
+        <exclusion>
+          <groupId>it.geosolutions.imageio-ext</groupId>
+          <artifactId>imageio-ext-gdalgeotiff</artifactId>
+        </exclusion>
+      </exclusions>
```
dromagnoli: Why the exclusion?
Does the gdalvrt module depend on gdalgeotiff?
If that's the case we can consider opening also a ticket in imageio-ext and make sure to avoid that dependency.
Probably it has been added as a dependency for tests, without <scope>test</scope>

dromagnoli:
@Damianofds please, make sure to backport too

PR #1471, opened 2017-02-09
[GEOT-5633] Add CSS Color Converter, enabled by Hints.COLOR_NAMES

mattkrusz:
Adds a Color Converter for converting String color names (from the W3C CSS4 Extended Color Keywords set: https://www.w3.org/TR/css-color-4/#named-colors) to Color objects. To avoid interfering with other string->color conversions, this Color Converter will only be used when a new Hints.COLOR_NAMES hint is provided (and set to "CSS").


jodygarnett:
This is a replacement for https://github.com/geotools/geotools/pull/1462

modules/library/main/src/main/java/org/geotools/util/ColorConverterFactory.java
```diff
@@ -43,66 +43,92 @@
  */
 public class ColorConverterFactory implements ConverterFactory {
 
+    /**
+     * Uses {@link Color#decode(String)} to convert String to Color.
+     */
+    public static Converter CONVERT_STRING = new Converter() {
```
jodygarnett: excellent, static converters keep us from making them each time.

modules/library/main/src/main/java/org/geotools/util/CssColorConverter.java
```diff
@@ -0,0 +1,202 @@
...
+/**
+ * 
+ * Converts CSS Color Module 4 names to {@link Color} objects.
+ * 
+ * @see <a href="https://www.w3.org/TR/css-color-4/#named-colors">https://www.w3.org/TR/css-color-4/#named-colors</a> 
+ *
+ */
+public class CssColorConverter implements Converter {
+
+    private static final Map<String, Color> CSS_COLORS;
```
jodygarnett: This should be a public field on ColorConverter so that the gt-ysld and gt-css modules can make use of this map (rather than have their own duplicate).

modules/library/main/src/main/java/org/geotools/util/CssColorConverter.java
```diff
@@ -0,0 +1,202 @@
...
+        colors.put("wheat", new Color(245, 222, 179));
+        colors.put("white", new Color(255, 255, 255));
+        colors.put("whitesmoke", new Color(245, 245, 245));
+        colors.put("yellow", new Color(255, 255, 0));
+        colors.put("yellowgreen", new Color(154, 205, 50));
+        CSS_COLORS = Collections.unmodifiableMap(colors);
+    }
+
+    @Override
+    public <T> T convert(Object source, Class<T> target) throws Exception {
```
jodygarnett: This is a small enough class it can be added directly to ColorConverterFactory.

modules/library/main/src/main/java/org/geotools/util/ColorConverterFactory.java
```diff
@@ -43,66 +43,92 @@
...
+
+    /**
+     * Converts color to hex representation.
+     */
+    private Converter CONVERT_COLOR_TO_STRING = new Converter() {
+
+        public <T> T convert(Object source, Class<T> target) throws Exception {
+            Color color = (Color) source;
+
+            String redCode = Integer.toHexString(color.getRed());
```
jodygarnett: This code should check the hint, and look up in the CSS_COLORS map to see if the provided color can be represented as a name.

mattkrusz: Not sure I understand -- this is a Converter and doesn't have access to the hint map, correct?

mattkrusz:
Updated in response to review.

mattkrusz:
As I write code to use this, I realize that it is missing the ability to fall back to other converters. I.e., suppose Hints.COLOR_NAMES="CSS" is provided, but the input is "#BEBEBE" -- the converter should successfully fall back to parsing the hex value, yes? I'll add this if no one objects.

modules/library/main/src/main/java/org/geotools/util/ColorConverterFactory.java
```diff
@@ -43,66 +46,259 @@
...
+
+            if (blueCode.length() == 1) {
+                blueCode = "0" + blueCode;
+            }
+            String hex = ("#" + redCode + greenCode + blueCode).toUpperCase();
+            return target.cast(hex);
+        }
+    };
+    
+    public static final Map<String, Color> CSS_COLORS;
```
jodygarnett: public field javadoc

modules/library/main/src/main/java/org/geotools/util/ColorConverterFactory.java
```diff
@@ -43,66 +46,259 @@
...
+        colors.put("paleturquoise", new Color(175, 238, 238));
+        colors.put("palevioletred", new Color(219, 112, 147));
+        colors.put("papayawhip", new Color(255, 239, 213));
+        colors.put("peachpuff", new Color(255, 218, 185));
+        colors.put("peru", new Color(205, 133, 63));
+        colors.put("pink", new Color(255, 192, 203));
+        colors.put("plum", new Color(221, 160, 221));
+        colors.put("powderblue", new Color(176, 224, 230));
+        colors.put("purple", new Color(128, 0, 128));
+        colors.put("rebeccapurple", new Color(102, 51, 153));
```
jodygarnett: CSS4 achievement unlocked

modules/library/main/src/main/java/org/geotools/util/ColorConverterFactory.java
```diff
@@ -43,66 +46,259 @@
...
+        colors.put("whitesmoke", new Color(245, 245, 245));
+        colors.put("yellow", new Color(255, 255, 0));
+        colors.put("yellowgreen", new Color(154, 205, 50));
+        CSS_COLORS = Collections.unmodifiableMap(colors);
+    }
+
+    /**
+     * Converts CSS Color Module 4 names to colors.
+     */
+    private static Converter CONVERT_CSS_NAME_TO_COLOR = new Converter() {
```
jodygarnett: Excellent, now that it can convert both CSS_NAME and hex we could just call it CONVERT_CSS_TO_COLOR ?

modules/library/main/src/main/java/org/geotools/util/ColorConverterFactory.java
jodygarnett: You are right, we will need to catch that at the factory level and do the same trick you pulled below.

modules/library/main/src/main/java/org/geotools/util/ColorConverterFactory.java
mattkrusz: Yes, although this brings up the broader issue -- the CSS Color Module (and MapBox Styles, for that matter) support not only color names, but also strings in the format of rgba(x,x,x,x) / hsla(x,x,x,x). Don't want this name to be misleading.

modules/library/main/src/main/java/org/geotools/util/ColorConverterFactory.java
mattkrusz: I'm not sure it's strictly necessary -- as it is now, when trying to convert a string to a color, this is what happens:

(1) If you provide `Hints.COLOR_NAMES="CSS"`: You get a converter that tries the color map and falls back to the ordinary `CONVERT_STRING` converter.
(2) If you don't provide the hint: You just get back the `CONVERT_STRING` converter.

An additional check in the factory would just duplicate the fallback logic in the css converter. Of course, I wouldn't mind it, if you want to be extra safe.

mprins:
Since css4 came up in this PR; should the convert method not handle alpha transparency when converting to hex? Not sure if that causes problems down the line when using the code... 
Now it supports converting to Color with alpha, but roundtripping will lose the transparency so that should at least be documented.

See: https://drafts.csswg.org/css-color/#hex-notation

jodygarnett:
It already covers alpha when converting from number --> color (so adding support for string hex alpha representation would not be too bad).

Still that is a nice to have; it is more important we get this API change in by the code freeze.

jodygarnett:
@mprins The CSS example provided conflicts with Java Color handling of hex:

* Example 3

  In other words, ``#0000ffcc`` represents the same color as ``rgb(0 0 100% / 80%)`` (a slightly-transparent blue).

In Java this would be ``new Color( 0xcc0000ff )`` - even though the parameter is called rgba the java Color class expects argb order :)

I cannot think of a safe way to untangle this that respects both Java and CSS expectations.

jodygarnett:
I think this work has been merged.

PR #1472, opened 2017-02-10
GEOT-5637 NetCDF: Support runtime, more than 4 dimensions

NielsCharlier:
https://osgeo-org.atlassian.net/browse/GEOT-5637


jodygarnett:
What is happening on this one @NielsCharlier ?

NielsCharlier:
I would like to see it reviewed, and then merged after the code freeze. I have sent an email about it on the list.

dromagnoli:
Hi @NielsCharlier.
I see some API breaks here and there: the ones switching from Slice2Dxxx to SliceNDxxx classes and related public methods. I know that this isn't a showstopper since, at the end, I don't think there are other implementations around using those classes and methods beside GeoTools itself. So it's not a critical issue in my opinion.
However, note that the term "Slice2D" has been chosen to refer the fact that, at the end, an ImageReader always returns a 2D image (width * height RenderedImage) being extracted from an N-Dimensional Hypercube. So, given an N-Dimensional hypercube, once you define a tuple of values for each dimension composing it (as an instance, Time=T3, Elevation=Z4, Reftime=RT1), you are able to individuate and refer a 2D slice. This is the concept of Slice2DIndexManager: it allows to access a 2D slice provided the indexes of other dimensions (excluding the x,y/lon,lat dimensions since they compose the 2D slice itself). Therefore, I think the concept of Slice2Dxxx still applies. What do you think about it? Does it make sense or the add of the referencetime support introduces a different point of view?

Additional note:
I see the axis switch cases construct in the NetCDFGeoreferenceManager has been modified so that now you are adding a dimension for the default case instead of logging an "unsupported axis message".
Wondering if that will introduce some regression when dealing with "exoteric" Netcdfs (not standard CF). Wouldn't be safer to recognize and handle the reftime in a specific way? (just asking in case it is possible)

NielsCharlier:
Thank you for having a look at this.

You are right. I mistakenly thought by changing the SliceIndex to support more than 2 dimensions the name became inconsistent. But I now see that is not the case, it is referring to the inner dimensions of the slice, which remains 2, rather than the outer dimensions. I will revert this change.

To answer your second question, I don't think it will cause regression because I made these changes so that any additional dimension, no matter which one, should in theory be supported. Runtime is not treated as a special case. Only time and elevation are special cases because they play a special role in geotools and geoserver.

Also a note: this PR, once approved, will immediately be followed by a new PR that will support 2D coordinates.

modules/plugin/coverage-multidim/netcdf/src/main/java/org/geotools/imageio/netcdf/Slice2DIndex.java
```diff
@@ -159,14 +175,17 @@ public synchronized Slice2DIndex getSlice2DIndex( int imageIndex ) throws IOExce
             long endDataPosition = raf.readLong();
 
             raf.seek(dataPosition);
-            int tIndex = raf.readInt();
-            int zIndex = raf.readInt();
+            int dimensions = raf.readInt();
```
dromagnoli: Sorry, I didn't notice that before:
will this change break already configured layer in GeoServer?
I'm wondering if the new code running against a .idx being configured by previous code will throws exceptions or refers wrong values.
OFF TOPIC: Andrea recently changed the sampleImage being stored by ImageMosaic and things stopped working on already configured layers, so he needed to introduce backward-compatibility fix and a different naming for the new file. 
Wondering if this slices change needs similar checks.

modules/plugin/coverage-multidim/netcdf/src/main/java/org/geotools/imageio/netcdf/Slice2DIndex.java
```diff
@@ -204,12 +223,14 @@ public static void writeIndexFile( File file, List<Slice2DIndex> indexList, int
...
 
                 for( int i = 0; i < size; i++ ) {
-                    Slice2DIndex slice2DIndex = indexList.get(i);
+                    Slice2DIndex sliceNDIndex = indexList.get(i);
                     long pos = raf.getFilePointer();
                     pointer[i] = pos;
-                    raf.writeInt(slice2DIndex.getTIndex());
-                    raf.writeInt(slice2DIndex.getZIndex());
-                    raf.write(slice2DIndex.getVariableName().getBytes());
+                    raf.writeInt(sliceNDIndex.getNCount());
```
dromagnoli: See above comment.

modules/plugin/coverage-multidim/netcdf/src/main/java/org/geotools/imageio/netcdf/Slice2DIndex.java
NielsCharlier: You are right. It would be necessary to delete the index and recreate it. I do not know how that can be avoided though.

NielsCharlier: Hmmm, I guess I could store the dimension size as a negative value. If it can't find a negative value it assumes it is an old file and that the dimension size = 2

NielsCharlier: I added a commit that supports backwards compatibility. The presence of a negative number indicates that there is a variable dimension size.

NielsCharlier:
So is the PR approved now?

dromagnoli:
+1 from me

NielsCharlier:
Thanks Daniele!

PR #1473, opened 2017-02-10
[GEOT-5641] Mollweide projection misses parameter

marpet:
[GEOT-5641] Mollweide projection misses parameter

### DIFF
--- a/modules/library/referencing/src/main/java/org/geotools/referencing/operation/projection/Mollweide.java
+++ b/modules/library/referencing/src/main/java/org/geotools/referencing/operation/projection/Mollweide.java
@@ -34,12 +34,12 @@ import org.opengis.referencing.operation.MathTransform;
 
 /**
  * Mollweide projection
- * 
+ *
  * @see <A HREF="http://mathworld.wolfram.com/MollweideProjection.html">Mollweide</A>
  * @see <A HREF="http://en.wikipedia.org/wiki/Mollweide_projection">"Mollweide" on Wikipedia</A>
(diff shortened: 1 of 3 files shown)


modules/library/referencing/src/main/java/org/geotools/referencing/operation/projection/Mollweide.java
```diff
@@ -34,12 +34,12 @@
 
 /**
  * Mollweide projection
- * 
+ *
```
aaime: Full of whitespace changes, please remove

modules/library/referencing/src/test/java/org/geotools/referencing/operation/projection/MollweideProviderTest.java
```diff
@@ -0,0 +1,66 @@
...
+import org.opengis.parameter.ParameterDescriptorGroup;
+import org.opengis.parameter.ParameterValueGroup;
+import org.opengis.referencing.operation.MathTransform;
+
+import static org.junit.Assert.assertNotNull;
+
+/**
+ * @author Marco Peters
+ */
+public class MollweideProviderTest {
```
aaime: Could you also add a script test checking the reprojection values against a reference generated by cs2cs? Check the existing Mollweide.txt and extend it.

marpet:
Done. White spaces are removed and the script is extended.

modules/library/referencing/src/test/java/org/geotools/referencing/operation/projection/MollweideProviderTest.java
marpet: The script is updated.

PR #1474, opened 2017-02-13
GEOT-5639: Adding ability to plug in callback interfaces for JDBC classes.

jdeolive:
See https://github.com/geotools/geotools/wiki/JDBC-Callbacks

### DIFF
--- a/modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCCallbackFactory.java
+++ b/modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCCallbackFactory.java
@@ -1,0 +1,41 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ *
+ *    (C) 2017, Open Source Geospatial Foundation (OSGeo)
+ */
+package org.geotools.jdbc;
+
+/**
+ * JDBC callback factory. 
+ * <p>
+ *   This interface is used to "inject" callbacks into the various JDBC classes. 
+ * </p>
+ */
+public interface JDBCCallbackFactory {
+
+  /**
+   * Null callback factory. 
+   */
+  JDBCCallbackFactory NULL = new JDBCCallbackFactory() {
+    @Override
+    public String getName() {
+      return "null";
+    }
+
+    @Override
+    public JDBCReaderCallback createReaderCallback() {
+      return JDBCReaderCallback.NULL;
+    }
+  };
+
+  /**
+   * Callback factory name.
+   */
+  String getName();
+
+  /**
+   * Creates a callback for {@link JDBCFeatureReader}.
+   */
+  JDBCReaderCallback createReaderCallback();
+}
--- a/modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCCallbackFactory.java
+++ b/modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCCallbackFactory.java
@@ -3,6 +3,16 @@
  *    http://geotools.org
  *
  *    (C) 2017, Open Source Geospatial Foundation (OSGeo)
+ *    
+ *    This library is free software; you can redistribute it and/or
+ *    modify it under the terms of the GNU Lesser General Public
+ *    License as published by the Free Software Foundation;
+ *    version 2.1 of the License.
+ *
+ *    This library is distributed in the hope that it will be useful,
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ *    Lesser General Public License for more details.
  */
 package org.geotools.jdbc;
 
@@ -26,6 +36,11 @@ public interface JDBCCallbackFactory {
     @Override
     public JDBCReaderCallback createReaderCallback() {
       return JDBCReaderCallback.NULL;
+    }
+
+    @Override
+    public String toString() {
+      return "null callback";
     }
   };
 
--- a/modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
+++ b/modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
@@ -306,7 +306,7 @@ public final class JDBCDataStore extends ContentDataStore
      */
     protected List<ConnectionLifecycleListener> connectionLifecycleListeners = new CopyOnWriteArrayList<ConnectionLifecycleListener>();
 
-    protected JDBCCallbackFactory callbackFactory;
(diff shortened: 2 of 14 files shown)


modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCCallbackFactory.java
```diff
@@ -0,0 +1,41 @@
...
+  JDBCCallbackFactory NULL = new JDBCCallbackFactory() {
+    @Override
+    public String getName() {
+      return "null";
+    }
+
+    @Override
+    public JDBCReaderCallback createReaderCallback() {
+      return JDBCReaderCallback.NULL;
+    }
```
jodygarnett: go on give it a toString() method for us poor saps using a debugger

modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
```diff
@@ -306,8 +306,22 @@
      */
     protected List<ConnectionLifecycleListener> connectionLifecycleListeners = new CopyOnWriteArrayList<ConnectionLifecycleListener>();
 
+    protected JDBCCallbackFactory callbackFactory;
```
jodygarnett: should this default to:
```java
protected JDBCCallbackFactory callbackFactory = JDBCCallbackFactory.NULL;
```

modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStoreFactory.java
```diff
@@ -146,7 +149,20 @@
             "SQL statement executed when the connection is released to the pool", false, null,
             Collections.singletonMap(Parameter.IS_LARGE_TEXT, Boolean.TRUE));
 
-    
+    /** JDBC callback factory */
+    public static final Param CALLBACK_FACTORY = new Param("Callback factory", String.class,
```
jodygarnett: Is this is an "advanced" parameter (ie programatic use only) in which case you could pass in the instance rather than supply a string...

modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStoreFactory.java
```diff
@@ -567,4 +593,24 @@ protected String getJDBCUrl(Map params) throws IOException {
...
         return url;
     }
+
+    /**
+     * Looks up the callback factory with the specified name.
+     * <p>
+     *   Will return {@link JDBCCallbackFactory#NULL} if the factory with the specified name can't be found.
+     * </p>
+     */
+    static JDBCCallbackFactory findCallbackFactory(String factoryName) {
```
jodygarnett: guess you did not want your own "factoryfinder" since this has limited use.

modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCReaderCallback.java
```diff
@@ -0,0 +1,76 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ *
+ *    (C) 2017, Open Source Geospatial Foundation (OSGeo)
+ */
```
jodygarnett: think our header is longer

jodygarnett:
I could not see any use of {{Query.getHandle()}} for the JDBCReaderCallback.

modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStoreFactory.java
jdeolive: While it is certainly an advanced parameter it's not just intended for programatic use, it's meant to be configured by the user as well.

modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStoreFactory.java
jdeolive: Not sure what you mean. I did add a static FactoryRegistry instance for looking up instances of the factory. This method uses it.

modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCCallbackFactory.java
```diff
@@ -0,0 +1,41 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ *
+ *    (C) 2017, Open Source Geospatial Foundation (OSGeo)
+ */
```
jdeolive: Right, my bad. Will fix all the headers.

modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCCallbackFactory.java
jdeolive: Will do.

modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
jdeolive: Good call.

jdeolive:
Thanks for the review @jodygarnett . Regarding the use of Query.handle the reader didn't already have direct access to the Query object in order to pass it along to the callback. That said it looks like most code does pass in the query hints, meaning it probably has the query object handy. I can take a shoot at doing that but I think if it leads down a rabbit hole I will ask that we leave that for a future improvement.

jdeolive:
Ok, the change to update JDBCFeatureReader to take the query object was pretty mechanical and has been made. The callback now has access to the query object (through the reader) in the reader init() method.

modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStoreFactory.java
```diff
@@ -146,7 +149,20 @@
...
 
-    
+    /** JDBC callback factory */
+    public static final Param CALLBACK_FACTORY = new Param("Callback factory", String.class, 
+        "Name of JDBCReaderCallbackFactory to enable on the data store", false);
+
+    /**
+     * system property  name used to set callback factory
+     */
+    static final String CALLBACK_PROPERTY = "gt2.jdbc.callback";
```
bencaradocdavies: Perhaps call this property ``geotools.jdbc.callback``? Or are you trying to be consistent with other property names?

jdeolive: @bencaradocdavies Yeah, I was going for consistency with the existing property "gt2.jdbc.trace".... but if folks like that style better I am happy to go that way.

PR #1478, opened 2017-02-16
GEOT-5537: pass NOT instead of AND when creating NOT filters

mtoonen:
Unit test and fix for GEOT-5537 added: pass NOT filtertype to createHighLevelLogicFilter() method instead of AND.

### DIFF
--- a/modules/library/xml/src/test/java/org/geotools/xml/filter/FilterCompliancePreProcessorTest.java
+++ b/modules/library/xml/src/test/java/org/geotools/xml/filter/FilterCompliancePreProcessorTest.java
@@ -1,0 +1,79 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ * 
+ *    (C) 2017, Open Source Geospatial Foundation (OSGeo)
+ *
+ *    This library is free software; you can redistribute it and/or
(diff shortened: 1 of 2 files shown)


modules/library/xml/src/test/java/org/geotools/xml/filter/FilterCompliancePreProcessorTest.java
```diff
@@ -0,0 +1,79 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ * 
+ *    (C) 2002-20017, Open Source Geospatial Foundation (OSGeo)
```
mprins: That's a long time in the future :-)

mprins:
Would be good to get this into Monday's? 16.2  as well

modules/library/xml/src/test/java/org/geotools/xml/filter/FilterCompliancePreProcessorTest.java
bradh: Also I don't think OSGeo existed that far back (incorporated in 2006), so perhaps just this year...

mtoonen:
License header adjusted.

PR #856, opened 2015-05-25
Remove gt-wfs module [GEOT-5128]

jodygarnett:
The gt-wfs-ng module is available as drop-in replacement. We can also close open issue reported against wfs datastore

PR #1479, opened 2017-02-16
Remove AbstractDataStore and support classes (GEOT-5649)

jodygarnett:
This change removes the following two unsupported modules:

* gt-wfs - the replacement gt-wfs-ng is available (GEOT-5128)
* gt-vpf - no replacement available, module was non functionality (had not migrated to new feature model let alone ContentDataStore).

See:

* https://osgeo-org.atlassian.net/browse/GEOT-5649
* https://osgeo-org.atlassian.net/browse/GEOT-5128


jodygarnett:
This is a replacement for https://github.com/geotools/geotools/pull/856

jdeolive:
The change set is huge but I did a quick scan through... nothing jumped out as off to me. Looks good Jody. Nice to remove some old code :)

aaime:
This merge broke the GeoServer build, the NetCDF output format uses AbstractDataStore somewhere.

tbarsballe:
Potential fix here: https://github.com/geoserver/geoserver/pull/2118 (still need to run tests; don't have netcdf-out on my machine)

@jodygarnett - could I get a quick review of the above?

PR #1497, opened 2017-03-06
Use synchronized map for memory datastore concurrent access [GEOT-2515]

jodygarnett:
See https://osgeo-org.atlassian.net/browse/GEOT-2515

### DIFF
--- a/modules/library/data/src/main/java/org/geotools/data/memory/MemoryEntry.java
+++ b/modules/library/data/src/main/java/org/geotools/data/memory/MemoryEntry.java
@@ -16,6 +16,7 @@
  */
 package org.geotools.data.memory;
 
+import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.Map;
 
@@ -60,7 +61,7 @@ public class MemoryEntry extends ContentEntry {
     MemoryEntry( MemoryDataStore store, SimpleFeatureType schema){
         super( store, schema.getName() );
         this.schema = schema;
-        memory = new LinkedHashMap<String, SimpleFeature>();
+        memory = Collections.synchronizedMap(new LinkedHashMap<String, SimpleFeature>());
     }
 
     protected MemoryState createContentState(ContentEntry entry) {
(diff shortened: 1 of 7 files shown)


jodygarnett:
This change is suitable for back port.

modules/library/data/src/main/java/org/geotools/data/memory/MemoryEntry.java
```diff
@@ -70,4 +71,24 @@ protected MemoryState createContentState(ContentEntry entry) {
...
         return "MemoryEntry '" + getTypeName()+"': "+memory.size() + " features";
     }
+    
+    /**
+     * Safely add feature to {@link #memory}.
+     * <p>
+     * Feature is required to be non-null, and of the expected {@link #schema}.
+     * @param feature
+     */
+    void addFeature(SimpleFeature feature) {
```
tbarsballe: Do we want/need an equivalently synchronized removeFeature method?

modules/library/data/src/main/java/org/geotools/data/memory/MemoryEntry.java
```diff
@@ -60,7 +61,7 @@
     MemoryEntry( MemoryDataStore store, SimpleFeatureType schema){
```
tbarsballe: To enforce the synchronized functionality, should the `memory` map be made private?

jodygarnett: tricky question, I would love to say "yes", but I understand that some developers have extended this class (for example to preserve feature ids on transaction commit).

It is currently package visible; so they are unlike to get access to it - so subclasses are unlikely to get direct access to it.

jodygarnett:
merged to master as 6bf42a0396b1b928eb16b9acee27bd108964840c

jodygarnett:
This change is a bug fix, due to the API change of MemoryEntry.getMemory() it will not be back ported.

PR #1498, opened 2017-03-06
[GEOT-5632] Default PointPlacement anchorPoint does not match SLD 1.0

boundlessgeo-yancy:
The default point is now X=0, Y=0.5.

See https://osgeo-org.atlassian.net/browse/GEOT-5632


modules/library/api/src/test/java/org/geotools/styling/AnchorPointTest.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+public class AnchorPointTest {
+
+  @Test
+  public void testDefaultAnchorPoint() {
+    AnchorPoint anchorPoint = AnchorPoint.DEFAULT;
+
+    assertEquals(0.0, anchorPoint.getAnchorPointX().evaluate(null, Double.class), 0.0);
+    assertEquals(0.5, anchorPoint.getAnchorPointY().evaluate(null, Double.class), 0.0);
+  }
```
jodygarnett: Test looks good, hold I thought only one of the values was 0.0.

From https://osgeo-org.atlassian.net/browse/GEOT-5632

> According to SLD v1.0 standard on page 47,
> The default point is X=0, Y=0.5, which is at the middle height of the lefthand
side of the label.

modules/library/render/src/main/java/org/geotools/renderer/style/SLDStyleFactory.java
```diff
@@ -578,12 +578,12 @@ Style2D createPointStyle(Object feature, Symbolizer symbolizer, Graphic sldGraph
             displacementY = evalToFloat(displacement
 					.getDisplacementY(), feature, 0f);
 		}
-        float anchorPointX = 0.5f;
-        float anchorPointY = 0.5f;
+        float anchorPointX = AnchorPoint.DEFAULT.getAnchorPointX().evaluate(null, Float.class);
+        float anchorPointY = AnchorPoint.DEFAULT.getAnchorPointY().evaluate(null, Float.class);
```
jodygarnett: Okay, making reuse of SLD defaults here.

modules/library/render/src/test/java/org/geotools/renderer/style/SLDStyleFactoryTest.java
```diff
@@ -111,6 +111,10 @@ public void testCreatePointStyle(){
         NumberRange<Double> scaleRange = NumberRange.create( 1.0,50000.0);
         Style2D style = sld.createPointStyle(null, pointsym, scaleRange);
         assertNotNull( style );
+
+        PointStyle2D ps = (PointStyle2D) style;
+        assertEquals(0.0f, ps.getAnchorPointX(), 0.0f);
+        assertEquals(0.5f, ps.getAnchorPointY(), 0.0f);
```
jodygarnett: okay so you already have this covered, no need for AnchorPointTest above?

modules/library/api/src/test/java/org/geotools/styling/AnchorPointTest.java
boundlessgeo-yancy: The expected values are 0.0 for x and 0.5 for y. The 0.0 at the end of those lines are the delta values for float comparisons.

modules/library/render/src/test/java/org/geotools/renderer/style/SLDStyleFactoryTest.java
boundlessgeo-yancy: I suppose so. That is fine with me to remove that other test.

modules/library/api/src/test/java/org/geotools/styling/AnchorPointTest.java
jodygarnett: sorry read the code wrong

jodygarnett:
Okay so we have the expected test failures:

````
...
You can add -Dorg.geotools.image.test.interactive=true to show a dialog comparing them (requires GUI support)
  VectorFillTest.testFontFill:160->runSingleLayerTest:81->runSingleLayerTest:87->runSingleLayerTest:105 Images are visibly different, found 13898 different pixels, against a threshold of 100
You can add -Dorg.geotools.image.test.interactive=true to show a dialog comparing them (requires GUI support)
  VectorFillTest.testCrossFill:111->runSingleLayerTest:87->runSingleLayerTest:105 Images are visibly different, found 30285 different pixels, against a threshold of 300
You can add -Dorg.geotools.image.test.interactive=true to show a dialog comparing them (requires GUI support)
  WrapPointSymbolizerTest.testWrapPointSymbolizer:66 expected:<java.awt.Color[r=255,g=0,b=0]> but was:<java.awt.Color[r=0,g=0,b=0]>

Tests run: 868, Failures: 143, Errors: 0, Skipped: 2
````

jodygarnett:
Okay, doubling back to the bug report - we have a different default value between SLD 1.0 and SE 1.1 (sigh). So there is not going to be a clear right answer. 

I think we could go for the SLD 1.0 default; and double check that the SE 1.1 parser fills in the correct value when it creates the LabelPlacement.

jodygarnett:
Thanks @boundlessgeo-yancy - adding "org.geotools.renderer.style.legacyAnchorPoint" takes some of the risk out of this change for those upgrading. You are a champ.

jodygarnett:
merged, thanks for taking part in the bug stomp.

PR #1499, opened 2017-03-07
[GEOT-5672] MongodDB Add support for 'Number' literal filter

newmanw:
If filter literal is a number leave it alone.

### DIFF
--- a/modules/unsupported/mongodb/src/main/java/org/geotools/data/mongodb/FilterToMongo.java
+++ b/modules/unsupported/mongodb/src/main/java/org/geotools/data/mongodb/FilterToMongo.java
@@ -621,6 +621,8 @@ public class FilterToMongo implements FilterVisitor, ExpressionVisitor {
             }
             // by default, return string as is
             return literal;
+        } else if (literal instanceof Integer || literal instanceof Long || literal instanceof Double) {
+            return literal;
         } else {
             return literal.toString();
         }
--- a/modules/unsupported/mongodb/src/test/java/org/geotools/data/mongodb/FilterToMongoTest.java
+++ b/modules/unsupported/mongodb/src/test/java/org/geotools/data/mongodb/FilterToMongoTest.java
@@ -64,6 +64,24 @@ public class FilterToMongoTest extends TestCase {
         assertEquals(1, obj.keySet().size());
         assertEquals("bar", obj.get("properties.foo"));
     }
+    
+    public void testEqualToInteger() throws Exception {
+        PropertyIsEqualTo equalTo = ff.equals(ff.property("foo"), ff.literal(10));
+        BasicDBObject obj = (BasicDBObject) equalTo.accept(filterToMongo, null);
+        assertNotNull(obj);
+
+        assertEquals(1, obj.keySet().size());
+        assertEquals(10, obj.get("properties.foo"));
+    }
+    
+    public void testEqualToLong() throws Exception {
+        PropertyIsEqualTo equalTo = ff.equals(ff.property("foo"), ff.literal(10L));
+        BasicDBObject obj = (BasicDBObject) equalTo.accept(filterToMongo, null);
+        assertNotNull(obj);
+
+        assertEquals(1, obj.keySet().size());
+        assertEquals(10L, obj.get("properties.foo"));
+    }
 
     public void testBBOX() throws Exception {
         BBOX bbox = ff.bbox("loc", 10d,10d,20d,20d, "epsg:4326");


jodygarnett:
Do you have a bug report for this fix? I am trying to sort out if it is a fix (so I can merge it during the code freeze) or a new feature (in which case we add it in next week).

modules/unsupported/mongodb/src/main/java/org/geotools/data/mongodb/FilterToMongo.java
```diff
@@ -621,6 +621,8 @@ Object encodeLiteral(Object literal, Class<?> targetType) {
             }
             // by default, return string as is
             return literal;
+        } else if (literal instanceof Number) {
+            return literal;
```
jodygarnett: I do not know what kinds of Numbers mongodb can hold, are we sure this won't trip up on BigDecimal or Byte? both of which are considered numbers.

newmanw:
@jodygarnett I did not write a ticket as I was unsure of the process for an unsupported module.  If there needs to be a ticket let me know and I can create one.

modules/unsupported/mongodb/src/main/java/org/geotools/data/mongodb/FilterToMongo.java
newmanw: Good point!

Looks like: 

double
32-bit integer
64-bit integer
decimal

From: https://docs.mongodb.com/manual/reference/operator/query/type/

Might be better suited to check for Integer, Double, and Long only.   Thoughts?

jodygarnett: I do not get a chance to work with it; if you check for Integer, Double, Long only ... throw an appropriate unsupported exception for the case you miss.

tbarsballe: Doing an Integer / Double / Long check might be a good idea. 

You could always add another test to see what happens to mongodb when you give it a filter literal containing a BigDecimal or Byte.

newmanw: Agreed, checked the driver docs and for now I think Int, Long Double only is the best option.

@jodygarnett you mention, 'throw an appropriate unsupported exception for the case you miss.'  That was my initial thought as well, but didn't fully understand the original intention of this, which made me leery of that change:

https://github.com/newmanw/geotools/blob/9c0f61669551b6e04b6a7ab6c81b2cd6efdbe482/modules/unsupported/mongodb/src/main/java/org/geotools/data/mongodb/FilterToMongo.java#l626-L628 

which was the original implementation.  Thoughts?

tbarsballe: Given the original implementation, returning doing a toString when there is no matching type is probably safer.

An UnsupportedException is likely more correct, but runs the risk of breaking backwards compatibility.

I'd say return a supported number where applicable, and otherwise do a toString().

newmanw: Agreed!  I have updated the fix to return the literal if Long, Integer or Double.  For backwards compatibility the else still returns literal.toString().

Thanks for all the help and input, and please let me know if this PR needs anything else.

jodygarnett:
this branch now has conflicts?

newmanw:
@jodygarnett, yeah looks like this commit from @nmco pulled in a converter:

07b2e2def9faeb3e649486e69fb5dad630688c40

I will need to look at the converters to see what this does.

nmco:
Converters try to find a converter able to convert the input type to the target type, if they don't succeed NULL is returned (if my memory serves me well).

newmanw:
@nmco no worries. 

'If the target type is not supported maybe we should not being using it or well at a specific handling for it like dates and geometries.' Agree, this pull request checks for Long, Integer or Double since mongo can handle those.  Returns literal.toString() if not a supported type to maintain backwards compatibility.

At this point I am not sure if best to kill off this pull request in favor of an issue that would outline that there may be a potential issue with unsupported data types.

nmco:
I would keep this pull request tests cases and create an issue describing the unsupported types problem.

A fix for the unsupported types problem could be:
- having a list of MongoDB supported [natives types](http://mongodb.github.io/mongo-java-driver/3.0/bson/documents/)
- ask the converted to try to convert to the target type
- if the conversion result is supported by MongoDB we use it otherwise we do a ``toString()``

newmanw:
@nmco the issue I have with this is that the code you added introduced this problem.  Your merge will now potentially crash if passed a BigDecimal.

This pull request did exactly what you outlined: 

> having a list of MongoDB supported natives types
> ask the converted to try to convert to the target type
> if the conversion result is supported by MongoDB we use it otherwise we do a toString()

IE in this pull request:
```
} else if (literal instanceof Integer || literal instanceof Long || literal instanceof Double) {
  return literal;
} else {	
   return literal.toString();	
}
```

You could certainly argue the the 'list' of supported data types that is defined within the else check is crude, but it does solve the problem, while maintaining backwards compatibility with the code before your merge.

nmco:
@newmanw thanks for the issue.

Pull request that fix the issues discussed here: https://github.com/geotools/geotools/pull/1555
I imported this PR tests cases, so this PR can be closed.

@newmanw let me know if this looks good to you.

newmanw:
@nmco thanks so very much for being so accommodating, your PR looks great.

PR #1500, opened 2017-03-08
[GEOT-5671] MongoDB Add support for polygon filter bigger than a hemisphere

newmanw:
Mongo recently added support for polygon queries that are bigger than a hemisphere:

https://docs.mongodb.com/manual/reference/operator/query/geometry/#op._S_geometry

Added code to support that feature.

### DIFF
--- a/modules/unsupported/mongodb/src/main/java/org/geotools/data/mongodb/FilterToMongo.java
+++ b/modules/unsupported/mongodb/src/main/java/org/geotools/data/mongodb/FilterToMongo.java
@@ -112,7 +112,8 @@ public class FilterToMongo implements FilterVisitor, ExpressionVisitor {
 
     private static final Logger LOGGER = Logging.getLogger(FilterToMongo.class);
 
-    static final SimpleDateFormat ISO8601_SDF = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSXXX");
+    static final SimpleDateFormat ISO8601_SDF = new SimpleDateFormat(
+            "yyyy-MM-dd'T'HH:mm:ss.SSSXXX");
 
     final CollectionMapper mapper;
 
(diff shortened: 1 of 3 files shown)


jodygarnett:
Thanks @newmanw - we have code freeze on this week and then we can get to this.

Can I ask that you make a bug report - so this change shows up in the release notes.

newmanw:
Sure np, issue created.  [GEOT-5671]

jodygarnett:
Code freeze is over and this can be merged to master :)

modules/unsupported/mongodb/src/main/java/org/geotools/data/mongodb/FilterToMongo.java
```diff
@@ -403,9 +403,14 @@ public Object visit(BBOX filter, Object extraData) {
...
-        DBObject dbo = BasicDBObjectBuilder.start().push("$geoIntersects")
-                .add("$geometry", geometryBuilder.toObject(envelope)).get();
+                
+        DBObject geometryDBObject = geometryBuilder.toObject(envelope);
+        addCrsToGeometryDBObject(geometryDBObject);
+        DBObject dbo = BasicDBObjectBuilder
+        		.start()
+        		.push("$geoIntersects")
+                .add("$geometry", geometryDBObject)
+                .get();
```
jodygarnett: did your code formatter get to this? Or did you do it by hand :)

bencaradocdavies: @newmanw hand-formatting will always attract attention, but is in my view acceptable when formatting a long method call chain because such a chain can be like a Domain Specific Language. Similarly, inline XML, SQL, or WKT can benefit from hand-formatting.

@newmanw are you using Eclipse? The project Eclipse formatter configuration supports comment tags to turn formatting off and on. I find these most useful when hand-formatting code:
```
    // @formatter:off
   [hand formatted code goes here]
    // @formatter:on
```

newmanw: Sorry, that was all me :).   I made sure I had the templates/formatter imported and did a re-format.  When you get a chance take a look and let me know if it looks better.  

Thanks!

PR #1501, opened 2017-03-08
[GEOT-5670] MongoDB Fix 'And' filter

newmanw:
* The And filter builder had this comment 'Expressions like ((A == 1) AND (B == 2)) are basically implied. So just build up all sub expressions'.
  Which is true for an AND expressions with 2 different properties, however for an AND on the same property '((A > 1) AND (A <= 2))'
  this expression will overwrite the first expression with the second since the key is the same.
  IE { "A" : { "$lte" : 2 } }, rather than the correct expression { "$and" : [{ "A" : { "$gt" : "0" } }, { "A" : { "$lt" : "10" } }] }

### DIFF
--- a/modules/unsupported/mongodb/src/main/java/org/geotools/data/mongodb/FilterToMongo.java
+++ b/modules/unsupported/mongodb/src/main/java/org/geotools/data/mongodb/FilterToMongo.java
@@ -189,17 +189,17 @@ public class FilterToMongo implements FilterVisitor, ExpressionVisitor {
     // logical
     //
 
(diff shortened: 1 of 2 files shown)


jodygarnett:
Can I ask that you report a bug so this change shows up in the release notes. Since this is a bug fix I would like to get it in *today* before the RC1 goes out.

newmanw:
@jodygarnett no problem, bug created.  [GEOT-5670]

tbarsballe:
Fix/code looks good

PR #1502, opened 2017-03-08
[GEOT-5667] Fix app-schema property substitution group support

braeckel:
https://osgeo-org.atlassian.net/browse/GEOT-5667

Per the email thread http://bit.ly/2neUxoT, when no properties are found for the declared XSD type, XSD substitutionGroups are then checked as well.  This fix specifically addresses an issue observed where substitutable types were not being utilized by appschema when generating XML/GML, but this change will also apply to other uses of this method.

I would like to submit a test for both this change and the original behavior, but this method requires a number of complex GeoTools objects.  AbstractFeatureType_getProperties() requires a Feature, a SchemaIndex, a Configuration, and an XSDElementDeclaration.  I've been able to construct all the appropriate objects except the Feature, which appears quite complex to build by hand.  If you have suggestions on how to simply construct a test Feature object without introducing new dependencies I am happy to create tests.

It is probably also desirable from my end to backport this into 2.9.x and/or 2.10.x maintenance releases.


bencaradocdavies:
@braeckel you are definitely going to need unit tests for this; your fix is plausible and straightforward, but you will also want to make sure we do not break it in the future.

Because of the complexity of constructing app-schema test fixtures, the most workable solution to create or extend a test case in GeoServer gs-app-schema-test. Yes, this does mean that failures caused by GeoTools changes will only be noticed in GeoServer, but they will be noticed. Your reviewer can make two local builds to check beforehand. There are many schemas available in gs-app-schema-test, both published schemas bundled in Maven artifacts and local synthetic schemas that you might be able to extend to demonstrate this case. The advantage of testing in GeoServer-land is that you can test the fully encoded GeoServer XML output. It is not unusual for app-schema changes to comprise two parallel pull requests, one each for GeoTools and GeoServer.

Please create a Jira issue for this improvement to ensure that it is recorded in release notes. It is common to mention GeoTools Jira issue tags in GeoServer commit messages and pull request titles, especially for app-schema changes. I would tend to use one Jira issue in the project with the implementation (and for app-schema that is entirely in GeoTools).

braeckel:
@bencaradocdavies I've created a GeoServer unit test and pull request: https://github.com/geoserver/geoserver/pull/2156 to complement this GeoTools pull request. Sorry for the mismatched branch names but it seems clearer to have the branch named after the JIRA issue overall.

I am happy to backport to 2.10.x and 2.9.x once this is finalized.

bencaradocdavies:
Jira and PR title updated to be a bit more release-note friendly.

I do not like merges on pull request branches as it makes it harder to see what it going on. I am testing a local squash (with your authorship preserved).

bencaradocdavies:
This PR seems to cause https://github.com/geoserver/geoserver/pull/2156 to fail in GeoServer gs-app-schema-test ```TimeSeriesWfsTest```.

bencaradocdavies:
@braeckel have you had a chance to look at this?

bencaradocdavies:
@braeckel I found the problem: the CSML schema used for GeoServer `TimeSeries*Test` uses a type `csml:PointSeriesCoverage` that both inherits a property `gml:domainSet` and has a direct property `csml:pointSeriesDomain` which is in the `gml:domainSet` substitution group. This PR changed the resolution order and broke some assumption in the encoder.

I manually squashed this PR into a commit on master:

[GEOT-5667] Added support in AbstractFeatureType_getProperties() for reading properties which are substitutable (in the sense of XSD substitutionGroups) for the type declared in the XSD.  When no properties are found with the declared type the substitutable types are checked and used when applicable
https://github.com/geotools/geotools/commit/8e6c652cf0fb894edebb91e153082b05d85320b3

I then added a fix:

[GEOT-5667] Ensure that substitution groups are only used to locate properties that cannot be found by other means
https://github.com/geotools/geotools/commit/d6528a31d8b03d973941f99b960c6a5e3922bb76

Would you like me to backport these changes?

braeckel:
A backport to GEOS 2.10 and the associated GEOT version (16?) would be very helpful, thanks.

bencaradocdavies:
16.x/2.10.x is the current maintenance. We will also need this on 17.x/2.11.x (stable). My one concern is the performance impact. I do not think it will be too noticeable. I am tempted to ask on the list. This is a bug fix so is eligible for backport to maintenance.

bencaradocdavies:
@braeckel I want to benchmark the performance impact before backporting to stable, and especially before backporting to maintenance.

PR #1503, opened 2017-03-09
[GEOT-5668] Tomcat classloader does not allow de-serialization of ImageMosaic SampleImage

aaime:
@dromagnoli this should fix the issue we were discussing yesterday. Could you double check it?


dromagnoli (approved):
Looks good to me.
This change allows to be able to deal with new style sampleImages as well as old ones (based on serializableRenderedImage) so it should guarantee retrocompatibility.

jodygarnett:
As this is a bug, will you be backporting to the stable branch?

aaime:
Noooooh... it breaks the geoserver build, I created the pull only to discuss with Daniele...

PR #1508, opened 2017-03-16
Use java 8 Base64 Encoder / Decoder [GEOT-5676]

jodygarnett:
Now that functionality is public we can avoid use of sun.misc.BASE64Decoder / sun.misc.BASE64Encoder classes.

This avoids a compile failure in Java 1.8.0_102 and Eclipse 4.6.2

See https://osgeo-org.atlassian.net/browse/GEOT-5676

### DIFF
--- a/modules/extension/xsd/xsd-core/pom.xml
+++ b/modules/extension/xsd/xsd-core/pom.xml
@@ -70,17 +70,6 @@
     </developer>
   </developers>
 
-
-  <dependencyManagement>
-    <dependencies>
-      <dependency>
-        <groupId>commons-collections</groupId>
-        <artifactId>commons-collections</artifactId>
-        <version>3.2.2</version>  <!-- Updated for security (CVE-2015-8103) binary compatible with 3.1 -->
-      </dependency>
-    </dependencies>
-  </dependencyManagement>
-
   <!-- =========================================================== -->
   <!--     Dependency Management                                   -->
   <!-- =========================================================== -->
@@ -158,7 +147,6 @@
       <scope>test</scope>
     </dependency>
   </dependencies>
-
 
 </project>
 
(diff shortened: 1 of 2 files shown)


jodygarnett:
This PR is suitable for backport.

tbarsballe:
Looks good, though I am a bit curious about the seemingly unrelated removal of the commons-collections dependency.
Also, is there a jira ticket for this issue?

jodygarnett:
dependency management was already covered at the top level, not sure what it was doing duplicated here.

modules/extension/xsd/xsd-core/pom.xml
```diff
@@ -70,17 +70,6 @@
...
-    <dependencies>
-      <dependency>
-        <groupId>commons-collections</groupId>
-        <artifactId>commons-collections</artifactId>
-        <version>3.2.2</version>  <!-- Updated for security (CVE-2015-8103) binary compatible with 3.1 -->
-      </dependency>
-    </dependencies>
-  </dependencyManagement>
-
   <!-- =========================================================== -->
```
jodygarnett: this dependency management version 3.2.2 is already covered in root pom.xml (also we do not use dependencyManagement in child pom.xml files right?)

tbarsballe: Makes sense. Only reason I could see doing so if this is the parent of a bunch of submodules, or if it needs a specific version of commons-collections different from everything else (which seems like a bad idea / not viable anyway)
Git blame shows this bit has been part of the code as long as it has been in git.

If this is already in the parent pom I have no issues (I had initially missed that it was in a dependencyManagement section rather than just being a dependency)

tbarsballe:
Backported to 17.x and 16.x

PR #1509, opened 2017-03-16
GEOT-5675: Division by zero in NetCDFPolyphemusTest

dromagnoli:
https://osgeo-org.atlassian.net/browse/GEOT-5675


bencaradocdavies:
Thanks @dromagnoli, that looks great! The removal of the ```catch``` ensures that any future exceptions will fail the build.

bencaradocdavies:
Once the ares build on master has passed, this fix could be backported to 17.x and 16.x.

PR #1512, opened 2017-03-17
[GEOT-5677] Band selection via request param and RBG expansion are incompatible

aaime:


aaime:
RGB expansion down the road won't happen because the color model after expansion does not become a IndexedColorModel anymore, and BaseSubmosaicProducer checks that.

nmco:
Looks good to me.

nmco:
One note, I don't think that the NetCDF reader supports palleted images otherwise the same exception will raise. In the future if new readers that support palleted images  start supporting bands selection they will need somehow to perform the RGB expansion.

aaime:
@nmco yep, or as an alternative, mosaic needs to have a way to figure a read can do band selection, and don't pass it down in case palette expansion is needed (after all, indexed means reading just one band, so there is no optimization to be done in that case, you read one anyways, expand, and then select, no point in replicating the behavior in all ancillary readers).

dromagnoli:
and/or at the upper level, get the ImageLayout from the reader and see if it has an IndexColorModel cm and in that case avoiding sending down the BANDS parameter (as part of the optimization)

aaime:
@dromagnoli I followed up with an extra commit, trying to implement your suggestion, did I get it right?

dromagnoli:
@aaime, @nmco  Looks good to me.
Do you think we should do similar checks on the other pieces of code applying the band selection, too?
I think on GridCoverageRenderer and/or in RenderedImageMapOutputFormat where the GridCoverage2DReader bands parameter is set.

aaime:
No external control needed, the image mosaic expands to RGB and should manage band selection accordingly.

aaime:
Based on our discussion I've also opened https://osgeo-org.atlassian.net/browse/GEOT-5681

PR #1517, opened 2017-03-21
[GEOT-5686] Shapefile dumper throws a NPE on NULL geometry values

nmco:
This PR fix the dump of features to shapefiles that don't have a geometry type defined (they use the generic Geometry type) and have NULL geometries.

A test case was improved a new test case added.

Associated issue: https://osgeo-org.atlassian.net/browse/GEOT-5686


simboss:
we need to backport this as well.

aaime:
+1

PR #1521, opened 2017-03-22
[GEOT-5684] List Feature collection bounding box fixes

goudine:
Ensure that correct bounding boxes are retained after deleting and adding a feature.

For more details:
- https://osgeo-org.atlassian.net/browse/GEOT-5684

### DIFF
--- a/modules/library/main/src/test/java/org/geotools/data/collection/ListFeatureCollectionTest.java
+++ b/modules/library/main/src/test/java/org/geotools/data/collection/ListFeatureCollectionTest.java
@@ -17,27 +17,33 @@
 
 package org.geotools.data.collection;
 
-import org.junit.Before;
-import java.util.Random;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+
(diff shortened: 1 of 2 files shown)


modules/library/main/src/test/java/org/geotools/data/collection/ListFeatureCollectionTest.java
```diff
@@ -132,6 +138,64 @@ public void iterator() {
...
+        featureCollection.add(f1);
+        featureCollection.add(f2);
+
+        // get the bounds with two features, remove and add another feature
+        ReferencedEnvelope origBounds = featureCollection.getBounds();
+        featureCollection.remove(f2);
+        featureCollection.add(f3);
+        ReferencedEnvelope postRemoveBounds = featureCollection.getBounds();
+        assertNotEquals(origBounds, postRemoveBounds);
+        assertNotEquals(postRemoveBounds, f3.getBounds());
```
jodygarnett: you may wish use contains to verify if f3 is contained by postRemoveBounds?

jodygarnett:
Thanks for addressing my feedback, merging now. I would ask that we hold off on backporting this fix until Monday (as there is a security vulnerability which may necessitate rereleasing the 2.11.x branch).

PR #1522, opened 2017-03-22
[GEOT-5683] fix MemoryFeature Store & Writer to preserve IDs

goudine:
Reinstates ability to preserve feature ID's.

For more information - https://osgeo-org.atlassian.net/browse/GEOT-5683

### DIFF
--- a/modules/library/data/src/main/java/org/geotools/data/memory/MemoryFeatureStore.java
+++ b/modules/library/data/src/main/java/org/geotools/data/memory/MemoryFeatureStore.java
@@ -96,9 +96,6 @@ public class MemoryFeatureStore extends ContentFeatureStore {
         return delegate.handleVisitor(query, visitor);
     }
 
-    /**
-     *  GEOT-5683] preserve FeatureIDd during add feature
-     */
     @Override
     protected QueryCapabilities buildQueryCapabilities() {
         return new QueryCapabilities() {
--- a/modules/library/data/src/main/java/org/geotools/data/memory/MemoryFeatureStore.java
+++ b/modules/library/data/src/main/java/org/geotools/data/memory/MemoryFeatureStore.java
@@ -21,6 +21,7 @@ import java.io.IOException;
 import org.geotools.data.FeatureReader;
 import org.geotools.data.FeatureWriter;
 import org.geotools.data.Query;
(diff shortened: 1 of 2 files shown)


modules/library/data/src/main/java/org/geotools/data/memory/MemoryFeatureStore.java
```diff
@@ -94,4 +95,17 @@ protected int getCountInternal(Query query) throws IOException {
     protected boolean handleVisitor(Query query, FeatureVisitor visitor) throws IOException {
         return delegate.handleVisitor(query, visitor);
     }
+
+    /**
+     *  GEOT-5683] preserve FeatureIDd during add feature
```
jodygarnett: I tend to only mention bug reports in test cases, GEOT-5683 will not add value to someone trying to learn how to use MemoryDataStore for the 1st time.

modules/library/data/src/main/java/org/geotools/data/memory/MemoryFeatureWriter.java
```diff
@@ -111,7 +115,15 @@ public void write() throws IOException {
...
         }
-    
+        //[GEOT-5683] preserve FeatureIDd during add feature
+        if (Boolean.TRUE.equals(current.getUserData().get(Hints.USE_PROVIDED_FID))) {
+            if (current.getUserData().containsKey(Hints.PROVIDED_FID)) {
+                String fid = (String) current.getUserData().get(Hints.PROVIDED_FID);
+                FeatureId id = new FeatureIdImpl(fid);
+                current = new SimpleFeatureImpl(current.getAttributes(), current.getFeatureType(), id);
+            }
+        }
```
jodygarnett: Where is the test case to verify this functionality?

mtCarto: The MemoryDataStoreTest contains two tests that provide coverage of this functionality, _testGetFeatureStoreTransactionSupport_ and _testGetFeatureWriterTransaction_. These tests use a DefaultTransaction that sets the **USE_PROVIDED_FID** flag to true. I can add another test that sets this flag manually if necessary

PR #1527, opened 2017-03-24
integration of checkstyle into geotools build process

wumpz:
As for I am having so much difficulties to produce correct code for geotools, mainly in setting up my coding environment to the right standards, it would be nice to have a checkstyle check, rightly configured to match geotools standards.

For now this simple integration only checks for tab characters within source code files. 

It is integrated within a separat non active profile **check.sources**.

One could start it at every submodule of geotools to get the validation or violation of the rules by

checkstyle:check -Pcheck.sources

As for now there are many "old" sources that do not follow geotools own coding convention (e.g. tab characters). But if it does, this checking style could be a part of the main build.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -214,9 +214,17 @@
                         <checkstyleRules>
                             <module name="Checker">
                                 <module name="FileTabCharacter" />
+                                <!-- git checkout may change linefeeds on the fly
                                 <module name="RegexpMultiline">
                                     <property name="format" value="(?s:(\r\n|\r).*)"/>
                                     <property name="message" value="CRLF and CR line endings are prohibited, but this file uses them."/>
+                                </module>
+                                -->
+                                <module name="TreeWalker">
+                                    <module name="ConstantName" />
+                                    <module name="EmptyStatement" />
+                                    <module name="NeedBraces" />
+                                    <module name="UnnecessaryParentheses" />
                                 </module>
                             </module>
                         </checkstyleRules>
--- a/pom.xml
+++ b/pom.xml
@@ -214,6 +214,10 @@
                         <checkstyleRules>
                             <module name="Checker">
                                 <module name="FileTabCharacter" />
+                                <module name="RegexpMultiline">
+                                    <property name="format" value="(?s:(\r\n|\r).*)"/>
+                                    <property name="message" value="CRLF and CR line endings are prohibited, but this file uses them."/>
+                                </module>
                             </module>
                         </checkstyleRules>
                     </configuration>
--- a/pom.xml
+++ b/pom.xml
@@ -192,6 +192,35 @@
         </site>
       </distributionManagement>
     </profile>
+    <profile>
+        <id>check.sources</id>
+        <build>
+            <plugins>
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-checkstyle-plugin</artifactId>
+                    <version>2.17</version>
+                    <executions>
+                        <execution>
+                            <id>verify-style</id>
+                            <phase>process-classes</phase>
+                            <goals>
+                                <goal>check</goal>
+                            </goals>
+                        </execution>
+                    </executions>
+                    <configuration>
+                        <logViolationsToConsole>true</logViolationsToConsole>
+                        <checkstyleRules>
+                            <module name="Checker">
+                                <module name="FileTabCharacter" />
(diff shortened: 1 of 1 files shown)


aaime:
I started a thread on this topic some time ago, did not get any traction:
http://osgeo-org.1560.x6.nabble.com/Formatting-formatting-getting-tired-of-all-complaints-td5307518.html

wumpz:
I see. But the situation now is very difficult without  such tools. 
The advantage of a maven plugin is that both developer and merger could use it with the same configuration. Since it is within a separate profile IMHO build time is not a problem.

With such a tool it can only be better than now.

aaime:
It's not up to me to accept this pull request, all maintainers need to agree to clean up their modules, otherwise it's just a dead tool that nobody will use. You might want to review the discussion again and share your opinion in it.

wumpz:
You are saying, that not all maintainers following geotools coding conventions? Why?

bencaradocdavies:
@wumpz GeoTools was founded in 1996. There is a lot of code that predates the coding conventions. Modules maintainers leave and new ones take over. The problem with reformatting is that it obscures history. There have been a few opportunities to reformat the entire code base, such as when we truncated history in the migration from Subversion to Git, but we did not reformat at that time.

There are also omissions: the style guide is silent on XML. The code base is full of poms and applicationContexts that are a sea of tabs and varying amounts of spaces, often in the same file.

bencaradocdavies:
@wompz what IDE are you using? The Eclipse formatter in the GeoTools project works well for me. It can also be applied to local parts of a file by selecting them before formatting.

wumpz:
@bencaradocdavies: I do not agree. 

If one installes coding conventions then they have to rule and should not be partially applied. 

You are right, that most of reformating does obscure history. The question is, to which level. But if you do not make a cut, then you will have mixed sourcecode files, if there was an update, and you are waiting til the last change, when this one concrete SourceCode file will match your coding conventions. Til when do you want to wait?

My main problem at the moment were tabs and spaces. Most diff tools are able to filter that out, so there is no visualy impact on history. So whats the point in not doing a complete "whitespace" reformat?

At my last PR I had also sld files with tabs in it. If you say, that these files are not covered by coding conventions, then that should be changed.

For now you have to decide what costs more time and nerves, 
* a look back into a at one specific point in time slightly obscured history, but a tool to check coding conventions nearly automatically for developers and mergers, 
* or mergers that have to check all those conventions manually? but with a very pretty history.

wumpz:
I invite you to run to run `checkstyle:check -Pcheck.sources` on **gt-main** or **gt-render**.

Sure this configuration should be more specific about file types and more formating stuff but it draws a clear picture.

wumpz:
By the way, why is travis messaging errors for this commit, while its profile is not active and the pom.xml is syntactically correct?

bencaradocdavies:
@wumpz [Travis CI Status](https://twitter.com/traviscistatus) reports an outage for scheduled maintenance. I cannot see any logs. I have restarted the build for this pull request.

wumpz:
I have tested some more rules. 

The linefeed check does not survived, because of gits possibility to change linefeeds on checkout.

Some more tests, like NeedBraces, which is I think forced by Geotools coding conventions, does deliver tons of problematic lines.

wumpz:
Just to add these investigations. These are two code snippets out of gt-main:
**CommonsConverterFactory**

```
if( source == null ) return null;
       String string = (String) source;
       if( string.length() > 0 ){
           return target.cast( string.charAt(0) );
       }
```

**ListenerList**

```
if (size == 0) {
     listeners = new Object[capacity];
} else {
```

If you look at the spaces used at braces ... 
A reformat of **ListenerList**, does provide good results and does not obscure history, regarding blaming so much. **CommonsConverterFactory** is the one, that seems to be much more problematic.

By the way I will register at geotools developers mailing list. As I saw the discussion is revived again ;).

aaime:
We have an incoming proposal to reformat and then check the code, but we won't use checkstyle, but the "google style" machinery instead, due to the amont of changes needed to make checkstyle work (automatic formatting using intellij or eclipse won't generate files that satisfy checkstyle, we'd have to make thousands of manual changes instead).
So, closing this one.

wumpz:
@aaime That is so good to hear. 👍 I hope you plan a complete reformat of the code and not one after the other and to put this reformating "machinery" into the build process.

PR #1531, opened 2017-03-28
GEOT-5702 NetCDF: support 2D coordinate variables, aggregate different sizes, f…

NielsCharlier:
…eature collections, auxiliary WFS

https://osgeo-org.atlassian.net/browse/GEOT-5702

GeoServer documentation: https://github.com/geoserver/geoserver/pull/2276


NielsCharlier:
@dromagnoli could you have a look at this one when you get the chance please?

aaime:
@dromagnoli is away in honeymoon, he will be back at the end of the month

modules/plugin/coverage-multidim/netcdf/src/main/java/org/geotools/imageio/netcdf/utilities/NetCDFUtilities.java
```diff
@@ -636,6 +659,93 @@ public static boolean isVariableAccepted(final String name,
             return true;
         }
     }
+        
+    public static FileFormat getFormat(URI uri) throws IOException {
```
bencaradocdavies: Note: this is refactored with the existing code pulled out into a method and some XML handling logic added.

modules/plugin/coverage-multidim/netcdf/src/main/java/org/geotools/imageio/netcdf/utilities/NetCDFUtilities.java
```diff
@@ -651,17 +761,20 @@ public static boolean isVariableAccepted(final String name,
...
         NetcdfDataset dataset = null;
-        if (input instanceof File) {
+        if (input instanceof URI) {
+            dataset = acquireDataset((URI) input);
+        }
+        else if (input instanceof File) {
             final File file= (File) input;
             if (!file.isDirectory()) {
-                dataset = NetcdfDataset.acquireDataset(file.getPath(), null);
+                dataset = acquireDataset(file.toURI());
```
bencaradocdavies: All these ```File.toURI()``` are an improvement.

modules/plugin/coverage-multidim/netcdf/src/test/java/org/geotools/coverage/io/netcdf/NetCDFMultiDimTest.java
```diff
@@ -158,7 +333,204 @@ private void checkFoursome(NetCDFReader reader, GeneralParameterValue[] pams, do
...
+        checkFoursome(reader, new GeneralParameterValue[] { zValue, timeValue, runtimeValue }, 32, 33, 34, 35);
+
+        //z = 1, time = 3, runtime = 3;
+        zValue.setValue(Collections.singletonList(1.0));
+        timeValue.setValue(Collections.singletonList(DF.parse("2012-04-01 03:00:00")));
+        runtimeValue.setValue(Collections.singletonList(DF.parse("2012-04-01 03:00:00")));
+
+        checkFoursome(reader, new GeneralParameterValue[] { zValue, timeValue, runtimeValue }, 36, 37, 38, 39);
+
+
```
bencaradocdavies: Multiple blank lines.

modules/plugin/coverage-multidim/netcdf/src/test/java/org/geotools/coverage/io/netcdf/NetCDFMultiDimTest.java
```diff
@@ -150,6 +162,169 @@ public void test() throws Exception {
         
     }
     
+    @SuppressWarnings("rawtypes")
```
bencaradocdavies: @NielsCharlier this is pretty thorough unit test coverage.

bencaradocdavies:
And I almost forgot to mention: a full build of this PR merged onto master and a subsequent full GeoServer build both pass for me locally.

NielsCharlier:
@bencaradocdavies Thanks for your review. You are right, this should have been a geotools issue. This must have been an error on my part.

Regarding the auxiliary WFS, you have to manually publish it as a vector data store, AFTER making the netcdf datastore

NielsCharlier:
Formatting errors are fixed, issue moved to geotools, title changed.

modules/plugin/coverage-multidim/netcdf/src/main/java/org/geotools/coverage/io/netcdf/NetCDFAuxiliaryStoreFactory.java
```diff
@@ -0,0 +1,236 @@
...
+    public boolean canProcess(Map<String, Serializable> params) {
+        try {
+            File file = (File) FILE_PARAM.lookUp(params);
+            String indexPath = (String) INDEX_PARAM.lookUp(params);
+            String dsPath = (String) DS_PARAM.lookUp(params);
+            AncillaryFileManager ancilaryFileManager = new AncillaryFileManager(file, indexPath, dsPath);
+            DataStoreConfiguration datastoreConfig = ancilaryFileManager.getDatastoreConfiguration();
+            return datastoreConfig.getDatastoreSpi().canProcess(datastoreConfig.getParams());            
+        } catch (NoSuchAlgorithmException | JAXBException | IOException e) {
+
```
bencaradocdavies: You should be able to move the ```return false``` to the ```catch``` block for clarity as all path then have a valid return.

bencaradocdavies:
@NielsCharlier thanks for those formatting fixes. This pull request still requires review from a maintainer such as @dromagnoli (when he gets back) or @simboss.

I still find the name and parameters for the auxiliary data store to be unclear. This store will be offered to all users of the GeoServer NetCDF plugin. There should be some documentation of the relationship between the auxiliary store and the main NetCDF store. I also wonder whether this data store should be hidden by default as it might confuse users.

NielsCharlier:
Thanks @bencaradocdavies 
I will make that last code change as well.

About the store: can you clarify - the parameters of the auxiliary store are identical to the regular store. Do you have an alternative for the paramters and the name?
How would you hide the store and still make it possible for users to unhide it?

bencaradocdavies:
@NielsCharlier one way to hide the store would be to make ```NetCDFAuxiliaryStoreFactory.isAvailable``` return ```true``` only if a system property is set. This change should only be made if we decide that it should be hidden by default.

The parameter names are fine; your explanation belongs in the GeoServer use documentation for the NetCDF plugin. Perhaps a small section in the NetCDF plugin section of the manual describing the auxiliary store and how to use it?

NielsCharlier:
@bencaradocdavies That was a good idea. I have made the changes, and I will do a documentation update as well later.

bencaradocdavies:
@dromagnoli there is also a pull request for GeoServer documentation: https://github.com/geoserver/geoserver/pull/2276

dromagnoli:
Hi @bencaradocdavies and @NielsCharlier.
I have spent some time checking the Pull Request and, in general, it looks good to me.
I agree with Ben concerns expressed in the GeoServer documentation PR, about having the SystemProperty based parameters min, max being global. (However, a different approach can be part of a future improvement).
I have also seen (correct me if I'm wrong) that with this PR, classes now deal with simple CoordinateAxis instead of CoordinateAxis1D. Will this have any impact with respect to, as an instance, 2D latitudes and longitudes (i.e. lat(ni,nj), lon(ni,nj)) which aren't currently supported?

NielsCharlier:
@dromagnoli While the tests focus on time/runtime dimension, any 2D coordinate variable should now be supported in the same way (converted to a single dimension on the geotools side of things).

dromagnoli:
What I wanted to say is that, before the change, a dataset with 2D longitudes,latitudes (latitude(ni,nj), longitude(ni,nj)) was not supported. Only datasets with 1-dimensional geocoordinates were supported. So a latitude(lat), longitude(lon) OR something like projection_x_coordinate(x), projection_y_coordinate(y).
So when you request a GT GridCoverage by specifying longitude and latitude spans, the reader will do proper NetCDF Sections/Ranges subsetting

dromagnoli:
Hi Niels,
sorry, Not sure I have fully understood. 
What I wanted to say is that, before the change, a dataset with 2D longitudes,latitudes (latitude(ni,nj), longitude(ni,nj)) was not supported. Only datasets with 1-dimensional geocoordinates were supported:  so a latitude(lat), longitude(lon) OR something like projection_x_coordinate(x), projection_y_coordinate(y). Therefore, when you request a GT GridCoverage by specifying longitude and latitude spans, the reader will do proper NetCDF Sections/Ranges subsetting. 
Is it now possible to access and expose coverages on top of datasets having each geospatial coordinate (lat,lon Or y,x) "function" of 2 dimensions each one?

http://cfconventions.org/cf-conventions/v1.6.0/cf-conventions.html#_two_dimensional_latitude_longitude_coordinate_variables

NielsCharlier:
yes, this should now be supported, with the caveat that the 2D coordinate variables are translated to a single dimension in geotools/geoserver. And that this has not been actually tested with this kind of use case. But in theory, yes, it should work.

bencaradocdavies:
There is a lot of special handling for spatial variables, so until it is tested, I would avoid claiming support for 2D spatial coordinate variables. It is also worth noting that the resulting output will be sparse after translation to two one-dimensional output coordinate variables. Consider a rotated grid: my understanding is that all the longitudes with data will be present in the output longitude, and likewise for all the latitudes, and the data will be sparse. This will surprise most users who might expect spatially dense output.

NielsCharlier:
Thanks @dromagnoli 
The documentation explains the sparse output with time/runtime as an example.

PR #1545, opened 2017-04-03
[GEOT-5692] Reproject feature reader swallows feature user data

nmco:
Makes the reproject feature reader copy the user data from the original feature. A test case for this is also added.

Associated issue: https://osgeo-org.atlassian.net/browse/GEOT-5692

### DIFF
--- a/modules/library/main/src/test/java/org/geotools/data/crs/ReprojectFeatureReaderTest.java
+++ b/modules/library/main/src/test/java/org/geotools/data/crs/ReprojectFeatureReaderTest.java
@@ -1,0 +1,75 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ *
+ *    (C) 2017, Open Source Geospatial Foundation (OSGeo)
+ *
+ *    This library is free software; you can redistribute it and/or
+ *    modify it under the terms of the GNU Lesser General Public
+ *    License as published by the Free Software Foundation;
+ *    version 2.1 of the License.
+ *
+ *    This library is distributed in the hope that it will be useful,
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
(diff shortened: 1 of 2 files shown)


modules/library/main/src/test/java/org/geotools/data/crs/ReprojectFeatureReaderTest.java
```diff
@@ -0,0 +1,75 @@
...
+        // create a feature collection wit a single feature
+        SimpleFeatureType featureType = DataUtilities.createType("feature", "id:string,geometry:Point:srid=4326");
+        SimpleFeature feature = DataUtilities.createFeature(featureType, "1|POINT(1 2)");
+        ListFeatureCollection features = new ListFeatureCollection(featureType);
+        features.add(feature);
+        // add some user data to the feature
+        feature.getUserData().put("someKey", "someValue");
+        // instantiate the reproject reader
+        CoordinateReferenceSystem sphericalMercator = CRS.decode("EPSG:3857");
+        FeatureReader<SimpleFeatureType, SimpleFeature> reader = DataUtilities.reader((SimpleFeatureCollection) features);
```
aaime: Is this cast actually needed? Just curious.

aaime:
Looks good to me.

modules/library/main/src/test/java/org/geotools/data/crs/ReprojectFeatureReaderTest.java
ianturton: apparently - "The method reader(Collection<SimpleFeature>) is ambiguous for the type DataUtilities" so I guess yes.

nmco: Yest the cast is need :disappointed:, ``ListFeatureCollection`` extends ``AbstractFeatureCollection`` which extends ``SimpleFeatureCollection`` but also implements ``Collection<SimpleFeature>``. The main reason as Ian stayed is that ``ListFeatureCollection`` can actually match both readers methods available in ``DataUtilities`` so Java ask us to do the disambiguation.

PR #1547, opened 2017-04-06
RangeLookupProcess: assigned no data property to same as coverage input

fedeliallalinea:
RangeLookupProcess assign 0 by default to no data property at coverage.
This patch assign no data property value same as the original coverage if exists, otherwise 0.


aaime:
This pull request cannot be accepted as it lacks tests, please review the contributors guide when making newer ones in the future: 

Timing is a bit funny, a colleague of mine made the same fix a week ago and I've just completed it yesterday by adding a fix, then had to switch to something else and the pull request has just been made:
https://github.com/geotools/geotools/pull/1548

Can you check it's covering your case as well?

aaime:
I've closed this one in the meantime. If the pull request I've reference does not fully cover your needs please open a new pull request following the project contribution guidelines.

PR #1552, opened 2017-04-18
Correct CQL encoding for color literals [GEOT-5704]

jodygarnett:
See https://osgeo-org.atlassian.net/browse/GEOT-5704

### DIFF
--- a/modules/library/cql/src/main/java/org/geotools/filter/text/commons/ExpressionToText.java
+++ b/modules/library/cql/src/main/java/org/geotools/filter/text/commons/ExpressionToText.java
@@ -16,6 +16,7 @@
  */
 package org.geotools.filter.text.commons;
 
+import java.awt.Color;
 import java.text.DateFormat;
 import java.text.SimpleDateFormat;
 import java.util.Arrays;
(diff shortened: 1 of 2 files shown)


modules/library/main/src/main/java/org/geotools/util/ColorConverterFactory.java
```diff
@@ -354,7 +354,7 @@ public String toString() {
      * <li>Hex representation of the form <code>#RRGGBB</code> and <code>#RRGGBBAA</code></li>
      * </ul>
      */
-    private static Converter CONVERT_CSS_TO_COLOR = new Converter() {
+    public static Converter CONVERT_CSS_TO_COLOR = new Converter() {
```
aaime: These two visibility switches seem to be a leftover from a previous attempt? They are not used elsewhere in the patch.

modules/library/cql/src/main/java/org/geotools/filter/text/commons/ExpressionToText.java
```diff
@@ -184,17 +186,40 @@ else if (literal instanceof Period){
...
             output = dateToText( period.getBeginning().getPosition().getDate(), output );
             output.append("/");
-    		output = dateToText( period.getEnding().getPosition().getDate(), output );
-    		
-    		return output;
+            output = dateToText(period.getEnding().getPosition().getDate(), output);
+
+            return output;
+        }
+        else if (literal instanceof Color){
```
tbarsballe: Doesn't adhere to formatting guidelines...

modules/library/cql/src/main/java/org/geotools/filter/text/commons/ExpressionToText.java
```diff
@@ -184,17 +186,40 @@ else if (literal instanceof Period){
...
+            Color color = (Color) literal;
+            
+            String redCode = Integer.toHexString(color.getRed());
+            String greenCode = Integer.toHexString(color.getGreen());
+            String blueCode = Integer.toHexString(color.getBlue());
```
tbarsballe: Should we support alpha?

jodygarnett: I duplicating what the color converter does exactly to prevent surprises, alpha is usually handled by a seperate opacity setting after all.

modules/library/main/src/main/java/org/geotools/util/ColorConverterFactory.java
jodygarnett: The intension was to make the converts public, not sure why it was missed ealier. You are correct that it is off topic for this bug fix.

jodygarnett:
Revised based on feedback, squashed to single commit for ease of backporting

jodygarnett:
Revised based on feedback, merging to master (and would like to backport this fix).

PR #1553, opened 2017-04-18
[GEOT-5708] Add CachingExternalGraphicFactory interface 

mattkrusz:
Adds a common interface for ExternalGraphicFactory implementations that use a cache, giving GeoServer's WMSLifcycleHandler::onReset() a general way to clear caches.

This commit does not change the API, because it only creates a new sub-interface of ExternalGraphicFactory, and it leaves the existing static resetCache() methods on SVGGraphicFactory and ImageGraphicFactory in place. (Java doesn't allow static methods to be moved up to be "abstract static" methods on an interface).

Issue: https://osgeo-org.atlassian.net/browse/GEOT-5708
Mailing list reference (geoserver-devel): https://sourceforge.net/p/geoserver/mailman/message/35747439/


bencaradocdavies:
Please create a Jira issue for this pull request and include the Jira tag in the pull request title and commit message.

tbarsballe:
This also seems to be missing a test case, could you ad one?

modules/plugin/svg/src/main/java/org/geotools/renderer/style/SVGGraphicFactory.java
```diff
@@ -433,5 +433,9 @@ public void paint(Graphics2D g, int width, int height, int x, int y) {
     public static void resetCache() {
         glyphCache.clear();
     }
+    
+    public void clearCache() {
```
bencaradocdavies: Javadoc or @Override?

bencaradocdavies:
Is this for mbstyles? Why is it needed? It adds API. Does it require an improvement proposal?

mattkrusz:
Hi Ben, thank you for your time and sorry for the lack of context here. Yes, you're right, this is for MapBox styles. It is motivated by the issue that although we can create new ExternalGraphicFactory implementations (in the case of Mapbox styles, a SpriteGraphicFactory), GeoServer does not support resetting their caches in a general way, but has hard-coded calls to the ImageGraphicFactory and SVGGraphicFactory resetCache methods.

I have created a Jira issue here: https://osgeo-org.atlassian.net/browse/GEOT-5708 and will reference it as appropriate in this PR.

Also for reference, there was a conversation on the GeoServer devel mailing list about the possibilities here: https://sourceforge.net/p/geoserver/mailman/message/35747439/

mattkrusz:
One more thing, -- we were hoping to minimize the effect of this change by implementing it as a sub-interface of ExternalGraphicFactory, which should have no effect on any code that currently references that interface.

mattkrusz:
@tbarsballe, I added some tests for ImageGraphicFactory / SVGGraphicFactory caching that use the new interface.

tbarsballe:
Overall, this looks good now. However, could you pleas squash your commits into a single commit before I merge?

mattkrusz:
@tbarsballe Of course -- squashed.

modules/library/render/src/main/java/org/geotools/renderer/style/CachingExternalGraphicFactory.java
```diff
@@ -0,0 +1,29 @@
...
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ *    Lesser General Public License for more details.
+ */
+package org.geotools.renderer.style;
+
+
+/**
+ * Interface for {@link ExternalGraphicFactory} classes that cache external graphics.
+ */
+public interface CachingExternalGraphicFactory extends ExternalGraphicFactory {
```
jodygarnett: Recommend making this available to both ExternalGraphicsFactory and MarkFactory

jodygarnett:
Merged to master, refactored so that clearCache method could be applied to ExternalGraphicFactory and MarkFactory

PR #1554, opened 2017-04-18
Fix DB2 online testcases for autogenerated key column - JIRA GEOT-5707

dwadler:
The basic problem is that the DB2 JDBC support for addBatch() and executeBatch() doesn't work correctly for key columns that are automatically generated as identity.
I've been in touch with DB2 JDBC development and it doesn't seem likely that these problems will be fixed so the failing testcases are being no-opt'd.

DB2SQLDialect wasn't implementing lookupGeneratedValuesPostInsert() so I added this. However, implementing it correctly causes sequence key tests to fail so it was changed to return 'false' and copious 
comments added about the problem.

All changed flagged with TODO to help find them in the future.


bencaradocdavies:
Please create a Jira issue for this pull request and include the Jira tag in the pull request title and commit message.

PR #1556, opened 2017-04-20
[GEOT-5711] Cannot set up a heterogeneous CRS mosaic if granules are all in local CRSs

aaime:
For bonus points the hetero CRS test is no more randomly choosing a target mosaic CRS, but the one it is told to use instead.
Due to the resolutions levels computation being fixed the old reference image has been replaced by a newer one that's one pixel off in width and height (but looks the same)


modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/ImageMosaicConfigHandler.java
```diff
@@ -1337,12 +1342,22 @@ public void updateConfiguration(GridCoverage2DReader coverageReader,
...
+                configBuilder.setCrs(actualCRS);
+            }
+
+            // get/compute the resolution levels
+            resolutionLevels = getResolutionLevels(coverageReader, inputCoverageName, configBuilder.getCrs());
+            numberOfLevels = resolutionLevels.length;
             configBuilder.setLevels(resolutionLevels);
-            configBuilder.setLevelsNum(numberOfLevels);
+            configBuilder.setLevelsNum(numberOfLevels)
+            ;
```
dromagnoli: small oversight here (semicolon being put on the next line)

aaime:
@dromagnoli formatting fixed, it seems I cannot shake off that habit around the ifs :-p

aaime:
Travis build failure in xsd wfs binding module, which is unrelated (and random, the module is trying to download xml.xsd and failing to do so sometimes, I did look at fixing it in the past but failed).

PR #1561, opened 2017-04-22
[GEOT-5712] Add a vendor option to enable kerning in text symbolizer (and have it on by default)

aaime:
https://osgeo-org.atlassian.net/browse/GEOT-5712

Mailing list discussion: http://osgeo-org.1560.x6.nabble.com/Adding-a-kerning-vendor-option-and-enable-it-by-default-td5314295.html

### DIFF
--- a/modules/library/render/src/main/java/org/geotools/renderer/style/SLDStyleFactory.java
+++ b/modules/library/render/src/main/java/org/geotools/renderer/style/SLDStyleFactory.java
@@ -27,12 +27,15 @@ import java.awt.RenderingHints;
 import java.awt.Shape;
 import java.awt.Stroke;
 import java.awt.TexturePaint;
+import java.awt.font.TextAttribute;
 import java.awt.geom.AffineTransform;
 import java.awt.geom.Rectangle2D;
 import java.awt.image.AffineTransformOp;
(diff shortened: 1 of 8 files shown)


modules/library/render/src/main/java/org/geotools/renderer/style/SLDStyleFactory.java
```diff
@@ -910,10 +913,14 @@ private Geometry findGeometry(final Object feature, String geomName) {
...
      *            An array of fonts dependent of the feature, the first that is
      *            found on the current machine is returned
+     * @param vendorOptions 
      *
      * @return The first of the specified fonts found on this machine (Serif 10 if none found)
      */
-    private java.awt.Font[] getFonts(Object feature, List<Font> fonts) {
+    private java.awt.Font[] getFonts(Object feature, List<Font> fonts, TextSymbolizer symbolizer) {
+        
+        // add kerning here!
```
bencaradocdavies: I do not understand why this comment is here.

aaime: Ah, it's a leftover, I should remove it

PR #1562, opened 2017-04-23
[GEOT-5713] Add support for lesscss color operation functions

aaime:


jodygarnett:
As per email discussion this look great. I reviewed quickly earlier and did not see anything amiss.

modules/library/main/src/main/java/org/geotools/filter/function/color/ConstrastFunction.java
```diff
@@ -0,0 +1,80 @@
...
+import org.geotools.filter.FunctionImpl;
+import org.geotools.filter.capability.FunctionNameImpl;
+import org.opengis.filter.capability.FunctionName;
+
+/**
+ * Contrast lesscss.org color function. Returns the color with highest contrast with a given reference
+ *
+ * @author Andrea Aime - GeoSolutions
+ */
+public class ConstrastFunction extends FunctionImpl {
```
jodygarnett: Thanks for using FunctionImpl, less overhead then FunctionExpressionImpl.

modules/library/main/src/main/java/org/geotools/filter/function/color/ConstrastFunction.java
```diff
@@ -0,0 +1,80 @@
...
+    private double luma(Color color) {
+        double r = color.getRed() / 255d;
+        double g = color.getGreen() / 255d;
+        double b = color.getBlue() / 255d;
+
+        r = (r <= 0.03928) ? r / 12.92 : Math.pow(((r + 0.055) / 1.055), 2.4);
+        g = (g <= 0.03928) ? g / 12.92 : Math.pow(((g + 0.055) / 1.055), 2.4);
+        b = (b <= 0.03928) ? b / 12.92 : Math.pow(((b + 0.055) / 1.055), 2.4);
+
+        return 0.2126 * r + 0.7152 * g + 0.0722 * b;
```
jodygarnett: gotta assume these numbers are from something specific, should they be constants? Readability seems fine as they are all in one method.

modules/library/main/src/main/java/org/geotools/filter/function/color/HSLColor.java
```diff
@@ -0,0 +1,194 @@
...
+
+import java.awt.Color;
+
+/**
+ * Support class for color functions, represents a color in HSL space (based on
+ * http://www.niwa.nu/2013/05/math-behind-colorspace-conversions-rgb-hsl/)
+ *
+ * @author Andrea Aime - GeoSolutions
+ */
+class HSLColor {
```
jodygarnett: Should HSLColor extend Color for consistency? Color is a normal class, with several extensions SystemColor, ColorUIResource, ...

modules/library/main/src/main/java/org/geotools/filter/function/color/HSLFunction.java
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * hsl lesscss.org color function.
+ *
+ * @author Andrea Aime - GeoSolutions
+ */
+public class HSLFunction extends FunctionImpl {
+
+    public static FunctionName NAME = new FunctionNameImpl("hsl", parameter("result", Color.class),
+            parameter("hue", Double.class), parameter("saturation", Double.class),
+            parameter("lightness", Double.class));
```
jodygarnett: Would not mind some more detail on the valid range of these three parameters.

modules/library/main/src/main/java/org/geotools/filter/function/color/HSLColor.java
jodygarnett: Digging into this more, it may be too complicated and not add value.

jodygarnett:
Finished review:

One design question:
- Should HSVColor extend Color?

And one request for the future (happy to help):
- Can we provide additional detail for the function parameters
- Specifically looking for min/max values to be defined for HSL parameters.

aaime:
Whoops, you said it was ok and I merged before you did the larger review.

modules/library/main/src/main/java/org/geotools/filter/function/color/HSLFunction.java
aaime: hue is an angle, so between 0 and 360, saturation and lightness are percentages, between 0 and 1

PR #1563, opened 2017-04-24
MBObjectParser function support

dvntucker:
- Move "object" utility method from MBFunctionTest to MapboxTestUtils for reuse
- Update MBObjectParser to support function objects + test cases
- Fix "tranform" spelling error

### DIFF
--- a/modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/parse/MBObjectParser.java
+++ b/modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/parse/MBObjectParser.java
@@ -642,8 +642,8 @@ public class MBObjectParser {
             throw new MBFormatException("\"" + tag + "\" percentage from boolean " + obj
                     + " not supported, expected value between 0 and 1");
         } else if (obj instanceof JSONObject) {
-            throw new UnsupportedOperationException(
-                    "\"" + tag + "\"precentage from Function not yet supported");
+            MBFunction function = new MBFunction(this, (JSONObject) obj);
+            return function.numeric();
         } else if (obj instanceof JSONArray) {
             throw new MBFormatException("\"" + tag
                     + "\" percentage from JSONArray not supported, expected value between 0 and 1");
@@ -671,8 +671,8 @@ public class MBObjectParser {
             throw new MBFormatException(
                     context + " number from Boolean " + bool + " not supported");
         } else if (obj instanceof JSONObject) {
-            throw new UnsupportedOperationException(
-                    context + " number from Function not yet supported");
+            MBFunction function = new MBFunction(this, (JSONObject) obj);
+            return function.numeric();
         } else if (obj instanceof JSONArray) {
             throw new MBFormatException(context + " number from JSONArray not supported");
         } else {
@@ -720,7 +720,7 @@ public class MBObjectParser {
      * @return Expression based on provided json, or null
      * @throws MBFormatException
      */
-    public  Expression number(JSONObject json, String tag) throws MBFormatException {
+    public Expression number(JSONObject json, String tag) throws MBFormatException {
         return number(json, tag, null);
     }
 
@@ -759,8 +759,8 @@ public class MBObjectParser {
             Boolean bool = (Boolean) obj;
             return ff.literal(bool.toString());
         } else if (obj instanceof JSONObject) {
-            throw new UnsupportedOperationException(
-                    context + " string from Function not yet supported");
+            MBFunction function = new MBFunction(this, (JSONObject) obj);
+            return function.function(String.class);
         } else if (obj instanceof JSONArray) {
             throw new MBFormatException(context + " string from JSONArray not supported");
         } else {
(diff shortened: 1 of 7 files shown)


modules/unsupported/mbstyle/src/test/java/org/geotools/mbstyle/parse/MBObjectParserTest.java
```diff
@@ -0,0 +1,69 @@
...
+        JSONObject object = MapboxTestUtils.object("{'expr': {'base': 1, 'stops': [[0, '{name-en}']]}}");
+        Expression expression = parser.string(object, "expr", "name");
+        assertTrue("Parse result must be a function",
+            Function.class.isAssignableFrom(expression.getClass()));
+    }
+
+    @Test
+    public void testBoolFromFunction() throws ParseException {
+        JSONObject object = MapboxTestUtils.object("{'expr': { 'base': 1, 'stops': [ [ 0, true ], [ 10, true ], [ 22, true ] ] }}");
+        Expression expression = parser.string(object, "expr", "name");
```
mattkrusz: This one should probably be parser.bool ?

dvntucker: Woops, yep.

modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/parse/MBObjectParser.java
```diff
@@ -639,8 +639,8 @@ public Expression percentage(JSONObject json, String tag, Number fallback)
             throw new MBFormatException("\"" + tag + "\" percentage from boolean " + obj
                     + " not supported, expected value between 0 and 1");
         } else if (obj instanceof JSONObject) {
-            throw new UnsupportedOperationException(
-                    "\"" + tag + "\"precentage from Function not yet supported");
+            MBFunction function = new MBFunction(this, (JSONObject) obj);
+            return function.numeric();
```
jodygarnett: Do we need a specific percentage function (much like we had a color function) to handle "0.5", "50%", "50" and produce an error for "1.1", "110%", "110" ?

modules/unsupported/mbstyle/src/test/java/org/geotools/mbstyle/MapboxTestUtils.java
```diff
@@ -71,4 +71,16 @@ public static BufferedImage showRender(String testName, GTRenderer renderer, lon
...
+     * @param json
+     * @return parsed JSONArray
+     * @throws ParseException
+     */
+    public static JSONObject object(String json) throws ParseException{
+        JSONParser parser = new JSONParser();
+        String text = json.replace('\'', '\"');
+        Object object = parser.parse(text);
+        return (JSONObject) object;
+    }
```
jodygarnett: Thanks, I love this method for making readable test cases.

modules/unsupported/mbstyle/src/test/java/org/geotools/mbstyle/parse/MBObjectParserTest.java
```diff
@@ -0,0 +1,69 @@
+package org.geotools.mbstyle.parse;
```
jodygarnett: Header?

modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/parse/MBObjectParser.java
jodygarnett: See below, the only example I could find of percentage was "hsl(100,50%,50%)" as part of a CSS color definition. 

This method could be removed and replaced with numeric if you would like to reduce code duplication.

mattkrusz: In the mapbox spec for these cases, the parameter value is generally of type 'number' and are supposed to be between 0 and 1. Unless we want to be extremely forgiving, I think 50% should fail (and it currently would).

mattkrusz: (Ah, you beat me to it.)

mattkrusz:
Thanks Devon, overall this looks great!

The only thing I am wondering is whether we're going to run into trouble on the individual layers' getXXX methods when they are specified as mapbox functions (b/c they currently want to return a concrete type). 
E.g., I bet that if a symbol layer's symbol-placement property was set to a function, that it would currently fail, because the style transformer tries to use a SymbolPlacement value directly. (I'll check on that soon if you don't get to it first). We should probably create a separate issue to make sure we don't forget about that.  

That said, I vote for merging this so we can keep moving forward on integration.

dvntucker:
- Added the header
- Fixed the mistaken call in the bool test, which also uncovered a small bug

vickdw:
@dvntucker before seeing your PR, I made the spelling error corrections and pushed them.  I hope I didn't create more conflicts.

mattkrusz:
If you haven't started on it already, Devon, this morning I am working on the getXXX issue and I'm breaking it out into its own sub-task -- I don't want it to hold up this merge.

PR #1565, opened 2017-04-26
[GEOT-5716] MongoFeatureSource to handle Min/Max Visitor

newmanw:
* Sort and limit query to handle Min/Max Visitor. This will prevent both visitors
  from looping the entire collection to find these values.

### DIFF
--- a/modules/unsupported/mongodb/pom.xml
+++ b/modules/unsupported/mongodb/pom.xml
@@ -147,13 +147,6 @@
           <!--systemPropertyVariables>
             <embedmongo.port>${embedmongo.port}</embedmongo.port>
           </systemPropertyVariables-->
-          <excludes>
-           <exclude>**/MongoSchemaDBStoreTest.java</exclude>
-           <exclude>**/MongoSchemaFileStoreTest.java</exclude>
-           <exclude>**/GeoJSONMongoDataStoreTest.java</exclude>
-           <exclude>**/GeoJSONMongoFeatureSourceTest.java</exclude>
-           <exclude>**/MongoTestUtil.java</exclude>
-          </excludes>
           <forkCount>1</forkCount>
         </configuration>
       </plugin>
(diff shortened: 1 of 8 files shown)


newmanw:
JIRA issue here:

https://osgeo-org.atlassian.net/browse/GEOT-5716

tbarsballe:
This change is missing a test case, could you add one? Thanks.

newmanw:
@tbarsballe gladly.  I do however need a bit of help.

I think my best bet is to add another OnlineTest case here:

https://github.com/geotools/geotools/blob/0e6fc022f395194e7670b63d2de785123a178cdf/modules/unsupported/mongodb/src/test/java/org/geotools/data/mongodb/MongoFeatureSourceTest.java

However I am not sure how to get the OnlineTest cases to run.

Any help would be much appreciated.

newmanw:
@tbarsballe gladly.  However I am having trouble getting the online tests to run:

I believe this should do it.

> mvn test -P online

However no luck.  I did check my ~/.geotools dir to see if an example fixture for mongodb was created and it was not.

Ideas?

tbarsballe:
First of all, a unit test (perhaps using Mocks?) ought to be sufficient. You probalby shouldn't need a full online integration test.
At current, I'm not sure if there are any online MongoDB tests, and we don't run them on the build server, so an online test is probably not going to be especially useful anyways.

However, if you still do want to try and add an online test:

`mvn test -P online` should be correct to run online tests. If a fixture file was not created automatically, then you will have to create one manually; mongodb has rather different connection parameters from most other databases, so that could be where you were running into issues with the one you created.

You will also need a properly configured local mongodb instance for the tests to run against.

I believe online tests use a rather different configuration from most other tests; what does your test file look like? (and what class is it extending?)

newmanw:
Ok, I have added a couple new online tests to test the min/max visitor.  I know this might not be optimal, but I did not find a good way to do this with mocks.

I figured out why the online tests were not running as they were being excluded in the pom.  I removed the exclusions and fixed all the online tests so that they are working correctly again.

Added the schema store uri to the fixture example file that is created when running online tests.  Most of the tests were failing as this is a required parameter.

tbarsballe:
Huh. Odd that there were specific exclusions in the pom. The tests look good, and seem to pass fine.

While we are currently in the middle of a code freeze, pending the 2.12-RC1 release, mongodb is a community module, so I will go ahead and merge this.

Thanks for the fix / improvement.

PR #1567, opened 2017-04-26
[GEOT-5715] Fix GeoPackage tile order

tbarsballe:
GeoPackage tile order along the y-axis is inverted relative to what it should be; the 0,0 tile corresponds to the lower-left corner. This PR fixes this issue.

According to the [OGC Spec](http://www.geopackage.org/spec/#tile_matrix):

> The tile coordinate (0,0) always refers to the tile in the upper left corner of the tile matrix at any zoom level, regardless of the actual availability of that tile

For more details:
https://osgeo-org.atlassian.net/browse/GEOT-5715
https://osgeo-org.atlassian.net/browse/GEOS-8112
https://trac.osgeo.org/gdal/ticket/6871


tbarsballe:
Squashed + Merged.

tbarsballe:
Backported to 17.x and 16.x

PR #1568, opened 2017-04-26
Quick fix for NPE error causing tests in StyleTransformerFunctionTest to fail.

vickdw:
@mattkrusz Please review.

### DIFF
--- a/modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/parse/MBObjectStops.java
+++ b/modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/parse/MBObjectStops.java
@@ -332,7 +332,7 @@ public class MBObjectStops {
                         JSONArray stop = (JSONArray) stops.get(i);
                         if (stop.get(0) instanceof Long) {
                             if (((Long) stop.get(0)).longValue() != range[0]) {
-                                objectsToRemove.add(stops.get(i));
+                                //objectsToRemove.add(stops.get(i));
                             }
                         }
                         if (stop.get(0) instanceof JSONObject) {
--- a/modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/parse/MBObjectStops.java
+++ b/modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/parse/MBObjectStops.java
@@ -95,6 +95,46 @@ public class MBObjectStops {
         return zoomLevels;
     }
 
+    public static MBLayer reducePropertyAndZoomFunctions(MBLayer layer) {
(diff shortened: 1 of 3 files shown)


modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/parse/MBObjectStops.java
```diff
@@ -95,6 +95,46 @@
         return zoomLevels;
     }
 
+    public static MBLayer reducePropertyAndZoomFunctions(MBLayer layer) {
+        JSONObject jsonObject = (JSONObject) layer.getJson();
+        MBLayer newLayer = MBLayer.create(jsonObject);
+
+        if (newLayer.getPaint() != null) {
+            getReducedProperty(newLayer.getPaint());
```
mattkrusz: Is this happening for layout somewhere too?

modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/parse/MBObjectStops.java
```diff
@@ -95,6 +95,46 @@
...
+            if (jsonObject.get(key) instanceof JSONObject) {
+                JSONObject child = (JSONObject) jsonObject.get(key);
+                if (child.containsKey("stops")) {
+                    JSONArray stops = (JSONArray) child.get("stops");
+                    for (int i = 0; i < stops.size(); i++) {
+                        JSONArray stop = (JSONArray) stops.get(i);
+                        if (stop.get(0) instanceof JSONObject) {
+                            reducedStops = new JSONArray();
+                            JSONArray x = (JSONArray) stops.get(i);
+                            reducedStops.add(0, ((Long)((JSONObject)x.get(0)).get("zoom")).longValue());
```
mattkrusz: I'm not sure I understand. Is this supposed to be reducing a zoom-and-property function to a property function? (But it's using the "zoom" here, not the property "value").

modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/transform/MBStyleTransformer.java
```diff
@@ -104,38 +101,45 @@ public StyledLayerDescriptor transform(MBStyle mbStyle) {
...
+//                            style.featureTypeStyles().add(featureTypeStyle);
+//                        }
+//                    } catch (ParseException e) {
+//
+//                    }
+
+//                } else {
+                    //featureTypeStyle = transform(layer, mbStyle);
+
+//                }
```
mattkrusz: Is all this not needed anymore?

PR #1569, opened 2017-05-01
MBStyle review

tbarsballe:
@mattkrusz / @vickdw - I am reviewing the MapBox Style stuff in chunks - here is part one (MBStyle, MBLayer, and children). I will be continuing to review the rest of the codebase, and add to this PR as I do.

I've done a bunch of fixup (mostly javadocs and formatting, but there is also some code changes that may or may not be correct).

I also have some general comments that should be addressed - feel free to add to this PR as these are dealt with.


jodygarnett:
**MBLayer.java**:

https://github.com/geotools/geotools/blob/master/modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/MBLayer.java#L137

MetaData is optional; should it really be throwing an exception?


https://github.com/geotools/geotools/blob/master/modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/MBLayer.java#L150

Q: Is ref optional? Should it really be throwing an exception?

A: Yes, we do not support the functionality yet so any transform method trying to make use of the value needs a reminder to implement.

https://github.com/geotools/geotools/blob/master/modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/MBLayer.java#L148

Q: Is updating this viable now?

A: Yes, ref should be implemented (as it should be just mixing information from several json sources and returning the result as a wrapper). 

The individual access methods, like layout() and paint() should be able to perform this work (leaving the rest of the codebase unchanged).

https://github.com/geotools/geotools/blob/master/modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/MBLayer.java#L271
https://github.com/geotools/geotools/blob/master/modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/MBLayer.java#L314

Q: Can this be done now?

A: Yes - this is where lookup of ref should be performed.

jodygarnett:
**LineMBLayer.java**:

https://github.com/geotools/geotools/blob/master/modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/LineMBLayer.java#L101
https://github.com/geotools/geotools/blob/master/modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/LineMBLayer.java#L112

Q: It looks like `lineCap()` supports functions, but `getLineCap()` doesn't. Could someone explain this? (Am I wrong?). Same for `lineJoin()`, etc.

The get methods act as assessors returning the contents of the wrapped data structure without needing to drill down through all the JSON.

The query functions, like lineCap() and lineJoin(), convert this into Java Objects and GeoTools expressions which we can use when implementing a Transform. This is done to make the transform code easier to follow.

https://github.com/geotools/geotools/blob/master/modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/LineMBLayer.java#L201

What units are this in?

https://github.com/geotools/geotools/blob/master/modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/LineMBLayer.java#L112

Should this use a parse function?

jodygarnett:
**SymbolMBLayer.java**

https://github.com/geotools/geotools/blob/master/modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/SymbolMBLayer.java#L493
https://github.com/geotools/geotools/blob/master/modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/SymbolMBLayer.java#L499

Undone TODOs

Several get__ methods don't have a counterpart that returns an expression.

Do you have an example? Not all things in the mapbox style spec support "data driven styling".

This class would benefit from either:
a) All text methods and all label methods each in separate sections (rather than interleaved), or
b) All text and label methods of the same type (e.g. textAllowOverlap, labelAllowOverlap) paired.


https://github.com/geotools/geotools/blob/master/modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/SymbolMBLayer.java#L1402

Move these alongside the corresponding get methods (and add javadocs where appropriate).

jodygarnett:
**General**:

`toDisplacement`: Unlike the other implementations, this returns null (instead of the default). This seems wrong, and has been changed.

Also, javadocs say this should handle functions, but implementation didn't / doesn't support this? Likewise, should the __translate (and textOffset) methods support mapbox functions?

Units also seem to be all over the place here, with a mix of doubles and ints.

LineMBLayer seems the most accurate, and still just throws an exception if it encounters a function.

The "(Optional)" prefix to several of the javadocs seems inconsistent (missing in some places). Is this even necessary / usefull?

https://github.com/geotools/geotools/blob/master/modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/MBStyle.java#L40

https://github.com/geotools/geotools/blob/master/modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/MBLayer.java#L47
etc. 

This is unclear; presuming this refers to the `layers()` (etc.) methods?


https://github.com/geotools/geotools/blob/master/modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/MBStyle.java#L153
https://github.com/geotools/geotools/blob/master/modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/MBLayer.java#L132
etc.

This example of prefixing is incomplete, and not especially helpful since this is a get method

jodygarnett:
**MBStyle.java**:

https://github.com/geotools/geotools/blob/master/modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/MBStyle.java#L43

Terminology ("Access methods") inconsistent with the above.

https://github.com/geotools/geotools/blob/master/modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/MBStyle.java#L48

Q: Is this complete now?

JG: No some work, glyphs, were placed out of scope.

https://github.com/geotools/geotools/blob/master/modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/MBStyle.java#L153

More detail about the "selector" needed here; this does not seem to match up with the implementation

tbarsballe:
Okay, most of my questions on terminology are answered. There are still some implementation stuff that needs to be fixed.

tbarsballe:
Updated comments with checkboxes.

tbarsballe:
Reviewed functions. Per-class comments as separate comments below.

tbarsballe:
**MBFunction**:

- [] General comment: Javadocs could use some tune-up - actual description of arguments + return values (Return type in javadoc is generally redundant, as this is already specified by the method signature)

tbarsballe:
**MBFunctionFactory**:

- [] https://github.com/boundlessgeo/geotools/blob/mbstyle-review/modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/function/MBFunctionFactory.java#L42

  Are we missing DefaultIfNullFunction here, or is that defined elsewhere? (I've added it, please revert if this is incorrect)
  Also, given that this was missing, perhaps we need a test case that uses this function?

tbarsballe:
**ZoomLevelFunction.java**:

- [] https://github.com/boundlessgeo/geotools/blob/mbstyle-review/modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/function/ZoomLevelFunction.java#L91

  Can we also support EPSG:4326 for the first release?

- [ ] https://github.com/boundlessgeo/geotools/blob/mbstyle-review/modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/function/ZoomLevelFunction.java#L93

  This should be a predefined constant somewhere (although I can't find it). Converted to a static constant on ZoomLevelFunction for now.

- [ ] I don't se a test case in MBFunctionFactoryTest for ZoomLevelFunction...

tbarsballe:
**ZoomRange**

- [ ] https://github.com/boundlessgeo/geotools/blob/mbstyle-review/modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/parse/ZoomRange.java#L101

  Does this only support EPSG:3857? If so, please add this to the javadoc.

- [ ] https://github.com/boundlessgeo/geotools/blob/mbstyle-review/modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/parse/ZoomRange.java#L177

  Please add a javadoc

- [ ] https://github.com/boundlessgeo/geotools/blob/mbstyle-review/modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/parse/ZoomRange.java#L228

  What is paint?

- [ ] https://github.com/boundlessgeo/geotools/blob/mbstyle-review/modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/parse/ZoomRange.java#L246

  Should these be the other way around? Can't we have type be something other than identity and still be a function

tbarsballe:
**MBObjectStops**:

- [] This is both a fairly complex method, and doesn't have much documentation. While no other javadocs are striclty needed (most functionality is not public), some more comments would be quite useful

tbarsballe:
**MBFilter**:

- [ ] https://github.com/boundlessgeo/geotools/blob/mbstyle-review/modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/parse/MBFilter.java#L120
  
  I can't see anywhere this is actually used other than test cases. Does the current code actually support type filters?

  https://github.com/boundlessgeo/geotools/blob/mbstyle-review/modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/parse/MBFilter.java#L120

  Shouldn't we be calling semanticTypeIdentifiers here?

One other comment: filter() is a bit on the long side, and has a bit more repeated code than I would like, but I don't see any particularily good alternatives to the current approach.

tbarsballe:
**MBObjectParser**:

- [ ] Javadocs are a bit sparse in places; there are a few public methods missing docs, and others missing @return. Most of the javadocs are good.

tbarsballe:
- [ ] One general comment: Like MBSource, we should probably move MBLayer and assocated classes to their own package, rather than leaving them at top-level. (Probably hold off on this until other issues are addressed, so as to avoid merge conflicts)

tbarsballe:
**MBSource**:

- [ ] https://github.com/boundlessgeo/geotools/blob/mbstyle-review/modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/source/MBSource.java#L93

  Subclasses use a string value here. MBLayer uses a very similar structure, but subclasses use a spring constant instead. Lets be consistent?

- [ ] General source question - what (if any) source types do we actually support? Or are they just 'decorative' right now.

tbarsballe:
**MapboxSpriteException**:

- [ ] https://github.com/boundlessgeo/geotools/blob/mbstyle-review/modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/sprite/MapboxSpriteException.java#L19

  No javadocs. Also, weren't we generally avoiding mentioning Mapbox in classnames for IP reasons?

tbarsballe:
**SpriteGraphicFactory**:

- [ ] https://github.com/boundlessgeo/geotools/blob/mbstyle-review/modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/sprite/SpriteGraphicFactory.java#L282

  This class seems substantial enough that is should probably be broken out to its own file

modules/unsupported/mbstyle/src/test/java/org/geotools/mbstyle/parse/StyleParseTests.java
```diff
@@ -283,4 +287,12 @@ public void testParseLineLayer() throws  IOException, ParseException {
         assertEquals(2.0, l.getLineBlur().doubleValue(), .00001);
     }
 
+    @Test
+    public void getRefTest() throws IOException, ParseException {
```
tbarsballe: This test is currently failing...

tbarsballe:
**MBStyleTransformer**

`transform()` functions now split out into each of the MBLayer classes.

- [ ] All of these are full of TODOs and commented out code. This should be tidied up a bit. Based on the subsequent code, it looks like some of these TODOs have since been completed. Where this is the case, they should be removed.

- [ ] Additionally, FillExtrusionMBLayer has an alternate implementation: `transform(MBStyle styleContext, Boolean useMultipleFeatureTypeStyles)`
  However, the boolean value here is never used - so my question is how does this differ from the implementation without the boolean?

modules/unsupported/mbstyle/src/test/java/org/geotools/mbstyle/parse/StyleParseTests.java
vickdw: This test has been fixed in a later commit.

modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/LineMBLayer.java
```diff
@@ -281,19 +268,8 @@ public Point lineTranslate() {
...
-        } else if (defn instanceof JSONArray) {
-            JSONArray array = (JSONArray) defn;
-            return sf.displacement(parse.number(array, 0, 0), parse.number(array, 1, 0));
-        } else if (defn instanceof JSONObject) {
-            throw new MBFormatException("\"line-translate\": Functions not supported yet.");
-        } else {
-            throw new MBFormatException("\"line-translate\": Expected array or function, but was "
-                    + defn.getClass().getSimpleName());
-        }
+    public Displacement lineTranslateDisplacement() {
```
tbarsballe: Inconsistent naming?

tbarsballe: Based on SymbolMBLayer below, this is probably the better naming scheme, so FillMBLayer and CircleMBLayer should be updated to conform to it.

modules/unsupported/mbstyle/src/test/java/org/geotools/mbstyle/transform/VisualTransformerTest.java
```diff
@@ -306,8 +306,8 @@ public void mbCircleLayerVisualTest() throws Exception {
         // Read file to JSONObject
         JSONObject jsonObject = MapboxTestUtils.parseTestStyle("circleStyleTest.json");        
         testVisualizeStyleWithPointFeatures(jsonObject, "Circle Style Test", "circle-style-test", true, 300, 300);
-    }
-    
+    }
```
tbarsballe: Trailing whitespace

modules/unsupported/mbstyle/src/test/java/org/geotools/mbstyle/transform/VisualTransformerTest.java
```diff
@@ -456,6 +466,6 @@ File file(String name) throws IOException {
         // -Dorg.geotools.image.test.interactive=true</code>
         return new File("src/test/resources/org/geotools/mbstyle/transform/test-data/rendered/"
                 + name + ".png");
-    }
+    }
```
tbarsballe: More trailing whitespace

modules/unsupported/mbstyle/src/test/java/org/geotools/mbstyle/transform/VisualTransformerTest.java
vickdw: Fixed in latest commit.

modules/unsupported/mbstyle/src/test/java/org/geotools/mbstyle/transform/VisualTransformerTest.java
vickdw: Fixed in latest commit.

modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/LineMBLayer.java
mattkrusz: That would be fine. The difference is that most layers only have one displacement -- Symbol layer has several.

mattkrusz: Fixed

vickdw: Changes applied.

tbarsballe:
Squashed and manually merged.

PR #1572, opened 2017-05-03
[GEOS-8070] - Update Rendering Transform BBOX math to use cell_corner 

dvntucker:
Fixes issues with BBOX calculations on the rendering transform path. See ticket for more details.


modules/library/render/src/main/java/org/geotools/renderer/lite/RenderingTransformationHelper.java
```diff
@@ -110,7 +110,9 @@ public Object applyRenderingTransformation(Expression transformation,
                     final GridCoverage2DReader reader = (GridCoverage2DReader) GRID_PROPERTY_NAME.evaluate(gridWrapper);
                     // don't read more than the native resolution (in case we are oversampling)
                     if(CRS.equalsIgnoreMetadata(reader.getCoordinateReferenceSystem(), gridGeometry.getCoordinateReferenceSystem())) {
-                         MathTransform g2w = reader.getOriginalGridToWorld(PixelInCell.CELL_CENTER);
+                        //GEOS-8070, changed the pixel anchor to corner. CENTER can cause issues with
+                        //the BBOX calculation and cause it to be incorrect
+                         MathTransform g2w = reader.getOriginalGridToWorld(PixelInCell.CELL_CORNER);
```
jodygarnett: I am scared of any changes due to possibility of regressions; still you say it fixes bbox calculations?

modules/library/render/src/test/java/org/geotools/renderer/lite/NoOpFunction.java
```diff
@@ -0,0 +1,23 @@
+package org.geotools.renderer.lite;
```
jodygarnett: header

modules/library/render/src/test/java/org/geotools/renderer/lite/RenderingTransformationTest.java
```diff
@@ -1,6 +1,7 @@
 package org.geotools.renderer.lite;
```
jodygarnett: header

modules/library/render/src/test/java/org/geotools/renderer/lite/RenderingTransformationTest.java
```diff
@@ -83,6 +86,33 @@ public void testTransformReprojectedGridReader() throws Exception {
...
+            9.848993475036622,
+            11.958867853088378,
+            40.74254816253662,
+            41.64941961090088,
+            crs);
+
+        BufferedImage image = RendererBaseTest.showRender("Transform BBOX", renderer,  4000, reWgs84);
+        //last pixel is white when doing a transformation and the rendering transform BBOX transforms are
+        //incorrect, it shouldn't be
+        assertNotEquals(Color.WHITE, getPixelColor(image, 299, 0));
```
jodygarnett: comment should be "if last pixel is white"

modules/library/render/src/main/java/org/geotools/renderer/lite/RenderingTransformationHelper.java
dvntucker: All the tests pass :D

TBH pixel orientation is a bit of a mystery to me, but cell_center causes quite noticeable errors in the BBOX calculation in my testing. You can see the screenshots in my comment.

sikeoka:
I was wondering if it would be possible to use to add 1 to the rectangle width and height instead of changing the code to use CELL_CORNER?  This bug currently affects all raster rendering transformations, potentially creating gaps in the bottom and right edges, but this particular fix leaves one process (ras:Contour) still with the bug whereas the +1 fix would fix the bug for all process.  

While changing it to CELL_CORNER significantly reduces the size of the gaps on the bottom and right edges of the image for the Contour process, the gaps are still going to be there, which is very noticeable when using tiling.  In addition, the CELL_CORNER changes introduces the possibility of gaps in the top and left edges of the images instead of just on the bottom and right edges.  

Considering that the Contour process is the very first example in the rendering transformation documentation, it's possible that CELL_CENTER was specifically chosen because it works better with the Contour process.  I know that the gaps did not exist in GeoServer 2.5.0 which would make this a regression that was introduced sometime after that.  In my opinion, a fix for all rendering transformations is better than a fix for all but one that also introduces new artifacts for the one that isn't fixed.

dvntucker:
@sikeoka I'd be fine with adding +1 to the requested coverage bounding box, but I'm a bit afraid of just papering over issues there might be in the contour process. I'm not familiar with how the rendering transforms happened in earlier versions of GeoServer, but looking at it it appears pretty significantly different so it's tough to say what could have changed since then.

@aaime or @simboss any thoughts on this?

modules/library/render/src/main/java/org/geotools/renderer/lite/RenderingTransformationHelper.java
aaime: It is consistent with other places where we do coverage rendering (they are using CELL_CORNER, not CELL_CENTER), I'm good with it.

sikeoka:
I have one last comment.  In order to prevent this pull request from creating gaps in the top and left edges when using the Contour process, would it be possible to change this line:
Rectangle rect = new Rectangle(minx, miny,

to:
Rectangle rect = new Rectangle(
    minx - (TRANSFORM_READ_BUFFER_PIXELS / 2),
    miny - (TRANSFORM_READ_BUFFER_PIXELS / 2),

dvntucker:
@sikeoka Can do. Realized almost immediately that my approach was wrong :D

dvntucker:
Ok, updated the read padding to work basically like the coverage reader helper does things, 10 pixels a side.

sikeoka:
Thank you.  It looks good to me with the latest changes.

aaime:
Yeah, should be fine, +1

dvntucker:
Ok, no objections with merging then?

aaime:
No objections, +1 means +1 :-)

aaime:
Ops, closed by mistake, reopened. Merge as you see fit.

PR #1574, opened 2017-05-05
[GEOT-5720] Heterogeneous CRS mosaic composes output with erroenous image positioning and scaling

aaime:


modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/RasterManager.java
```diff
@@ -1644,4 +1650,86 @@ public boolean isExpandMe() {
     public ImageMosaicReader getParentReader() {
         return parentReader;
     }
+    
+    /**
+     * Builds a RasterManager for the sub mosaic of a given template granule, and withing a given search bounds
```
dromagnoli: trivial TYPO here: withing = within?

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/RasterManager.java
```diff
@@ -1644,4 +1650,86 @@ public boolean isExpandMe() {
...
+            ReferencedEnvelope requestBounds) throws IOException {
+        
+        String crsAttribute = configuration.getCRSAttribute();
+        if (crsAttribute == null) {
+            crsAttribute = CRSExtractor.DEFAULT_ATTRIBUTE_NAME;
+        }
+        GranuleSource granuleSource = getGranuleSource(true, null);
+        if(granuleSource.getSchema().getDescriptor(crsAttribute) == null) {
+            throw new IllegalStateException("Invalid heterogeneous mosaic configuration, "
+                    + "the crs property is missing collector must be configured");
```
dromagnoli: Do we need a comma or a period somewhere in this error message to make it more clear?
(Maybe between "missing" and "collector"?)

aaime:
Build is failing, will have a look Monday as to why it's happening (does not happen locally...)

dvntucker:
I'm able to recreate the issue with the test case failure, but I'm surprised that it would be system dependent. It looks like the target grid space of the read has changed from what it was before? Maybe we just need different results images and/or to make sure to request a fixed target grid.

aaime:
Double checked, builds locally, no uncommitted files... probably something related to file system scan order, I'll have another look

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/RasterManager.java
aaime: fixing

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/RasterLayerResponse.java
```diff
@@ -313,9 +321,21 @@ public void visit(GranuleDescriptor granuleDescriptor, SimpleFeature sf) {
...
+                    if(!CRS.equalsIgnoreMetadata(granuleCRS, mosaicCRS)) {
+                        MathTransform mt = CRS.findMathTransform(granuleCRS, mosaicCRS);
+                        inclusionGeometry = JTS.transform(inclusionGeometry, mt);
+                    }
+                } catch (FactoryException | MismatchedDimensionException | TransformException e) {
+                    // in case there was a reprojection issue assume intersection
+                    intersects = true;
+                }
+
                 intersects = inclusionGeometry.intersects(bb);
```
aaime: right

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/RasterManager.java
aaime: yep

dvntucker:
I did some debugging on Friday, but didn't get too far. What I did notice is that the levels were coming back differently on your branch vs. master:

https://github.com/geotools/geotools/blob/master/modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/ImageMosaicReader.java#L830

Unfortunately I'm not working today, but if there are still issues I can try to look tomorrow.

PR #1576, opened 2017-05-06
mbspec

jodygarnett:
This is a work in progress, porting spec to RST so we can document what is supported and what does not.


jodygarnett:
This is a port of the mbstyle spec to RST.

There is some overlap with #1579 which provides example for java developer, and an example of each kind of layer.

jodygarnett:
This work has been taken up by @vickdw

PR #1577, opened 2017-05-07
[GEOT-5722] Updated temporal component "T" for case sensitive

sp-suresh:
IGNORING_CASE for "T" temporal component not allowing to query with attribute name "t" throwing exception :org.geotools.filter.text.cql2.CQLException:


aaime:
Mind, @ianturton 's is not a suggestion, it's a rule, github linked you to this document when you made the pull request: 

PR #1578, opened 2017-05-07
Docs fix so they build

ianturton:
I have added the repositories back into the tutorial poms **AGAIN** removing them breaks the docs and prevents new users making use of them. Please stop deleting them!!!!


mprins:
I'm pretty sure the Java.net repo's (http://download.java.net/maven/*) can be removed as they don't exist anymore and will just cause any builds to slow down

PR #1580, opened 2017-05-09
MBStyle - Support $type filters

tbarsballe:
The code currently generates semantic type correctly; that is not used by the rendering engine to restrict content.

Looking at a default style in geoserver the ``dimension( geometry() )`` function is used to check for point, line or polygon content, along wiht ``isCoverage()``.

### DIFF
--- a/modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/parse/MBFilter.java
+++ b/modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/parse/MBFilter.java
@@ -88,6 +88,10 @@ import org.opengis.style.SemanticType;
  * @see MBFunction
  */
 public class MBFilter {
+
+    public static final String TYPE_POINT = "Point";
+    public static final String TYPE_LINE = "LineString";
+    public static final String TYPE_POLYGON = "Polygon";
     
     /** Default semanticType (or null for "geometry"). */
     final protected SemanticType semanticType;
@@ -225,6 +229,25 @@ public class MBFilter {
         return Collections.emptySet();
     }
 
+    private Filter translateType(String jsonText) {
+        final FilterFactory2 ff = parse.getFilterFactory();
+        //TODO: How to wildcard geometry
+        Expression dimension = ff.function("dimension", ff.function("geometry"));
+
+        switch (jsonText) {
+            case TYPE_POINT:
+                return ff.equals(dimension, ff.literal(0));
+            case TYPE_LINE:
+                return ff.equals(dimension, ff.literal(1));
+            case TYPE_POLYGON:
+                return ff.and(
+                        ff.equals(dimension, ff.literal(2)),
+                        ff.not(ff.equals(ff.function("isCoverage"), ff.literal(true))));
+            default:
+                return null;
+        }
+    }
+
     /**
      * Translate from json "Point", "LineString", and "Polygon".
      * @param jsonText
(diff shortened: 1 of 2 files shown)


modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/parse/MBFilter.java
```diff
@@ -225,6 +225,25 @@ public MBFilter(JSONArray json, MBObjectParser parse, SemanticType semanticType)
         return Collections.emptySet();
     }
 
+    private Filter translateType(String jsonText) {
+        final FilterFactory2 ff = parse.getFilterFactory();
+        //TODO: How to wildcard geometry
+        Expression dimension = ff.function("dimension", ff.function("geometry"));
```
tbarsballe: There may be an argument for using the `geometryType` function instead of the `dimension` function here. Dimension is simpler, but I am not certain it captures all cases - in the case of time or elevation data, does dimension still map to the correct geometry type?

jodygarnett: I am happy with either, does geometryType capture coverage cleanly?

modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/parse/MBFilter.java
```diff
@@ -225,6 +225,25 @@ public MBFilter(JSONArray json, MBObjectParser parse, SemanticType semanticType)
...
+
+        switch (jsonText) {
+            case "Point":
+                return ff.equals(dimension, ff.literal(0));
+            case "LineString":
+                return ff.equals(dimension, ff.literal(1));
+            case "Polygon":
```
jodygarnett: do we have constants for these?

modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/parse/MBFilter.java
```diff
@@ -266,9 +285,33 @@ public Filter filter() {
...
                 "$type".equals(parse.get(json, 1))) {
 
-            // handled by semanticsTypes() method
-            // (unsure if #type can be used with all/any/none - if so we will process the json)
-            return Filter.INCLUDE;
+            List<Filter> typeFilters = new ArrayList<>();
+            List<?> types = json.subList(2, json.size());
+            for (Object type : types ) {
+                if (type instanceof String) {
+                    typeFilters.add(translateType((String) type));
```
jodygarnett: you may want to check for translateType returning a {{null}}.

modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/parse/MBFilter.java
tbarsballe: Nope, would be good to add

tbarsballe:
Updated based on feedback

modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/parse/MBFilter.java
mattkrusz: The spec does consistently refer to these with this casing, but I would vote for equalsIgnoreCase.

mattkrusz:
Looks good to me! Thanks Torben.

tbarsballe:
Backported to 17.x

PR #1581, opened 2017-05-11
[GEOT-5726] Parse an empty CSSParameter as an empty string instead of null

tbarsballe:
See https://osgeo-org.atlassian.net/browse/GEOT-5726

### DIFF
--- a/modules/library/main/src/main/java/org/geotools/styling/SLDParser.java
+++ b/modules/library/main/src/main/java/org/geotools/styling/SLDParser.java
@@ -2308,6 +2308,10 @@ public class SLDParser {
         for (Expression expression : expressions) {
             ret = manageMixed(ret, expression);
         }
+        //If the expression list is empty, we have an empty tag, and should return an empty string.
+        if (ret == null) {
+            return ff.literal("");
+        }
 
         return ret;
     }
--- a/modules/library/main/src/test/java/org/geotools/styling/SLDTransformerTest.java
+++ b/modules/library/main/src/test/java/org/geotools/styling/SLDTransformerTest.java
@@ -21,11 +21,7 @@ import static org.custommonkey.xmlunit.XMLAssert.assertXpathExists;
 import static org.custommonkey.xmlunit.XMLAssert.assertXpathNotExists;
 import static org.custommonkey.xmlunit.XMLUnit.buildTestDocument;
 import static org.custommonkey.xmlunit.XMLUnit.setXpathNamespaceContext;
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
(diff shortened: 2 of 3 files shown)


modules/library/main/src/main/java/org/geotools/styling/SLDParser.java
```diff
@@ -2308,6 +2308,9 @@ private Expression parseCssParameter(Node root, boolean trimWhiteSpace) {
         for (Expression expression : expressions) {
             ret = manageMixed(ret, expression);
         }
+        if (ret == null) {
+            return ff.literal("");
```
jodygarnett: I would prefer to use null for something that is not there?

jodygarnett: Double check the GTXML parser handles an empty tag in the same way or we will get into inconsistencies in our model.

modules/library/main/src/test/java/org/geotools/styling/SLDTransformerTest.java
```diff
@@ -834,6 +830,50 @@ public void testTextSymbolizerTransformOutAndInAndOutAgain() throws Exception {
...
+
+        SLDTransformer st = new SLDTransformer();
+        String firstExport = st.transform(style);
+        SLDParser sldp = new SLDParser(CommonFactoryFinder.getStyleFactory(null));
+        sldp.setInput(new StringReader(firstExport));
+        Style[] firstImport = sldp.readXML();
+
+        assertNotNull(firstImport[0]);
+
+        // NPE here
```
jodygarnett: there is no longer a NPE exception here

modules/library/main/src/test/java/org/geotools/styling/SLDTransformerTest.java
```diff
@@ -834,6 +830,50 @@ public void testTextSymbolizerTransformOutAndInAndOutAgain() throws Exception {
         String secondExport = st.transform(firstImport);
 
     }
+
+    @Test
+    public void testTextSymbolizerTransformOutAndInAndOutAndInAgainWithEmptyFontFamily() throws Exception {
```
jodygarnett: for a test case this specific quote the bug number or people will think you are nuts

tbarsballe:
Addressed feedback - added / improved comments.

I added a test case for consistency with the GTXML parser. Turns out it doesn't match the either the current behavior (getting `null`) or the new behavior (getting `""`). Instead, it fills in the default, in this case `"Serif"`. I've added this test case, but set it as ignored, in case we revisit this behavior in the future

tbarsballe:
Backported to 17.x

PR #1582, opened 2017-05-11
MBstyle - Supported named (ttf) fonts

tbarsballe:
Adds support for named system fonts to MBStyle (not using glyphs).

### DIFF
--- a/modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/layer/SymbolMBLayer.java
+++ b/modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/layer/SymbolMBLayer.java
@@ -1663,13 +1663,13 @@ public class SymbolMBLayer extends MBLayer {
         Fill fill = sf.fill(null, textColor(), textOpacity());
 
 
-        //Font not supported until glyphs are working
-        Font font = sb.createFont(ff.literal("" /* transformer.getDefaultFonts() */), ff.literal("normal"),
-                ff.literal("normal"), textSize());
-        /* else {
-            font = sb.createFont(ff.literal(getTextFont()), ff.literal("normal"),
-                    ff.literal("normal"), textSize());
-        } */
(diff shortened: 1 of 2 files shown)


modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/layer/SymbolMBLayer.java
```diff
@@ -1663,13 +1663,13 @@ public FeatureTypeStyle transformInternal(MBStyle styleContext) {
...
 
-        //Font not supported until glyphs are working
-        Font font = sb.createFont(ff.literal("" /* transformer.getDefaultFonts() */), ff.literal("normal"),
-                ff.literal("normal"), textSize());
-        /* else {
-            font = sb.createFont(ff.literal(getTextFont()), ff.literal("normal"),
-                    ff.literal("normal"), textSize());
-        } */
+        Font font = sb.createFont(ff.literal(""), ff.literal("normal"), ff.literal("normal"), textSize());
+        if (getTextFont() != null) {
```
mattkrusz: I may be mistaken, but this will probably die if fonts are specified as a function in the style. It may be a little premature to support function-defined fonts at this point -- just something to be aware of.

tbarsballe: Yep, that makes sense. For fonts as functions, do we know if the font array as a whole can be a function or just individual entries?

If the latter, this should be fairly simple to change once we start supporting functions for fonts. Otherwise, it could be a bit more difficult.

mattkrusz: The spec isn't exactly clear on that, but I interpret it to say that the array as a whole can be a function.

tbarsballe:
Ok. I'll merge this one, and we can implement functions whenever we get to them...

tbarsballe:
Backported to 17.x

PR #1585, opened 2017-05-12
[GEOT-5727] Heterogeneous CRS mosaic fails to display when there are granules crossing the dateline

aaime:
So, this one is a beast... right now trying to reproject data in UTM 60 and UTM 1 fails because those two zones _cross_ the dateline. Here is, visually, what happens with a heterogeneous mosaic of that kind, each polygon is a image footprint, with zoomed in details at west and east:

world

west

east

As you can see some rasters are literally sitting across the dateline.
Before the pull request, reprojection would just fail, with it:

result

To make this work I needed the "advanced projection handling" classes, so they got moved to gt-main, and then basically each envelope/geometry reprojection got managed though them.

Now, it's not all roses, reprojection across the dateline is really slow because the crossing kills the linearized warp operation, when zooming out a lot the output is not perfect and somtimes missing a column of pixels, but ... those improvements can come later imho :-p


dvntucker:
Looks good to me. Have you had a chance to look at the test failure? If I had to guess I'd say it's similar to last time, different resolution being used by default.

aaime:
The issue was caused by sample result images being in the same folder as the inputs, moving them to a different folder fixed the problem.

PR #1586, opened 2017-05-12
Changed API to return a List of FeatureTypeStyles instead of a single object.

vickdw:


docs/user/unsupported/mbstyle.rst
```diff
@@ -3,6 +3,11 @@ Mapbox Styles Module
 
 The **gt-mbstyle** module is an unsupported module that provides a parser/encoder to convert between Mapbox Styles and GeoTools style objects. These docs are under active development, along with the module itself.
 
+* Human readable alternative to machine readable XML
```
tbarsballe: This file seems like it should be part of a different PR?

modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/MBStyle.java
```diff
@@ -358,12 +358,12 @@ public StyledLayerDescriptor transform() {
                         }
 
                         featureTypeStyle = l.transform(this, minScaleDenominator, maxScaleDenominator);
-                        style.featureTypeStyles().add(featureTypeStyle);
+                        style.featureTypeStyles().add(featureTypeStyle.get(0));
```
tbarsballe: This should be `addAll`

modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/MBStyle.java
```diff
@@ -358,12 +358,12 @@ public StyledLayerDescriptor transform() {
...
                         i++;
                     }
                 } else {
                     featureTypeStyle = layer.transform(this, layerMinScaleDenominator, layerMaxScaleDenominator);
-                    style.featureTypeStyles().add(featureTypeStyle);
+                    style.featureTypeStyles().add(featureTypeStyle.get(0));
```
tbarsballe: This should be `addAll`

modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/layer/BackgroundMBLayer.java
```diff
@@ -179,12 +179,14 @@ public FeatureTypeStyle transformInternal(MBStyle styleContext) {
         rule.setLegendGraphic(new Graphic[0]);
 
         rules.add(rule);
-        return sf.featureTypeStyle(getId(),
+        List<FeatureTypeStyle> style = new ArrayList<>();
```
tbarsballe: You can use `Collections.singletonList(sf.featureTypeStyle(...))` here instead of creating a new array list. Same for most of the other MBLayer implementations.

modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/layer/FillExtrusionMBLayer.java
```diff
@@ -278,17 +278,17 @@ public FeatureTypeStyle transformInternal(MBStyle styleContext) {
...
-                getId(),
-                sf.description(
-                        Text.text("MBStyle "+getId()),
-                        Text.text("Generated for "+getSourceLayer())),
-                null, // (unused)
-                Collections.emptySet(),
-                filter.semanticTypeIdentifiers(),
-                Arrays.asList(rule)
-        );
+        return null;
```
tbarsballe: If we aren't returning anything here, we should remove the whole method rather than just the return statement.

We are probably fine just returning a fill for now?

modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/layer/MBLayer.java
```diff
@@ -329,26 +332,27 @@ public JSONObject getPaintProperties(){
     /**
      * Transforms a given {@link MBLayer} to a GeoTools {@link FeatureTypeStyle}.
      *
-     * @param layer The MBLayer to transform.
```
tbarsballe: Missing `@param styleContext`

tbarsballe:
Travis failure, looks like a syntax error in the RST:

```
     [exec] Warning, treated as error:
     [exec] /home/travis/build/geotools/geotools/docs/user/unsupported/mbstyle.rst:43: SEVERE: Unexpected section title.

     [exec] Color
     [exec] '''''
```

Once that commit is removed, that shouldn't be a problem anymore.

tbarsballe:
I'll manually remove the extraneous doc commit, and merge.

tbarsballe:
Manually merged, and backported to 17.x

PR #1588, opened 2017-05-16
GEOT-5730 check if sources are supplied by 'Sources' parameter

jdries:
https://osgeo-org.atlassian.net/browse/GEOT-5730

check if sources are supplied by 'Sources' parameter, which is what the WPS process, and possibily also other JAI-EXT code does. This makes switching between jai and jai-ext more transparant (for the multiply operation). 
Also included unit test, only could not find a way to properly test both jai-ext and jai in one test.


bencaradocdavies:
@jdries changes must first target master. The Travis CI build fails for the PR.

PR #1589, opened 2017-05-16
text-max-width(mbstyle) converts to autoWrap(sld)

johngschulz:


modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/layer/SymbolMBLayer.java
```diff
@@ -1731,7 +1731,12 @@ public Expression textTranslateAnchor() {
         // layer.textHaloBlur();
 
         // auto wrap
-        // symbolizer.getOptions().put("autoWrap", layer.textMaxWidth()); //  Pixels (GS) vs ems (MB); Vendor options with expressions?
+        // uses getTextSize to convert text-max-width from ems to pixels
+        if(getTextMaxWidth() != null) {
+            symbolizer.getOptions().put("autoWrap", String.valueOf(getTextMaxWidth().intValue() * getTextSize().intValue())); 
+        } else{
+            symbolizer.getOptions().put("autoWrap", String.valueOf(getTextSize().intValue() * 10));
```
tbarsballe: Where is the 10 coming from here? Can we perhaps instead try and grab the default text size from StyleFactory?

johngschulz: 10 is the default value for ems in the Mapbox docs, I'll look into the StyleFactory and see what I can find, and work on a test case thanks!

tbarsballe: If 10 is the MapBox default, just add a comment mentioning that and leave it as-is then.

(StyleFactory would just give us the SLD default - best to stick to the MapBox default where it exists)

modules/unsupported/mbstyle/src/test/java/org/geotools/mbstyle/SymbolMBLayerTest.java
```diff
@@ -36,4 +52,17 @@ public void testTextAnchorEnum(){
...
         assertEquals( 0.5, TextAnchor.getAnchorY("center"), 0.0 );
     }
-}
+
+    @Test
+    public void testTextMaxWidth(){
+        //Test default MBStyle value
+        assertEquals(10, testLayerDefault.getTextMaxWidth().intValue());
+        assertEquals(16, testLayerDefault.getTextSize().intValue());
+        assertEquals(160, testLayerDefault.getTextMaxWidth().intValue() * testLayerDefault.getTextSize().intValue());
```
tbarsballe: This line seems a bit redundant, but perhaps do a call to `testLayerDefault.transform()` and ensure that the correct value shows up in the resulting FeatureTypeStyle?

tbarsballe:
I left a minor comment on the test case, but otherwise the changes look good.

Once this is ready to merge, can you squash the commits together to make backporting easier?
Thanks.

johngschulz:
Squashed

PR #1590, opened 2017-05-16
Mbspec fixup

tbarsballe:
Fixed up `info` and `note` admonitions.

Fixed up headings in the root section.

Still todo:

- [x] Headings (with reference) all sections after the **Sources** section
- [x] Supported versions in all sections after the **Sources** section


tbarsballe:
Manually squashed and merged

PR #1595, opened 2017-05-23
stop newlines being eaten in XML parsing of CDATA

ianturton:
Initial fix to prevent GeoTools eating the newlines in CDATA blocks of WPS requests. Needs two minor fixes to existing tests as I can't limit the behaviour to just CDATA blocks as that information seems to be lost by the time GT currently collapses the white space.

### DIFF
--- a/modules/extension/xsd/xsd-core/src/main/java/org/geotools/xml/impl/ParseExecutor.java
+++ b/modules/extension/xsd/xsd-core/src/main/java/org/geotools/xml/impl/ParseExecutor.java
@@ -187,9 +187,6 @@ public class ParseExecutor implements Visitor {
             type = (XSDSimpleTypeDefinition) typeDefinition;
         } else {
             XSDComplexTypeDefinition complexType = (XSDComplexTypeDefinition) typeDefinition;
-            if (complexType.getContentType() instanceof XSDParticle) {
-                
-            }
             if (complexType.getContentType() instanceof XSDSimpleTypeDefinition) {
                 type = (XSDSimpleTypeDefinition) complexType.getContentType();
             }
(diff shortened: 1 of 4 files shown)


aaime:
Would be nice to have a test that checks CDATA though ;)

ianturton:
I just realised that the test in the GS commit - I'll move it to GT tomorrow (or later this evening)

modules/extension/xsd/xsd-core/src/test/java/org/geotools/xml/impl/ParseExecutorTest.java
```diff
@@ -0,0 +1,84 @@
...
+        Configuration configuration = new MLConfiguration();
+        Parser parser = new Parser(configuration);
+        parser.setValidating(true);
+
+        HashMap<String,String> parsed;
+
+        InputStream reader = new ByteArrayInputStream(data.getBytes());
+        parsed = (HashMap<String, String>) parser.parse(reader );
+        for(Entry<String, String> e:parsed.entrySet()) {
+            System.out.println(e);
```
aaime: Leftover printout

modules/extension/xsd/xsd-core/src/test/java/org/geotools/xml/impl/ParseExecutorTest.java
```diff
@@ -0,0 +1,84 @@
...
+            if(e.getKey()!=null && e.getKey().equals("mimeType")) {
```
aaime: Formatting is not conforming with our guidelines (@dromagno always gets me on these too :-p )

modules/extension/xsd/xsd-core/src/main/java/org/geotools/xml/impl/ParseExecutor.java
```diff
@@ -326,10 +327,13 @@ public XSDNamedComponent getDeclaration() {
         } else {
             //type is not simple, or complex with simple content, do a check 
             // for mixed
-            if (instance.getTypeDefinition() instanceof XSDComplexTypeDefinition
-                    && ((XSDComplexTypeDefinition) instance.getTypeDefinition()).isMixed()) {
-                //collape the text
-                text = Whitespace.COLLAPSE.preparse(text);
+            //TODO: If this came from a CDATA block we should not collapse it,
```
aaime: This is a TODO, but still, it removes whitespace collapsing which is normally required (outside of CDATA)

aaime: And still, this is the fix right?

ianturton:
seems as if part of the issue was curl eating new lines if you send data as ASCII - may be other issues but I'm starting again from scratch on this!

PR #1597, opened 2017-05-25
function metadata for between function parameters

jodygarnett:


docs/user/build/maven/eclipse.rst
```diff
@@ -5,21 +5,29 @@ Maven can be used to work with the Eclipse IDE. While direct integration is curr
...
    
    C:\java\geotools\trunk\>mvn eclipse:eclipse
 
 Because maven and eclipse will both use **target/classes** you will need to perform a clean when switching between maven and eclipse for building.
 
+.. note::
+   
+   If you would like maven command line and eclipse IDE to use different output directories::
+    
+      mvn eclipse:eclipse -DoutputDirectory=bin -Dall
```
tbarsballe: If you have -Dall here, you should have it above as well. `-Dall` just makes sure you are building everything

modules/library/main/src/main/java/org/geotools/filter/function/FilterFunction_between.java
```diff
@@ -32,10 +32,10 @@
...
     
     public static FunctionName NAME = new FunctionNameImpl("between",
-            parameter("between", Boolean.class),
-            parameter("value", Object.class), // they can be numbers or strings...
-            parameter("low", Object.class),
-            parameter("high", Object.class));
+            parameter("between", Boolean.class,"Between","Returns true if between min and max."),
+            parameter("value", Object.class, "Value", "Number or string value"),
+            parameter("min", Object.class,"Min","Minimum number or string"),
+            parameter("max", Object.class,"Max","Maximum number or string"));
```
tbarsballe: Does this need a corresponding doc change, or are the docs autogenerated?

jodygarnett: I will generate the docs by hand once we have the commits in from today.

PR #1598, opened 2017-05-25
Adding comments for function 'Contains'

shortycake:
Adding code comments to "contains" function


modules/library/main/src/main/java/org/geotools/filter/function/FilterFunction_contains.java
```diff
@@ -34,10 +34,10 @@
 public class FilterFunction_contains extends FunctionExpressionImpl {
 
     public static FunctionName NAME = new FunctionNameImpl("contains",
-            parameter("contains", Boolean.class),
+            parameter("contains", Boolean.class, "Contains","Returns true if the geometry a contains b"), 
             parameter("geometry1", Geometry.class),
             parameter("geometry2", Geometry.class));
```
jodygarnett: It would be good to add tile and description to geometry1, geometry2

jodygarnett:
Thanks @shortycake welcome to open source development

PR #1602, opened 2017-05-29
[GEOT-5737] DataUtilities.fileToUrl does not percent-encode non-ASCII characters (master)

bencaradocdavies:
https://osgeo-org.atlassian.net/browse/GEOT-5737

This PR also ports updates to the cut-down version of `DataUtilities` in gt-referencing. Further changes to that class will be the subject of another PR. See: https://osgeo-org.atlassian.net/browse/GEOT-5739.

### DIFF
--- a/modules/library/referencing/src/main/java/org/geotools/referencing/factory/gridshift/DataUtilities.java
+++ b/modules/library/referencing/src/main/java/org/geotools/referencing/factory/gridshift/DataUtilities.java
@@ -35,15 +35,33 @@ import java.net.URLDecoder;
  */
 public class DataUtilities {
 
+    private static final boolean IS_WINDOWS_OS = System.getProperty("os.name").toUpperCase()
+            .contains("WINDOWS");
+
     /**
-     * Copy of DataUtilities.urlToFile(URL url) in gt-main, which cannot be used here because of
-     * cyclic Maven dependencies.
+     * Takes a URL and converts it to a File. The attempts to deal with Windows UNC format specific
+     * problems, specifically files located on network shares and different drives.
+     * 
+     * If the URL.getAuthority() returns null or is empty, then only the url's path property is used
+     * to construct the file. Otherwise, the authority is prefixed before the path.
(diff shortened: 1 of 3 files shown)


modules/library/referencing/src/main/java/org/geotools/referencing/factory/gridshift/DataUtilities.java
```diff
@@ -35,15 +35,33 @@
  */
 public class DataUtilities {
 
+    private static final boolean IS_WINDOWS_OS = System.getProperty("os.name").toUpperCase()
+            .contains("WINDOWS");
+
     /**
-     * Copy of DataUtilities.urlToFile(URL url) in gt-main, which cannot be used here because of
-     * cyclic Maven dependencies.
+     * Takes a URL and converts it to a File. The attempts to deal with Windows UNC format specific
```
jodygarnett: Can we make a Files utility method then for this stuff, and then deprecate the method in the DataUtilities class? Having two classes with the same name in different packages is ... confusing.

modules/library/referencing/src/main/java/org/geotools/referencing/factory/gridshift/DataUtilities.java
```diff
@@ -75,27 +100,23 @@ public static File urlToFile(URL url) {
...
+     * on Mac.
      * 
      * @param file
      * @return URL
      */
     public static URL fileToURL(File file) {
         try {
-            URL url = file.toURI().toURL();
-            String string = url.toExternalForm();
+            String string = file.toURI().toASCIIString();
```
jodygarnett: can you add a comment here about why this method? since this is a messed up story we may as well tell it

bencaradocdavies:
As discussed on Gitter, I will pull the {{fileToURL}} and {{urlToFile}} methods up into gt-metadata.

bencaradocdavies:
I will make a new PR after #1604 is merged.

PR #1603, opened 2017-05-29
Changing HeatMapProcess access modifiers to allow for subclassing

elahrvivaz:


aaime:
Why not contribute the new subclassed process while you're at it?

elahrvivaz:
Our current class is here, but we had to duplicate some of the functionality, which we'd like to just subclass instead: https://github.com/locationtech/geomesa/blob/master/geomesa-process/geomesa-process-vector/src/main/scala/org/locationtech/geomesa/process/analytic/DensityProcess.scala
It sets hints in the query, which are picked up by our data store implementation, and specially encoded features are returned which have serialized pre-computed weight grids in the attribute. Does any of that seem worth contributing? (it's also in scala)

aaime:
You are most likely right, but I needed to check :-)

aaime:
[Formally](https://github.com/geotools/geotools/blob/master/CONTRIBUTING.md) I could not merge this one, you changed two files and I don't know if you have shared a contribution agreement with OSGeo, and there is no JIRA ticket tracking this, but given the changes are really small and the module officially unsupported, oh well, I'll go ahead anyways.

elahrvivaz:
thanks!

jnh5y:
@aaime @elahrvivaz as a note, CCRi filled out the OSGeo contribution agreement.  @elahrvivaz is one of the developers listed on that agreement.  

Totally understand about the JIRA ticket.  Would it be possible to backport this change to the 11.x branch?

PR #1604, opened 2017-05-30
[GEOT-5740] Move DataUtilities URL methods to new URLs class in gt-metadata (master)

bencaradocdavies:
https://osgeo-org.atlassian.net/browse/GEOT-5740

### DIFF
--- a/docs/developer/conventions/code/url.rst
+++ b/docs/developer/conventions/code/url.rst
@@ -15,13 +15,13 @@ C:\direct with a space\file.txt     file://C:/directory%20with%20a%20space/file.
 /Users/Jody/java                    file:///Users/Jody/java/
 =================================== ==================================================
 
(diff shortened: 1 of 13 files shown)


jodygarnett:
Please follow java naming conventions Urls renames to URLs

bencaradocdavies:
@jodygarnett we might need to put `Urls` vs `URLs` to a vote on the mailing list. Looking through the dependencies we have (visible from GeoServer gs-web-app Start) we have a huge mix of `URL` versus `Url` prefixes. There is at least one other `Urls` class out there. I think we ask the developers what they want. There is a modern style that three-letter acronyms get lowercase tail letters when forming CamelCase/PascalCase. Does this rule apply when adding an "s" in the style of `Lists`, `Collections`, and `Streams`? I thought yes, but happy to change it to match majority preference.

bencaradocdavies:
`Urls` renamed to `URLs` throughout.

PR #1606, opened 2017-05-30
Mbstyle max angle

johngschulz:
Add text max angle and test to mbstyles.

### DIFF
--- a/modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/layer/SymbolMBLayer.java
+++ b/modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/layer/SymbolMBLayer.java
@@ -16,13 +16,22 @@
  */
 package org.geotools.mbstyle.layer;
 
+import java.awt.Color;
+import java.awt.Point;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.Set;
+
+import javax.measure.unit.NonSI;
+
 import org.geotools.mbstyle.MBStyle;
 import org.geotools.mbstyle.parse.MBFilter;
 import org.geotools.mbstyle.parse.MBFormatException;
 import org.geotools.mbstyle.parse.MBObjectParser;
 import org.geotools.mbstyle.sprite.SpriteGraphicFactory;
 import org.geotools.mbstyle.transform.MBStyleTransformer;
-import org.geotools.styling.*;
 import org.geotools.styling.AnchorPoint;
 import org.geotools.styling.Displacement;
 import org.geotools.styling.ExternalGraphic;
@@ -37,23 +46,18 @@ import org.geotools.styling.Mark;
 import org.geotools.styling.PointPlacement;
 import org.geotools.styling.Rule;
 import org.geotools.styling.Stroke;
+import org.geotools.styling.StyleBuilder;
+import org.geotools.styling.TextSymbolizer2;
 import org.geotools.text.Text;
 import org.json.simple.JSONArray;
 import org.json.simple.JSONObject;
 import org.opengis.filter.expression.Expression;
 import org.opengis.filter.expression.Literal;
-import org.opengis.style.*;
+import org.opengis.style.GraphicalSymbol;
+import org.opengis.style.SemanticType;
 import org.opengis.style.Symbolizer;
 
 import com.google.common.collect.ImmutableSet;
-
-import javax.measure.unit.NonSI;
-import java.awt.*;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.List;
-import java.util.Set;
 
 /**
  * A symbol.
@@ -1047,6 +1051,7 @@ public class SymbolMBLayer extends MBLayer {
         }
         return sf.anchorPoint(ff.literal(anchor.getX()), ff.literal(anchor.getY()));
     }
+    
     /**
      * (Optional) Units in degrees. Defaults to 45. Requires text-field. Requires symbol-placement = line.
      * 
@@ -1067,6 +1072,14 @@ public class SymbolMBLayer extends MBLayer {
      */
     public Expression textMaxAngle() {
         return parse.percentage(layout, "text-max-angle", 45.0);
+    }
+
+    /**
+     *
+     * @return True if the layer has a text-max-angle explicitly provided.
+     */
+    private boolean hasTextMaxAngle() throws MBFormatException {
+        return parse.isPropertyDefined(layout, "text-max-angle");
     }
 
     /**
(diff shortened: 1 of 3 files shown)


modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/layer/SymbolMBLayer.java
```diff
@@ -1663,7 +1676,7 @@ public Expression textTranslateAnchor() {
         List<Symbolizer> symbolizers = new ArrayList<Symbolizer>();
 
         LabelPlacement labelPlacement;
-
+        Boolean followLine = false;
```
tbarsballe: As far as I can tell, this doesn't do anything, whenever we have a linePlacement, followline is always set to true.

modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/layer/SymbolMBLayer.java
```diff
@@ -1732,10 +1746,15 @@ public Expression textTranslateAnchor() {
...
 
-        // text max angle
-        // layer.getTextMaxAngle();
-        // symbolizer.getOptions().put("maxAngleDelta", "40");
-
+        // text max angle - only for line placement
+        // throw MBFormatException if point placement
+        if(labelPlacement instanceof LinePlacement){
+            // will be true if line placement, false else.
+            symbolizer.getOptions().put("followLine", String.valueOf(followLine));
```
tbarsballe: Is there any way this can ever be false? From what I can see, the only time labelPlacement is a LinePlacement is when we have `symbol-placement = "line"`, in which case `followLine` always is true.

modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/layer/SymbolMBLayer.java
```diff
@@ -1732,10 +1746,15 @@ public Expression textTranslateAnchor() {
...
+            symbolizer.getOptions().put("maxAngleDelta", String.valueOf(getTextMaxAngle()));
+        }else if(labelPlacement instanceof PointPlacement && hasTextMaxAngle()){
```
tbarsballe: `labelPlacement instanceof PointPlacement` is redundant. Just need `} else if (hasTextMaxAngle()) {`

Also, take a look at your formatting - there should be spaces as in my example above.

modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/layer/SymbolMBLayer.java
```diff
@@ -1047,6 +1051,15 @@ public AnchorPoint anchorPoint() {
...
         return sf.anchorPoint(ff.literal(anchor.getX()), ff.literal(anchor.getY()));
     }
+    
+    /**
+     * 
+     * @return True if the layer has a text-max-angle explicitly provided.
+     */
+    public boolean hasTextMaxAngle() throws MBFormatException {
+        return parse.isPropertyDefined(layout, "text-max-angle");
+    }
```
tbarsballe: Maybe group this with wherever getTextMaxAngle is defined, just to make it easier to find related methods?

Also, does this need to be public? Probably better if it is protected or private.

modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/layer/SymbolMBLayer.java
```diff
@@ -1732,10 +1746,15 @@ public Expression textTranslateAnchor() {
...
-        // symbolizer.getOptions().put("maxAngleDelta", "40");
-
+        // text max angle - only for line placement
+        // throw MBFormatException if point placement
+        if(labelPlacement instanceof LinePlacement){
+            // will be true if line placement, false else.
+            symbolizer.getOptions().put("followLine", String.valueOf(followLine));
+            symbolizer.getOptions().put("maxAngleDelta", String.valueOf(getTextMaxAngle()));
+        }else if(labelPlacement instanceof PointPlacement && hasTextMaxAngle()){
+        	throw new MBFormatException("Text labels using point placement cannot utilize maxAngleDelta");
```
tbarsballe: Looking at the code, the format exceptions seem to typically be thrown by parse. However, since this depends on two properties, I think it is fine here.

Maybe update the wording a bit to better match the error messages in MBObjectParser and the parameters defined in the MapBox spec. How about:

```
"Property text-max-angle requires symbol-placement = line but symbol-placement = " + symbolPlacementVal
```

We'll probably need to revisit this when symbol placement supports functions.

modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/layer/SymbolMBLayer.java
johngschulz: I moved has `TextMaxAngle()` below the other textMaxAngle methods, it was above them before.  As far as public vs. private or protected, I was following suit for the surrounding code, which all appear to be public as well.

johngschulz:
Fixed and squashed.

modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/layer/SymbolMBLayer.java
tbarsballe: For future reference, most of the surrounding code was core API - namely the getters for the actual parameters in the MBStyle spec, hence why it was public. This is more of a utility method, so doesn't need to be public.

modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/layer/SymbolMBLayer.java
```diff
@@ -1732,10 +1744,15 @@ public Expression textTranslateAnchor() {
...
-        // symbolizer.getOptions().put("maxAngleDelta", "40");
-
+        // text max angle - only for line placement
+        // throw MBFormatException if point placement
+        if(labelPlacement instanceof LinePlacement){
+            // followLine will be true if line placement, it is an implied default of MBstyles.
+            symbolizer.getOptions().put("followLine", "true");
+            symbolizer.getOptions().put("maxAngleDelta", String.valueOf(getTextMaxAngle()));
+        } else if (hasTextMaxAngle()) {
+        	throw new MBFormatException("Property text-max-angle requires symbol-placement = line but symbol-placement = " + symbolPlacementVal + "\"");
```
tbarsballe: You've got a trailing escaped quote, and no opening quote here: `\"`. It should be removed.

tbarsballe:
Looking better. I've got a couple more minor comments.

modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/layer/SymbolMBLayer.java
johngschulz: Noted and changed.

PR #1610, opened 2017-06-01
Initial commit of S3 geotiff unsupported module.

dvntucker:
This PR creates a new community module to support serving geotiff over s3.


modules/unsupported/pom.xml
```diff
@@ -195,7 +195,13 @@
...
                 <module>mbstyle</module>
             </modules>
-        </profile>        
+        </profile>
+		<profile>
+			<id>s3-geotiff</id>
+			<modules>
+				<module>s3-geotiff</module>
+			</modules>
+		</profile>
```
jodygarnett: Are these tabs rather than spaces

modules/unsupported/s3-geotiff/pom.xml
```diff
@@ -0,0 +1,66 @@
...
+    </dependency>
+    <dependency>
+      <groupId>net.sf.ehcache</groupId>
+      <artifactId>ehcache</artifactId>
+      <version>2.10.3</version>
+    </dependency>
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <version>4.12</version>
```
jodygarnett: un comfortable listing specific junit dependency here, should be define by parent or parameter?

dvntucker:
Fixed those issues.

PR #1615, opened 2017-06-03
Css units related improvements

aaime:
- Usage of metrix prefixes (that's how SI calls them, because they normally come before the unit, while in CSS case they are suffix to the value because a scale is just a ratio, it's unitless)
- Allow using "@sd" instead of "@scale" in scale denominator rules


bradh:
Looks good. Only consideration would be if we want to keep "m" (for a future milli, if that becomes important in the future) and "M" (mega) distinct. Trade-off for consistency (where other units are case insensitive) vs future flexibility.

aaime:
Restricted to "kMG", merging.

PR #1617, opened 2017-06-07
[GEOT-5754] Support data uris as external graphics

hwbllmnn:
I tried to use the generic jdk mechanism to support custom protocols, but that is unusable when working in a servlet environment with eg. Tomcat or Jetty.

### DIFF
--- a/modules/library/main/src/main/java/org/geotools/util/CommonsConverterFactory.java
+++ b/modules/library/main/src/main/java/org/geotools/util/CommonsConverterFactory.java
@@ -22,7 +22,6 @@ import java.math.BigInteger;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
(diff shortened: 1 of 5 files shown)


aaime:
Note, please discuss on the list (next time, suggestion, before doing the coding), but what you're after here seems to be already supported as part of SLD 1.1 external graphic "inline content", see this test dataset used in the rendering module: 
https://github.com/geotools/geotools/blob/master/modules/library/render/src/test/resources/org/geotools/renderer/lite/test-data/base64.sld

In any case, please let's start a conversation on the devel list, where there is a full audience, instead of inside the pull request or jira ticket.

aaime:
Looks like the discussion is over and no issues were found, merged

aaime:
Ah, since this is not a bug fix, if you need a backport you can propose one in a month from the merge on master.

PR #1618, opened 2017-06-07
Implemented MBBackgroundLayer with initial testing

johngschulz:
BackgroundMBlayer testing

implementing background and testing for mbstyle

Implemented MBBackgroundLayer and added testing


johngschulz:
In the request and rendering in geoserver all look good in manual testing and the tiles line up as expected with no border.  I'm not really sure why the border is in the background.png - I thought it might just be a bi-product of swing.  It's a good question though.

tbarsballe:
Ok. It could just be an artifact of the image comparison library. Merging.

PR #1619, opened 2017-06-08
Assorted mixed CRS mosaic fixes

aaime:
Cumulative fixes for three code paths used in mixed CRS mosaisc with footprints:
*    [GEOT-5755] Mosaic with mixed CRS can miss areas when footprint management is set to transparent
*    [GEOT-5757] ImageWorker.mosaic mail fail with NPE if some granules have a ROI while others do not
*   [GEOT-5756] Coverage Crop can fail with TopologyException


modules/library/coverage/src/main/java/org/geotools/image/ImageWorker.java
```diff
@@ -4465,6 +4465,16 @@ private ROI mosaicROIs(Vector sources, ROI... roiArray) {
                 rois.add(r);
             }
         }
+        // handle possible null inputs
+        for (int i = 0; i < roiArray.length; i++) {
+            ROI roi = roiArray[i];
+            if(roi == null) {
+                RenderedImage img = (RenderedImage) sources.get(i);
+                ROI r = new ROIShape(new Rectangle(img.getMinX(), img.getMinY(), img.getWidth(), img.getHeight()));
```
dromagnoli: Wondering if would be better to use ROIGeometry instead.
However, in the previous lines we are still using ROIShape indeed, so no objection on using ROIShape in the new code too, to avoid mixing types.

aaime:
You're right about ROIGeometry vs ROIShape usage, fixed it. And formatting as well.

PR #1620, opened 2017-06-08
Mbstyle text-keep-upright

johngschulz:
Also changed naming in the symbol layer tests for clarification purposes.


tbarsballe:
This looks like it unintentionally contains the commit from https://github.com/geotools/geotools/pull/1618

johngschulz:
I think it should be good to go now.

modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/layer/SymbolMBLayer.java
```diff
@@ -1748,6 +1748,7 @@ public Expression textTranslateAnchor() {
         // throw MBFormatException if point placement
         if(labelPlacement instanceof LinePlacement){
             // followLine will be true if line placement, it is an implied default of MBstyles.
+        	symbolizer.getOptions().put("forceLeftToRight", String.valueOf(textKeepUpright()));
```
tbarsballe: Looks like there is a tab here - should be using spaces.

johngschulz:
Fixed and squashed thanks!

PR #1621, opened 2017-06-08
Implementation of text-padding

johngschulz:
Also includes the .png comparison file for text-keep-upright visual transformer test.  Sorry for the overlap.


johngschulz:
Have to adjust implementation to include icon-padding

tbarsballe:
Looks good to me.

PR #1624, opened 2017-06-15
[GEOT-5697] fix encoding the filter Not( is null ) as XML

mprins:
- [x] Add a failing testcase to illustrate [GEOT-5697](https://osgeo-org.atlassian.net/browse/GEOT-5697)
  - [x] Add a fix


mprins:
I Would like to have feedback the correctness of this, it took me some time to get a grasp of the way the testing works, and still not 100% convinced...

Will prepare backports if this is OK.

aaime:
@mprins, I'm just as surprised as you by the diff :-)
@jdeolive do you have time for a quick look?

jdeolive:
@mprins @aaime Change looks good to me.

mprins:
Anything I can do to get this merged? it would be nice to have this fixed in 17.2 (having missed the 16.4 release)

PR #1625, opened 2017-06-16
[GEOT-5697] fix encoding the filter Not( is null ) as XML (16.x backport)

mprins:
backport of #1624 to 16.x


mprins:
it seems this failed on Travis-CI because the build job was terminated `The job exceeded the maximum time limit for jobs, and has been terminated.`

aaime:
Restarting build

aaime:
Oh well, it did not finish but was killed on the documentation generator, good enough.

PR #1628, opened 2017-06-16
[GEOT-5232] Added a toString method for temporal filters.

sikeoka:
This pull request adds a toString method for temporal filters that is similar to what is currently used for other types of filters.  It also modifies the toString method for temporal objects that are used in the temporal filters to remove their newlines.  The newlines make the logs difficult to read, especially in some GeoServer requests where the filter string may be displayed multiple times per request.

Unit tests were added that simply checks that the default Object toString method is not being used.

The following CQL filter in a GeoServer request:
> issuetime DURING 2017-06-15T11:55:00.000Z/2017-06-15T15:15:00.000Z

instead of displaying a useless string like:
> org.geotools.filter.temporal.DuringImpl@e998726e

now displays the following in the logs:
> [ issuetime During Period{begin:Instant{position:Position{Thu Jun 15 11:55:00 UTC 2017}}, end:Instant{position:Position{Thu Jun 15 15:15:00 UTC 2017}}} ]

This pull request can be backported to 17.x and 16.x.

### DIFF
--- a/modules/library/main/src/main/java/org/geotools/filter/temporal/BinaryTemporalOperatorImpl.java
+++ b/modules/library/main/src/main/java/org/geotools/filter/temporal/BinaryTemporalOperatorImpl.java
@@ -9,6 +9,7 @@
  */
 package org.geotools.filter.temporal;
 
+import org.geotools.filter.visitor.OperatorNameFilterVisitor;
 import org.opengis.filter.expression.Expression;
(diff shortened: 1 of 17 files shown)


sikeoka:
The Travis build timed out:
The job exceeded the maximum time limit for jobs, and has been terminated.

PR #1633, opened 2017-06-26
Implementing text-font interval stop functionality for MBSymbolLayer

johngschulz:
added functions for font to MBfunction and MBObject parser

debugging new implementations of text-font

more testing

updated test file and removed extranious textExpression logic

Added initial testing

working tested text-font for first-font in array

### DIFF
--- a/modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/layer/SymbolMBLayer.java
+++ b/modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/layer/SymbolMBLayer.java
@@ -34,7 +34,6 @@ import org.opengis.filter.expression.Literal;
 import org.opengis.style.GraphicalSymbol;
 import org.opengis.style.SemanticType;
 import org.opengis.style.Symbolizer;
-
 import javax.measure.unit.NonSI;
 import java.awt.*;
 import java.util.*;
@@ -828,16 +827,16 @@ public class SymbolMBLayer extends MBLayer {
      *
      * @return The font to use for the label
      */
+
     public List<String> getTextFont() {
         String[] fonts;
         if (layout.get("text-font") instanceof JSONObject) {
-            JSONArray stops = (JSONArray) ((JSONObject) layout.get("text-font")).get("stops");
-            fonts = ((JSONArray)((JSONArray) stops.get(0)).get(1)).get(0).toString().split(",");
+            return null;
         } else {
             fonts = parse.array(String.class, layout, "text-font",
-                    new String[] { "Open Sans Regular", "Arial Unicode MS Regular" });
+                    new String[]{"Open Sans Regular", "Arial Unicode MS Regular"});
+            return Arrays.asList(fonts);
         }
-        return Arrays.asList(fonts);
     }
 
     /**
@@ -845,13 +844,9 @@ public class SymbolMBLayer extends MBLayer {
      *
      * @return The font to use for the label
      */
-    public List<Expression> textFont() {
-        List<Expression> fontExpressions = new ArrayList<>();
-        String[] fonts = parse.array(String.class, layout, "text-font", new String[] {"Open Sans Regular","Arial Unicode MS Regular"});
-        for (int i = 0; i < fonts.length; i++) {
-            fontExpressions.add(ff.literal(fonts[i]));
-        }
-        return fontExpressions;
+
+    public Expression textFont() throws MBFormatException {
+        return parse.font(layout, "text-font");
     }
 
     /**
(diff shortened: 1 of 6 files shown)


modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/layer/SymbolMBLayer.java
```diff
@@ -828,30 +827,26 @@ private boolean hasTextField() throws MBFormatException {
...
      * @return The font to use for the label
      */
+
     public List<String> getTextFont() {
         String[] fonts;
-        if (layout.get("text-font") instanceof JSONObject) {
-            JSONArray stops = (JSONArray) ((JSONObject) layout.get("text-font")).get("stops");
-            fonts = ((JSONArray)((JSONArray) stops.get(0)).get(1)).get(0).toString().split(",");
-        } else {
+        if (!(layout.get("text-font") instanceof JSONObject)) {
```
tbarsballe: Bit of a nitpick, but since we already have an if-else, using:

```
if (layout.get("text-font") instanceof JSONObject) {
    return null
} else {
    ...
}
```

would be a bit clearer.

modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/parse/MBObjectStops.java
```diff
@@ -199,7 +199,7 @@ boolean containsStops(JSONObject jsonObject) {
             String key = (String) keys.next();
             if (jsonObject.get(key) instanceof JSONObject) {
                 JSONObject child = (JSONObject) jsonObject.get(key);
-                if (child.containsKey("stops")) {
+                if (child.containsKey("stops") && ((JSONArray)((JSONArray)child.get("stops")).get(0)).get(0) instanceof JSONObject) {
```
tbarsballe: This looks brittle, and possibly wrong.

Taking one of the examples below
```
"stops": [
             [10000, "red"],
             [100000, "blue"],
             [1000000, "green"],
             [1000000000, "black"]
           ]
```

stops.get(0) = [10000, "red"]
stops.get(0).get(0) = 10000, which should be a number, not a JSONObject.

If you did want to do more checks, you should probably do some `instanceof JSONArray` checks too, rather than just blindly casting.

modules/unsupported/mbstyle/src/test/java/org/geotools/mbstyle/SymbolMBLayerTest.java
```diff
@@ -115,5 +123,25 @@ public void testIconPadding(){
...
+    @Test
+    public void testTextFont(){
+        MBLayer fontLayer = (SymbolMBLayer) fontStyle.layer("text-font");
+        List<FeatureTypeStyle> featureTypeFont = fontLayer.transformInternal(fontStyle);
+        assertEquals(true, ((JSONObject) fontLayer.getLayout().get("text-font")).containsKey("stops"));
+        assertEquals("Apple-Chancery", ((JSONArray)((JSONArray)((JSONArray)((JSONObject) fontLayer.getLayout().get("text-font")).get("stops")).get(0)).get(1)).get(0));
+        assertEquals("Apple-Chancery", ((CategorizeFunction)((TextSymbolizer)featureTypeFont.get(0).rules().get(0).getSymbolizers()[0]).fonts().get(0).getFamily().get(0)).getParameters().get(1).toString());
+        System.out.println(((TextSymbolizer) featureTypeDefaults.get(0).rules().get(0).getSymbolizers()[0]).fonts().get(0).getFamily());
+        assertEquals("Open Sans Regular", ((TextSymbolizer) featureTypeDefaults.get(0).rules().get(0).getSymbolizers()[0]).fonts().get(0).getFamily().get(0).toString());
+//        StyledLayerDescriptor sld = fontStyle.transform();
```
tbarsballe: Shouldn't be commiting commented out code

modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/parse/MBObjectStops.java
johngschulz: That was what was in place before David Vicks last change to this class and appeared to me to have maybe broken the SLD transformation - will revisit and see if I can correct it.

johngschulz:
I fixed the if/else statement (appreciate the "nit-pick" I want to use/learn best practices).
The MBObjectStops had a ! that was undoing all of the corrections for the containsStops() removing it seems to have negated the reversal and should be corrected.
Removed commented out code from the tests.

johngschulz:
Upon further investigation it looked like the boolean variable in MBObjectStops pre-processing kept conflicting with itself, I refactored it - let me know if it looks okay to you.

modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/parse/MBObjectStops.java
```diff
@@ -49,13 +49,7 @@
      */
     public MBObjectStops(MBLayer layer) {
         try {
-            if (layer.getPaint() != null) {
```
tbarsballe: We still need these null checks, else getStops will throw an NPE id paint or layout are not present (which is perfectly valid for some layers)

PR #1634, opened 2017-06-27
[GEOT-5765]: contrastEnhancement in channelSelection is ignored when …

dromagnoli:
…bandSelection is performed


aaime:
Looks like a simple fix, +1

PR #1637, opened 2017-06-28
Factoring out some centroid code from label cache to RendererUtiliites.

jdeolive:
- Adds getGeometryCentroid method for “safe” centroid calculation for polygon
- Adds sampleForCentralPoint for finding “centroid” that lies within
the polygon

This change is in support of [GEOS-8033](https://osgeo-org.atlassian.net/browse/GEOS-8033) where the goal is to add some better placemark placement for polygons. The idea is to factor out some of the rendering code so it can be reused.

### DIFF
--- a/modules/library/render/src/main/java/org/geotools/renderer/label/LabelCacheImpl.java
+++ b/modules/library/render/src/main/java/org/geotools/renderer/label/LabelCacheImpl.java
@@ -1270,7 +1270,7 @@ public final class LabelCacheImpl implements LabelCache {
         if(!pg.contains(centroid)) {
             // resort to sampling, computing the intersection is slow and
             // due invalid geometries can easily break with an exception
-            Point central = RendererUtilities.sampleForCentralPoint(geom, centroid, pg, gf, 5d, -1);
+            Point central = RendererUtilities.sampleForInternalPoint(geom, centroid, pg, gf, 5d, -1);
             if (central != null) {
                 centroid = central;
             }
--- a/modules/library/render/src/main/java/org/geotools/renderer/label/LabelCacheImpl.java
+++ b/modules/library/render/src/main/java/org/geotools/renderer/label/LabelCacheImpl.java
@@ -1261,6 +1261,9 @@ public final class LabelCacheImpl implements LabelCache {
         }
         
         Point centroid = RendererUtilities.getPolygonCentroid(geom);
+        if (centroid == null) {
+            return false;
+        }
         
         // check we're inside, if not, use a different approach
         PreparedGeometry pg = PreparedGeometryFactory.prepare(geom);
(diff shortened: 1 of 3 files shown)


modules/library/render/src/main/java/org/geotools/renderer/label/LabelCacheImpl.java
```diff
@@ -1259,59 +1260,18 @@ private boolean paintPolygonLabel(LabelPainter painter, AffineTransform tempTran
...
-                centroid = geom.getExteriorRing().getCentroid();
-            } catch (Exception ee) {
-                try {
-                    centroid = geom.getFactory().createPoint(geom.getCoordinate());
-                } catch (Exception eee) {
-                    return false; // we're hooped
-                }
-            }
-        }
+        Point centroid = RendererUtilities.getPolygonCentroid(geom);
```
aaime: The code used to return false in case of failure, now centroid gets set to null and then will NPE on the prepared geometry check?

modules/library/render/src/main/java/org/geotools/renderer/lite/RendererUtilities.java
```diff
@@ -874,4 +876,105 @@ static FeatureCollection fixFeatureCollectionReferencing(FeatureCollection featu
...
+     * </p>
+     * 
+     * @param geom The polygon.
+     * @param centroid The centroid of the polygon, can be null in which case it will be computed from {@link #getPolygonCentroid(Polygon)}.
+     * @param pg The prepared version of geom, can be null in which case it will be computed on demand.
+     * @param gf The geometry factory, can be null in which case the polygons factory will be used.
+     * 
+     * @return A central point that lies inside of the polygon, or null if one could not be found.
+     */
+    public static Point sampleForCentralPoint(Polygon geom, Point centroid, PreparedGeometry pg, GeometryFactory gf,
```
aaime: sampleForInternalPoint? sampleForLabelPoint?

modules/library/render/src/main/java/org/geotools/renderer/label/LabelCacheImpl.java
jdeolive: Whoops :) Good catch.

modules/library/render/src/main/java/org/geotools/renderer/lite/RendererUtilities.java
jdeolive: Both of those names make sense to me... I guess I'll go with sampleForInternalPoint, let me know if you like sampleForLabelPoint better and I'll happily rename.

aaime: sampleForInternalPoints sounds good

jdeolive:
Thanks for the review @aaime. Now handling the failure properly and renamed the method.

jdeolive:
@aaime Went ahead and merged this assuming you were ok with the changes.

PR #1640, opened 2017-07-04
GEOT-5769: GridCoverageRenderer BandSelect optmization being applied …

dromagnoli:
…2 times in a row

### DIFF
--- a/modules/library/render/src/main/java/org/geotools/renderer/lite/gridcoverage2d/GridCoverageRenderer.java
+++ b/modules/library/render/src/main/java/org/geotools/renderer/lite/gridcoverage2d/GridCoverageRenderer.java
@@ -1198,18 +1198,9 @@ public final class GridCoverageRenderer {
                     .toString());
 
         setupInterpolationHints(interpolation);
-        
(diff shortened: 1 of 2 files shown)


modules/library/render/src/main/java/org/geotools/renderer/lite/gridcoverage2d/GridCoverageRenderer.java
```diff
@@ -1198,18 +1198,9 @@ public void paint(final Graphics2D graphics, final GridCoverage2DReader gridCove
...
         setupInterpolationHints(interpolation);
-        
-        RasterSymbolizer finalSymbolizer = symbolizer;
-        //
-        // Band selection
-        //
-        if (isBandsSelectionApplicable(gridCoverageReader, symbolizer)){
-            readParams = applyBandsSelectionParameter(gridCoverageReader, readParams, symbolizer);
-            finalSymbolizer = setupSymbolizerForBandsSelection(symbolizer);
-        }
```
dromagnoli: This check is already made in the renderImage call of the next line

PR #1648, opened 2017-07-12
[GEOS-8034] Fixed ysld color conversion

hwbllmnn:
This fixes GEOS-8034. I'm unsure why the literals were using java.awt.Color objects before, even the tests were checking for that. I tested with Geoserver 2.11 and ysld styles work just fine with this fix. I'll also provide a fix for the master branch shortly.

### DIFF
--- a/modules/extension/ysld/src/main/java/org/geotools/ysld/parse/Util.java
+++ b/modules/extension/ysld/src/main/java/org/geotools/ysld/parse/Util.java
@@ -218,13 +218,7 @@ public class Util {
         if (value != null)
             value = value.toString();
 
-        String hex = null;
-        if (color != null) {
-            hex = String.format("#%02x%02x%02x", color.getRed(), color.getGreen(), color.getBlue());
-            hex = hex.toUpperCase();
-        }
-
-        return color != null ? factory.filter.literal(hex) : expression((String) value, factory);
+        return color != null ? factory.filter.literal(color) : expression((String) value, factory);
     }
 
     static Color parseColorAsHex(Matcher m) {
--- a/modules/extension/ysld/src/main/java/org/geotools/ysld/parse/Util.java
+++ b/modules/extension/ysld/src/main/java/org/geotools/ysld/parse/Util.java
@@ -218,7 +218,13 @@ public class Util {
         if (value != null)
             value = value.toString();
 
-        return color != null ? factory.filter.literal(color) : expression((String) value, factory);
+        String hex = null;
+        if (color != null) {
+            hex = String.format("#%02x%02x%02x", color.getRed(), color.getGreen(), color.getBlue());
+            hex = hex.toUpperCase();
+        }
+
+        return color != null ? factory.filter.literal(hex) : expression((String) value, factory);
     }
 
     static Color parseColorAsHex(Matcher m) {
--- a/modules/extension/ysld/src/test/java/org/geotools/ysld/parse/UtilTest.java
+++ b/modules/extension/ysld/src/test/java/org/geotools/ysld/parse/UtilTest.java
@@ -19,8 +19,6 @@ package org.geotools.ysld.parse;
 
 import static org.junit.Assert.assertEquals;
 
-import java.awt.Color;
-
 import org.junit.Test;
 import org.opengis.filter.expression.Literal;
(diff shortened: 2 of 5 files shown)


aaime:
Hey, I'm not sure about the merits of the pull request (unfamiliar with ysld and the code backing it), but from a procedure point of view, every change needs to land on master first, and then be backported. At this point I'd probably wait for the module maintainers review.

hwbllmnn:
Hi, sure, see #1649 . Am I at least correct in the assumption that OGC literals for color values should be created with a hex color string as their value? Or can/should they be created with java.awt.Color objects as their value?

aaime:
The rendering subsystem should be happy with java.awt.Color as well, as far as I know. Maybe if you're trying to covert ysld to SLD, it can cause issues in the code dumping down the style.

modules/extension/ysld/src/main/java/org/geotools/ysld/parse/Util.java
```diff
@@ -218,7 +218,13 @@ public static Expression color(Object value, Factory factory) {
         if (value != null)
             value = value.toString();
 
-        return color != null ? factory.filter.literal(color) : expression((String) value, factory);
```
tbarsballe: Rather than changing the behavior within YSLD, you should instead update the GeoTools [color converter factory] (https://github.com/geotools/geotools/blob/master/modules/library/main/src/main/java/org/geotools/util/ColorConverterFactory.java) to properly handle these values.

modules/extension/ysld/src/test/java/org/geotools/ysld/parse/UtilTest.java
```diff
@@ -29,12 +27,12 @@
...
-        assertEquals(new Color(0xFFFF00), ((Literal) Util.color("0xFFFF00", factory)).getValue());
-        assertEquals(new Color(0x00FFFF),
+        assertEquals("#FF0000", ((Literal) Util.color("FF0000", factory)).getValue());
+        assertEquals("#00FF00", ((Literal) Util.color("00ff00", factory)).getValue());
+        assertEquals("#0000FF", ((Literal) Util.color("#0000FF", factory)).getValue());
+        assertEquals("#FFFF00", ((Literal) Util.color("0xFFFF00", factory)).getValue());
+        assertEquals("#00FFFF",
                 ((Literal) Util.color("rgb(0, 255, 255)", factory)).getValue());
-        assertEquals(new Color(0xFF00FF), ((Literal) Util.color("f0f", factory)).getValue());
+        assertEquals("#FF00FF", ((Literal) Util.color("f0f", factory)).getValue());
```
tbarsballe: These tests should still work as-is. If they don't, you may have broken some other behaviour in YSLD.

modules/extension/ysld/src/test/java/org/geotools/ysld/parse/YsldParseTest.java
```diff
@@ -1240,13 +1235,13 @@ public void testColourMapEmpty() throws Exception {
...
+        String colour1 = (String) map.getColorMapEntry(0).getColor().evaluate(null);
+        String colour2 = (String) map.getColorMapEntry(1).getColor().evaluate(null);
+        String colour3 = (String) map.getColorMapEntry(2).getColor().evaluate(null);
 
-        assertThat(colour1, is(Color.RED));
-        assertThat(colour2, is(Color.GREEN));
-        assertThat(colour3, is(Color.BLUE));
+        assertThat(colour1, is("#FF0000"));
+        assertThat(colour2, is("#00FF00"));
+        assertThat(colour3, is("#0000FF"));
```
tbarsballe: These tests should still work as-is. If they don't, you may have broken some other behaviour in YSLD.

hwbllmnn:
I think I found the real culprit, the YsdlEncodeHandler converted all objects to String (and tries to convert back to number if possible). I added an exception in the case of a literal with a single Object, in that case it is returned directly. I also added a test for this.

Do you think that's the proper solution?

hwbllmnn:
Backported requested changes from master PR.

tbarsballe:
#1649 was merged to master, and manually backported to 18.x and 17.x

PR #1649, opened 2017-07-12
[GEOS-8034] Fix ysld color conversion master

hwbllmnn:
From #1648 

> This fixes GEOS-8034. I'm unsure why the literals were using java.awt.Color objects before, even the tests were checking for that. I tested with Geoserver 2.11 and ysld styles work just fine with this fix. I'll also provide a fix for the master branch shortly.

Bug report https://osgeo-org.atlassian.net/browse/GEOS-8034


modules/extension/ysld/src/main/java/org/geotools/ysld/parse/Util.java
```diff
@@ -218,7 +218,13 @@ public static Expression color(Object value, Factory factory) {
         if (value != null)
             value = value.toString();
 
-        return color != null ? factory.filter.literal(color) : expression((String) value, factory);
+        String hex = null;
+        if (color != null) {
+            hex = String.format("#%02x%02x%02x", color.getRed(), color.getGreen(), color.getBlue());
+            hex = hex.toUpperCase();
+        }
```
jodygarnett: Okay what is this logic trying to do? Is there a bug report?
 
There is a bunch of color converters that cover these cases, would be good to avoid duplication.

jodygarnett:
@smithkm do you want to have a look at this?

jodygarnett:
We have a tension between design (using converters configured with Hints) and YSLD expected functionality.

Functionality to check:

- three digit hex {{'#00F'}}
- octal colors  {{'0x007'}}

modules/extension/ysld/src/test/java/org/geotools/ysld/parse/UtilTest.java
```diff
@@ -29,12 +27,12 @@
     public void testColor() {
         Factory factory = new Factory();
 
-        assertEquals(new Color(0xFF0000), ((Literal) Util.color("FF0000", factory)).getValue());
-        assertEquals(new Color(0x00FF00), ((Literal) Util.color("00ff00", factory)).getValue());
-        assertEquals(new Color(0x0000FF), ((Literal) Util.color("#0000FF", factory)).getValue());
-        assertEquals(new Color(0xFFFF00), ((Literal) Util.color("0xFFFF00", factory)).getValue());
-        assertEquals(new Color(0x00FFFF),
+        assertEquals("#FF0000", ((Literal) Util.color("FF0000", factory)).getValue());
```
jodygarnett: We need to make sure that litearls with color objects are still supported, please do not remove the above tests, but you can add to them :)

hwbllmnn:
Please see my comment on #1648 , I ported the fix here as well.

modules/extension/ysld/src/main/java/org/geotools/ysld/encode/YsldEncodeHandler.java
```diff
@@ -225,6 +225,17 @@ Object toObjOrNull(Expression expr, boolean isname) {
...
+        if (subExpressions.size() == 1) {
+            Expression subExpr = subExpressions.get(0);
+            if (subExpr instanceof Literal) {
+                Object obj = ((Literal) subExpr).getValue();
+                if (!(obj instanceof String)) {
+                    return obj;
+                }
+            }
+        }
+
```
tbarsballe: This seems quite brittle.
Tracing through the existing YSLD logic, it looks like the only time this gets called with a Color an an argument is when it is called by toColorOrNull (line 170 of this class). Rather than risk breaking the behaviour of toObjOrNull, it would be better to modify that method to return the Color object instead.

A simple check prior to calling this method should be sufficient (replace line 170 with the following):

```
Object obj;
if (expr instanceof Literal) {
  obj = ((Literal)expr).getValue();
  if (obj instanceof Color) {
    return obj;
  }
}
obj = toObjOrNull(expr, false);
```

(Additionally, this change violates the api defined in the javadocs for this method - it should return either a String or a Number; if you modify toColorOrNull instead you wont have that issue, as that method is expected to return a Color)

modules/extension/ysld/src/test/java/org/geotools/ysld/encode/YsldEncodeTest.java
```diff
@@ -1320,4 +1322,12 @@ public void testSymbolizerUoMOther() throws Exception {
...
         Ysld.encode(sld(fts), out);
     }
+
+    @Test
+    public void testColorObject() throws Exception {
+        RasterSymbolizerEncoder encoder = new RasterSymbolizerEncoder(null);
+        Literal literal = new FilterFactoryImpl().literal(Color.BLACK);
+        Object obj = encoder.toObjOrNull(literal);
+        assertEquals(obj.getClass(), Color.class);
+    }
```
tbarsballe: With the change suggested above, this test will no longer be valid. Something that tests a full style would probably be better anyway; something like this should work (using the style example reported in the ticket):

```
    @Test
    public void testEncodeColorMapEntry() throws IOException {
        StyledLayerDescriptor style = new YsldParser(new ByteArrayInputStream(
                    ("name:  Test\n" +
                    "title: Test Style title\n" +
                    "abstract: Styling of Test layer\n" +
                    "feature-styles:\n" +
                    "- rules:\n" +
                    "  - title: raster\n" +
                    "    symbolizers:\n" +
                    "      - raster:\n" +
                    "          opacity: 1.0\n" +
                    "          color-map:\n" +
                    "            type: values\n" +
                    "            entries:\n" +
                    "            - ['#e20374', 1.0, 1, Lorem Ipsum (magenta = covered)]").getBytes())).parse();

        RasterSymbolizer symbolizer = (RasterSymbolizer)((NamedLayer)style.getStyledLayers()[0]).styles().get(0).featureTypeStyles().get(0).rules().get(0).symbolizers().get(0);

        ColorMap colorMap = symbolizer.getColorMap();
        RasterSymbolizerEncoder.ColorMapEntryIterator iterator = new RasterSymbolizerEncoder(symbolizer).new ColorMapEntryIterator(colorMap);
        Tuple map = iterator.next();
        assertEquals("('#E20374',1.0,1,Lorem Ipsum (magenta = covered))", map.toString());
    }
```

hwbllmnn:
Ok, I've incorporated your changes, will backport to 17.x now.

tbarsballe:
I've squashed and merged, and will backport manually, so that the changes are encapsulated within a single commit.

PR #1651, opened 2017-07-13
[GEOT-5766] Add a WMTS CoverageReader and related WMTS model

etj:
Proposal:
- https://github.com/geotools/geotools/wiki/Add-a-WMTS-CoverageReader

GEOT ticket:
- https://osgeo-org.atlassian.net/browse/GEOT-5766

Related GeoServer work:
- https://github.com/geoserver/geoserver/wiki/GSIP-162


aaime:
The EMF OGC module should really be called net.opengis.wmts (without the v1), the other modules do not have a version, they host all relevant protocol versions inside the module instead (e.g., the wfs one has v 1.0, 1.1 and 2.0)

aaime:
The wmts OGC module likely contains too much of the GML schema (seems like a fully copy), but I understand that EMF painted us in a dark corner, with nobody really able to control the model generation (I had to change a single bean for WCS in the last month, adding a single module, and it took me a day to figure out a suitable eclipse version, and then I had to make several manual adjustments).
So I'd say let's go with a fat module for the moment, and see later if we can put it on a diet.

modules/extension/wmts/src/main/java/org/geotools/data/wmts/model/TileMatrixSet.java
```diff
@@ -0,0 +1,131 @@
...
+     * @return the crs
+     */
+    public String getCrs() {
+        return crs;
+    }
+
+    public CoordinateReferenceSystem getCoordinateReferenceSystem()
+            throws NoSuchAuthorityCodeException, FactoryException {
+        
+        //TODO: Kill who ever is still using this hack!
```
aaime: Well well now!
I fully agree we should stop using it, but GWC still defaults to it. A more constructive approach:

- In the comment state that it's compatibility with the legacy code used before EPSG recognized the web mercator CRS
- Go to GWC and suggest changing the default to a sane and standard EPSG:3857 instead

modules/extension/wmts/src/main/java/org/geotools/data/wmts/model/TileMatrixSet.java
```diff
@@ -0,0 +1,131 @@
...
+import java.util.ArrayList;
+import org.geotools.data.ows.CRSEnvelope;
+
+import org.geotools.referencing.CRS;
+import org.opengis.referencing.FactoryException;
+import org.opengis.referencing.NoSuchAuthorityCodeException;
+import org.opengis.referencing.crs.CoordinateReferenceSystem;
+
+
+public class TileMatrixSet {
```
aaime: Public class without javadoc. Really, go around and check every new public class, make sure it has a javadoc if it needs to be public, or make it package private otherwise.

modules/extension/wmts/src/main/java/org/geotools/data/wmts/model/WMTSCapabilities.java
```diff
@@ -0,0 +1,537 @@
...
+public class WMTSCapabilities extends Capabilities {
+
+    static public final Logger LOGGER = org.geotools.util.logging.Logging
+            .getLogger("org.geotools.data.wmts");
+
+    private static CoordinateReferenceSystem CRS84;
+    static {
+        try {
+            CRS84 = CRS.decode("CRS:84");
+        } catch (FactoryException e) {
```
aaime: No, please don't each exception and/or write to stdout/stderr

aaime: Btw, if you want a WGS84 lon/lat oriented, DefaultGeographicCRS.WGS84 should do the trick (hopefully)

modules/extension/wmts/src/main/java/org/geotools/data/wmts/model/WMTSCapabilities.java
```diff
@@ -0,0 +1,537 @@
...
+
+                    layer.setLatLonBoundingBox(boundingBoxes.get("CRS:84"));
+
+
+                }
+                layer.setBoundingBoxes(boundingBoxes);
+
+                EList<URLTemplateType> resourceURL = layerType.getResourceURL();
+                if (resourceURL != null && !resourceURL.isEmpty()) {
+                    for (URLTemplateType resource : resourceURL) {
```
aaime: "resource" is not used in the loop, resourceURL is instead... does not seem right.

modules/extension/wmts/src/main/java/org/geotools/data/wmts/model/WMTSCapabilities.java
```diff
@@ -0,0 +1,537 @@
...
+        Map<String, CoordinateReferenceSystem> names = new HashMap<>();
+        for (TileMatrixSet tms : matrixes) {
+
+            try {
+                CoordinateReferenceSystem coordinateReferenceSystem = tms
+                        .getCoordinateReferenceSystem();
+                crs.add(coordinateReferenceSystem);
+                names.put(tms.getIdentifier(), coordinateReferenceSystem);
+            } catch (FactoryException e) {
+                // LOGGER.log(Level.FINER, e.getMessage(), e);
```
aaime: Exception eaten away

modules/extension/wmts/src/main/java/org/geotools/data/wmts/model/WMTSCapabilities.java
```diff
@@ -0,0 +1,537 @@
...
+                        // Take the first good bbox
+                        // TODO: refer a bbox which is natively wgs84
+                        ReferencedEnvelope re = new ReferencedEnvelope(tms.getBbox());
+                        try {
+                            ReferencedEnvelope wgs84re = re.transform(CRS84, true);
+                            wmtsLayer.setLatLonBoundingBox(new CRSEnvelope(wgs84re));
+                            break;
+                        } catch (Exception ex) {
+                            // the RE can't be projected on WGS84, so let's try another one
+                            LOGGER.fine("Can't use " + tms.getIdentifier() +" for bbox");
```
aaime: Guard log...

modules/extension/wmts/src/main/java/org/geotools/data/wmts/model/WMTSService.java
```diff
@@ -0,0 +1,57 @@
...
+ *    This library is distributed in the hope that it will be useful,
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ *    Lesser General Public License for more details.
+ */
+package org.geotools.data.wmts.model;
+
+import java.util.ArrayList;
+import java.util.List;
+import org.eclipse.emf.common.util.EList;
```
aaime: Unused import, please clean them up

modules/extension/wmts/src/main/java/org/geotools/data/wmts/request/AbstractGetTileRequest.java
```diff
@@ -0,0 +1,401 @@
...
+ * @author Emanuele Tajariol (etj at geo-solutions dot it)
+ */
+public abstract class AbstractGetTileRequest extends AbstractWMTSRequest
+        implements GetTileRequest {
+
+    /** MAXTILES */
+    private static final int MAXTILES = 256;
+
+    /** DPI */
+    private static final double DPI = 96.0;
```
aaime: This is not the standard OGC DPI (which would be 91.something), where is it coming from?

modules/extension/wmts/src/main/java/org/geotools/data/wmts/request/AbstractGetTileRequest.java
```diff
@@ -0,0 +1,401 @@
...
+        Tile first = tiles.iterator().next();
+        int z = first.getTileIdentifier().getZ();
+        List<TileMatrixLimits> limits = layer.getTileMatrixLinks().get(matrixSet.getIdentifier())
+                .getLimits();
+        TileMatrixLimits limit;
+        if (!limits.isEmpty()) {
+
+            limit = limits.get(z);
+        } else {
+            // seems that MapProxy (and all REST APIs?) don't create limits
```
aaime: TileMatrixLimits are not required, please change the comment accordingly (no point blaming a software for being within the spec limits)

modules/extension/wmts/src/main/java/org/geotools/data/wmts/request/AbstractGetTileRequest.java
```diff
@@ -0,0 +1,401 @@
...
+            limit = new TileMatrixLimits();
+            limit.setMaxCol(matrixSet.getMatrices().get(z).getMatrixWidth() - 1);
+            limit.setMaxRow(matrixSet.getMatrices().get(z).getMatrixHeight() - 1);
+            limit.setMinCol(0);
+            limit.setMinRow(0);
+            limit.setTileMatix(matrixSet.getIdentifier());
+        }
+        ArrayList<Tile> remove = new ArrayList<>();
+        for (Tile tile : tiles) {
```
aaime: I'm confused, we are asking for some extra tiles just to throw them away afterwards? Wouldn't it be better to ask only for the tiles we really need?

modules/extension/wmts/src/main/java/org/geotools/data/wmts/request/AbstractGetTileRequest.java
```diff
@@ -0,0 +1,401 @@
...
+//    }
+
+    private TileMatrixSet selectMatrixSet() throws ServiceException, RuntimeException {
+        TileMatrixSet retMatrixSet = null;
+
+        Map<String, TileMatrixSetLink> links = layer.getTileMatrixLinks();
+        CoordinateReferenceSystem requestCRS = getCrs();
+        LOGGER.fine("request CRS " + requestCRS.getName());
+        if (requestCRS == null) {
+            try {
```
aaime: Hmm... Eclipse claims the whole content of this if block is dead code...

modules/extension/wmts/src/main/java/org/geotools/data/wmts/request/GetTileRequest.java
```diff
@@ -0,0 +1,57 @@
...
+import org.geotools.data.wmts.model.WMTSLayer;
+import org.geotools.geometry.jts.ReferencedEnvelope;
+import org.geotools.ows.ServiceException;
+import org.geotools.tile.Tile;
+import org.opengis.referencing.crs.CoordinateReferenceSystem;
+
+/**
+ * Construct a WMTS getTile request.
+ *
+ * DOC TODO
```
aaime: Please DO! :-p

modules/extension/wmts/src/main/java/org/geotools/data/wmts/response/WMTSGetCapabilitiesResponse.java
```diff
@@ -0,0 +1,82 @@
...
+    public WMTSGetCapabilitiesResponse(HTTPResponse response, Map<String, Object> hints) throws ServiceException, IOException {
+        super(response);
+
+        try {
+
+            Object object;
+            InputStream inputStream = null;
+            try {
+                inputStream = response.getResponseStream();
+                Parser parser = new Parser(new WMTSConfiguration());
```
aaime: The parser is a throw away, but the configuration is shareable, please move it to a class field. (or at least, it should be, in GeoServer most configurations are Spring singletons)

modules/extension/wmts/src/main/java/org/geotools/map/WMTSCoverageReader.java
```diff
@@ -0,0 +1,521 @@
...
+        tileRequest.setRequestedBBox(gridEnvelope); // should be requestEnvelope?
+        tileRequest.setRequestedTime(time);
+
+        try {
+            this.requestCRS = CRS.decode(requestSrs);
+        } catch (Exception e) {
+            throw new IOException("Could not decode request SRS " + requestSrs);
+        }
+
+        ReferencedEnvelope requestEnvelope = gridEnvelope;
```
aaime: Unused variable

modules/extension/wmts/src/main/java/org/geotools/map/WMTSMapLayer.java
```diff
@@ -0,0 +1,142 @@
...
+    private String rawTime;
+
+    /**
+     * Builds a new WMS layer
+     *
+     * @param wmts
+     * @param layer
+     */
+    public WMTSMapLayer(WebMapTileServer wmts, Layer layer) {
+        super( new WMTSCoverageReader(wmts, layer), STYLE );
```
aaime: Danger danger, Style objects are mutable, if someone gets the style and changes it, it will affect the static variable and all subsequent requests. Better create the style every time.

modules/extension/xsd/xsd-wmts/src/main/java/org/geotools/wmts/bindings/AcceptedFormatsTypeBinding.java
```diff
@@ -0,0 +1,72 @@
+package org.geotools.wmts.bindings;
```
aaime: Not picking on this class in particular, but the module does not have a single test. As a supported module, it should instead. You can have a look at other OGC modules, checking the parsing tests.

aaime:
Ah, forgot, there should be a documentation page for the two new supported modules, and proof they can get into supported land (gold star requirements)

ianturton:
I'll work on the docs soon, maybe this week after I do the workshop review
for foss4g.

Ian

modules/extension/wms/src/main/java/org/geotools/data/ows/Layer.java
```diff
@@ -591,6 +597,8 @@ public void setLatLonBoundingBox(CRSEnvelope latLonBoundingBox) {
             if( !srsName.equals("CRS:84")){
                 throw new IllegalStateException("Layer LatLonBoundingBox srsName required to be null or CRS:84");
             }
+        }else {
```
etj: Ok

modules/extension/wms/src/main/java/org/geotools/data/ows/Layer.java
```diff
@@ -866,6 +863,7 @@ public GeneralEnvelope getEnvelope(CoordinateReferenceSystem crs) {
                 env = new GeneralEnvelope(new double[] { tempBBox.getMinX(),tempBBox.getMinY() },
                         new double[] { tempBBox.getMaxX(), tempBBox.getMaxY() });
                 env.setCoordinateReferenceSystem(crs);
+                LOGGER.warning("Forcing bbox as " + env);
```
etj: The warning is in a catch which is a last resort fallback. 
This fallback is a hack which takes coords from one side, crs from another and puts them together into a GeneralEnvelope; this is very probably going to generate bad data, and really deserves a warning IMHO.

modules/extension/wmts/src/main/java/org/geotools/data/wmts/client/WMTSTile.java
```diff
@@ -0,0 +1,232 @@
...
+        this(new WMTSTileIdentifier(x, y, zoomLevel, service.getName()), service);
+    }
+
+    /**
+     * @param tileIdentifier
+     * @param service
+     */
+    public WMTSTile(WMTSTileIdentifier tileIdentifier, TileService service) {
+        super(tileIdentifier, WMTSTileFactory.getExtentFromTileName(tileIdentifier, service),
+                DEFAULT_TILE_SIZE);
```
etj: Gotcha!

modules/extension/wmts/src/main/java/org/geotools/data/wmts/client/WMTSTile.java
```diff
@@ -0,0 +1,232 @@
...
+        } finally {
+            IOUtils.closeQuietly(is);
+        }
+    }
+
+    private InputStream setupInputStream(URL url, Map<String, String> headers) throws IOException
+    {
+        HttpClient client = new HttpClient();
+        String uri = url.toExternalForm();
+        uri = uri.replaceAll("\\{Time\\}", "20140101");
```
etj: Whoa, old stubbed value. By the time the uri reaches this stage, all the tokens should already have been replaced since long.

bencaradocdavies:
Close/open to retrigger Travis with new settings (failure looks like a timeout).

modules/extension/wms/src/main/java/org/geotools/data/ows/Layer.java
aaime: Ok, leave it there then :)

modules/extension/wmts/src/main/java/org/geotools/data/wmts/request/AbstractGetTileRequest.java
etj: The initial set of tiles is only filtered according to the tilematrix and the requested bbox.
The layer may have further limits, so we are removing tiles which do not contain layer data.
No data is fetched here: we are providing the set of tiles that shall be retrieved in a next step.

etj: Anyway confusion should be avoided :) so I added some more comment lines about what is being done there

modules/extension/wmts/src/main/java/org/geotools/data/wmts/request/AbstractGetTileRequest.java
etj: Yep, in case requestCRS == null, line 347 would throw

modules/extension/wmts/src/main/java/org/geotools/map/WMTSMapLayer.java
etj: Ok

modules/extension/wmts/src/main/java/org/geotools/data/wmts/client/WMTSTileFactory.java
```diff
@@ -0,0 +1,199 @@
...
+            tileMatrixMaxY = tileMatrix.getTopLeft().getX();
+            tileMatrixMinX = tileMatrix.getTopLeft().getY();
+        }
+        // to compensate for floating point computation inaccuracies
+        double epsilon = 1e-6;
+        long xTile = (int) Math.floor((lon - tileMatrixMinX) / tileSpanX + epsilon);
+        long yTile = (int) Math.floor((tileMatrixMaxY - lat) / tileSpanY + epsilon);
+        // to avoid requesting out-of-range tiles
+
+        if (xTile >= tileMatrixLimits.getMaxcol())
```
etj: OK: This part of the code worked perfectly fine for its purposes, where input filtering is properly performed , but it was not general enough indeed.
Refactored / cleaned all that was needed to.

modules/extension/wmts/src/main/java/org/geotools/data/wmts/client/WMTSTileFactory.java
```diff
@@ -0,0 +1,199 @@
...
+            xTile = tileMatrixLimits.getMaxcol() - 1;
+        if (yTile >= tileMatrixLimits.getMaxrow())
+            yTile = tileMatrixLimits.getMaxrow() - 1;
+
+        if (xTile < tileMatrixLimits.getMincol())
```
etj: Same as above :)

modules/extension/tile-client/src/main/java/org/geotools/tile/ImageLoader.java
```diff
@@ -22,13 +22,10 @@
 /**
  * The ImageLoader is a delegate used by a tile to load the image. Its purpose
  * is to provide a tile image cache opportunity to speed up image loading and
- * prevent tile to be repeatedly loaded off the network.
+ * and prevent tile to be repeatedly loaded off the network.
```
bradh: This looks like you now have repeating "and"

modules/extension/tile-client/src/main/java/org/geotools/tile/TileService.java
```diff
@@ -322,27 +318,33 @@ public int getMaxZoomLevel() {
         return tileList;
     }
 
-    private boolean listContainsTile(String tileId) {
-        return !(tiles.peek(tileId) == null || tiles.get(tileId) == null);
-    }
-
+    /**
+     *
```
bradh: Fill in javadoc?

aaime:
Checked, looks good to me. Merging.

PR #1653, opened 2017-07-13
Implementation or raster-brightness-min and raster-brightness-max

johngschulz:
converted min/max brightness to integers within 255 for SLD processing

changed parsed min and max values to type long to fix casting error

Cleanup for testing

### DIFF
--- a/modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/layer/RasterMBLayer.java
+++ b/modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/layer/RasterMBLayer.java
@@ -225,10 +225,17 @@ public class RasterMBLayer extends MBLayer {
 
         // Use of builder is easier for code examples; but fills in SLD defaults
         // Currently only applies the opacity.
+        if(getBrightnessMax() != null || getBrightnessMin() != null){
+            ce.setMethod(ContrastMethod.NORMALIZE);
+            ce.addOption("algorithm", ff.literal("StretchToMinimumMaximum"));
+            Double newMin = (Double)getBrightnessMin() * 255;
+            Double newMax = (Double)getBrightnessMax() * 255;
+            ce.addOption("minValue", ff.literal(String.valueOf(newMin.intValue())));
+            ce.addOption("maxValue", ff.literal(String.valueOf(newMax.intValue())));
+        }
         RasterSymbolizer symbolizer = sf.rasterSymbolizer(getId(), null,
                 sf.description(Text.text("raster"), null), NonSI.PIXEL, opacity(), null,
                 null, null, ce, null, null);
-
         List<Rule> rules = new ArrayList<>();
         MBFilter filter = getFilter();
         org.geotools.styling.Rule rule = sf.rule(getId(), null, null, 0.0, Double.MAX_VALUE,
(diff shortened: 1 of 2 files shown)


modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/layer/RasterMBLayer.java
```diff
@@ -225,10 +225,17 @@ public Number getFadeDuration() {
 
         // Use of builder is easier for code examples; but fills in SLD defaults
         // Currently only applies the opacity.
+        if(getBrightnessMax() != null || getBrightnessMin() != null){
```
bradh: Is this test looking for both of those to not be null? I assume you're trying to prevent multiplying by null in 231/232. If so, you probably wanted `&&`.

I'd also suggest brackets to make it clearer.

Maybe some more spaces here to be consistent with other code too.

tbarsballe: You're using an OR here, but assuming both are not null later. Should add individual null checks and initialize them to their default values in case (for example) max is set but min is not.

modules/unsupported/mbstyle/src/test/java/org/geotools/mbstyle/RasterMBLayerTest.java
```diff
@@ -0,0 +1,54 @@
...
+public class RasterMBLayerTest {
+
+    @Test
+    public void testRasterLayer() throws IOException, ParseException {
+        JSONObject json = MapboxTestUtils.parseTestStyle("rasterStyleTest.json");
+        MBStyle rasterStyle = MBStyle.create(json);
+        RasterMBLayer rasterLayer = (RasterMBLayer) rasterStyle.layer("testid");
+        List<org.geotools.styling.FeatureTypeStyle> featureTypeRaster = rasterLayer.transformInternal(rasterStyle);
+
+        assertEquals(0.59, rasterLayer.getOpacity());
```
bradh: adding some comparison tolerances here would probably make the test less brittle.

modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/layer/RasterMBLayer.java
tbarsballe: You seem to be using the raster-brightness-min and raster-brightness-max to define a contrast function, without handling the mbstyle raster-contrast property. I think you need to take into account both here.

In fact, I am not entirely certain the MapBox brightness contrast settings map especially well to ContrastMethod.Normalize in GeoTools (although I can't seem to find a better match for the functionality).

Given the poor documentation of the ContrastMethod stuff, you'll probably want to do some manual testing to see that this is behaving how you expect - have you done this yet?

johngschulz: Ah, I do think that you are right after testing vs. the intended results on mapbox studio.  In this respect I believe that this would be better used for the contrast property.  That being said, I am still having trouble finding a good example where a whole image is transformed. It looks like currently it can only work band by band - resulting in grayscale images.

jodygarnett:
This is for a comunity module so it is not holding up the code freeze, but the last comment is July 14th.

@johngschulz do you need any additional direction here or can @tbarsballe merge.

tbarsballe:
There are still outstanding changes requested, so this is not ready to merge.

johngschulz:
I've had to switch up the initial approach.  I'm working on tracking some issues I'm having with rendering. I'm able to pull in all of the raster's bands, and have avoided NPEs that were keeping me back.

PR #1656, opened 2017-07-19
GEOT-5773 avoid NPE in GridCoverageRenderer when colormodel is null

jdries:
@dromagnoli 
This can occur with non-byte images. (Remote sensing images commonly use shorts as data type, both in netcdf and geotiff files.)

### DIFF
--- a/modules/library/coverage/src/main/java/org/geotools/coverage/TypeMap.java
+++ b/modules/library/coverage/src/main/java/org/geotools/coverage/TypeMap.java
@@ -510,6 +510,9 @@ public final class TypeMap {
 	public static ColorInterpretation getColorInterpretation(final ColorModel model, final int band)
             throws IllegalArgumentException
     {
+        if (model == null) {
+            return ColorInterpretation.UNDEFINED;
+        }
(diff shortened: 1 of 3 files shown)


dromagnoli:
Hi @jdries 
The PR looks good to me. Thanks for the fix and test
I think that the current travis failure is due to https://osgeo-org.atlassian.net/browse/GEOT-5784
so you have my +1 since it's an unrelated issue, hopefully.

dvntucker:
Looks like the build failure was unrelated.

simboss:
Does this one need a backport? @dromagnoli ?

dromagnoli:
@dvntucker can you backport that and update the JIRA accordingly?

jdries:
@dvntucker Is there still a chance of backporting this bugfix to the 16.x branch?

aaime:
A note, the 16.x branch is "dead", there are no more planned releases on it (there is one in progress right now, but it's already built, just need announcements)

jdries:
Is a backport into 17.x an option then? (I'm upgrading my setup to that version now.)

PR #1658, opened 2017-07-21
Allow injecting different LabelPainters in LabelCacheImpl

smithkm:
Allows the call to `LabelPainter`'s constructor in `LabelCacheImpl` to be replaced with a `BiFunction`.

Injecting an extended `LabelPainter` allows downstream usage of the label placement code at a higher level (as in the GeoServer Vector Tiles module).  It should also be useful for more targeted unit tests.

### DIFF
--- a/modules/library/render/src/main/java/org/geotools/renderer/label/LabelCacheImpl.java
+++ b/modules/library/render/src/main/java/org/geotools/renderer/label/LabelCacheImpl.java
@@ -37,6 +37,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.function.BiFunction;
 import java.util.logging.Level;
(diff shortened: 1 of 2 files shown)


modules/library/render/src/main/java/org/geotools/renderer/label/LabelCacheImpl.java
```diff
@@ -166,6 +167,8 @@
     private VendorOptionParser voParser = new VendorOptionParser();
     
     private List<RenderListener> renderListeners = new CopyOnWriteArrayList<RenderListener>();
+    
+    private BiFunction<Graphics2D, LabelRenderingMode, LabelPainter> constructPainter = LabelPainter::new;
```
jodygarnett: I am so not use to java 8 syntax.

dvntucker:
Looks good, should make downstream use simpler.

PR #1659, opened 2017-07-28
[GEOT-5790] dbtype and filetype should be marked as 'program'-level parameters

tbarsballe:
See https://osgeo-org.atlassian.net/browse/GEOT-5790

Adds LEVEL="program" to DBTYPE / DSTYPE / FILETYPE parameters for all the various datastore implementations.

Ultimately intended to fix https://osgeo-org.atlassian.net/browse/GEOS-8010

### DIFF
--- a/modules/library/jdbc/src/main/java/org/geotools/data/jdbc/datasource/DBCPDataSourceFactory.java
+++ b/modules/library/jdbc/src/main/java/org/geotools/data/jdbc/datasource/DBCPDataSourceFactory.java
@@ -19,6 +19,7 @@ package org.geotools.data.jdbc.datasource;
 import java.io.IOException;
 import java.sql.Connection;
 import java.sql.SQLException;
+import java.util.Collections;
 import java.util.Map;
 
 import javax.sql.DataSource;
@@ -26,6 +27,7 @@ import javax.sql.DataSource;
 import org.apache.commons.dbcp.BasicDataSource;
 import org.geotools.data.DataSourceException;
 import org.geotools.data.DataAccessFactory.Param;
+import org.geotools.data.Parameter;
 
 /**
  * A datasource factory using DBCP connection pool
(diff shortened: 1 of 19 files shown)


modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/ShapefileDataStoreFactory.java
```diff
@@ -82,7 +82,7 @@
      */
     public static final Param FILE_TYPE = new Param("filetype", String.class,
             "Discriminator for directory stores", false, "shapefile", new KVP(Param.LEVEL,
-                    "advanced"));
+                    "program"));
```
tbarsballe: This was advanced before, should it really be changed to program?

modules/library/jdbc/src/main/java/org/geotools/data/jdbc/datasource/DBCPDataSourceFactory.java
```diff
@@ -39,8 +41,8 @@
  */
 public class DBCPDataSourceFactory extends AbstractDataSourceFactorySpi {
 
-    public static final Param DSTYPE = new Param("dstype", String.class,
-            "Must be DBCP", false);
+    public static final Param DSTYPE = new Param("dstype", String.class, "Must be DBCP", false,
+            null, Collections.singletonMap(Parameter.LEVEL, "program"));
```
tbarsballe: Not "DBTYPE", but seems to have the same purpose; presuming we should add this one too.

modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/ShapefileDataStoreFactory.java
jodygarnett: yes, sorry I added program later when it became obvious some things were really only machine editable

bradh:
Would those labels ("advanced", "program", ....) be better as constants?

PR #1662, opened 2017-07-31
GEOT-5793 MultiGeometry not working in GML 3.2 parser/encoded

NielsCharlier:


bencaradocdavies:
Jira issue?

modules/extension/xsd/xsd-gml3/src/main/java/org/geotools/gml3/bindings/MultiGeometryTypeBinding.java
```diff
@@ -112,7 +112,7 @@ public Object parse(ElementInstance instance, Node node, Object value)
     
     @Override
     public Object getProperty(Object object, QName name) throws Exception {
-        if (GML.geometryMember.equals(name)) {
+        if ("geometryMember".equals(name.getLocalPart())) {
```
bencaradocdavies: Bare string "geometryMember"? How about `GML.geometryMember.getLocalPart().equals(name.getLocalPart())`? Is this change to make it work for both GML 3.1 and 3.2?

bencaradocdavies:
@NielsCharlier this looks good and seems to make sense but needs a Jira issue.

NielsCharlier:
Turns out I did make a jira issue, I just forgot to mention it here GEOT-5793

modules/extension/xsd/xsd-gml3/src/main/java/org/geotools/gml3/bindings/MultiGeometryTypeBinding.java
NielsCharlier: Yes, I followed the same pattern of all the other classes which use literals.
But I agree that would be better.

NielsCharlier:
Thanks for your review @bencaradocdavies 
I have made the suggested change

bencaradocdavies:
Looks good @NielsCharlier.

PR #1663, opened 2017-07-31
[GEOT-5794] Add a label strikethrough vendor option

aaime:
@nmco this is underline evil brother! 
Can you have a look? :-)


modules/library/render/src/main/java/org/geotools/renderer/label/LabelPainter.java
```diff
@@ -672,19 +691,28 @@ public void paintCurvedLabel(LineStringCursor cursor) {
             }
 
             // draw halo and label
+            // extracting label first line component and compute is metrics
```
bradh: Can there be multiple lines? If so, can subsequent lines have "worse" metrics here?

Also nit: "is metrics" -> "its metrics".

modules/library/render/src/test/java/org/geotools/renderer/lite/AbstractLabelLineTest.java
```diff
@@ -0,0 +1,89 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ *
+ *    (C) 2016, Open Source Geospatial Foundation (OSGeo)
```
bradh: 2017?

modules/library/render/src/test/resources/org/geotools/renderer/lite/test-data/strikethroughStyle.sld
```diff
@@ -0,0 +1,146 @@
...
+                            <LinePlacement/>
+                        </LabelPlacement>
+                        <VendorOption name="autoWrap">3</VendorOption>
+                        <VendorOption name="strikethroughText">true</VendorOption>
+                    </TextSymbolizer>
+                </Rule>
+            </FeatureTypeStyle>
+        </UserStyle>
+    </NamedLayer>
+</StyledLayerDescriptor>
```
bradh: Maybe a `<Vendor Option name="strikethroughText">false</VendorOption>`, just for variety?

Also newline at end of file.

modules/unsupported/css/src/test/resources/css/cookbook/line/strikethrough.css
```diff
@@ -0,0 +1,6 @@
+* {
+  stroke: red;
+  label: [name];
+  font-fill: black;
+  label-strikethrough-text: true;
+}
```
bradh: Newline at end of file?

modules/library/main/src/main/java/org/geotools/styling/AbstractSymbolizer.java
```diff
@@ -149,9 +149,9 @@ public boolean equals(Object obj) {
         if (options == null) {
             if (other.options != null && !other.options.isEmpty())
                 return false;
-        } else if (!options.equals(other.options))
-            if (options.isEmpty() && other.options != null) 
-                return false;
+        } else if (!options.equals(other.options)) {
+            return false;
+        }
         return true;
```
nmco: This is a very confusing piece of code, so before we had this:
```
if (options == null) {
	if (other.options != null && !other.options.isEmpty())
		return false;
} else if (!options.equals(other.options))
	if (options.isEmpty() && other.options != null) 
		return false;
}
```
and now we have this:
```
if (options == null) {
	if (other.options != null && !other.options.isEmpty())
		return false;
} else if (!options.equals(other.options)) {
	return false;
}
```

I think that old behavior and the new one differ, let's consider this situation:
```
This options and other options are not the same object,
but this options is empty and the other options is NULL.
```
It seems to me that the old code would not return ``FALSE`` but the new code will. I'm not sure if this was the indented behavior or not.

If the intention of this code is to consider two maps equal if the ``equals(...)`` method returns ``TRUE`` or if both are ``NULL`` or empty then I would have wrote something like this:
```
if (options == null || options.isEmpty()) {
	// this options are NULL or empty
	if (other.options != null && !other.options.isEmpty()) {
		// the other options are neither NULL or empty
		return false;
	}
} else if (!options.equals(other.options)) {
	// options are not considered the same
	return false;
}
```

nmco:
Quality work: code was refactored to avoid code duplication, some issues were fixed, tests were added and documentation updated.

I have manually tested this and it seems to be working as expected.

**Strikethrough with halo**

image

**Strikethrough with halo and follow line**

image

**Strikethrough with halo, follow line and underline**

image

I suppose documentation will also be added on GeoServer side, right ?

docs/user/library/render/style.rst
```diff
@@ -21,7 +21,7 @@ StyleFactory
 We have three style factories offering various levels of standards compliance vs pragmatism.
 
 ========== ================ ================ ======== =======================================
-Module     Class            Capability       Scope    Description
+Module     Class            Capability       Scope    Descriptionunder
```
aaime: Nope, I don't, good catch :-)

modules/library/render/src/main/java/org/geotools/renderer/label/LabelPainter.java
aaime: I'm not sure about the line metrics changing per line. Maybe? But until we see an actual issue, I'd let it stay (the code was already there in @nmco version)

modules/library/render/src/test/java/org/geotools/renderer/lite/AbstractLabelLineTest.java
aaime: The file is actually a copy and refactor of a tests @nmco made in 2016, so that's the right date.

modules/library/render/src/test/resources/org/geotools/renderer/lite/test-data/strikethroughStyle.sld
aaime: I was surprised by this comment and looked around, seems like a unix convention that some tools require to consider a line.. a line. As far as I know, it's not part of the GeoTools coding conventions, and files in the IDE look like having an extra line that is there for nothing (they look wrong to me) but... oh well, it's an easy change.

modules/library/main/src/main/java/org/geotools/styling/AbstractSymbolizer.java
aaime: Yes, the old behavior differs, but I haven't found a good reason for it (build still works) and normal java equality does not consider null and empty as equals as equals so I just removed it.
However, someone outside of geotools/geoserver might be depending on this, it's observable behavior indeed. So I've picked your suggestion and merged it in.

modules/library/render/src/test/resources/org/geotools/renderer/lite/test-data/strikethroughStyle.sld
aaime: Hum... I'm confused, I added the newline, my IDE shows it, git tells me the file is committed, and yet it seems the extra newline did not reach github....

bencaradocdavies: [This discussion on Stack Overflow](https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline) references the the POSIX definition of a line as ending with a newline and the requirement of ANSI and ISO C that source files end with a newline. Files that do not end with a newline are defined as non-text files under POSIX because they do not consist of solely of lines.

bencaradocdavies: And the "blank line" at the end of a file in Eclipse (etc) is not a blank line: it is Eclipse showing you where your next line would be should you start typing it.  :-)

aaime:
@nmco yes I will follow up with docs update on the GeoServer side

modules/library/render/src/test/resources/org/geotools/renderer/lite/test-data/strikethroughStyle.sld
aaime: @bencaradocdavies so what is a Eclipse user to do in order to ensure newlines at the end of the file? :-)

aaime: Ah, ok: https://stackoverflow.com/questions/13478373/how-to-remove-empty-line-from-the-end-of-new-class-eclipse

PR #1664, opened 2017-07-31
Fixed test to remove '' from true false assertions

johngschulz:


tbarsballe:
Looks good, fixes failing MBFilterTests for me

PR #1666, opened 2017-08-01
fixed failing tests due to test data changes, logic bug in MBObjectSt…

vickdw:
…ops, missing test images, and wrong assertions.


modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/parse/MBObjectStops.java
```diff
@@ -370,4 +378,41 @@ JSONObject reduceJsonForRange(JSONObject jsonObject, long[] range) {
...
+    public class LayerStops {
+        public boolean propertyStops = false;
+        public boolean zoomStops = false;
+        public boolean zoomPropertyStops = false;
+        public boolean hasStops = false;
+
+        public LayerStops() {
+            if (zoomPropertyStops || zoomStops || propertyStops) {
+                this.hasStops = true;
+            }
```
tbarsballe: Since all three of `zoomPropertyStops || zoomStops || propertyStops` are initialized to false, this code will never get executed; you can't set them any other way before calling this constructor.

Consequently, there is no reason to even define a constructor, the default empty constructor is functionally identical.

modules/unsupported/mbstyle/src/test/java/org/geotools/mbstyle/transform/StyleTransformFunctionTest.java
```diff
@@ -16,15 +16,15 @@
...
-
 import org.geotools.mbstyle.MBStyle;
 import org.geotools.mbstyle.MapboxTestUtils;
 import org.geotools.styling.StyledLayerDescriptor;
 import org.json.simple.JSONObject;
 import org.json.simple.parser.ParseException;
 import org.junit.Test;
 
+import java.io.IOException;
+
```
tbarsballe: Umm, why was this file included? There isn't anything functionally different here?

vickdw: Only explanation is the difference in IDE's used, IntelliJ vs Eclipse.  looks like one of them rearranged imports and since git reported it as being modified, I included it.

PR #1667, opened 2017-08-01
[GEOT-5795] remove services file org.geotools.xml.schema.Schema from wfs-ng

mprins:
because this lists a non-existent class, also remove commented code that used this class.

see: https://osgeo-org.atlassian.net/browse/GEOT-5795

### DIFF
--- a/modules/extension/wms/src/main/java/org/geotools/data/wms/xml/ogcComplexTypes.java
+++ b/modules/extension/wms/src/main/java/org/geotools/data/wms/xml/ogcComplexTypes.java
@@ -246,13 +246,6 @@ public class ogcComplexTypes {
             
             if (canEncode(element, value, hints)) {
                 AttributesImpl attributes = new AttributesImpl();
-//                attributes.addAttribute(WFSSchema.NAMESPACE.toString(),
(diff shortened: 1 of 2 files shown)


tbarsballe:
It looks like this class was included in the old gt-wfs module, was removed in [pull#1479](https://github.com/geotools/geotools/pull/1479/files#diff-b946f3716d79c88d4332b8560a17879a), and did not get ported over to the gt-wfs-ng module.

I think the correct thing to do here is to port that WFSSchema class over to gt-wfs-ng (I expect a bunch of support classes, such as WFSSchemaFactory will also need to be ported). However, I am not especially familiar with this module - @NielsCharlier or @jodygarnett may know more.

mprins:
not sure about the port, cursory scan of the removed WFSSchema.java file give me the impression that is is basically a utility class that wraps up a java version of the WFS 1.0.0 schema. 
Since all that referenced code has also been removed and WFS-NG works just fine I tend to disagree on porting this and everything it references into WFS-NG.

NielsCharlier:
I agree with @mprins . I think it can just be removed. It isn't used any longer.

tbarsballe:
Ok, thanks for the info @NielsCharlier. Merging.

PR #1668, opened 2017-08-01
[GEOT-5795] remove services file org.geotools.xml.schema.Schema from wfs-ng (17.x)

mprins:
because this lists a non-existent class, also remove commented code that used this class.

see: 
  - https://osgeo-org.atlassian.net/browse/GEOT-5795
  -  #1667


jodygarnett:
So what happened to this class, is it a typo or was it removed?

mprins:
This class was part of the old wfs module which was removed

PR #1670, opened 2017-08-03
[GEOT-5289] Make FactoryRegistry Java 9 compatible

nicolaiparlog:
Fixes [GEOT-5289](https://osgeo-org.atlassian.net/browse/GEOT-5289) by extending `FactoryRegistry` to implement the `ServiceRegistry` API. Work similar to this has been done in #1064 but was not merged due to legal concerns. This reimplementation is solely based on the calls GeoTools classes make to the `ServiceRegistry` API - no JDK code was copied.

This PR is by no means complete. The main classes (`FactoryRegistry`, `CategoryRegistry`, and `PartiallyOrderedSet`) are littered with TODOs and I made some design decisions that need to be revisited. But GeoTools builds (on Java 8) and so does our project using it (compiling/testing on Java 9), so it fulfills the minimum requirements.

I decided not to add the open issues as line comments because it is harder to keep track of them, particularly when they get folded due to new commits. Instead I put all my concerns here - sorry for the wall of text.

## Call Hierarchy Driven Design

I only implemented those `ServiceRegistry` methods that are called in the GeoTools code base as well as some overloads that seemed to make sense (mostly to make registration and deregistration symmetrical).

This could break clients that call `ServiceRegistry` methods that GeoTools does not call. Someone might have to investigate this.

## Exceptions

The new code in `FactoryRegistry` can throw exception. At least one of the existing methods does some fancy exception creation. Maybe the new ones should do that, too?

## Registrable Services

The service registry calls providers when they are (de)registered if they implement the [`RegisterableService` interface](https://docs.oracle.com/javase/8/docs/api/javax/imageio/spi/RegisterableService.html). The methods expect a `ServiceRegistry` as parameter, though, and the `FactoryRegistry` no longer fufills that requirement.

My temporary solution was to create a `RegistrableFactory` interface and have registrable GeoTools classes implement it.

This change is in so far backwards incompatible that user implementations of `RegisterableService` will silently stop getting called. A weird way to address this could be to throw an exception when an implementation of `RegisterableService` is encountered.

## Naming

Since this change is not 100% backwards compatible, it might make sense to go all the way and rename the methods inherited from `ServiceRegistry`. Since this class seems to be concerned with factories, I would replace `ServiceProvider` with `Factory` in all new methods. More explicit breakage might actually improve error-free migration because it highlights the changes and gives a good opportunity to inform the caller of the change to registrable services/factories.

Backwards-compatible, deprecated overloads could be provided.

## Replace `Filter` With `Predicate`

Since we're breaking code anyway (maybe), why not replace `Filter` With `Predicate`? Would make integration with Java 8 more seamless.

Backwards-compatible, deprecated overloads could be provided.

## Partially Ordered Set

I decided to reuse the existing `PartiallyOrderedSet`. To get it to where I can use it, I simply copied it into the _gt-main_ project. It is entirely possible that a better place exists.

The implementation also had at least two bugs and while fixing them I ended up changing the code quite a lot. I'm happy to provide information on why exactly I picked the current design.


modules/library/metadata/src/main/java/org/geotools/factory/FactoryRegistry.java
```diff
@@ -152,15 +158,52 @@ public FactoryRegistry(final Class<?>[] categories) {
...
+    public <T> T getServiceProviderByClass(Class<T> providerClass) {
+        if (providerClass == null) {
+            // TODO: do something fancy like in getServiceProvider(Class, Filter, Hints, Hints.Key) ?
+            throw new IllegalArgumentException("The providerClass must not be null.");
+        }
+        return registry.getInstanceOfType(providerClass).orElse(null);
+    }
+
+    // TODO: document
+    public <T> Iterator<T> getServiceProviders(final Class<T> category, final boolean useOrdering) {
```
ansell: Javadoc the useOrdering parameter with its workflow particularly. Ditto for the filter version below also.

modules/library/metadata/src/main/java/org/geotools/factory/CategoryRegistry.java
```diff
@@ -0,0 +1,200 @@
...
+	private final FactoryRegistry factoryRegistry;
+	private final Map<Class<?>, InstanceRegistry<?>> categories = new HashMap<>();
+
+	public CategoryRegistry(FactoryRegistry factoryRegistry) {
+		this.factoryRegistry = factoryRegistry;
+	}
+
+	public void registerCategory(final Class<?> category) {
+		// TODO: should anything happen if the category is already registered?
+		categories.put(category, new InstanceRegistry<>(factoryRegistry, category));
```
nicolaiparlog: So registering the same category multiple times should not lead to an error, right?

modules/library/metadata/src/main/java/org/geotools/factory/FactoryRegistry.java
```diff
@@ -152,15 +158,52 @@ public FactoryRegistry(final Class<?>[] categories) {
...
+    /*
+     * TODO: discuss how much API can be broken
+     *
+     * Breakage:
+     *
+     *  - not all methods of `ServiceRegistry` API were implemented
+     *
+     * Possible breakage:
+     *
+     *  - in general nomenclature, replace "ServiceProvider" with "Factory"
```
nicolaiparlog: > If the naming is going to change, it would be best done at the same time as this, not in steps over time.

I think so, too. So should I change the names now? I could put in deprecated, forwarding variants with the old name.

modules/library/metadata/src/main/java/org/geotools/factory/FactoryRegistry.java
```diff
@@ -152,15 +158,52 @@ public FactoryRegistry(final Class<?>[] categories) {
...
+     *  - RegisterableService interface can no longer be used because it expects a `ServiceProvider`
+     *  - replace `Filter` with `Predicate<? super T>`
```
nicolaiparlog: I don't see a way to get rid of the generics by subtyping. There is no sensible `T` from which all factories extend (or did I miss it?). The `? super T` _could_ be replaced by `T` but that would make using the API more painful.

Should I do this together with renaming the methods? Then the new methods with their shiny new names take predicates and the old ones accept filter?

modules/library/metadata/src/main/java/org/geotools/factory/FactoryRegistry.java
```diff
@@ -152,15 +158,52 @@ public FactoryRegistry(final Class<?>[] categories) {
...
+    public Stream<Class<?>> streamCategories() {
+        return registry.streamCategories();
+    }
+
+    // TODO: ensure uniform null handling here and in CategoryRegistry
+
+    // TODO: document
+    public <T> T getServiceProviderByClass(Class<T> providerClass) {
+        if (providerClass == null) {
+            // TODO: do something fancy like in getServiceProvider(Class, Filter, Hints, Hints.Key) ?
```
nicolaiparlog: I meant something like [in `registerFromSystemProperty`](https://github.com/nicolaiparlog/geotools/blob/aefb899a7e092f6cd749e58bfa768d449a07d7dc/modules/library/metadata/src/main/java/org/geotools/factory/FactoryRegistry.java#L990):

```
throw new FactoryRegistryException(Errors.format(
ErrorKeys.CANT_CREATE_FACTORY_$1, classname), exception);
```

I thought that may be GeoTools' preferred way of creating error messages and wondered whether I should do something similar.

nicolaiparlog:
Thanks for the feedback, will work on it next week.

modules/library/metadata/src/main/java/org/geotools/factory/CategoryRegistry.java
ansell: That detail would need to be clarified by someone who is familiar with the current behaviour, and then added as a Javadoc to this method.

``Map.put`` will overwrite any existing ``InstanceRegistry``, and return the previous value as a result, but the result is being ignored right now. ``computeIfAbsent`` attempts to avoid that behaviour by checking if the key is there before the put occurs.

If you want thread-safe behaviour you would need to switch to a ``ConcurrentMap``, as although ``Map`` supports ``computeIfAbsent``, it doesn't have a thread-safe contract.

modules/library/metadata/src/main/java/org/geotools/factory/FactoryRegistry.java
ansell: The strategy for that may require some more discussion, but deprecations with the functionality still available, particularly with the goal of fitting with the naming conventions used by the rest of the GeoTools API wouldn't worry me.

modules/library/metadata/src/main/java/org/geotools/factory/FactoryRegistry.java
ansell: This may lead down a rabbit hole... Maybe best to avoid it for now given that ``Filter`` is still public and probably will be for a long time.

modules/library/metadata/src/main/java/org/geotools/factory/FactoryRegistry.java
ansell: Ahh, thats a good idea I think.

modules/library/metadata/src/main/java/org/geotools/factory/CategoryRegistry.java
nicolaiparlog: #### Multiple Registration

I think the original code just `put` the specified provider, i.e. it overwrote the previous one without comment. I'll check which way it was, replicate the behavior and document in JavaDoc.

#### Thread Safety

`FactoryRegistry` Javadoc says:

> NOTE: This class is not thread safe

I think we'd need to do more to guarantee thread-safety than just replacing `Map` with `ConcurrentMap` - I think that should be a separate issue/PR.

nicolaiparlog: Yep, original was `put`, so I'll just do the same and leave a comment.

modules/library/metadata/src/main/java/org/geotools/factory/CategoryRegistry.java
```diff
@@ -0,0 +1,200 @@
...
+		InstanceRegistry<T> registry = (InstanceRegistry<T>) categories.get(category);
+		if (registry == null) {
+			// TODO: do something fancy like in FactoryRegistry#getServiceProvider(Class, Predicate, Hints, Hints.Key) ?
+			throw new IllegalArgumentException("The category '" + category + "' is not registered");
+		}
+		return registry;
+	}
+
+	public Stream<Class<?>> streamCategories() {
+		return categories.keySet().stream();
```
nicolaiparlog: Good point, but there is no code registering/deregistering categories after construction of `FactoryRegistry`. After realizing that I decided to put that into code, so now the `Map` is an `ImmutableMap` and created during construction of `CategoryRegistry`.

modules/library/metadata/src/main/java/org/geotools/factory/FactoryRegistry.java
```diff
@@ -102,7 +109,7 @@
      * On initialization, all categories need to be scanned for plugins. After a category
      * has been first used, it is removed from this set so we don't scan for plugins again.
      */
-    private Set<Class<?>> needScanForPlugins;
+    private final Set<Class<?>> needScanForPlugins = new HashSet<>();
```
nicolaiparlog: Good catch! :+1:

modules/library/metadata/src/main/java/org/geotools/factory/FactoryRegistry.java
nicolaiparlog: I thought about this and decided to go with `Predicate`. True, `Filter` is still public but I find it weird to depend on some arbitrary interface instead of one that was intended for it. It also simplified code a little because the `filter` (nor `test`) method no longer has to handle `Object` but can be written to only deal with the intended factory type.

modules/library/metadata/src/main/java/org/geotools/factory/CategoryRegistry.java
```diff
@@ -0,0 +1,200 @@
...
+			Iterator<T> values = instancesByType.values().iterator();
+			while (values.hasNext()) {
+				T instance = values.next();
+				values.remove();
+				instances.remove(instance);
+				notifyDeregistered(instance);
+			}
+		}
+
+		public Iterator<T> iterate(final boolean useOrder) {
```
nicolaiparlog: I documented that on the method `CategoryRegistry` exposes.

nicolaiparlog:
I fixed many TODOs and included much of @ansell's feedback. I introduced methods with `Factory` instead of `ServiceProvider` in their name - they use `Predicate` instead of `Filter`. I also started having the new methods return `Stream` and `Optional` but that proved to be a non-trivial endeavor (and I ran out of time), so I didn't want to follow through with it until I get a thumbs-up that that's a desired change.

I'm a little short on time - is there any chance somebody else could jump in and document the rest of the methods from the `ServiceRegistry API` as to the GeoTool team's preference.

nicolaiparlog:
@geotools Any progress on this? Anything I need to do to get this merged?

(We kinda need this sooner rather than later because it blocks our Java 9 migration, but I'm hesitant to build a locally patched version.)

jodygarnett:
Thanks we will explore this after the code freeze, really appreciate your taking on this topic.

modules/library/main/src/main/java/org/geotools/filter/expression/PropertyAccessors.java
```diff
@@ -44,18 +45,15 @@
...
-            cache.add((PropertyAccessorFactory) factory);
-         }
-         FACTORY_CACHE = (PropertyAccessorFactory[]) cache.toArray(new PropertyAccessorFactory[cache.size()]);
+        cache.add( new DirectPropertyAccessorFactory());
+        for (PropertyAccessorFactory factory : ServiceLoader.load(PropertyAccessorFactory.class)) {
+            boolean nonCached = factory instanceof SimpleFeaturePropertyAccessorFactory
+                    || factory instanceof DirectPropertyAccessorFactory
+                    || factory instanceof NullPropertyAccessorFactory;
+            if (!nonCached)
+                cache.add(factory);
```
jodygarnett: Thanks that is a clear way to describe the cache condition. We tend to use braces even for single line if statements

modules/library/metadata/pom.xml
```diff
@@ -130,6 +130,10 @@
       <scope>provided</scope>  <!-- Not required for default execution. -->
       <optional>true</optional>
     </dependency>
+    <dependency>
+        <groupId>com.google.guava</groupId>
+        <artifactId>guava</artifactId>
+    </dependency>
```
jodygarnett: I am a bit uncomfortable with the use of guava as it continues to produces incompatibilities with downstream applications. Most of the benefits we were seeking from guava are now available as part of Java.

What specific features of Guava are you looking to leverage here?

modules/library/metadata/src/main/java/org/geotools/factory/RegistrableFactory.java
```diff
@@ -0,0 +1,10 @@
+package org.geotools.factory;
```
jodygarnett: header :) New file so I gotta ask

jodygarnett:
@nicolaiparlog can I ask if you (or your employer) have signed an Open Source Geospatial Foundation (OSGeo) contributor license agreement?

It looks like you are a new contributor to GeoTools (welcome!) GeoTools is managed by OSGeo and we like to make sure that the foundation has permission to distribute contributions under an open source license at the direction of the project management committee. Occasionally we get requests to donate code to other projects (most recently JTS Topology Suite) and we can only do this when everyone has filled in a CLA.

For more information, and links, see https://github.com/geotools/geotools/blob/master/CONTRIBUTING.md

The actual CLA can be handled electronically, download as a PDF, sign with a paint program, and email to secretary@osgeo.org.

modules/library/metadata/pom.xml
aaime: Agreed, having Guava as a core dependency is likely going to be a blocker. We have core devs trying to promote a "un-guava-ifycation" of our code base (not as easy as it seems, but I agree we should not have it everywhere).

modules/library/metadata/src/main/java/org/geotools/factory/AbstractFactory.java
```diff
@@ -121,7 +119,7 @@
  * @version $Id$
  * @author Martin Desruisseaux
  */
-public class AbstractFactory implements Factory, RegisterableService {
+public class AbstractFactory implements Factory, RegistrableFactory {
```
jodygarnett: I wonder if we can repurpose our Factory interface and avoid multiple inheritance here?

modules/library/metadata/src/main/java/org/geotools/factory/CategoryRegistry.java
```diff
@@ -0,0 +1,301 @@
...
+
+import static java.util.Objects.requireNonNull;
+import static org.geotools.util.Utilities.stream;
+
+/**
+ * The category registry holds multiple instances per category. Categories are
+ * {@link Class classes} and instances are also accessible by the class they implement.
+ * Note that instances have to implement/extend the category they are filed under.
+ */
+class CategoryRegistry {
```
jodygarnett: Sorry misread this, my initial comment was incorrect.

For class javadocs this low level in the library we try and communicate Class, Responsibility, Collaboration information. This class is package visible rather than public so we do not expect direct access

modules/library/metadata/src/main/java/org/geotools/factory/CategoryRegistry.java
```diff
@@ -0,0 +1,301 @@
...
+	 *
+	 * @param instance The instance to register.
+	 */
+	@SuppressWarnings("unchecked")
+	public <T> void registerInstance(final T instance) {
+		requireNonNull(instance);
+		streamCategories()
+				.filter(category -> category.isAssignableFrom(instance.getClass()))
+				// the cast is correct because the filter above only leaves categories that are supertypes of `instance`
+				.map(category -> (InstanceRegistry<? super T>) instanceRegistry(category))
```
jodygarnett: You may consider passing in Class<T> type as a parameter to avoid this cast

modules/library/metadata/src/main/java/org/geotools/factory/FactoryCreator.java
```diff
@@ -21,6 +21,8 @@
 import java.lang.ref.WeakReference;
 import java.lang.reflect.Modifier;
 import java.lang.reflect.InvocationTargetException;
+import java.util.function.Predicate;
+import javax.imageio.spi.ServiceRegistry.Filter;
```
jodygarnett: We may need to make our own copy of this interface, so that this infrastructure can be used on systems where imageio is not available.\

(this is a blocker for people trying to use GeoTools on Andriod for example)

modules/library/metadata/src/main/java/org/geotools/factory/FactoryCreator.java
```diff
@@ -134,15 +136,27 @@ public FactoryCreator(final Collection<Class<?>> categories) {
      * @throws FactoryRegistryException if the factory can't be created for some other reason.
      */
     @Override
+    @Deprecated // TODO document
```
jodygarnett: I guess you are deprecating this to move towards Predicate use rather than javax.imageio.spi.ServiceRegistry.Filter.

I would not mind checking with the GeoTools PMC about making this a breaking change (forcing the use of getFactory( category, filter, hints, key).

modules/library/metadata/src/main/java/org/geotools/util/Utilities.java
```diff
@@ -670,4 +672,16 @@ public static void ensureNonNull(final String name, final Object object)
         }
     }
 
+    // TODO: document
+    public static <T> Stream<T> stream(Optional<T> optional) {
```
jodygarnett: we will of course need javadocs for public methods

modules/library/metadata/src/main/java/org/geotools/util/PartiallyOrderedSet.java
```diff
@@ -14,10 +14,9 @@
  *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
  *    Lesser General Public License for more details.
  */
-package org.geotools;
+package org.geotools.util;
```
jodygarnett: So what is the reason to change the package here? That may (or may not) break down stream code...

jodygarnett:
To answer one of the questions raised in your introduction. I am okay remaining from ServiceRegistry to FactoryRegistry - the only hold over is how Java refers to these things (making us use the ServiceLocator). 

I agree for a GeoTools focused solution being really clear that we are managing factories helps with readability.

ansell:
@jodygarnett I am not sure what your overall comment is referring to when you say:

> "I am okay remaining from ServiceRegistry to FactoryRegistry - the only hold over is how Java refers to these things (making us use the ServiceLocator)."

The confusing part of your comment is "I am okay remaining from ServiceRegistry to FactoryRegistry", which doesn't seem to grasp the overall reason for the PR, which is to completely remove the ``ServiceFactory`` inheritance from ``FactoryRegistry``.

To clarify, this PR is about making it possible to use GeoTools in Java-9. Starting from Java-9 (implemented since Java-9 build 80 http://bugs.java.com/view_bug.do?bug_id=JDK-8068749 ), there is no way for a user to extend ``javax.imageio.spi.ServiceRegistry`` for non imageio service classes, which has broken the entire GeoTools dynamic loading model as it was completely based on extending the ``ServiceRegistry`` class for use with GeoTools service classes. The alternative is ``java.util.ServiceLoader`` which has been available and recommended since Java-6. ``java.util.ServiceLoader`` isn't based on inheritance, as it was created as a final class with static methods for calling into it.

jodygarnett:
Checking [online](http://www.oracle.com/technetwork/java/eol-135779.html) we have until September 2018 (at least) until Java 8 is EOL.

modules/library/main/src/main/java/org/geotools/filter/expression/PropertyAccessors.java
nicolaiparlog: Done.

modules/library/metadata/pom.xml
nicolaiparlog: I used it to filter an Iterator and for an `ImmutableMap`. Removed that and thus the dependency in aa015e8.

modules/library/metadata/src/main/java/org/geotools/factory/RegistrableFactory.java
nicolaiparlog: Done in e0c8e75. I simply copied some other header - I hope that's ok.

modules/library/metadata/src/main/java/org/geotools/factory/AbstractFactory.java
nicolaiparlog: One thing should only extend the other if one thing _is_ the other. I can not decide whether that's the case here. Regarding implementation details, having one interface extend the other had these consequences:

* If `RegistrableFactory extends Factory` GeoTools users that want to be called back on registration would have to implement `getImplementationHints`.
* If `Factory extends RegistrableFactory` (leaving aside that then the more specific type would have a less specific name), lots of GeoTools classes would have to implement the registration functions even though they don't need them.
* If the interfaces were to be merged (e.g. into `Factory`), we had both downsides.

All situations could be improved by providing default implementations for some or all of these methods.

Your call.

modules/library/metadata/src/main/java/org/geotools/factory/CategoryRegistry.java
```diff
@@ -0,0 +1,301 @@
+package org.geotools.factory;
```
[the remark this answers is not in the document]
nicolaiparlog: Done in e0c8e75.

modules/library/metadata/src/main/java/org/geotools/factory/CategoryRegistry.java
nicolaiparlog: I didn't read the initial comment - does that mean the Javadoc needs improvement is it fine?

modules/library/metadata/src/main/java/org/geotools/factory/CategoryRegistry.java
nicolaiparlog: Do you mean have a second parameter `Class<T>`? I don't see how that would spare me the cast.

Also, I don't see how we could pass that parameter. The calling method `FactoryRegistry::registerFactory` only has the instance as an `Object`. Even if we generified that method...

    public <T> void registerFactory(final T factory)

... it wouldn't work because `factory.getClass()` returns `Class<? extends Object>`.

modules/library/metadata/src/main/java/org/geotools/factory/FactoryCreator.java
nicolaiparlog: Seems like a good time to do that. But what exactly do you mean by "our own copy"? Does that include the namespace/package or just the interface's functionality?

modules/library/metadata/src/main/java/org/geotools/factory/FactoryCreator.java
nicolaiparlog: Sure, let's discuss that at the call.

modules/library/metadata/src/main/java/org/geotools/util/PartiallyOrderedSet.java
nicolaiparlog: I actually moved the class from `library/xml` to `library/metadata`. The `util´ just seemed the most fitting in Metadata. Note that there is currently no class in `org.geotools` in Metadata.

nicolaiparlog:
> Checking online we have until September 2018 (at least) until Java 8 is EOL.

While that is true, I hope it doesn't mean "we have until September 2018 before we need to be Java 9 compatible". :smiley:

modules/library/metadata/src/main/java/org/geotools/util/Utilities.java
nicolaiparlog: I just documented the utility methods, but would really appreciate if someone else took a shot at the `FactoryRegistry` methods. I would mostly be guessing about the details...

nicolaiparlog:
I finally found time to create [a proposal for this change](https://github.com/geotools/geotools/wiki/FactoryRegistry-Refactoring-for-Java-9-Compatibility) - I did not yet add it to [the list of proposals](https://github.com/geotools/geotools/wiki/Proposals) because I don't know which version to put it under (18.x I assume, but there is no such entry yet - were there no proposals for 17.x?).

Will inform the mailing list later today.

don-vip:
So Java 9 has been released today as expected, and GeoTools is not yet ready.
Now that Nicolai has undertaken all the coding and administrative work, could you please tell us when you're planning to release a compatible version?
Also now that Java is speeding up with a release every 6 months, I strongly advise you to join the [OpenJDK Quality Outreach](https://wiki.openjdk.java.net/display/quality/Quality+Outreach) and setup continuous integration with latest Java builds if you want to keep pace.

nicolaiparlog:
> Now that Nicolai has undertaken all the coding and administrative work, could you please tell us when you're planning to release a compatible version?

Apparently, GeoTools evolves a little more deliberate than many other projects on GitHub. Besides a proposal, there also needs to be a discussion on the mailing list, [which I initiated just now](https://sourceforge.net/p/geotools/mailman/message/36049360/). It might take a few more weeks before it is clear which version will be the first to run on Java 9.

> Also now that Java is speeding up with a release every 6 months, I strongly advise you to join the [OpenJDK Quality Outreach](https://wiki.openjdk.java.net/display/quality/Quality+Outreach) and setup continuous integration with latest Java builds if you want to keep pace.

Yes, :+1:!

nicolaiparlog:
Had some time to patch the PR and I think it's getting ready. All that's missing is the documentation and **I really hope someone can help me out**! Remember, I don't actually know what I'm doing here, so I have no clue what exactly these methods do and what their contract is.

ineedyou

If anybody's interested to contribute, just mention it here and I will invite you to my fork, so you can push to this branch.

modules/library/metadata/src/main/java/org/geotools/util/Utilities.java
```diff
@@ -670,4 +678,100 @@ public static void ensureNonNull(final String name, final Object object)
...
+     * @return a stream over the iterable's elements
+     */
+    public static <T> Stream<T> stream(Iterable<T> iterable) {
+        return StreamSupport.stream(iterable.spliterator(), false);
+    }
+
+    /**
+     * Creates a stream over the specified {@link Iterator}'s elements.
+     *
+     * This operation is lazy in the sense that it does not iterate over the elements.
```
ansell: Change this to "does not iterate over the elements until the Stream is consumed"

modules/unsupported/efeature/src/main/java/org/geotools/data/efeature/EFeatureFactoryFinder.java
```diff
@@ -80,19 +80,15 @@ public static EFeatureContextFactory getContextFactory()
             // Do the lousy slow system scan
             //
             synchronized (EFeatureFactoryFinder.class) {
```
ansell: Repeating the if condition inside of the synchronized block would minimize the impact of the slow scan, particularly if the scan is so slow that other threads get stuck here while an iteration is executed.

modules/library/metadata/src/main/java/org/geotools/factory/CategoryRegistry.java
```diff
@@ -0,0 +1,319 @@
...
+	public CategoryRegistry(final FactoryRegistry factoryRegistry, final Iterable<Class<?>> categories) {
+		ensureArgumentNonNull("factoryRegistry", factoryRegistry);
+		ensureArgumentNonNull("categories", categories);
+		// use an unmodifiable map to guarantee immutability
+		this.categories = stream(categories)
+				.collect(collectingAndThen(
+						toMap(
+								category -> category,
+								category -> new InstanceRegistry<>(factoryRegistry, category),
+								(firstCategory, secondCategories) -> secondCategories),
```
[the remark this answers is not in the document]
nicolaiparlog: `Collectors::toMap` enforces explicit handling of duplicate keys. The two-parameter variant throws an `IllegalStateException` if the same key is encountered twice. The variant I call here requires me to define what happens with the values of duplicate keys. Since both registries are empty, it does not matter which one I choose. (My naming is off: It should be `firstRegistry`/`secondRegistry`, not `firstCategory`/`secondCategory`.)

modules/library/metadata/src/main/java/org/geotools/factory/CategoryRegistry.java
```diff
@@ -0,0 +1,319 @@
...
+		ensureArgumentNonNull("secondInstance", secondInstance);
+		return instanceRegistry(category).clearOrder(firstInstance, secondInstance);
+	}
+
+	private static class InstanceRegistry<T> {
+
+		private final FactoryRegistry factoryRegistry;
+		private final Class<?> category;
+
+		private final Map<Class<?>, T> instancesByType = new HashMap<>();
```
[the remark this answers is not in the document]
nicolaiparlog: Yes, this class is not thread-safe (because `FactoryRegistry` isn't either). I will add that to the class Javadoc.

modules/library/metadata/src/main/java/org/geotools/factory/CategoryRegistry.java
```diff
@@ -0,0 +1,319 @@
...
+		private final PartiallyOrderedSet<T> instances = new PartiallyOrderedSet<>(false);
+
+		private InstanceRegistry(FactoryRegistry factoryRegistry, Class<?> category) {
+			this.factoryRegistry = factoryRegistry;
+			this.category = category;
+		}
+
+		public boolean register(final T instance) {
```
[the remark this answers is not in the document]
nicolaiparlog: This code is internal and hence not documented. The public `CategoryRegistry::registerInstance` is, though, and it describes the contract. I'd rather not duplicate that information here. Do you think it is necessary?

modules/library/metadata/src/main/java/org/geotools/factory/CategoryRegistry.java
```diff
@@ -0,0 +1,319 @@
...
+				((RegistrableFactory) instance).onRegistration(factoryRegistry, category);
+			}
+		}
+
+		private void registerInternal(final T instance) {
+			instancesByType.put(instance.getClass(), instance);
+			instances.add(instance);
+		}
+
+		public boolean deregister(final T instance) {
```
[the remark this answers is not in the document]
nicolaiparlog: This code is internal and hence not documented. The public `CategoryRegistry::deregisterInstance` is, though, and it describes the contract. I'd rather not duplicate that information here. Do you think it is necessary?

modules/library/metadata/src/main/java/org/geotools/factory/CategoryRegistry.java
```diff
@@ -0,0 +1,319 @@
...
+				deregisterByType(instance);
+				return true;
+			} else {
+				return false;
+			}
+		}
+
+		// documentation hint: returns true if an instance of the same type was previously registered
+		private boolean deregisterByType(final T instance) {
+			T removed = instancesByType.remove(instance.getClass());
```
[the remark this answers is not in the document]
nicolaiparlog: The current code is perfectly safe, because `PartiallyOrderedSet::remove` and `notifyDeregistered` handle `null` well. But the `null` check could guard against future implementation errors, so it does make some sense. Do you want me to add it?

modules/library/metadata/src/main/java/org/geotools/factory/CategoryRegistry.java
```diff
@@ -0,0 +1,319 @@
...
+		}
+
+		public <S> Optional<S> getInstanceOfType(Class<S> type) {
+			ensureArgumentNonNull("type", type);
+			@SuppressWarnings("unchecked")
+			S instance = (S) instancesByType.get(type);
+			return Optional.ofNullable(instance);
+		}
+
+		public boolean setOrder(T firstInstance, T secondInstance) {
```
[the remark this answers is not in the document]
nicolaiparlog: See `CategoryRegistry::setOrder` - should it be repeated here?

modules/library/metadata/src/main/java/org/geotools/factory/CategoryRegistry.java
```diff
@@ -0,0 +1,319 @@
...
+			return instancesByType.containsKey(firstInstance.getClass())
+					&& instancesByType.containsKey(secondInstance.getClass())
+					// if both are contained, set the order
+					&& instances.setOrder(firstInstance, secondInstance);
+		}
+
+		public boolean clearOrder(T firstInstance, T secondInstance) {
```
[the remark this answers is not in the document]
nicolaiparlog: See `CategoryRegistry::clearOrder` - should it be repeated here?

modules/library/metadata/src/main/java/org/geotools/factory/FactoryRegistry.java
```diff
@@ -797,17 +837,55 @@ public void scanForPlugins() {
...
+        registry.registerInstance(factory);
+    }
+
+    // TODO: document
+    @Deprecated
+    public <T> boolean registerServiceProvider(final T provider, final Class<T> category) {
+        return registerFactory(provider, category);
+    }
+
+    public <T> boolean registerFactory(final T factory, final Class<T> category) {
```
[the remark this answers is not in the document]
nicolaiparlog: I struggled a little with where exactly to put the `null` checks:

1. `CategoryRegistry` has its own API and should be safe ~> check `null` there
2. checking in `FactoryRegistry` gives clearer stack traces ~> check `null` there
3. don't go overboard ~> check only once for each code path

We can't have all three, so I decided to pick 1. and 3. That means all arguments for `FactoryRegistry` methods are eventually `null` checked (or at least should be), but for many that happens in `CategoryRegistry`.

We can change this if you want and move those checks into `FactoryRegistry`, but then I need advice whether they should be duplicated or removed from `CategoryRegistry`.

modules/unsupported/efeature/src/main/java/org/geotools/data/efeature/EFeatureFactoryFinder.java
nicolaiparlog: I'd rather not have my fingerprints on code that does concurrency _and_ interacts with GC if I didn't make an effort to properly understand it. It's also not related to what this PR is about.

modules/library/metadata/src/main/java/org/geotools/factory/CategoryRegistry.java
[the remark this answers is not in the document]
nicolaiparlog: To make the method itself easier to grasp, I added a line of Javadoc describing the return value.

modules/library/metadata/src/main/java/org/geotools/factory/CategoryRegistry.java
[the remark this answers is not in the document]
nicolaiparlog: To make the method itself easier to grasp, I added a line of Javadoc describing the return value.

modules/library/metadata/src/main/java/org/geotools/factory/CategoryRegistry.java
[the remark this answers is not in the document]
nicolaiparlog: The API is not duplicated. The inner class operates on one specific category, which simplifies the requirements and makes the code easier to read (I hope). The outer class handles multiple categories and provides method overloads.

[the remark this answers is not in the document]
nicolaiparlog: To make navigation easier, I added Javadoc `@see` tags.

nicolaiparlog:
Forgot to mention that last time: I updated many GeoTools-internal calls to use the new `Predicate`-eating and `Stream`-spewing methods, but not all of them. Many needed an `Iterator` and returned an `Iterator` and I felt it would be cheating to update them to accept a `Stream` and return an `Iterator`. I'm sure many of those could be updated to instead return a `Stream` as well, but didn't want to go down that road of updating all of them, too.

Instead I left the calls in - deprecation warnings should help identify them.

nicolaiparlog:
As I see it, there are three open issues. I need some guidance on how to resolve them.

## Documentation

There's still no JavaDoc but if [Ansell's comment on documentation](https://github.com/geotools/geotools/pull/1670#pullrequestreview-70703660) stands, that's not a problem.

## Compatibility vs `RegisterableService`

The current changes are fully compatible at compile time and should even be for bytecode. At runtime, though, they cause a slightly different behavior because implementations of [`RegisterableService`](https://docs.oracle.com/javase/9/docs/api/javax/imageio/spi/RegisterableService.html) won't be called when they get registered. At the moment, this is silent.

Alternatively, a warning could be logged or even an exception thrown if the service implements `RegisterableService`, but not `RegistrableFactory` (our replacement). This could be done with `Class.forName("javax.imageio.spi.RegisterableService").isInstance(...)` (plus checks) to prevent a new compile-time dependency.

## Compatibility vs `javax.imageio.spi.ServiceRegistry.Filter`

The compatibility described above is achieved by keeping the "old" methods, which accept `Filter`, around. If the class is removed, the overloads need to be removed as well, which means callers of FactoryRegistry will most likely see compile errors.

mipastgt:
This should be merged into the latest snapshot as soon as possible so that we can find out whether there are even more hidden problems with JavaFX.

jodygarnett:
Pinged the list to accept the proposal, thanks for detailed review @aaime.

@nicolaiparlog can I take you up on the opportunity to contribute to your branch please write up javadocs etc... my remaining feedback is pretty stock, use of tabs rather than spaces etc...

modules/library/coverage/src/main/java/org/geotools/coverage/grid/io/GridFormatFinder.java
```diff
@@ -83,16 +85,10 @@ private GridFormatFinder() {
...
-			final GridFormatFactorySpi spi = it.next();
-			if (spi.isAvailable())
-				formats.add(spi);
-
-		}
-		return Collections.unmodifiableSet(formats);
+		return getServiceRegistry()
+				.getFactories(GridFormatFactorySpi.class, true)
+				.filter(GridFormatFactorySpi::isAvailable)
+				.collect(toUnmodifiableSet());
```
jodygarnett: Bleck old code with tabs rather than spaces :(

jodygarnett:
I have commented out deprecated code, but not deleted yet.

The email discussion referenced replacing GeoServer code:

      Iterator i = FactoryRegistry.lookupProviders(extensionPoint);

But I did not see lookupProvider being removed yet.

jodygarnett:
Okay based on email list feedback the deprecated methods are restored, also had a chance to go over java docs for references to service providers etc...

don-vip:
Are you able to run all tests? When I run `mvn clean install` (with Java 8) I have test failures in "Dynamic symbolizers for SVG symbols" component (apparently caused by an NPE at org.apache.batik.svggen.ImageHandlerBase64Encoder.encodeImage)

bencaradocdavies:
@don-vip in what module? I just ran a full local build with `-Dall` and it passed.

bencaradocdavies:
@don-vip I have restarted the failed Travis CI build: https://travis-ci.org/geotools/geotools/builds/301096616

That failure looks familiar and it may be one of those nondeterministic failures that we see from time time.

bencaradocdavies:
Here is the previous failure in `JDBCTransactionStateTest.testSetTransactionNullWithInternalConnection` but it looks different:
https://osgeo-org.atlassian.net/browse/GEOT-5603

bencaradocdavies:
Travis CI failed again:
```
Tests run: 2, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.094 sec <<< FAILURE! - in org.geotools.jdbc.JDBCTransactionStateTest
testSetTransactionNullWithInternalConnection(org.geotools.jdbc.JDBCTransactionStateTest)  Time elapsed: 0.04 sec  <<< FAILURE!
java.lang.AssertionError: Regular close, no warnings expected. expected:<0> but was:<1>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:743)
	at org.junit.Assert.assertEquals(Assert.java:118)
	at org.junit.Assert.assertEquals(Assert.java:555)
	at org.geotools.jdbc.JDBCTransactionStateTest.testLogWarnings(JDBCTransactionStateTest.java:118)
	at org.geotools.jdbc.JDBCTransactionStateTest.testSetTransactionNullWithInternalConnection(JDBCTransactionStateTest.java:100)
```

ansell:
All of the tests pass on master for me on my dev machine:

```
[INFO] BUILD SUCCESS
[INFO] Total time: 19:05 min
[INFO] Finished at: 2017-11-13T12:34:07+11:00
[INFO] Final Memory: 136M/404M
```

jodygarnett:
Follow on PR here https://github.com/geotools/geotools/pull/1738

nicolaiparlog:
Thank you very much, @jodygarnett, for patching this up and merging it! :smiley: I'm looking forward to the first Java 9 compatible release.

mipastgt:
Well, may be the patch is not yet all right. In a simple Quickstart example which uses the OSMService I get the following exception which seems to be related to this work: (19-snapshot 19-20171112.221426-77)
Caught and handled this exception :
```
java.lang.NumberFormatException: For input string: ""
	at java.base/java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
	at java.base/java.lang.Integer.parseInt(Integer.java:662)
	at java.base/java.lang.Integer.parseInt(Integer.java:770)
	at com.sun.media.imageioimpl.common.ImageUtil.processOnRegistration(ImageUtil.java:1408)
	at com.sun.media.imageioimpl.plugins.wbmp.WBMPImageWriterSpi.onRegistration(WBMPImageWriterSpi.java:106)
	at java.desktop/javax.imageio.spi.SubRegistry.registerServiceProvider(ServiceRegistry.java:788)
	at java.desktop/javax.imageio.spi.ServiceRegistry.registerServiceProvider(ServiceRegistry.java:330)
	at java.desktop/javax.imageio.spi.IIORegistry.registerApplicationClasspathSpis(IIORegistry.java:212)
	at java.desktop/javax.imageio.spi.IIORegistry.<init>(IIORegistry.java:136)
	at java.desktop/javax.imageio.spi.IIORegistry.getDefaultInstance(IIORegistry.java:157)
	at java.desktop/javax.imageio.ImageIO.<clinit>(ImageIO.java:66)
	at org.geotools.tile.Tile.loadImageTileImage(Tile.java:194)
	at org.geotools.tile.Tile.getBufferedImage(Tile.java:182)
	at org.geotools.tile.util.TileLayer.getTileImage(TileLayer.java:138)
	at org.geotools.tile.util.TileLayer.renderTile(TileLayer.java:131)
	at org.geotools.tile.util.TileLayer.renderTiles(TileLayer.java:125)
	at org.geotools.tile.util.TileLayer.draw(TileLayer.java:90)
	at org.geotools.renderer.lite.CompositingGroup$WrappingDirectLayer.draw(CompositingGroup.java:227)
	at org.geotools.renderer.lite.StreamingRenderer$RenderDirectLayerRequest.execute(StreamingRenderer.java:3521)
	at org.geotools.renderer.lite.StreamingRenderer$PainterThread.run(StreamingRenderer.java:3582)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:514)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
	at java.base/java.lang.Thread.run(Thread.java:844)
```

nicolaiparlog:
That's a JAI problem. [JAI core tries to parse the Java version string](https://github.com/jai-imageio/jai-imageio-core/blob/3f9c4019d0596d1589282a2ca9d5864fe2b8cfd2/src/main/java/com/github/jaiimageio/impl/common/ImageUtil.java#L1388) and fails because it no longer starts with `1.`.

mipastgt:
You are right. It has been reported already more than a year ago but JAI does not seem to be actively maintained anymore and nobody seems to be willing to apply the proposed patch :-(
https://github.com/jai-imageio/jai-imageio-core/issues/24
I am wondering what other Java 9 related surprises are still slumbering in the code.

don-vip:
@aaime is aware of this, see my other issue: https://github.com/geosolutions-it/jai-ext/issues/161

PR #1672, opened 2017-08-06
[GEOT-5797] Pre-generalized data store spams the log with WARN messages when wrapping JDBC data store

aaime:
The code is logging at warn/info level messages that are debug support, should be at "fine" level instead

### DIFF
--- a/modules/plugin/feature-pregeneralized/src/main/java/org/geotools/data/gen/PreGeneralizedFeatureSource.java
+++ b/modules/plugin/feature-pregeneralized/src/main/java/org/geotools/data/gen/PreGeneralizedFeatureSource.java
@@ -29,6 +29,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.TreeSet;
 import java.util.Map.Entry;
+import java.util.logging.Level;
 import java.util.logging.Logger;
 
 import org.geotools.data.DataAccess;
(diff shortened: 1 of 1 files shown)


bencaradocdavies:
Close/open to retrigger Travis with new settings.

modules/plugin/feature-pregeneralized/src/main/java/org/geotools/data/gen/PreGeneralizedFeatureSource.java
```diff
@@ -465,13 +466,14 @@ private Double getRequestedDistance(Query query) {
...
         if (result == null) {
-            log.warning("No hint for geometry distance in query, fallback to base feature" );
-        }
-        else {
-            log.info("Hint geometry distance: "+result);
+            log.fine("No hint for geometry distance in query, fallback to base feature" );
+        } else {
+            if(log.isLoggable(Level.FINE)) {
+                log.fine("Hint geometry distance: " +result);
+            }
```
tbarsballe: My one comment here is why do you have a `if(log.isLoggable(Level.FINE)) {` wrapping the one `log.fine`  but not the other?

aaime: Because one is building a string, the other is not (one causes garbage even if the log is not logged, but the other is just playing with a constant).

tbarsballe: Right, makes sense.

PR #1674, opened 2017-08-10
[GEOT-5798] New vendor option to control label character spacing

aaime:


docs/user/library/render/style.rst
```diff
@@ -548,6 +548,8 @@ Considerable vendor options are provided for working with TextSymbolizers:
   
 * spaceAround(50) The minimum distance between two labels, in pixels
 
+* charSpacing(0) The extra space between chars, in pixels
```
bradh: Since this is in the docs, I'd suggest expanding "chars" -> "characters".

modules/library/render/src/main/java/org/geotools/renderer/style/SLDStyleFactory.java
```diff
@@ -983,6 +987,12 @@ private Geometry findGeometry(final Object feature, String geomName) {
         if(kerning) {
             javaFont = applyKerning(javaFont);
         }
+        double spacing = voParser.getDoubleOption(symbolizer, TextSymbolizer.CHAR_SPACING_KEY, 
+                TextSymbolizer.DEFAULT_CHAR_SPACING);
+        if(spacing != 0) {
```
bradh: Is this just a common-case optimisation? It should still work with spacing == 0?

aaime: It would, but then kerning would be set and java2d might end up doing more work (depending on its implementation). Since spacing is going to be 0 almost always, I thought it was worth it having the extra if.

modules/library/render/src/test/java/org/geotools/renderer/lite/LabelSpacingTest.java
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * Tests labels underling.
+ */
+public class LabelSpacingTest extends AbstractLabelLineTest {
+    
+    @Test
+    public void testLabelsIncreaseSpacing() throws Exception {
+        // load the style that will underline the labels
+        Style style = loadParametricStyle(this, "lineStyleTemplate.sld", "VENDOR_KEY", "charSpacing", "VENDOR_VALUE", "10");
```
nmco: I guess you mean spacing not underline :stuck_out_tongue:

modules/library/render/src/test/java/org/geotools/renderer/lite/LabelSpacingTest.java
```diff
@@ -0,0 +1,54 @@
...
+        // let's see if the result image match our expectations
+        File reference = new File(
+                "./src/test/resources/org/geotools/renderer/lite/test-data/lineLabelsIncreaseSpacing.png");
+        ImageAssert.assertEquals(reference, image, 3000);
+    }
+    
+    @Test
+    public void testLabelsDecreaseSpacing() throws Exception {
+        // load the style that will underline the labels
+        Style style = loadParametricStyle(this, "lineStyleTemplate.sld", "VENDOR_KEY", "charSpacing", "VENDOR_VALUE", "-5");
```
nmco: Same here.

modules/library/render/src/test/java/org/geotools/renderer/lite/AbstractLabelLineTest.java
```diff
@@ -67,6 +65,23 @@ public void setUp() throws Exception {
         bounds = featureSource.getBounds();
         bounds.expandBy(1, 1);
     }
+    
+    protected Style loadParametricStyle(Object loader, String sldFilename, String... parameters) throws IOException {
+        StyleFactory factory = CommonFactoryFinder.getStyleFactory(null);
```
nmco: Ah! it will help indeed test different variants without having to repeat the styles, thanks.

modules/library/render/src/test/java/org/geotools/renderer/lite/LabelStrikethoughtTest.java
```diff
@@ -29,14 +30,26 @@
...
         // let's see if the result image match our expectations
         File reference = new File(
                 "./src/test/resources/org/geotools/renderer/lite/test-data/strikethroughStyle.sld.png");
         ImageAssert.assertEquals(reference, image, 3000);
     }
+    
+    @Test
+    public void testLabelsNotStrikethrough() throws Exception {
+        // load the style that will underline the labels
+        Style style = loadParametricStyle(this, "lineStyleTemplate.sld", "VENDOR_KEY", TextSymbolizer.STRIKETHROUGH_TEXT_KEY, "VENDOR_VALUE", "false");
```
nmco: Same here.

aaime:
Handled both @bradh and @nmco feedback, merging.

PR #1687, opened 2017-08-24
fixed icon padding test

johngschulz:
The test result changed due to the removal of the icon field in the .json test file.  15 would be the desired result now.


tbarsballe:
In the future, please be sure to run the tests locally before committing when dealing with community modules that are not yet added to the geotools build (MBStyle is not yet added, but will be soon)

PR #1689, opened 2017-08-27
Update GeoPackage format name to make it friendlier to WCS 1.0.

bradh:


modules/plugin/geopkg/src/main/java/org/geotools/geopkg/mosaic/GeoPackageFormat.java
```diff
@@ -130,7 +130,7 @@ public GeoPackageFormat() {
      */
     private void setInfo() {
         final HashMap<String,String> info = new HashMap<String,String> ();
-        info.put("name", "GeoPackage (mosaic)");
+        info.put("name", "GeoPackageMosaic");
```
jodygarnett: How is this name used? Does it actually show up in the WCS protocol at any point - if so this change would be public facing as part of the protocol.

Test case requested? Or is this just a chance that is seen downstream.

aaime:
Closing, it was part of the pull request you just merged .

PR #1695, opened 2017-09-06
[GEOT-5819] Provide access to WFS Transaction exception reports

awaterme:
Proposal of a solution for [GEOT-5819](https://osgeo-org.atlassian.net/browse/GEOT-5819).

Aim: WFS Transaction ExceptionReports should be accessible for API consumers. 

In the current proposal 
- the textual content of the exception is available for all WFS versions.
- structured access (error codes and messages separately) through new fields in WFSException is available for WFS 1.1 and 2.0, but not for 1.0.

Reason for not fully supporting WFS 1.0: The GT parser fails to parse the XML in this case. Probably because the WFS 1.0 parser configuration misses bindings for the exceptions. 
Any hints on how to solve this are welcome.

I have already asked for support on the mailing list but received no response.


modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/internal/parsers/AbstractGetFeatureResponseParserFactory.java
```diff
@@ -92,134 +77,25 @@ public boolean canProcess(final WFSRequest request, final String contentType) {
...
-        // CubeWerx, upon a successful GetFeature request, set the response's content-type
-        // header to plain "text/xml" instead of "text/xml;subtype=gml/3.1.1". So we'll do a bit
-        // of heuristics to find out what it actually returned
-        final int buffSize;
-        if (LOGGER.isLoggable(Level.FINER)) {
-            buffSize = 4096;
-        } else {
-            buffSize = 512;
-        }
-        PushbackInputStream pushbackIn = new PushbackInputStream(response.getResponseStream(), buffSize);
```
jodygarnett: Looks like this logic moved to AbstractWFSResponseFactory.java below

modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/internal/WFSException.java
```diff
@@ -25,9 +29,54 @@
  */
 public class WFSException extends IOException {
     private static final long serialVersionUID = -2828901359361793862L;
+    
+    public static final class ExceptionDetails implements Serializable {
```
jodygarnett: Please javadoc public class as this forms API, looking for class name, responsibility and relationships (how it is used by whom).

I am not sure why this needs to be an inner class, is it so the fields of WFSException can be serialized as a distinct object?

modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/internal/parsers/AbstractGetFeatureResponseParserFactory.java
awaterme: Yes, right.

modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/internal/WFSException.java
awaterme: I will add some javadocs, as you suggested.

I have chosen an inner class because the ExceptionDetails are so closely related to the WFSException, like a composition. So everything is in one place. The ExceptionDetails are not expected to be used outside of this class, by the gt-wfs-ng module.
It has nothing to do with serialization.
But I am fine to make the class top level, if you like.

awaterme:
Many thanks for the review Jody! When you want to make a the ExceptionDetails top level, let me know. I will amend it all together then.
Best regards,
Andreas

aaime:
I cannot tell if this needs further action or not? "I will add some javadocs, as you suggested" but I don't see any extra commit.

aaime:
Hi, I see that nothing moved after a month since my last comment. Would be a pity to lose this pull request thought.

awaterme:
Hi Andrea, 
I refactored the `ExceptionDetails` to be a top-level class, as requested. As far as I can see there are no more open issues.
Please let me know otherwise; I also don't want to loose this.
Best regards,
Andreas

aaime:
Merged! :-)

PR #1696, opened 2017-09-06
[GEOT-5817] StreamingRenderer can result in map with white lines at tile borders against a heterogeneous CRS mosaic

aaime:


modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/Utils.java
```diff
@@ -125,6 +125,7 @@
 import it.geosolutions.imageio.pam.PAMDataset.PAMRasterBand.Metadata.MDI;
 import net.sf.ehcache.Cache;
 import net.sf.ehcache.Element;
+import net.sf.ehcache.store.disk.ods.AATreeSet;
```
dromagnoli: Spurious import

PR #1699, opened 2017-09-11
Fix for mbstyle visual-transformer font tests

johngschulz:
Imported and registered Vera.ttf
Once the font is registered in the set up it works across platforms
There are continued problems with symbol-icon-line-placement test


modules/unsupported/mbstyle/src/test/java/org/geotools/mbstyle/transform/VisualTransformerTest.java
```diff
@@ -108,6 +111,11 @@ public void setUp() throws Exception {
         lineZigFS = ds.getFeatureSource("testlinezigs");
         pointsWithMarksFS = ds.getFeatureSource("testmarks");
         bounds = new ReferencedEnvelope(0, 10, 0, 10, CRS.decode("EPSG:4326"));
+        
+        
+        FontCache.getDefaultInstance().registerFont(
+        Font.createFont(Font.TRUETYPE_FONT, TestData.getResource(LabelShieldTest.class, "Vera.ttf")
+        		 .openStream()));
```
tbarsballe: So, looks like there are some additional steps to get the fonts working, good to know.

johngschulz:
All of the font related test are working locally - still looking into the issue with the non-font related test where the goldfish sprites are shifting a pixel or two in linux vs mac os.

modules/unsupported/mbstyle/src/test/java/org/geotools/mbstyle/transform/VisualTransformerTest.java
johngschulz: Jody helped me through here, also of note as you may have seen in the community emails vera is the go to for any visual tests going forward as well.

PR #1700, opened 2017-09-12
[GEOT-5821]: TimeCoordinateVariable management fix

dromagnoli:


dromagnoli:
Hi @NielsCharlier,
we have a case of a NetCDF variable made of time,lat,lon dimensions containing the pixel time of each pixel having units = "seconds since ...:"
The NetCDF DefaultConvention class reports a variable having a Date unit attribute as a Time CoordinateAxis.
https://github.com/Unidata/thredds/blob/5.0.0/cdm/src/main/java/ucar/nc2/dataset/conv/DefaultConvention.java#L140

Before the netcdf aggregation / runtime support, all coordinate axis were checked to be 1D.
Right now, each type of CoordinateAxis is supported so you can have a 2D coordinateAxis with runtime,time as per your needs.  

In that case we are getting a CoordinateAxis with 3 dimensions since the variable being interpreted by the UCAR lib as a CoordinateAxis has time, lat, lon.
The result was that the new CoordinateAxisGeneralHelper, performing a coordinateAxis.read() which may make sense when you have runtime,time aggregation made of a few values, made the reader loading in memory a 3600x1800 (full world extent) doubles array.

Please, let me know what do you think about this fix.
Regards,
Daniele

NielsCharlier:
you have already merged, but yeah that looks fine.

dromagnoli:
yes. we were needing a quick deploy to solve a performance issue :) 
Thx for the feedback

PR #1701, opened 2017-09-13
GEOT-5816: ClassifyByRange vector process

mbarto:


modules/unsupported/process-feature/src/main/java/org/geotools/process/vector/ClassifyByRangeProcess.java
```diff
@@ -0,0 +1,144 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ *
+ *    (C) 2011-2017, Open Source Geospatial Foundation (OSGeo)
```
aaime: Copyright should be only 2017 to OSGeo

modules/unsupported/process-feature/src/main/java/org/geotools/process/vector/ClassifyByRangeProcess.java
```diff
@@ -0,0 +1,144 @@
...
+@DescribeProcess(title = "ClassifyByRange", description = "Computes a new attribute to classify another attribute by intervals over vector data sets.")
+public class ClassifyByRangeProcess implements VectorProcess {
+    
+    static FilterFactory filterFactory = CommonFactoryFinder.getFilterFactory(null);
+    
+    @DescribeResult(name = "result", description = "Classified feature collection")
+    public SimpleFeatureCollection execute(
+            @DescribeParameter(name = "features", description = "Input feature collection") SimpleFeatureCollection features,
+            @DescribeParameter(name = "classifyOnAttribute", description = "Attribute to be classified using intervals of values.") String classifyOnAttribute,
+            @DescribeParameter(name = "thresholds", min = 0, description = "Comma delimited list of thresholds (use this one to specify custom intervals).") String thresholds,
```
aaime: A list of things should be either String[] or List<String> (the latter requires a one more annotation attribute to specify what's inside the list

modules/unsupported/process-feature/src/main/java/org/geotools/process/vector/ClassifyByRangeProcess.java
```diff
@@ -0,0 +1,144 @@
...
+            @DescribeParameter(name = "classifier", min = 0, description = "Classifier type (EqualInterval, Quantile, Jenks). Use with classes to calculate intervals automatically") String classifier,
+            @DescribeParameter(name = "classes", min = 0, description = "Classifier # of classes, use with classifier (defaults to 5).") Integer classes,
+            @DescribeParameter(name = "include", min = 0, defaultValue = "FALSE", description = "Include or exclude current threshold in the interval.") Boolean include,
+            @DescribeParameter(name = "outputAttribute", min = 0, description = "Name of the output attribute with class values (defaults to class).") String outputAttribute,
+            @DescribeParameter(name = "outputValues", min = 0, description = "Comma delimited list of class values for each given threshold (+1 for out of range).") String outputValues,
```
aaime: Same here. Imho the output values should be close to the thresholds so that they can be read toghether (at least in a Java call, in a WPS call the convention of repeating the input n times will make that hard, but it's how the protocol is designed)

modules/unsupported/process-feature/src/main/java/org/geotools/process/vector/ClassifyByRangeProcess.java
```diff
@@ -0,0 +1,144 @@
...
+            @DescribeParameter(name = "outputType", min = 0, description = "Optional binding type for output values (defaults to String).") String outputType
```
aaime: Output type should be close to output name

modules/unsupported/process-feature/src/main/java/org/geotools/process/vector/ClassifyByRangeProcess.java
```diff
@@ -0,0 +1,144 @@
...
+            )
+            throws ProcessException {
+        
+        if(features != null) {
```
aaime: Formatting does not conform to convention

modules/unsupported/process-feature/src/main/java/org/geotools/process/vector/ClassifyByRangeProcess.java
```diff
@@ -0,0 +1,144 @@
...
+ * Computes a new attribute to classify another attribute by intervals over vector data sets.
+ * 
+ * @author Mauro Bartolomeoli
+ *
+ * @source $URL$
+ */
+@DescribeProcess(title = "ClassifyByRange", description = "Computes a new attribute to classify another attribute by intervals over vector data sets.")
+public class ClassifyByRangeProcess implements VectorProcess {
+    
+    static FilterFactory filterFactory = CommonFactoryFinder.getFilterFactory(null);
```
aaime: Should be also final, and UPPERCASE

modules/unsupported/process-feature/src/main/java/org/geotools/process/vector/ClassifyByRangeProcess.java
```diff
@@ -0,0 +1,144 @@
...
+    
+    @DescribeResult(name = "result", description = "Classified feature collection")
+    public SimpleFeatureCollection execute(
+            @DescribeParameter(name = "features", description = "Input feature collection") SimpleFeatureCollection features,
+            @DescribeParameter(name = "classifyOnAttribute", description = "Attribute to be classified using intervals of values.") String classifyOnAttribute,
+            @DescribeParameter(name = "thresholds", min = 0, description = "Comma delimited list of thresholds (use this one to specify custom intervals).") String thresholds,
+            @DescribeParameter(name = "classifier", min = 0, description = "Classifier type (EqualInterval, Quantile, Jenks). Use with classes to calculate intervals automatically") String classifier,
```
aaime: If the list if finite best to use an enumeration, it will show up in DescribeProcess output

modules/unsupported/process-feature/src/main/java/org/geotools/process/vector/ClassifyByRangeProcess.java
aaime: Also, best to specify what happens if both custom thresholds and classes count are specified at the same time

modules/unsupported/process-feature/src/main/java/org/geotools/process/vector/ClassifyByRangeProcess.java
```diff
@@ -0,0 +1,144 @@
...
+                    outputValues += count + ",";
+                }
+                outputValues += (ranges.size() + 1 + "");
+            }
+            String[] rangesValues = outputValues.split(",");
+            if (rangesValues.length != (ranges.size() + 1)) {
+                throw new ProcessException("values are not consistent with thresholds (should be +1)");
+            }
+            for(int count = 0; count < ranges.size(); count++) {
+                params.add(filterFactory.literal(Converters.convert(rangesValues[count], classify.binding)));
```
aaime: The converters will return null if the value cannot be converted, I guess we'd want an exception with a suitable error message if that happens?

modules/unsupported/process-feature/src/main/java/org/geotools/process/vector/ClassifyByRangeProcess.java
```diff
@@ -0,0 +1,144 @@
...
+            }
+            params.add(filterFactory.literal(Converters.convert(rangesValues[rangesValues.length - 1], classify.binding)));
+            params.add(filterFactory.literal(include ? CategorizeFunction.PRECEDING : CategorizeFunction.SUCCEEDING));
+            classify.expression = new CategorizeFunction(params, null);
+            
+            
+            transform.add(classify);
+            return new TransformProcess().executeList(features, transform);
+        }
+        throw new ProcessException("features input cannot be null!");
```
aaime: Small style bit, I'd flip the logic and avoid keeping the core of the process code inside an if block (flatter code is normally easier to read)

modules/unsupported/process-feature/src/test/java/org/geotools/process/vector/ClassifyByRangeProcessTest.java
```diff
@@ -0,0 +1,205 @@
...
+        assertEquals(3, result.size());
+        SimpleFeatureIterator iterator = result.features();
+        SimpleFeature ft = iterator.next();
+        assertEquals("A", ft.getAttribute("class"));
+        ft = iterator.next();
+        assertEquals("B", ft.getAttribute("class"));
+        ft = iterator.next();
+        assertEquals("C", ft.getAttribute("class"));
+    }
+}
```
aaime: Maybe add a test for invalid classifier function, in case using an enum for it is not suitable

modules/unsupported/process-feature/src/main/java/org/geotools/process/vector/ClassifyByRangeProcess.java
mbarto: It's not really finite, it could hypothetically use any geotools function, I would leave it open for future improvfements

PR #1702, opened 2017-09-13
[GEOT-5823] Mapping java.sql.Timestamp to TIMESTAMP database type

kzr-pzr:
We need to register a new type mapping to force java.sql.Timestamp to
be mapped to Oracle TIMESTAMP type (not DATE type) so time information
won't be lost when inserting features with java.sql.Timestamp attributes
(GeoTools maps DATE to java.sql.Date and strips time information before
inserting to database).

### DIFF
--- a/modules/plugin/jdbc/jdbc-oracle/src/main/java/org/geotools/data/oracle/OracleDialect.java
+++ b/modules/plugin/jdbc/jdbc-oracle/src/main/java/org/geotools/data/oracle/OracleDialect.java
@@ -163,6 +163,7 @@ public class OracleDialect extends PreparedStatementSQLDialect {
     		put("NVARCHAR", String.class);
     		put("NVARCHAR2", String.class);
             put("DATE", java.sql.Date.class);
+            put("TIMESTAMP", java.sql.Timestamp.class);
     	}
     };
     
@@ -1436,6 +1437,8 @@ public class OracleDialect extends PreparedStatementSQLDialect {
         // starting with Oracle 11 + recent JDBC drivers the DATE type does not have a mapping
         // anymore in the JDBC driver, manually register it instead
         overrides.put(Types.DATE, "DATE");
+        // overriding default java.sql.Timestamp to Oracle DATE mapping
+        overrides.put(Types.TIMESTAMP, "TIMESTAMP");
     }
     
     @Override
--- a/modules/plugin/jdbc/jdbc-oracle/src/test/java/org/geotools/data/oracle/OracleDateOnlineTest.java
+++ b/modules/plugin/jdbc/jdbc-oracle/src/test/java/org/geotools/data/oracle/OracleDateOnlineTest.java
@@ -2,10 +2,20 @@ package org.geotools.data.oracle;
 
 import java.sql.Date;
 import java.sql.Timestamp;
+import java.util.TimeZone;
 
+import org.geotools.data.DataUtilities;
+import org.geotools.data.DefaultTransaction;
+import org.geotools.data.Query;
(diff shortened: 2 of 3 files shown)


aaime:
Needs a test to be considered for merging

kzr-pzr:
I'm aware of it, however writing a test turned out to be more complicated than I thought and I need help/advice from other developers. So I created the PR first to have some grounds for discussion.

The general idea of testing this new mapping is to insert a feature with a `java.sql.Timestamp` attribute into Oracle database table and then read it and verify that the time information is not lost. 

So, I tried to add the following method to the `OracleDateOnlineTest`:

    public void testInsertTemporal() throws Exception {
        SimpleFeatureType ft = dataStore.getSchema( tname("dates") );
        SimpleFeatureBuilder builder = new SimpleFeatureBuilder(ft);
        long theTimestamp = 1503926476000L; // August 28, 2017 13:21:16 GMT
        builder.add(new Date(theTimestamp));
        builder.add(new Timestamp(theTimestamp));
        builder.add(new Timestamp(theTimestamp));
        SimpleFeature feature = builder.buildFeature(null);

        JDBCFeatureStore fs = (JDBCFeatureStore) dataStore.getFeatureSource(tname("dates"));
        Transaction t = new DefaultTransaction("add");
        fs.setTransaction(t);
        try {
            fs.addFeatures(DataUtilities.collection(feature));
            t.commit();
        } catch (Exception ex) {
            t.rollback();
            throw ex;
        } finally  {
            t.close();
        }

        FilterFactory ff = dataStore.getFilterFactory();

        Filter f = ff.equals(ff.property(aname("dt")),
                ff.literal("2017-08-28 13:21:16"));
        assertEquals(1, fs.getCount(new Query(tname("dates"), f)));
    }

The problem is that this test does not pass because of the cast
`JDBCFeatureStore fs = (JDBCFeatureStore) dataStore.getFeatureSource(tname("dates"));`

It blows with a `ClassCastException`. As I studied GeoTools code the explanation for such exception is that I don't have write access for the [`dates`](https://github.com/geotools/geotools/blob/master/modules/plugin/jdbc/jdbc-oracle/src/test/java/org/geotools/data/oracle/OracleDateTestSetup.java) table. The reason why I don't have write access to the table is because the `JDBCFeatureSource` class in `buildFeatureType()` method examines the table, [can't find primary key and therefore marks it as read-only](https://github.com/geotools/geotools/blob/master/modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCFeatureSource.java#L174). So I cannot use `dates` table to insert features, at least not without major modifications in existing code.

An alternative test case would be extending the `ft1` testing table with a TIMESTAMP column and then reusing the [`testAddFeatures`](https://github.com/geotools/geotools/blob/master/modules/library/jdbc/src/test/java/org/geotools/jdbc/JDBCFeatureStoreOnlineTest.java#L75) in the `JDBCFeatureStoreOnlineTest`, but that would then affect all supported databases and I don't have access to all of them (I can test only Oracle and Postgis).

Another possibility is to create a whole new testing table but I think reusing and tweaking existing code is a better idea.

Any thoughts, comments? 

Thanks

aaime:
Add a new small and writable table :-)

kzr-pzr:
Is it OK now?

aaime:
Aah yep. Thankfully fever got me enough time to install Oracle locally to try it out (I can also VPN into the office one, but running the test suite takes over an hour that way). Works fine, merging.

aaime:
Also backported on 17.x (squashed as a single commit). 16.x series is dead, so I did not backport any further.

PR #1703, opened 2017-09-13
Fixed test style and updated image

johngschulz:
This updated the test style to use Bitstream Sera Sans font to address visual test issues resulting from unregistered fonts.
commented out system interactive, true property


modules/unsupported/mbstyle/src/test/java/org/geotools/mbstyle/transform/VisualTransformerTest.java
```diff
@@ -118,7 +118,7 @@ public void setUp() throws Exception {
         		 .openStream()));
 
         // UNCOMMENT THE BELOW LINE TO DISPLAY VISUAL TESTS
-         System.setProperty("org.geotools.test.interactive", "true");
+        // System.setProperty("org.geotools.test.interactive", "true");
```
jodygarnett: I am not sure these are required, we have instructions in our docs about using -D command line options for interactive tests. You should be able to do the same thing from your IDE.

jodygarnett:
Small question good to merge; thanks for the debugging

modules/unsupported/mbstyle/src/test/java/org/geotools/mbstyle/transform/VisualTransformerTest.java
johngschulz: These were in place in other visual test as well, I think it's just there to make testing easier.  At any rate I had it uncommented while working through it and just remembered to re-comment out.

aaime: I use that a lot but don't put it in the code anymore, all launch configurations for rendering unit tests in my IDE have ``-Dorg.geotools.image.test.interactive=true``, this way one cannot forget uncommenting.

PR #1705, opened 2017-09-17
Geopackage spatial index support

aaime:


aaime:
Btw, I've tried to code this one with IntelliJ and noticed it's doing whitespace changes without me asking for them... how do I stop the sucker from doing so? ;-)

jratike80:
What I understand looks good. There is some evedence that using JOIN in the SQL could be faster than sub-query but I have not verified that.

I would add to this comment that current GeoPackage standard allows only one geometry field in a table or view.
    // This implementation only supports figuring a bbox in case there is a single
    // indexed spatial attribute (could be extended to use multiple spatial attributes if need be)

Reference: standard version 1.2
Requirement 30
A feature table SHALL have only one geometry column.

aaime:
Hi @jratike80 thanks for the feedback. Adding the spatial filter as a join is hard, we'd have to redesign filter encoding, currently it works under the assumption that it can be fully encoded inside the WHERE clause.

aaime:
Wondering if @ianturton has a minute to look at this pull request? :-)

aaime:
I'm merging this one so that it gets in the RC, not having a spatial index at all is problematic...

PR #1707, opened 2017-09-20
[GEOT-5829] Font antialiasing differences cause gt-mbstyle VisualTransformerTest failures on OpenJDK (master)

bencaradocdavies:
https://osgeo-org.atlassian.net/browse/GEOT-5829

### DIFF
--- a/modules/unsupported/mbstyle/src/test/java/org/geotools/mbstyle/transform/VisualTransformerTest.java
+++ b/modules/unsupported/mbstyle/src/test/java/org/geotools/mbstyle/transform/VisualTransformerTest.java
@@ -684,7 +684,7 @@ public class VisualTransformerTest {
         renderer.setJava2DHints(new RenderingHints(KEY_ANTIALIASING, VALUE_ANTIALIAS_ON));        
         BufferedImage image = MapboxTestUtils.showRender(renderTitle, renderer, DISPLAY_TIME,
                 new ReferencedEnvelope[] { bounds }, null, width, height);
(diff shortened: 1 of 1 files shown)


bencaradocdavies:
@johngschulz @tbarsballe @jodygarnett this pull request increases the perceptualdiff tolerance from 50 to 900 pixels (1% of the total number of pixels in each test image). Because the failures only seem to affect tests with a lot of text, this suite could be refactored so that the higher tolerance is applied only to tests that rended text.

tbarsballe:
Seems like a perfectly tolerable interim fix, although it would be nice to eventually figure out and fix the root cause of the issue.

Thanks; I'll merge after Travis passes.

bencaradocdavies:
Thanks @tbarsballe, Travis passed so I merged.

I am sure that this is a font antialiasing difference and not a kerning issue. Please have a look at any of the image pairs attached to [the Jira issue](https://osgeo-org.atlassian.net/browse/GEOT-5829). Flicking between them, the letter spacing appears to be the same, but the greyscale rendering of the glyphs is substantially different. Looks like font antialiasing to me, and when we saw this same problem in gt-render when we first started supporting OpenJDK, font antialiasing was the culprit.

tbarsballe:
Ok, I see what you mean. It does indeed look like an antialiasing issue.

bencaradocdavies:
At least we know that Vera is being used as intended and that the test fixture font setup by @johngschulz is working as intended!

johngschulz:
Good to know, sorry - took me a minute to get thing working for testing - I just transferred everything to a new machine and took me a minute to get virtualbox back up and running.

PR #1709, opened 2017-09-27
[GEOT-5833] Image mosaic won't respond to requests outside of the coverage bbox it cached in memory

aaime:
Hi @dromagnoli you should be already familiar with this pull request (it's the GeoTools side of it, we'll have a GeoServer one as well).

Update: the error I was previously seeing was due to a side effect of a geometric operation. The bbox returned was slightly larger than requested, the GetCoverage did a crop, and during the crop the wrapped dimensions were lost. This could probably be the subject of a different bug report.


aaime:
@dromagnoli pull request updated.

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/RasterLayerResponse.java
```diff
@@ -714,8 +683,15 @@ private MosaicOutput prepareResponse() throws DataSourceException {
                 }
             }
 
-            // prepare a blank response
-            return createBlankResponse();
+            // do we return a null (outside of the coverage) or a blank? The choice is "hard" as we
+            // might be in a hole of the coverage and not know it
+            if(!mosaicBBox.intersects((BoundingBox) ReferencedEnvelope.reference(coverageEnvelope)) &&
+                    !mosaicBBox.intersects((BoundingBox) ReferencedEnvelope.reference(rasterManager.spatialDomainManager.coverageBBox))) {
+                return null;
```
dromagnoli: Do you think it would make sense to add a log at FINE*** level reporting about no intersections at all?

aaime:
The test failure is something I cannot reproduce using neither maven nor IDE locally, really weird...

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/RasterLayerResponse.java
aaime: I'll add it

PR #1710, opened 2017-10-02
[GEOT-5836] Update GeoPackage table schemas to match spec.

bradh:
Should fix GeoServer CITE failures for GeoPackage (ETS-10).

@aaime should fix https://osgeo-org.atlassian.net/browse/GEOT-5546 which is the same as the tiles part of GEOT-5836.

### DIFF
--- a/modules/plugin/geopkg/src/main/java/org/geotools/geopkg/GeoPackage.java
+++ b/modules/plugin/geopkg/src/main/java/org/geotools/geopkg/GeoPackage.java
@@ -1161,11 +1161,11 @@ public class GeoPackage {
                 }
                 //create the tile table itself
                 st = cx.prepareStatement(format("CREATE TABLE %s (" +
-                    "id INTEGER PRIMARY KEY AUTOINCREMENT," +
-                    "zoom_level INTEGER NOT NULL DEFAULT 0," +
-                    "tile_column INTEGER NOT NULL DEFAULT 0," +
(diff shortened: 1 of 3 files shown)


modules/plugin/geopkg/src/main/java/org/geotools/geopkg/GeoPackage.java
```diff
@@ -1161,11 +1161,11 @@ public void create(TileEntry entry) throws IOException {
                 }
                 //create the tile table itself
                 st = cx.prepareStatement(format("CREATE TABLE %s (" +
-                    "id INTEGER PRIMARY KEY AUTOINCREMENT," +
-                    "zoom_level INTEGER NOT NULL DEFAULT 0," +
-                    "tile_column INTEGER NOT NULL DEFAULT 0," +
-                    "tile_row INTEGER NOT NULL DEFAULT 0," +
-                    "tile_data BLOB NOT NULL DEFAULT (zeroblob(4)))", e.getTableName()));
+                    "id INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT," +
```
bradh: This change is arguable. Table 11 in the released GeoPackage 1.1 spec and Table 10 in the GeoPackage 1.2 spec say not null, but the CITE suite says null. Will see what the results of a change to the ETS show.

aaime:
I did not have time to check the spec, but overall looks good to me.

PR #1712, opened 2017-10-04
[GEOT-5841] Fix "addFeatures(FeatureCollection)" to iterate over features

emerkle826:
This should address https://osgeo-org.atlassian.net/browse/GEOT-5841.

### DIFF
--- a/modules/library/data/src/test/java/org/geotools/data/memory/MemoryDataStoreTest.java
+++ b/modules/library/data/src/test/java/org/geotools/data/memory/MemoryDataStoreTest.java
@@ -164,6 +164,19 @@ public class MemoryDataStoreTest extends DataTestCase {
     }
 
     /*
+     * Test for void MemoryDataStore.addFeatures(FeatureCollection<SimpleFeatureType,SimpleFeature> collection)
+     */
+    public void testMemoryDataStoreAddFeatures() throws Exception {
+        MemoryDataStore store = new MemoryDataStore();
+        assertNotNull(store);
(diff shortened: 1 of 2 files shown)


modules/library/data/src/test/java/org/geotools/data/memory/MemoryDataStoreTest.java
```diff
@@ -161,6 +161,10 @@ public void testMemoryDataStore() throws Exception {
     public void testMemoryDataStoreFeatureCollection() throws IOException {
         DataStore store = new MemoryDataStore(DataUtilities.collection(roadFeatures));
```
aaime: Looking at it, it did not seem a valid test as there is no call to addFeatures. I see the constructor calls in internally, but this could change in the future. Maybe add an explicit test too?

emerkle826: I can add an explicit one for addFeatures as well.

emerkle826:
@aaime Thank you for the feedback. I've reverted the original test (in case the implicit functionality changes in the future) and added a separate explicit test for `addFeatures(FeatureCollection)`.

tbarsballe:
Does this need to be backported?

emerkle826:
@tbarsballe @aaime This can be backported to every version of GeoTool from 16.0 and up. GeoTools 15.x does not have this issue. I'm sure some versions of GeoTools since 16 are no longer supported, so please backport to as few or as many as you wish.

tbarsballe:
Backported to 18.x and 17.x.

PR #1713, opened 2017-10-05
[GEOT-5842] Maintain nodata in image mosaic when no granules are read and when JAI deferred loading is in use

aaime:


modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/ImageMosaicWalker.java
```diff
@@ -26,6 +26,7 @@
 import java.util.logging.Logger;
 
 import javax.imageio.ImageIO;
+import javax.imageio.ImageReader;
```
dromagnoli: looks like an unneeded import?

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/ImageMosaicConfigHandler.java
```diff
@@ -1439,6 +1450,10 @@ public void updateConfiguration(GridCoverage2DReader coverageReader,
                 currentConfigurationBean.getCatalogConfigurationBean().setHeterogeneousCRS(true);
             }
 
+            // nodata treatment
+            
+
+
```
dromagnoli: Don't hate me for this review: why 3 whitelines? :)

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/ImageMosaicConfigHandler.java
```diff
@@ -1379,6 +1379,17 @@ public void updateConfiguration(GridCoverage2DReader coverageReader,
                 configBuilder.setCrs(actualCRS);
             }
 
+            String noData = IndexerUtils.getParameter(Prop.NO_DATA, indexer);
+            if(noData != null && !noData.isEmpty()) {
+                try {
+                    double noDataValue = Double.parseDouble(noData);
+                    configBuilder.setNoData(noDataValue);
+                } catch (NumberFormatException e) {
+                    throw new RuntimeException("Invalid NoData specification " + noData + ", was expecting a number", e);
```
dromagnoli: Question on my side:
I see in the Utils class that a FINE log level will be raised reporting an unparseable nodata in case a NumberFormatException occurs. 
Should we adopt a similar approach here (maybe WARNING) or do you think it is better to stop the mosaic configuration at all, instead?

dromagnoli (commented):
I have spotted a potentially useless import added.
I have also added some questions which are not a blocker for the Approval.

dromagnoli (changes requested):
Ah. I have just noticed that the travis build is failed for a compiling issue. 
You may have forget a new class OR you just need to update the pom to depend on a new imageio-ext version

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/ImageMosaicWalker.java
aaime: Will fix

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/ImageMosaicConfigHandler.java
aaime: Personally I have a preference for fail fast, at least against configuration, otherwise one just looks around without understanding. Maybe doing both (log and exception) is better

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/ImageMosaicConfigHandler.java
aaime: Because I like having space! (will fix ;-) )

aaime:
Eh, 1.1.18 did not contain a necessary new method, upgraded to 1.1.19, let's see if it fixes the build

dromagnoli (approved):
Looks good to me

PR #1714, opened 2017-10-06
[GEOT-5840] Fix handling WellKnownScaleSet in WMTS getCapabilities

etj:
Using the capabilitites document at
https://openlayers.org/en/v4.3.4/examples/data/WMTSCapabilities.xml
an exception is thrown:
```
java.lang.RuntimeException: Parsing failed for TileMatrixSet: java.lang.ClassCastException: java.net.URI cannot be cast to java.lang.String
at org.geotools.wmts.bindings.TileMatrixSetBinding.parse(TileMatrixSetBinding.java:123)
...

### DIFF
--- a/modules/extension/wmts/src/test/java/org/geotools/data/wmts/model/WMTSCapabilitiesTest.java
+++ b/modules/extension/wmts/src/test/java/org/geotools/data/wmts/model/WMTSCapabilitiesTest.java
@@ -231,6 +231,30 @@ public class WMTSCapabilitiesTest extends TestCase {
         }
     }
 
+    public void testParserOLSample() throws Exception {
+        WMTSCapabilities capabilities = createCapabilities("ol.getcapa.xml");
+        try {
+            assertEquals("1.0.0", capabilities.getVersion());
+
+            WMTSService service = (WMTSService) capabilities.getService();
+            assertEquals("Koordinates Labs", service.getTitle());
+
+            List<TileMatrixSet> matrixSets = capabilities.getMatrixSets();
+            assertNotNull(matrixSets);
+            assertFalse(matrixSets.isEmpty());
+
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: modules/extension/wmts/src/test/java/org/geotools/data/wmts/model/WMTSCapabilitiesTest.java [hunk 1]
```diff
@@ -231,6 +231,30 @@ public void testParser3() throws Exception {
...
+            assertEquals("Koordinates Labs", service.getTitle());
+
+            List<TileMatrixSet> matrixSets = capabilities.getMatrixSets();
+            assertNotNull(matrixSets);
+            assertFalse(matrixSets.isEmpty());
+
+            assertEquals("urn:ogc:def:wkss:OGC:1.0:GoogleMapsCompatible", matrixSets.get(0).getWellKnownScaleSet());
+
+        } catch (Exception e) {
+            e.printStackTrace();
```
aaime: What is happening here?

modules/extension/xsd/xsd-wmts/src/main/java/org/geotools/wmts/bindings/TileMatrixSetBinding.java
```diff
@@ -120,7 +120,11 @@ public Object parse(ElementInstance instance, Node node, Object value) throws Ex
         matrixSet.setBoundingBox((BoundingBoxType) node.getChildValue("BoundingBox"));
         matrixSet.setIdentifier((CodeType) node.getChildValue("Identifier"));
         matrixSet.setSupportedCRS(((URI) node.getChildValue("SupportedCRS")).toString());
-        matrixSet.setWellKnownScaleSet((String) node.getChildValue("WellKnownScaleSet"));
+
+        URI wkss = (URI)node.getChildValue("WellKnownScaleSet");
+        if(wkss != null) {
```
aaime: Formatting, the if should be "if (wkss != null)" (with a space). I make this mistake a lot myself, @dromagnoli catches it every time :-p

PR #1717, opened 2017-10-11
GEOT-5844: Implemented the SumArea function for the Aggregate WPS pro…

mbarto:
…cess

### DIFF
--- a/modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
+++ b/modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
@@ -70,9 +70,12 @@ import org.geotools.factory.Hints;
 import org.geotools.feature.NameImpl;
 import org.geotools.feature.simple.SimpleFeatureBuilder;
 import org.geotools.feature.visitor.CountVisitor;
+import org.geotools.feature.visitor.FeatureAttributeVisitor;
 import org.geotools.feature.visitor.GroupByVisitor;
 import org.geotools.feature.visitor.LimitingVisitor;
+import org.geotools.filter.FilterAttributeExtractor;
 import org.geotools.filter.FilterCapabilities;
+import org.geotools.filter.FunctionExpression;
 import org.geotools.geometry.jts.ReferencedEnvelope;
 import org.geotools.jdbc.JoinInfo.JoinPart;
 import org.geotools.referencing.CRS;
@@ -263,6 +266,11 @@ public final class JDBCDataStore extends ContentDataStore
      * Feature visitor to aggregate function name
      */
     protected HashMap<Class<? extends FeatureVisitor>,String> aggregateFunctions;
+    
+    /**
+     * java supported filter function mappings to dialect name;
+     */
+    protected HashMap<String, String> supportedFunctions;
     
     /**
      * flag controlling if the datastore is supporting feature and geometry
@@ -672,6 +680,17 @@ public final class JDBCDataStore extends ContentDataStore
             dialect.registerAggregateFunctions(aggregateFunctions);
         }
         return aggregateFunctions;
+    }
+    
+    /**
+     * Returns the supported functions and the dialect name they map to.
+     */
+    public Map<String,String> getSupportedFunctions() {
+        if ( supportedFunctions == null ) {
+            supportedFunctions = new HashMap();
+            dialect.registerFunctions(supportedFunctions);
+        }
+        return supportedFunctions;
     }
     
     /**
@@ -1450,10 +1469,15 @@ public final class JDBCDataStore extends ContentDataStore
             return null;
         }
         // try to extract an aggregate attribute from the visitor
-        AttributeDescriptor att = extractAggregateAttribute(visitor, featureType);
-        if (att == null && !isCountVisitor(visitor)) {
-            return null; // aggregate function optimization only supported for PropertyName expression
+        Expression aggregateExpression = getAggregateExpression(visitor);
+        // check if aggregation expression is supported
+        if (aggregateExpression instanceof FunctionExpression) {
+            FunctionExpression func = (FunctionExpression)aggregateExpression;
+            if(!getSupportedFunctions().containsKey(func.getFunctionName().getName())) {
+                return null;
+            }
         }
+        
         // if the visitor is limiting the result to a given start - max, we will
         // try to apply limits to the aggregate query
         LimitingVisitor limitingVisitor = null;
@@ -1471,11 +1495,11 @@ public final class JDBCDataStore extends ContentDataStore
             
             try {
                 if ( dialect instanceof PreparedStatementSQLDialect ) {
-                    st = selectAggregateSQLPS(function, att, groupByAttributes, featureType, query, limitingVisitor,  cx);
+                    st = selectAggregateSQLPS(function, aggregateExpression, groupByAttributes, featureType, query, limitingVisitor,  cx);
                     rs = ((PreparedStatement)st).executeQuery();
                 } 
                 else {
-                    String sql = selectAggregateSQL(function, att, groupByAttributes, featureType, query, limitingVisitor);
+                    String sql = selectAggregateSQL(function, aggregateExpression, groupByAttributes, featureType, query, limitingVisitor);
                     LOGGER.fine( sql );
                     
                     st = cx.createStatement();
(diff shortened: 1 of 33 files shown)


modules/library/jdbc/src/main/java/org/geotools/jdbc/SQLDialect.java
```diff
@@ -907,7 +912,7 @@ public void encodeJoin(Type joinType, StringBuffer sql) {
     /**
      * Encodes the CREATE TABLE statement.
      * <p>
-     * Default implementation adds “CREATE TABLE” to the sql buffer.
+     * Default implementation adds â€œCREATE TABLEâ€� to the sql buffer.
```
bradh: Looks like some UTF-8 corruption.

modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
```diff
@@ -675,6 +683,17 @@ public void setAssociations(boolean foreignKeyGeometries) {
     }
     
     /**
+     * Returns the supported functions and the dialect name they map to.
+     */
+    public Map<String,String> getSupportedFunctions() {
+        if ( supportedFunctions == null ) {
+            supportedFunctions = new HashMap();
```
aaime: Generics, new HashMap<>();

modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
```diff
@@ -1450,10 +1469,15 @@ protected Object getAggregateValue(FeatureVisitor visitor, SimpleFeatureType fea
             return null;
         }
         // try to extract an aggregate attribute from the visitor
-        AttributeDescriptor att = extractAggregateAttribute(visitor, featureType);
-        if (att == null && !isCountVisitor(visitor)) {
-            return null; // aggregate function optimization only supported for PropertyName expression
+        Expression aggregateExpression = getAggregateExpression(visitor);
+        // check if aggregation expression is supported
+        if (aggregateExpression instanceof FunctionExpression) {
+            FunctionExpression func = (FunctionExpression)aggregateExpression;
```
aaime: Code conventions (space between cast and variable)

modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
```diff
@@ -1450,10 +1469,15 @@ protected Object getAggregateValue(FeatureVisitor visitor, SimpleFeatureType fea
...
+            if(!getSupportedFunctions().containsKey(func.getFunctionName().getName())) {
```
aaime: Code conventions (space between if and opening parenthesis)

modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
```diff
@@ -3934,23 +3970,33 @@ protected void encodeAttribute(AttributeDescriptor attribute, Query query, Strin
...
         }
-        else {
+        else if (expression instanceof PropertyName){
             if ( att instanceof GeometryDescriptor ) {
                 encodeGeometryColumn((GeometryDescriptor)att, sql,query.getHints());
             }
             else {
                 dialect.encodeColumnName( att.getLocalName(), sql);
             }
+        } else if(expression instanceof FunctionExpression) {
```
aaime: Coding conventions (same as above)

modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
```diff
@@ -3934,23 +3970,33 @@ protected void encodeAttribute(AttributeDescriptor attribute, Query query, Strin
...
+            dialect.encodeFunction(getFunctioName(((FunctionExpression)expression).getName()), att.getLocalName(), sql);
+        } else {
+            sql.append( "*" );
         }
         
         sql.append(")");
     }
     
+    private String getFunctioName(String function) {
```
aaime: Confused, how do we reach here with a function name that is not supported? Wouldn't that be an error?

modules/library/jdbc/src/main/java/org/geotools/jdbc/SQLDialect.java
```diff
@@ -383,6 +384,10 @@ public void registerAggregateFunctions(Map<Class<? extends FeatureVisitor>,Strin
         aggregates.put( SumVisitor.class, "sum");
     }
     
+    public void registerFunctions(Map<String ,String> functions) {
```
aaime: Public method without javadoc

modules/library/jdbc/src/main/java/org/geotools/jdbc/SQLDialect.java
```diff
@@ -1463,4 +1468,18 @@ protected PrimaryKey getPrimaryKey(String typeName) throws IOException {
         SimpleFeatureType featureType = dataStore.getSchema(typeName);
         return dataStore.getPrimaryKey(featureType);
     }
+    
+    protected void encodeFunction( String function, String column, StringBuffer sql ) {
```
aaime: Functions in general have several parameters. See also https://github.com/geotools/geotools/blob/870e3ab8eece49c54186241e74adf6645b925f1c/modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/FilterToSqlHelper.java#L536 for existing support on encoding filter functions (not aggregate functions).

modules/library/main/src/main/java/org/geotools/feature/visitor/SumAreaVisitor.java
```diff
@@ -0,0 +1,80 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ * 
+ *    (C) 2005-2017, Open Source Geospatial Foundation (OSGeo)
```
aaime: Just 2017

modules/library/main/src/main/java/org/geotools/feature/visitor/SumAreaVisitor.java
```diff
@@ -0,0 +1,80 @@
...
+
+import org.geotools.factory.CommonFactoryFinder;
+import org.geotools.filter.IllegalFilterException;
+import org.opengis.feature.Feature;
+import org.opengis.feature.simple.SimpleFeature;
+import org.opengis.feature.simple.SimpleFeatureType;
+import org.opengis.filter.FilterFactory;
+import org.opengis.filter.expression.Expression;
+
+public class SumAreaVisitor extends SumVisitor {
```
aaime: New public class without javadocs

modules/library/main/src/test/java/org/geotools/feature/visitor/SumAreaVisitorTest.java
```diff
@@ -0,0 +1,69 @@
...
+import org.opengis.feature.simple.SimpleFeatureType;
+
+import com.vividsolutions.jts.geom.Coordinate;
+import com.vividsolutions.jts.geom.GeometryFactory;
+import com.vividsolutions.jts.geom.LinearRing;
+import com.vividsolutions.jts.geom.Polygon;
+
+
+@RunWith(Parameterized.class)
+public class SumAreaVisitorTest<T> extends VisitorTestCase<T, T> {
```
aaime: Reformat file, several formatting issues (can be done fully since it's new)

modules/library/main/src/test/java/org/geotools/feature/visitor/SumAreaVisitorTest.java
```diff
@@ -0,0 +1,69 @@
...
+    private static final GeometryFactory GF = new GeometryFactory();
+    
+    public SumAreaVisitorTest(Class<T> valueClass, List<T> values, T expectedValue) {
+        super(valueClass, values, expectedValue);
+    }
+
+    @Parameterized.Parameters
+    public static Collection<Object[]> parameters() {
+
+        return Collections.singleton(new Object[]{Polygon.class, Arrays.asList(polygon( new int[]{ 12,6, 14,8, 16,6, 16,4, 12,4, 12,6} )), 12.0});
```
aaime: Just one test with one polygon, it's not checking for actual aggregation. I'd add a few more tests here, empty set, more than one, maybe see what happens with invalid polygons.

modules/library/main/src/main/java/org/geotools/feature/visitor/SumAreaVisitor.java
aaime: Btw, did not check the other new files, but in case, it would be the same

aaime:
Seeing lots of test and implementation of aggregation in several databases, that's good, which ones have you actually tested? Are the paths of native aggregation and in memory aggregation both tested?

mbarto:
@aaime, I tested (run online tests) all 3 that have native support at the moment (PostGIS, SQLServer, Oracle). I also tested H2 that has no native support. The other ones should behave as H2.

testSumArea automatically checks for native support and tests if in memory or native aggregation happened (looking at the visited flag).

modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
mbarto: You are right, it cannot happen, I will refactor the method

tbarsballe:
This looks ready for merge?

@aaime - any other feedback, or are you happy with the changes?

modules/library/jdbc/src/main/java/org/geotools/jdbc/SQLDialect.java
```diff
@@ -384,6 +385,16 @@ public void registerAggregateFunctions(Map<Class<? extends FeatureVisitor>,Strin
     }
     
     /**
+     * Register the dialect mappings between Geotools FilterFunction names and database
+     * related function.
+     * 
+     * @param functions mappings from GT FilterFunction name to sql function name
+     */
+    public void registerFunctions(Map<String ,String> functions) {
```
jodygarnett: This is a new public method for SQLDialect opening up an ability to extend for client applications; do we need a proposal here to consider this API change?

PR #1724, opened 2017-10-20
[GEOT-5851] Fixed line gap functionality for MBStyle line layers

johngschulz:


modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/layer/LineMBLayer.java
```diff
@@ -533,23 +533,17 @@ public boolean hasLinePattern() {
...
-                    sf.description(Text.text("line"), null), NonSI.PIXEL, gapStroke, lineOffset());
-
-            org.geotools.styling.Stroke blankStroke = sf.stroke(ff.literal(Color.WHITE), lineOpacity(),
-                    ff.subtract(ff.add(lineWidth(), lineGapWidth()), ff.literal(1)),
-                    lineJoin(), lineCap(), null, null);
-            LineSymbolizer blankLine = sf.lineSymbolizer(getId(), null,
-                    sf.description(Text.text("line"), null), NonSI.PIXEL, blankStroke, lineOffset());
-            symbolizers.add(gapOuterLine);
-            symbolizers.add(blankLine);
+        	    Double topOffset = getLineOffset().doubleValue() + (getLineGapWidth().doubleValue() / 2d) + getLineWidth().doubleValue()/2d ;
```
bradh: Possible issue with indentation here?

johngschulz:
I addressed the indentation issue, and updated the commit to include GEOT-5851 in the title, as it appears to be a duplicate of GEOS-8346 and this should fix both.

bradh:
Travis is failing to build the MBStyles module. You might want to take a look at that.

johngschulz:
I've had some issues with timing out on the build locally - but it looks like I forgot to update the visual transformation tests to reflect the new changes.

tbarsballe:
On the GEOS vs GEOT issue, I had incorrectly reported this issue against GEOS instead of GEOT.
I moved the ticket over on Friday, so any confusion there is my fault.

modules/unsupported/mbstyle/src/test/java/org/geotools/mbstyle/transform/StyleTransformTest.java
```diff
@@ -256,6 +256,16 @@ public void testLineGapWidth() throws IOException, ParseException {
...
+        StyledLayerDescriptor sld = mbStyle.transform();
+        SLDTransformer styleTransform = new SLDTransformer();
+        String xml;
+		try {
+			xml = styleTransform.transform(sld);
+	        System.out.print(xml);
+		} catch (TransformerException e) {
+			// TODO Auto-generated catch block
+			e.printStackTrace();
+		}
```
tbarsballe: Looks like tabs are being used here - should be spaces.

johngschulz: I actually meant to remove that before pushing - I was just using that to debug/test the sld output.

tbarsballe:
Looks good. One last request - please include the JIRA ticket number in your commit message. It makes navigating the git history easier.

johngschulz:
Updated commit message with Jira ticket number.

tbarsballe:
Backported to 18.x

PR #1725, opened 2017-10-22
Release improvements

jodygarnett:
* Update tutorials so rename.xml works, they now use 19-SNAPSHOT so that that can be replaced with the release version consistently
* Note the need to fix README links in the developers guide

This change primarily effects the quick-starts, where users will end up using the latest snapshot unless they carefully read the advice to use the stable geotools.version in production. When the documentation is generated for the release the correct tag will be shown.


aaime:
Thanks!

PR #1727, opened 2017-10-27
[GEOT-5852] names with . break GeoPackage Spatial Indexes

ianturton:
GeoPackages with . in name fail when using the Spatial Index with 
[SQLITE_ERROR] SQL error or missing database (no such table: rtree_mosselzaadinvanginstallaties.mosselzaadinvanginstallaties_geom).

This PR correctly escapes the spatial index name.


modules/plugin/geopkg/src/main/java/org/geotools/geopkg/GeoPackage.java
```diff
@@ -1297,8 +1298,10 @@ public boolean hasSpatialIndex(FeatureEntry entry) throws IOException {
             Connection cx = connPool.getConnection();
 
             try {
+                //since this will be a parameter not a table name in this query remove quotes;
+                String tableName = getSpatialIndexName(entry).replace("\"", "");
```
aaime: Instead of doing and then undoing, would it make sense to add a "boolean quoted" parameter to the getSpatialIndexName method? A simple refactoring in the IDE should do the trick, with default to true, and then only this call point gets fixed. Also, what if by any stupid accident one manages to get a quote in the actual name? The above would get rid of it.

ianturton: Turns out this is used in only two places - so I'll just quote the 2nd point instead.

aaime:
Thanks for checking on redistribution :-)

PR #1728, opened 2017-11-03
[GEOT-5855] ImageMosaic granule source may fail when coverage and underlying feature type have different names

aaime:


dromagnoli (approved):
Looks good to me

PR #1729, opened 2017-11-03
[GEOS-8227] - improve whitespace handling in XML Parser

ianturton:
Previously the XML parser ignored CDATA blocks and trimmed all whitespace from XML Strings. The XML spec actually says that elements extending xs:String should preserve **all** whitespace.  See https://www.w3.org/TR/xmlschema-2/#dt-whiteSpace for more details.

This PR introduces the CDATA handling and removes the excessive trimming that was going on.


modules/extension/xsd/xsd-core/src/main/java/org/geotools/xml/Parser.java
```diff
@@ -570,7 +572,8 @@ protected SAXParser parser(boolean validate) throws ParserConfigurationException
         //set hte property to map namespaces to schema locations
         parser.setProperty("http://apache.org/xml/properties/schema/external-schemaLocation",
             schemaLocation.toString());
-
+        //add the handler as a LexivalHandler too.
```
aaime: Formatting ;-)

modules/extension/xsd/xsd-core/src/main/java/org/geotools/xml/Parser.java
```diff
@@ -50,6 +50,8 @@
 import org.xml.sax.SAXException;
 import org.xml.sax.helpers.NamespaceSupport;
 
+import com.sun.org.apache.xerces.internal.impl.Constants;
```
aaime: Wondering out loud here, will this be an issue with Java 9? However I could not find a replacement that is not in com.sun...

modules/extension/xsd/xsd-core/src/main/java/org/geotools/xml/impl/ElementHandlerImpl.java
```diff
@@ -29,6 +29,8 @@
 import org.picocontainer.defaults.DefaultPicoContainer;
 import org.xml.sax.Attributes;
 import org.xml.sax.SAXException;
+import org.xml.sax.ext.LexicalHandler;
```
aaime: Seems like the changes in this file could be omitted

modules/extension/xsd/xsd-core/src/main/java/org/geotools/xml/impl/ParseExecutor.java
```diff
@@ -316,7 +318,11 @@ public XSDNamedComponent getDeclaration() {
                         }
 
                         if (whitespace.getValue() == XSDWhiteSpace.PRESERVE_LITERAL) {
-                            //do nothing
+                            //With GEOS-8227 fix now assumes that text is trimmed before return
```
aaime: Commented out sections here and there?

modules/extension/xsd/xsd-core/src/main/java/org/geotools/xml/impl/ParseExecutor.java
```diff
@@ -328,11 +334,16 @@ public XSDNamedComponent getDeclaration() {
             // for mixed
             if (instance.getTypeDefinition() instanceof XSDComplexTypeDefinition
                     && ((XSDComplexTypeDefinition) instance.getTypeDefinition()).isMixed()) {
-                //collape the text
-                text = Whitespace.COLLAPSE.preparse(text);
+                //Collapse the text
+                //Seems to be an issue for transaction updates - See GEOS-8227
```
aaime: I'd say "Collapse the text, but don't do so for CDATA where it's meant to be preserved"
Reference to GeoServer at this point is probably not required

modules/extension/xsd/xsd-core/src/main/java/org/geotools/xml/impl/ParseExecutor.java
```diff
@@ -328,11 +334,16 @@ public XSDNamedComponent getDeclaration() {
...
+                if(!parser.isCDATA()) {
+                    text = Whitespace.COLLAPSE.preparse(text);
+                }
             }
         }
-
+        if(!parser.isCDATA()) {
```
aaime: Formatting ifs (I do that too, @dromagnoli  catches me every time): ``if (!parser.isCDATA())``

modules/extension/xsd/xsd-core/src/main/java/org/geotools/xml/impl/ParserHandler.java
```diff
@@ -152,6 +153,10 @@
     
     /** context customizer **/
     ContextCustomizer contextCustomizer;
+
+    private boolean inCDATA=false;
```
aaime: Formatting ;-)
`` private boolean inCDATA = false;``

modules/extension/xsd/xsd-core/src/main/java/org/geotools/xml/impl/ParserHandler.java
```diff
@@ -848,4 +856,51 @@ protected void configure(Configuration config) {
...
         return (XSDSchemaLocationResolver[]) l.toArray(new XSDSchemaLocationResolver[l.size()]);
     }
+
+    
+
+
+    @Override
+    public InputSource resolveEntity(String name, String publicId, String baseURI, String systemId)
+            throws SAXException, IOException {
+        if(entityResolver!=null) {
```
aaime: Formatting ;-)

modules/extension/xsd/xsd-core/src/main/java/org/geotools/xs/bindings/XSAnyTypeBinding.java
```diff
@@ -115,7 +115,8 @@ public Object parse(ElementInstance instance, Node node, Object value)
         String text = null;
 
         if ((value != null) && value instanceof String) {
-            text = ((String) value).trim();
+          //GEOS-8227 don't trim text it will already be trimmed unless in CDATA
```
aaime: I'd just remove these comments, people can use history to check what happened and tests should protect the code

modules/extension/xsd/xsd-core/src/main/java/org/geotools/xs/bindings/XSAnyTypeBinding.java
```diff
@@ -138,33 +139,34 @@ public Object parse(ElementInstance instance, Node node, Object value)
         }
 
         //create a map of the elements and attributes
-        Map map = new HashMap();
+        Map<String, Object> map = new HashMap<String, Object>();
```
aaime: ``new HashMap<>();``

modules/extension/xsd/xsd-core/src/main/java/org/geotools/xs/bindings/XSAnyTypeBinding.java
```diff
@@ -138,33 +139,34 @@ public Object parse(ElementInstance instance, Node node, Object value)
...
         List attributes = node.getAttributes();
         List children = node.getChildren();
         mapBinding(map, attributes);
         mapBinding(map, children);
 
         if ((text != null) && !"".equals(text.trim())) {
-            map.put(null, text.trim());
+            //GEOS-8227 don't trim text it will already be trimmed unless in CDATA
```
aaime: Same here, no need to have the whole history of the code in comments

modules/extension/xsd/xsd-core/src/main/java/org/geotools/xs/bindings/XSAnyTypeBinding.java
```diff
@@ -138,33 +139,34 @@ public Object parse(ElementInstance instance, Node node, Object value)
...
+            map.put(null, text/*.trim()*/);
```
aaime: Avoid commented out bits.

modules/extension/xsd/xsd-wfs/src/test/java/org/geotools/wfs/v2_0/bindings/DescribeStoredQueriesTypeBindingTest.java
```diff
@@ -56,7 +56,7 @@ public void testParse() throws Exception {
         assertNotNull(desc);
         
         assertEquals("Hirlam Pressure Grid", desc.getTitle().get(0).getValue());
-        assertEquals("Hirlam forecast model's pressure levels as a grid data encoded in GRIB format.", desc.getAbstract().get(0).getValue());
+        assertEquals("\n            Hirlam forecast model's pressure levels as a grid data encoded in GRIB format.\n        ", desc.getAbstract().get(0).getValue());
```
aaime: This does not seem right, the XML does not have a CDATA section, this text should be trimmed:

````
        <Title>Hirlam Pressure Grid</Title>
        <Abstract>
            Hirlam forecast model&apos;s pressure levels as a grid data encoded in GRIB format.
        </Abstract>
````

File: modules/extension/xsd/xsd-wfs/src/test/resources/org/geotools/wfs/bindings/test-data/QueryTypeBindingTest.xml [hunk 14]
```diff
@@ -18,9 +18,7 @@
 	<wfs:PropertyName>property2</wfs:PropertyName>
 	<wfs:XlinkPropertyName
 		traverseXlinkDepth="*"
-		traverseXlinkExpiry="10">
-		gt:propertyA/gt:propertyB
-	</wfs:XlinkPropertyName>
+		traverseXlinkExpiry="10">gt:propertyA/gt:propertyB</wfs:XlinkPropertyName>
```
aaime: Same here, no need to change, the trimming here should happen

modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/internal/parsers/DescribeFeatureTypeResponseFactory.java
```diff
@@ -40,7 +40,7 @@ public boolean isAvailable() {
     public boolean canProcess(WFSRequest originatingRequest, String contentType) {
         return originatingRequest instanceof DescribeFeatureTypeRequest
                 && (contentType == null || contentType.startsWith("text/xml")
-                || contentType.startsWith("application/gml+xml"));
+                || contentType.startsWith("application/gml+xml")|| contentType.startsWith("application/xml"));
```
aaime: This seems unrelated

modules/extension/xsd/xsd-core/src/main/java/org/geotools/xml/Parser.java
bradh: also spelling LexicalHandler, and hte -> the (a couple of lines up) if you're going to make edits anyway.

modules/extension/xsd/xsd-wfs/src/test/java/org/geotools/wfs/v2_0/bindings/DescribeStoredQueriesTypeBindingTest.java
ianturton: The XML spec actually says that elements extending xs:String should preserve all whitespace. See https://www.w3.org/TR/xmlschema-2/#dt-whiteSpace for more details.

modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/internal/parsers/DescribeFeatureTypeResponseFactory.java
ianturton: It is but for some reason the test was failing here.

modules/extension/xsd/xsd-wfs/src/test/java/org/geotools/wfs/v2_0/bindings/DescribeStoredQueriesTypeBindingTest.java
aaime: Indeed I'm reading the same, but find it surprising :-)
@bencaradocdavies am I missing something here, or is indeed whitespace meant to be preserved there?

modules/extension/xsd/xsd-wfs/src/test/resources/org/geotools/wfs/bindings/test-data/QueryTypeBindingTest.xml
ianturton: I've put the trimming back in - I'm not sure it is 100% correct but it is the expected behaviour.

aaime: It's another xs:string, so the comment above is still valid. Might as well be that it was wrong before, let's hear from @bencaradocdavies

PR #1730, opened 2017-11-03
[GEOT-5858] Allow geometry spatial reference specification in CQL using EWKT syntax

aaime:
https://osgeo-org.atlassian.net/browse/GEOT-5858

@bencaradocdavies you showed interest for the topic, do you have time for a review?

### DIFF
--- a/docs/src/main/java/org/geotools/cql/ECQLExamples.java
+++ b/docs/src/main/java/org/geotools/cql/ECQLExamples.java
@@ -54,8 +54,9 @@ public class ECQLExamples {
                 System.out.println("15 - Temporal During: lastEarthQuake DURING 1700-01-01T00:00:00Z/2011-01-01T00:00:00Z");
                 System.out.println("16 - In predicate: principalMineralResource IN ('silver','oil', 'gold' )");
                 System.out.println("17 - Temporal During using UTC Zone +3: 2006-11-30T01:00:00+03:00 DURING 2006-11-30T00:30:00+03:00/2006-11-30T01:30:00+03:00");
-                
-                System.out.println("0 - quite");
+                System.out.println("18 - Spatial Operation using the contains DE-9IM: RELATE(geometry, SRID=4326;LINESTRING (-134.921387 58.687767, -135.303391 59.092838), T*****FF*)");
+
+                System.out.println("0 - Quit");
                 System.out.print(">");
 
                 String line = reader.readLine();
@@ -115,7 +116,12 @@ public class ECQLExamples {
                 case 17: 
                 	utcTimeZone();
                 	break;
-                default:
+
+                case 18:
+                    referencedRelatePattern();
+                    break;
+
+                    default:
                     System.out.println("invalid option");
                 }
                 System.out.println("Press a key to continue.");
@@ -144,6 +150,25 @@ public class ECQLExamples {
         System.out.println("Result of filter evaluation: " + result);
 		
 	}
+
+    /**
+     * Example using the "contains" intersection matrix.
+     *
+     * @throws Exception
+     */
+    private static void referencedRelatePattern() throws Exception {
+
+        // ecql referencedRelatePattern start
+        Filter filter = ECQL.toFilter("RELATE(geometry, SRID=4326;LINESTRING (-134.921387 58.687767, -135.303391 59.092838), T*****FF*)");
+        // ecql referencedRelatePattern end
+        Utility.prittyPrintFilter(filter);
+
+        SimpleFeature usa = DataExamples.getInstanceOfCountry();
+
+        Boolean result = filter.evaluate(usa);
+        System.out.println("Result of filter evaluation: " + result);
+
+    }
 
 
 	private static void afterPredicateWithLefHandtExpression() throws Exception{
--- a/docs/user/library/cql/ecql.rst
+++ b/docs/user/library/cql/ecql.rst
@@ -5,8 +5,14 @@ The ECQL language is intended as an extension of CQL, thus you can write all pre
 
 References
 
-* `ECQL Parser Design <http://old.geotools.org/ECQL-Parser-Design_110493908.html>`__ (design doc with BNF)
+* `ECQL Parser Design <http://old.geotools.org/ECQL-Parser-Design_110493908.html>`__ (design doc with BNF, note, in addition to WKT syntax for geometries 
+  GeoTools now supports also Extended WKT, same as PostGIS, see example below)
 * `GeoServer CQL Examples <http://docs.geoserver.org/latest/en/user/tutorials/cql/cql_tutorial.html>`_ (geoserver)
+
+.. note::
+
+  Starting with GeoTools 19.0 Geometry objects carrying a CoordinateReferenceSystem among user data get encoded as EWKT.
+  If you do not desire so, set the ``Hints.ENCODE_EWKT`` system hint to false (e..g, ``Hints.putSystemDefault(Hints.ENCODE_EWKT, false);``).
 
 ECQL Utility Class
 ^^^^^^^^^^^^^^^^^^
@@ -133,6 +139,15 @@ Examples
      :language: java
      :start-after: ecql relatePattern start
      :end-before: ecql relatePattern end
+     
+  The following variant shows the same, but giving the geometry a coordinate reference system using the EWKT convention of 
+  preceding it with "SRID=epsgCode;":
+
+  .. literalinclude:: /../src/main/java/org/geotools/cql/ECQLExamples.java
+     :language: java
+     :start-after: ecql referencedRelatePattern start
+     :end-before: ecql referencedRelatePattern end
+
 
 
 * Filter by temporal relation:
--- a/modules/library/cql/src/main/java/org/geotools/filter/text/commons/BuildResultStack.java
+++ b/modules/library/cql/src/main/java/org/geotools/filter/text/commons/BuildResultStack.java
@@ -185,11 +185,11 @@ public final class BuildResultStack {
     public int popIntegerValue() throws CQLException {
         try {
             Literal expr = this.popLiteral();
-            Integer number = (Integer) expr.getValue();
+            Number number = (Number) expr.getValue();
 
             return number.intValue();
         } catch (ClassCastException cce) {
-            throw new CQLException("Expected double");
+            throw new CQLException("Expected integer");
         }
     }
 
--- a/modules/library/cql/src/main/java/org/geotools/filter/text/commons/ExpressionToText.java
+++ b/modules/library/cql/src/main/java/org/geotools/filter/text/commons/ExpressionToText.java
@@ -16,19 +16,12 @@
  */
 package org.geotools.filter.text.commons;
 
-import java.awt.Color;
-import java.text.DateFormat;
-import java.text.SimpleDateFormat;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.Date;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Set;
-import java.util.TimeZone;
-import java.util.TreeSet;
-import java.util.regex.Pattern;
-
+import com.vividsolutions.jts.geom.Geometry;
+import com.vividsolutions.jts.io.WKTWriter;
+import org.geotools.factory.Hints;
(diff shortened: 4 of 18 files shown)


docs/src/main/java/org/geotools/cql/ECQLExamples.java
```diff
@@ -115,7 +116,12 @@ public static void main(String[] args) {
                 case 17: 
                 	utcTimeZone();
```
bencaradocdavies: Pre-existing tabs here could be expanded.

docs/src/main/java/org/geotools/cql/ECQLExamples.java
```diff
@@ -145,6 +151,25 @@ private static void relatePattern() throws Exception{
 		
 	}
 
+    /**
+     * Example using the "contains" intersection matrix.
+     *
+     * @throws Exception
+     */
+    private static void referencedRelatePattern() throws Exception{
```
bencaradocdavies: Should be a space between "Exception" and "{". (Cut and paste from another method in this class, I am sure.)

docs/src/main/java/org/geotools/cql/ECQLExamples.java
```diff
@@ -145,6 +151,25 @@ private static void relatePattern() throws Exception{
...
+
+        // ecql referencedRelatePattern start
+        Filter filter = ECQL.toFilter( "RELATE(geometry, SRID=4326;LINESTRING (-134.921387 58.687767, -135.303391 59.092838), T*****FF*)");
```
bencaradocdavies: Unwanted space between first "(" and '"".

docs/src/main/java/org/geotools/cql/ECQLExamples.java
```diff
@@ -145,6 +151,25 @@ private static void relatePattern() throws Exception{
...
+        // ecql referencedRelatePattern end
+        Utility.prittyPrintFilter(filter);
```
bencaradocdavies: Ugh, typo in method name: `pritty` should be `pretty`, but too late now and not the fault of this PR. Just reviewer griping.

docs/user/library/cql/ecql.rst
```diff
@@ -133,6 +134,15 @@ Examples
      :language: java
      :start-after: ecql relatePattern start
      :end-before: ecql relatePattern end
+     
+  The following variant shows the same, but giving the geometry a coordinate reference system using the EWKT convention of 
+  preceding it with "SRID=epsgCode;":
+
+  .. literalinclude:: /../src/main/java/org/geotools/cql/ECQLExamples.java
```
bencaradocdavies: Good use of this macro: docs from example code.

modules/library/cql/src/main/java/org/geotools/filter/text/commons/BuildResultStack.java
```diff
@@ -185,11 +185,11 @@ public double popDoubleValue() throws CQLException {
     public int popIntegerValue() throws CQLException {
         try {
             Literal expr = this.popLiteral();
-            Integer number = (Integer) expr.getValue();
+            Number number = (Number) expr.getValue();
 
             return number.intValue();
         } catch (ClassCastException cce) {
-            throw new CQLException("Expected double");
+            throw new CQLException("Expected integer");
```
bencaradocdavies: Clearly the exception message mention of "double" was wrong, but why is the type now `Number` not `Integer`?

modules/library/cql/src/main/java/org/geotools/filter/text/commons/ExpressionToText.java
```diff
@@ -168,6 +176,17 @@ public Object visit(Literal expression, Object extraData) {
         Object literal = expression.getValue();
         if (literal instanceof Geometry) {
             Geometry geometry = (Geometry) literal;
+            if (geometry.getUserData() instanceof CoordinateReferenceSystem) {
```
bencaradocdavies: @aaime this is the biggest problem that I have with this PR. If I understand this correctly, every geometry that has a CRS set will be encoded with `SRID=epsgCode`. What if GeoTools is being used to encode CQL that is then passed to third-party software that does not support this extension? If this is the case, this behavioural change risks breaking interoperability. Could application code use `ExpressionToText` for this purpose? Do we need a switch to control this behaviour?

modules/library/cql/src/test/java/org/geotools/filter/text/ecql/ECQLGeoOperationTest.java
```diff
@@ -154,43 +161,42 @@ public void functionAsFirstArgument() throws CQLException {
...
-     */
-    @Ignore
-    public void bboxWithFunctionOnFirstArgument(){
-        
-        //  TODO function on first arg BBOX( buffer( the_geom , 10), 10,20,30,40 )
-        //  TODO function on second argument BBOX( buffer( the_geom , 10), buffer( POINT( 15,15), 10) )
-        //  TODO Geometry on first and second arguments BBOX( POLYGON( .... ), POLYGON( .... ), 10) )
-        //  TODO Proposal in geotools wiki for this
+    @Test
+    public void IntersectsWithReferencedGeometry() throws CQLException, FactoryException {
```
bencaradocdavies: Method name should start with lowercase "i".

modules/library/cql/src/test/java/org/geotools/filter/text/ecql/ECQLLiteralTest.java
```diff
@@ -1,17 +1,5 @@
 package org.geotools.filter.text.ecql;
```
bencaradocdavies: Needs a copyright header.

modules/library/cql/src/test/java/org/geotools/filter/text/ecql/ECQLLiteralTest.java
```diff
@@ -45,40 +46,20 @@ public ECQLLiteralTest(){
...
-                resultFilter instanceof Literal);
-
-        Literal filter = (Literal) resultFilter;
-        Object actualGeometry = filter.getValue();
-        
-        Assert.assertTrue("LineString class was expected",  actualGeometry instanceof LineString);
-
-        assertEqualsGeometries(expectedGeometry, (LineString) actualGeometry);
+        String wkt = "LINESTRING (1 2, 3 4)";
+        assertGeometryParse(wkt, LineString.class, null);
```
bencaradocdavies: You could save much duplication with a method `assertGeometryParse(wkt, geometryClass)` that calls `assertGeometryParse(wkt, geometryClass, null)` and `assertGeometryParse(wkt, geometryClass, 4326)`.

And it would also be nice to make `4326` a constant for clarity.

modules/library/cql/src/test/java/org/geotools/filter/text/ecql/ECQLLiteralTest.java
```diff
@@ -208,19 +127,40 @@ public void geometryCollection()throws Exception{
...
 
+        assertThat(expression, instanceOf(Literal.class));
         Literal literal = (Literal) expression;
         Object actualGeometry = literal.getValue();
-        
-        Assert.assertTrue( actualGeometry instanceof MultiPolygon);
+        assertThat(actualGeometry, instanceOf(expectedGeometryClass));
+
+        if (srid != null) {
+            assertEqualsReferencedGeometries(expectedWkt, (Geometry) actualGeometry, 4326);
```
bencaradocdavies: Now if `4326` was a constant, you could use it here.

bencaradocdavies (changes requested):
@aaime this is a useful improvement. I like it. I had a few tiny nits to pick that you might like to address. I have one substantive concern: could `ExpressionToText` break existing applications by adding `SRID=epsgCode` to CQL geometries with no way to turn it off?

modules/library/cql/src/main/java/org/geotools/filter/text/commons/ExpressionToText.java
aaime: I don't think many apps do round trip CQL (you cannot even round trip a multipoint geometry for example, it expects a different syntax in read and write) but one is @jnh5y GeoMesa. Jim, do you have an opinion on this one?
Ideas:
* Limit it to ECQL (disable in CQL)
* Guard it with a system variable?

jnh5y: Let's see...  in general, I'm for being more clear about what's going on.  To say that a geometry intersects a polygon leaves the context (CRS) unspecified and hence open to interpretation.  

I think either of your suggestions would help.  

For what it is worth, I think Ben's issue is separate from ECQL.toFilter(ECQL.toCQL) and ECQL.toCQL(ECQL.toFilter) each being 'well-behaved'.  (Tersely, the compositions should be the identity function on equivalence classes of CQL strings and GeoTools Filter instances.)

On the plus side, I think that disabling the SRID output in the CQL version of the visitor knocks out Ben's concern.  

For folks who want the 'round-trip' property (or as close as possible), then the ECQL version should work.

docs/src/main/java/org/geotools/cql/ECQLExamples.java
aaime: Yep

docs/src/main/java/org/geotools/cql/ECQLExamples.java
aaime: The file is full of them... I would need a reformat

docs/src/main/java/org/geotools/cql/ECQLExamples.java
aaime: Yep

docs/src/main/java/org/geotools/cql/ECQLExamples.java
aaime: Hehe

modules/library/cql/src/main/java/org/geotools/filter/text/commons/BuildResultStack.java
aaime: Because the code in question was apparently never used, and the parsing actually emits a Long instead of a Integer. Since I was not sure, I switched to the base class for extra safety.

modules/library/cql/src/test/java/org/geotools/filter/text/ecql/ECQLLiteralTest.java
aaime: Yes

modules/library/cql/src/test/java/org/geotools/filter/text/ecql/ECQLGeoOperationTest.java
aaime: Whoops (copy paste, many other methods starting with a capital in that class)

modules/library/cql/src/test/java/org/geotools/filter/text/ecql/ECQLLiteralTest.java
aaime: Constant being added, refactored out method doing the two calls.

jnh5y:
Looks like the force push nuked the discussion about compatibility.  

@aaime, Which approach did you go for?

aaime:
Belts and suspenders:

- CQL does not do EWKT, period
- ECQL has a system hint (linked to an env var) that can be used to turn off EWKT encoding

(many many tests added to check these flags)

docs/src/main/java/org/geotools/cql/ECQLExamples.java
bencaradocdavies: Your discretion. I often give in to temptation and fix formatting adjacent to changes, but this is not required, and mass reformatting is not in general desirable.

modules/library/cql/src/main/java/org/geotools/filter/text/commons/BuildResultStack.java
bencaradocdavies: OK. The exception makes sense when understood as little-i integer, the category not a data type.

bencaradocdavies (approved):

bencaradocdavies:
Thanks @aaime, that looks great. I will paste a docs snipped into the Jira issue because this improvement is worth mentioning in the 19.x release notes.

PR #1613, opened 2017-06-02
[GEOT-5750] Add support for latest MySQL DB enhancements in jdbc-mysql module

nprigour:
Modifications of jdbc-mysql module to support latest MySQL DB  enhancements (version 5.6 and above)
This include mainly support for precise object shape spatial computations (use of ST_xxx() operations). 

In order to be backward compatible support is enabled by appropriate Map parameter which can be set during creation of JDBCDataStore or if not set it is computed based on MySQL DB version. 
 
Signed-off-by: Nikolaos Pringouris <nprigour@gmail.com>

PR #1739, opened 2017-11-13
[GEOT-5750] Add support for latest MySQL DB enhancements (MySql 5.6 or greater)

nprigour:
Pull request to replace  #1613. It includes all changes of  #1613 in a single commit so that it can be easily backported to previous releases of geotools.

### DIFF
--- a/modules/plugin/jdbc/jdbc-mysql/src/main/java/org/geotools/data/mysql/MySQLDataStoreFactory.java
+++ b/modules/plugin/jdbc/jdbc-mysql/src/main/java/org/geotools/data/mysql/MySQLDataStoreFactory.java
@@ -108,8 +108,8 @@ public class MySQLDataStoreFactory extends JDBCDataStoreFactory {
         	//version is at least 5.6.
         	enhancedSpatialSupport = isMySqlVersion56(dataStore);
         } else if (enhancedSpatialFlag && !isMySqlVersion56(dataStore)) {
-        		System.out.println("MySQL version does not support enhancedSpatialSupport. Disabling it");
-        		enhancedSpatialSupport = false;
+        	dataStore.getLogger().info("MySQL version does not support enhancedSpatialSupport. Disabling it");
+        	enhancedSpatialSupport = false;
         }
         
         SQLDialect dialect = dataStore.getSQLDialect();
--- a/modules/plugin/jdbc/jdbc-mysql/src/main/java/org/geotools/data/mysql/MySQLDataStoreFactory.java
+++ b/modules/plugin/jdbc/jdbc-mysql/src/main/java/org/geotools/data/mysql/MySQLDataStoreFactory.java
@@ -17,12 +17,11 @@
 package org.geotools.data.mysql;
 
 import java.io.IOException;
-import java.io.Serializable;
+import java.sql.Connection;
+import java.sql.SQLException;
 import java.util.Collections;
 import java.util.Map;
 
-import org.geotools.data.DataStore;
-import org.geotools.data.DataAccessFactory.Param;
 import org.geotools.data.Parameter;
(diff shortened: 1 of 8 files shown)


modules/plugin/jdbc/jdbc-mysql/src/main/java/org/geotools/data/mysql/MySQLDataStoreFactory.java
```diff
@@ -92,14 +101,58 @@ protected JDBCDataStore createDataStoreInternal(JDBCDataStore dataStore, Map par
...
             storageEngine = (String) STORAGE_ENGINE.sample;
         }
+
+        Boolean enhancedSpatialFlag = (Boolean) ENHANCED_SPATIAL_SUPPORT.lookUp( params );
+        if (enhancedSpatialFlag == null) {
+        	//enhanced spatial support should be enabled if MySQL 
+        	//version is at least 5.6.
+        	enhancedSpatialSupport = isMySqlVersion56(dataStore);
+        } else if (enhancedSpatialFlag && !isMySqlVersion56(dataStore)) {
+        		System.out.println("MySQL version does not support enhancedSpatialSupport. Disabling it");
```
aaime: Eeek, a system out! 
Also, don't use tabs, indentation is done only using spaces.
Was it here already? Did not see it before sorry!

nprigour: Hi @aaime,
I was not aware that it was such a hard constraint to avoid system out since there are quite a lot of them in other modules! Anyway check the latest commit where I have removed it based on your requested change.
By the way, I have and a couple of System.out in the test packages. Is it a requirement to also remove them?

aaime: @nprigour the code base is 15+ years old and there are indeed many bad things in there from a time when "anything went" (system outs, un-closed resources, free formatting for all, eaten exceptions, you name it... wish we could do a "clean up" code sprint, but we systematically fail to sponsor one... we try to clean up things a bit while we edit a certain code area for other reasons).
It test packages normally system outs are commented out prior to commit, or better, replaced with logging.

nprigour:
Ok I also replaced the system out with geotools logging in MySQLSpatialFiltersOnlineTest. It should be OK for merged and back-porting now!!

aaime:
It is not... well, it was not. The code you changed had tabs instead of spaces, the headers for new files w were copy/pasted and reporting the wrong years.
Normally we ask people to fix them, but since you waited for so long on this pull request, I made the fixes myself, manually pushed the changes, and then backported on 18.x and 17.x ;-)

nprigour:
Hi @aaime ,
I do not use tabs. I work in eclipse IDE so the code style and formatting adheres to eclipse buid-in templates and is automatically formatted when I press return. 
Regarding headers indeed I copy pasted them from other files of the mysql module so they might be outdated. 

If there is a code formatting template that I can use (i.e. in udig where I am also a contributor there is one) please let me now.

bencaradocdavies:
@nprigour as mentioned in the [Coding Style page in the GeoTools Developer Guide](http://docs.geotools.org/latest/developer/conventions/code/style.html), we provide an Eclipse formatter configuration in `build/eclipse/formatter.xml`. The default Eclipse formatter is quite different (the tabs @aaime mentioned). Our conventions are based on the Sun Java conventions but with a few adjustments, including using spaces not tabs and wider lines.

nprigour:
Hi @aaime,

In the corresponding JIRA issue (and in your comment above) you mention that you have also back-ported this to v17.4 (v17.x) but checking the corresponding branch it seems that this is not the case.

PR #1742, opened 2017-11-15
[GEOT-5862] sets border opacity to zero for mbstyle fill layers using fill-pattern

johngschulz:
…attern mbstyle


tbarsballe:
Looks good to me, seems like existing test cases already cover this functionality.

PR #1743, opened 2017-11-15
[GEOT-5870] - gt-svg: make renderer.lite.DrawTest pass on Windows

don-vip:
This PR fixes all problems encountered in https://osgeo-org.atlassian.net/browse/GEOT-5870:

- DrawTest.testParametricOnlyFill fails on Windows, because it is not possible to open a file URL with a query part
- DrawTest.testParametricAllValues fails on Windows for the same reason
- DrawTest.testRenderingHints does nothing useful (no assert) except polluting the console with stack traces (NPE in Batik). Now the test properly checks its output, like all other tests.
- I was unable to run the test from Eclipse, because TestData relies on specific ClassLoader behavior. It seems Maven and Eclipse behave differently, so I had to rename square.properties file to avoid conflict with another file named identically.

I also have added a listener to all tests, so that errors are not silently ignored.


aaime:
Yep, looks good. Thank you.

aaime:
Backported to 18.x and 17.x too

don-vip:
Thank you! :)

PR #1744, opened 2017-11-16
[GEOT-5872]: PostGIS HSTORE column reading support

dromagnoli:


modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/HStoreConverterFactory.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+package org.geotools.data.postgis;
+
+import java.util.Map;
+
+import org.geotools.factory.Hints;
+import org.geotools.util.Converter;
+import org.geotools.util.ConverterFactory;
+
+public class HStoreConverterFactory implements ConverterFactory {
```
aaime: Public class without javadoc

modules/plugin/jdbc/jdbc-postgis/src/test/java/org/geotools/data/postgis/PostGISHStoreTestSetup.java
```diff
@@ -0,0 +1,51 @@
...
+
+    @Override
+    protected void setUpData() throws Exception {
+        dropTestHStoreTable();
+        createTestHStoreTable();
+    }
+
+    private void createTestHStoreTable() throws Exception {
+        String sql = "CREATE TABLE \"hstoretest\" ("
+                + "\"id\" INT, \"name\" VARCHAR, \"mapping\" HSTORE, PRIMARY KEY(id))";
```
aaime: I believe one has to run "create extension hstore if not exists" in the database to have hstore run correctly, I'd add that to the setup commands (and it should be there pretty much always, it's an old one and afaik it's built-in but not enabled by default).

dromagnoli: Right. I think I did that on PGAdmin to play with hstore before doing any coding.
Going to add that.

aaime:
Looks good thanks

bencaradocdavies:
Causes: [[GEOT-5876] PostGISHStoreOnlineTest fails when database user not superuser](https://osgeo-org.atlassian.net/browse/GEOT-5876)

bencaradocdavies:
Fixture configurations for non-superuser database users can be fixed by the one-time procedure of connecting to the database as a superuser and running "create extension hstore;"

PR #1745, opened 2017-11-17
[GEOT-5861] fixed membership mbfilters to use PropertyIsEqualTo compa…

johngschulz:
…rison operator

And updated test to reflect new change.


tbarsballe:
Backported to 18.x

PR #1754, opened 2017-11-23
[GEOT-5885] GML 3.2 encoding misses mandatory gml:id on geometry elements - fast path

aaime:
https://osgeo-org.atlassian.net/browse/GEOT-5885

### DIFF
--- a/modules/extension/xsd/xsd-gml3/src/main/java/org/geotools/gml3/simple/CurveEncoder.java
+++ b/modules/extension/xsd/xsd-gml3/src/main/java/org/geotools/gml3/simple/CurveEncoder.java
@@ -60,8 +60,12 @@ class CurveEncoder extends GeometryEncoder<LineString> {
         this.arcString = ARC_STRING.derive(gmlPrefix, gmlUri);
     }
 
-    public void encode(LineString geometry, AttributesImpl atts, GMLWriter handler)
+    public void encode(LineString geometry, AttributesImpl atts, GMLWriter handler, String gmlId)
             throws Exception {
+        if ( gmlId != null) {
+            atts = cloneWithGmlId(atts, gmlId);
+        }
+        
         handler.startElement(curve, atts);
         handler.startElement(segments, null);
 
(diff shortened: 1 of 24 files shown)


bencaradocdavies:
I will try a local merge to check that GeoServer app-schema tests still pass.

modules/extension/xsd/xsd-gml3/src/main/java/org/geotools/gml3/simple/CurveEncoder.java
```diff
@@ -60,8 +60,12 @@ protected CurveEncoder(Encoder e, String gmlPrefix, String gmlUri) {
         this.arcString = ARC_STRING.derive(gmlPrefix, gmlUri);
     }
 
-    public void encode(LineString geometry, AttributesImpl atts, GMLWriter handler)
+    public void encode(LineString geometry, AttributesImpl atts, GMLWriter handler, String gmlId)
```
bencaradocdavies: Public API change without backwards compatibility. Please add a method
```
public void encode(LineString geometry, AttributesImpl atts, GMLWriter handler)
```
that calls
```
encode(geometry, atts, handler, null)
```

bencaradocdavies: Oh, I see, you did add the backwards-compatible `encode`, pulled up into the base class.

modules/extension/xsd/xsd-gml3/src/test/java/org/geotools/gml3/simple/CurveEncoderTest.java
```diff
@@ -29,22 +30,24 @@ public void testEncodeCircle() throws Exception {
         CurveEncoder encoder = new CurveEncoder(gtEncoder, "gml", GML.NAMESPACE);
         Geometry geometry = new WKTReader2()
                 .read("CIRCULARSTRING(-10 0, -8 2, -6 0, -8 -2, -10 0)");
-        Document doc = encode(encoder, geometry);
-        // XMLTestSupport.print(doc);
+        Document doc = encode(encoder, geometry, "circle.abc");
+        // MLTestSupport.print(doc);
```
bencaradocdavies: I think you lost an "X" in this commented-out line.

PR #1759, opened 2017-11-26
Maven Javadoc module: Activate JDK8 profiles only with Java 8

don-vip:
I try to build GeoTools with Java 9. The first build failure is in Maven Javadoc module, as it relies on the existence of `tools.jar` in JDK directory. This is no longer true starting from JDK9, so the current profiles must be restricted to JDK8. The obsolete pre-JDK7 profile has been dropped.

### DIFF
--- a/build/maven/javadoc/pom.xml
+++ b/build/maven/javadoc/pom.xml
@@ -92,34 +92,12 @@
 
     <profiles>
         <!-- =========================================================== -->
-        <!--     Sun's tools dependency (pre JDK 1.7)                    -->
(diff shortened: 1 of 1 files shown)


don-vip:
Next Java 9 failure is in `org.geotools.styling.SLDTransformerTest` and linked to http://java9.wtf/xml-transformer/ Thank you @nicolaiparlog (again !) for this initiative, it will be helpful.

don-vip:
Thank you! For the next failure I have created https://osgeo-org.atlassian.net/browse/GEOT-5900 as I don't see how to fix it.

PR #1760, opened 2017-11-26
[GEOT-5878] Very long symbols can cause an integer overflow when rendered

ianturton:
Adds a check to see if the size of the graphic * 3 will overflow MAX_INTEGER, prints a warning instead of a NegativeArraySizeException and Stacktrace.

### DIFF
--- a/modules/library/render/src/main/java/org/geotools/renderer/style/SLDStyleFactory.java
+++ b/modules/library/render/src/main/java/org/geotools/renderer/style/SLDStyleFactory.java
@@ -1273,32 +1273,39 @@ public class SLDStyleFactory {
    
         double size = evalToDouble(gr.getSize(), feature, 16);
         final double sizeX = size * shapeAspectRatio; // apply the aspect
-        												// ratio to fix the
-        												// sample's width.
+      	 					      // ratio to fix the
+        					      // sample's width.
         final double sizeY = size;
         
         // we need to paint the mark in a 3x3 grid to account for border effects
         // due to antialiasing (e.g., even if the mark is 10 pixels wide, due to the 
         // antialiasing graphically it occupies 12 or so pixels)
-        image = new BufferedImage((int) Math.ceil(sizeX * 3), (int) Math
-        		.ceil(sizeY * 3), BufferedImage.TYPE_INT_ARGB);
-        Graphics2D g2d = image.createGraphics();
-        g2d.setRenderingHints(renderingHints);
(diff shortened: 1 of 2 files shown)


ianturton:
@aaime only real question is what to return if we can't draw the graphic requested, I've gone with a small blank image, what do you think?

modules/library/render/src/main/java/org/geotools/renderer/style/SLDStyleFactory.java
```diff
@@ -1273,32 +1273,39 @@ private BufferedImage markToTilableImage(org.geotools.styling.Graphic gr, Object
...
-        double rotation = Math.toRadians(evalToDouble(gr.getRotation(), feature, 0.0)); // fix for GEOS-6217
-        for (int i = -1; i < 2; i++) {
-        	for (int j = -1; j < 2; j++) {
-        		double tx = sizeX * 1.5 + sizeX * i;
-        		double ty = sizeY * 1.5 + sizeY * j;
-        		fillDrawMark(g2d, tx, ty, mark, size, rotation, feature);
-        	}
+        
+        //check if this will cause an overflow in the image creation
+        if ((sizeX * 3) * (sizeY * 3) > Integer.MAX_VALUE) {
```
aaime: Since the 3x3 pattern is built to avoid small but visible artifacts in tiled polygon filling, I'd do this:

- If the sizeX * sizeY still < Integer.MAX_VALUE, return the symbol as is and print a warning
- If not even that, ok to return nothing and suggest them to go see a doctor :-)

modules/library/render/src/main/java/org/geotools/renderer/style/SLDStyleFactory.java
```diff
@@ -1255,57 +1255,66 @@ TexturePaint getTexturePaint(org.geotools.styling.Graphic gr, Object feature,
...
+        // check if this will cause an overflow in the image creation GEOT-5878
+        int repeat = 3;
+        if ((sizeX * repeat) * (sizeY * repeat) > Integer.MAX_VALUE) {
+            LOGGER.warning("Size of graphic (" + sizeX + " * " + sizeY + ") is too large");
+            if (sizeX * sizeY > Integer.MAX_VALUE) {
+                LOGGER.severe("Size of graphic (" + sizeX + " * " + sizeY
+                        + ") is too large will not draw");
+                return new BufferedImage(10, 10, BufferedImage.TYPE_INT_ARGB);
+            } else {
+                LOGGER.fine("Size of metatiled graphic (" + (3 * sizeX) + " * " + (3 * sizeY)
```
bradh: Use `repeat` in here?

PR #1761, opened 2017-11-28
Encode in functions

aaime:
I had the doubt of whether to allow encoding of "in" functions in sqlserver and oracle since they don't accept them as expressions like other databases but... I don't think it's worth throwing away the good use case (in(att, v1, v2, v3)=true) because those database don't support the one that does not make sense (in(att, v1, v2, v3) > false),  even if evaluated in memory it actually works.

### DIFF
--- a/modules/library/jdbc/src/main/java/org/geotools/data/jdbc/FilterToSQL.java
+++ b/modules/library/jdbc/src/main/java/org/geotools/data/jdbc/FilterToSQL.java
@@ -25,6 +25,7 @@ import org.geotools.filter.FilterCapabilities;
 import org.geotools.filter.FunctionImpl;
 import org.geotools.filter.LikeFilterImpl;
 import org.geotools.filter.capability.FunctionNameImpl;
+import org.geotools.filter.function.InFunction;
 import org.geotools.jdbc.JDBCDataStore;
 import org.geotools.jdbc.JoinPropertyName;
 import org.geotools.jdbc.PrimaryKey;
@@ -95,6 +96,7 @@ import org.opengis.filter.temporal.OverlappedBy;
 import org.opengis.filter.temporal.TContains;
 import org.opengis.filter.temporal.TEquals;
 import org.opengis.filter.temporal.TOverlaps;
+import org.opengis.parameter.Parameter;
 import org.opengis.referencing.crs.CoordinateReferenceSystem;
 import org.opengis.referencing.crs.GeographicCRS;
 import org.opengis.temporal.Instant;
@@ -238,6 +240,9 @@ public class FilterToSQL implements FilterVisitor, ExpressionVisitor {
     /** inline flag, controlling whether "WHERE" will prefix the SQL encoded filter */
     protected boolean inline = false;
 
+    /** Whether the encoder should try to encode "in" function into a SQL IN operator */
+    protected boolean inEncodingEnabled = true;
+
     /**
      * Default constructor
      */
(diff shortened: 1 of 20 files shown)


modules/library/jdbc/src/main/java/org/geotools/data/jdbc/FilterToSQL.java
```diff
@@ -260,6 +265,22 @@ public void setInline(boolean inline) {
         this.inline = inline;
     }
 
+    /**
+     * Whether "in" like functions are encoded in SQL or not
+     * @return
```
nmco: ``@return`` description is missing.

modules/library/jdbc/src/main/java/org/geotools/data/jdbc/FilterToSQL.java
```diff
@@ -746,6 +770,21 @@ protected void visitBinaryComparisonOperator(BinaryComparisonOperator filter, Ob
 
         Expression left = filter.getExpression1();
         Expression right = filter.getExpression2();
+
+        // shortcut for IN case, to have a more natural encoding for the common/sane case of 
+        // "in(property, v1, v2, ... vn) = true" into "property in (v1, v2, ..., vn) 
+        if ("=".equals(extraData) || "!=".equals(extraData)) {
+            if (left instanceof Function && right instanceof Literal && InFunction.isInFunction((Function) left) &&
```
nmco: Method ``isInFunction`` could accept an ``Expression`` directly and perform the ``instance of`` internally.

modules/library/jdbc/src/main/java/org/geotools/data/jdbc/FilterToSQL.java
```diff
@@ -746,6 +770,21 @@ protected void visitBinaryComparisonOperator(BinaryComparisonOperator filter, Ob
 
         Expression left = filter.getExpression1();
         Expression right = filter.getExpression2();
+
+        // shortcut for IN case, to have a more natural encoding for the common/sane case of 
+        // "in(property, v1, v2, ... vn) = true" into "property in (v1, v2, ..., vn) 
+        if ("=".equals(extraData) || "!=".equals(extraData)) {
```
nmco: This should not also check if ``inEncodeEnable`` is set to ``TRUE`` ? As far I can see ``encodeInComparison`` will directly encode the IN clause. I'm sure I miss something but is not clear how we differentiate for the use case were we need to encode or not the IN SQL caluse.

modules/library/main/src/main/java/org/geotools/filter/function/InFunction.java
```diff
@@ -41,6 +43,44 @@
...
+                || function instanceof FilterFunction_in6
+                || function instanceof FilterFunction_in7
+                || function instanceof FilterFunction_in8
+                || function instanceof FilterFunction_in9
+                || function instanceof FilterFunction_in10;
+    }
+
+    /**
+     * Returns filter capabilities for all the "in" functions
+     * @return
```
nmco: ``@return`` misses a description.

modules/library/jdbc/src/main/java/org/geotools/data/jdbc/FilterToSQL.java
aaime: Will add

modules/library/jdbc/src/main/java/org/geotools/data/jdbc/FilterToSQL.java
aaime: Good catch

modules/library/jdbc/src/main/java/org/geotools/data/jdbc/FilterToSQL.java
aaime: Hmm.. yep

modules/library/main/src/main/java/org/geotools/filter/function/InFunction.java
aaime: Will add

PR #1763, opened 2017-12-03
Upgrade geobuf's protobuf dependency from 2 to 3

jericks:
The unsupported Geobuf module depends on an old version of Google's protobuf library. I would like to upgrade from 2.x to 3.x because the 3.x versions are binary compatible with all other 3.x versions. 

JIRA:
https://osgeo-org.atlassian.net/browse/GEOT-5901


bencaradocdavies:
@jericks after this merge, `Geobuf.java` no longer compiles in Eclipse Oxygen.1 (4.7.1). See [GEOT-5910](https://osgeo-org.atlassian.net/browse/GEOT-5910).

PR #1764, opened 2017-12-04
[GEOT-5902] Fixed Encoder setters

awaterme:
Pull request for [GEOT-5902](https://osgeo-org.atlassian.net/browse/GEOT-5902)


aaime:
Found this change breaks a test in GeoServer. Merging and fixing on the GeoServer side.

aaime:
Backported too (too forever to run all builds to check it was actually good)

PR #1769, opened 2017-12-11
Downgrade protobuf to 3.4.0 to fix eclipse compiling issues

jericks:
The upgrade to protobuf 3.5.0 was causing problems for the Eclipse compiler. Downgrading to 3.4.0 works with maven and eclipse.
  
https://osgeo-org.atlassian.net/browse/GEOT-5910


bencaradocdavies:
Thanks very much @jericks that fixes the problem.

PR #1773, opened 2017-12-15
[GEOT-5904] Fixed ClassCastException when writing a Boolean to geopac…

vickdw:
…kage.  Fix for https://osgeo-org.atlassian.net/browse/GEOT-5904


modules/plugin/geopkg/src/test/java/org/geotools/geopkg/GeoPackageTest.java
```diff
@@ -86,12 +58,24 @@
...
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.Statement;
+import java.text.SimpleDateFormat;
+import java.util.*;
+import java.util.List;
+import java.util.logging.Level;
+
+import static org.junit.Assert.*;
```
tbarsballe: As per the style guide, please avoid using * imports. (also, looks like there is a bunch of unnecessary import changes in general - that is probably something you can turn off in your IDE)

aaime:
Since it's shared test data, can we have a less obnoxious name than "ogrgeojson_8b0dc300" ;-)
Besides that, there is a lot of import reformats, IntelliJ gets me as well with those, locally I disabled automatic import optimization to avoid it.

vickdw:
I just fixed those am working on squashing commits right now.

David Vick

Software Developer | Boundless <http://www.boundlessgeo.com>
dvick@boundlessgeo.com
mobile: +1-636-698-3174

vickdw:
yes, I can change the name.

David Vick

Software Developer | Boundless <http://www.boundlessgeo.com>
dvick@boundlessgeo.com
mobile: +1-636-698-3174

modules/plugin/geopkg/src/main/java/org/geotools/geopkg/GeoPackage.java
```diff
@@ -662,8 +644,15 @@ public void add(FeatureEntry entry, SimpleFeatureCollection collection) throws I
                     SimpleFeature f = it.next();
                     SimpleFeature g = w.next();
                     for (PropertyDescriptor pd : collection.getSchema().getDescriptors()) {
+                        /* geopkg spec requires booleans to be stored as SQLite integers this fixes
+                         * bug reported by https://osgeo-org.atlassian.net/browse/GEOT-5904 */
                         String name = pd.getName().getLocalPart();
-                        g.setAttribute(name, f.getAttribute(name));
+                        if (pd.getType().getBinding() == Boolean.class) {
+                            int bool = (Boolean)(f.getAttribute(name)) ? 1 : 0;
```
tbarsballe: Few questions about this change:
1. What happens if the value is null? You probably need a null check here.
2. What happens if you get a list of Features from a GeoPackage store? Will they be converted back to Booleans?
3. Is it possible for there to be GeoPackage files written by GeoTools before this fix that contain boolean values, and does GeoTools still read these correctly?
4. Does this affect the GeoServer GeoPackage output format; does this bug show up there?

modules/plugin/geopkg/src/main/java/org/geotools/geopkg/GeoPackage.java
```diff
@@ -662,8 +644,15 @@ public void add(FeatureEntry entry, SimpleFeatureCollection collection) throws I
                     SimpleFeature f = it.next();
                     SimpleFeature g = w.next();
                     for (PropertyDescriptor pd : collection.getSchema().getDescriptors()) {
+                        /* geopkg spec requires booleans to be stored as SQLite integers this fixes
+                         * bug reported by https://osgeo-org.atlassian.net/browse/GEOT-5904 */
```
tbarsballe: The full link to the issue is unnecessary, just the ticket ID (GEOT-5904) is fine; eg:
> geopkg spec requires booleans to be stored as SQLite integers; this fixes GEOT-5904

modules/plugin/geopkg/src/test/java/org/geotools/geopkg/GeoPackageTest.java
```diff
@@ -790,6 +788,25 @@ URL setUpShapefile() throws Exception {
...
+    URL setUpBoolShapefile() throws Exception {
+        File d = File.createTempFile("ogrgeojson_8b0dc300", "shp", new File("target"));
+        d.delete();
+        d.mkdirs();
+
+        String[] exts = new String[]{"shp", "shx", "dbf", "prj"};
+        for (String ext : exts) {
+            if("prj".equals(ext)) {
+                String wkt = CRS.decode("EPSG:4326", true).toWKT();
+                FileUtils.writeStringToFile(new File(d, "ogrgeojson_8b0dc300.prj"), wkt);
```
tbarsballe: There seems to be a bunch of unneccessary manual work (Decoding the WKT, creating the projection file, etc.) going on here. Can't you just copy all the test data to the temp folder?

modules/plugin/geopkg/src/main/java/org/geotools/geopkg/GeoPackage.java
vickdw: 1. Null value check added and tested.
2.Values from a geopackage, according to the spec, will already be of type Integer and will remain that way when being written to another geopackage.
3. While it was possible for GeoTools to write GeoPackages containing Boolean fields, the data in those fields would have been Integers and not Strings.  This fix allows GeoTools to convert a Boolean in String form either 'T' 'True' 'true' etc.. to 1 or 0 as integers.
4. This bug was found by exporting a Layer that was originally stored as a shapefile as a geopackage.  In tracing the process, the problem was discovered in the GeoTools GeoPackage.java class where the fix now resides.  This fix as been tested with a running instance of GeoServer and proven to resolve the issue.

tbarsballe:
Changes look good. Currently waiting on Travis CI to verify.

mrcnc:
👍 Thanks @vickdw

tbarsballe:
Backported to 18.x (Ran into compile issues when trying to backport it further, if you want this on 17.x please create a seperate pull request)

PR #1774, opened 2017-12-17
 [GEOT-5919] Disable by default H2 server mode, allow users to choose it as an option

aaime:


aaime:
I haven't seen the random failures in some time now, I'd rather close this one (it won't be lost if we need it) instead of spending more time on it.

aaime:
One more failure occurred, changing the pull request to satisfy formal requirements and merging

aaime:
Well, did not work, the first build triggered by this change failed the usual way...

PR #1777, opened 2017-12-21
[GEOT-5916]: NoData to Alpha may generate N alpha channels

dromagnoli:


modules/library/coverage/src/test/java/org/geotools/image/ImageWorkerTest.java
```diff
@@ -1694,7 +1695,39 @@ public void testMosaicBackgroundColor() {
...
     }
-    
+
+    @Test
+    public void testPrepareRenderingAlphaOnNoData() {
+        BufferedImage red = getSyntheticRGB(Color.RED);
+        final int w = red.getWidth();
+        final int h = red.getHeight();
+
+        // Setting some image pixels to all zeros
```
aaime: The test seems to be a bit weak, what about a combination of pixels that are zero only in one band, and/or in two bands,  and then in three bands, and making sure that only those with 3 zeroes have alpha set?

dromagnoli:
Tests updated with some more corner cases having different combinations of 1 or 2 different bands being set to zero.
Waiting for travis build before merging

aaime:
Thanks, +1

PR #1779, opened 2017-12-24
[GEOT-5666] Avoid parsing the LookAt as the placemark geometry.

bradh:
From https://osgeo-org.atlassian.net/browse/GEOT-5666  - fix LookAt coordinate order and avoid parsing LookAt as the placemark geometry.

### DIFF
--- a/modules/extension/xsd/xsd-kml/src/main/java/org/geotools/kml/bindings/LookAtTypeBinding.java
+++ b/modules/extension/xsd/xsd-kml/src/main/java/org/geotools/kml/bindings/LookAtTypeBinding.java
@@ -97,13 +97,13 @@ public class LookAtTypeBinding extends AbstractComplexBinding {
         Coordinate c = new Coordinate();
 
         //&lt;element default="0" minOccurs="0" name="longitude" type="kml:angle180"/&gt;
-        c.y = ((Double) node.getChildValue("longitude", Double.valueOf(0d))).doubleValue();
+        c.x = (Double) node.getChildValue("longitude", 0d);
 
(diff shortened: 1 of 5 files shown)


bradh:
From https://osgeo-org.atlassian.net/browse/GEOT-5666

aaime:
Hi Brad, I think it looks good, I just have a doubt about coordinate order (I wasn't the original author of this module).

What CRS is attached to the geometry definitions? Is it ``DefaultGeographicCRS.WGS84`` or ``CRS.decode("EPSG:4326", true)`` then it's good to force ordinates in eart/north order.
If instead it's CRS.decode("EPSG:4326") then it depends on how GeoTools is configured.
Long story short, some check to get a match between the returned geometry CRS order and the ordinates order should be made.

Looking around it seems the CRS is simply not defined in FeatureTypes, if that's indeed the case I'd be inclined to move forward as you did, with no reference in memory probably one should follow the native order of the format.

Wondering if @jdeolive has any hint/info to share? :-)

bradh:
For the coordinate order, I don't see a specified CRS. I decided to go with consistency with the coordinate order we use for the placemark geometry (e.g. https://github.com/geotools/geotools/blob/master/modules/extension/xsd/xsd-kml/src/test/java/org/geotools/kml/v22/KMLParsingTest.java#L44).

Would you like to see a Jira ticket / follow-up PR to be explicit about the CRS?

aaime:
@bradh the follow up ticket might be a good idea

bradh:
Added ticket as https://osgeo-org.atlassian.net/browse/GEOT-5922

PR #1783, opened 2018-01-04
Add tolerance to CRSTest coordinate comparaisons to support Java 9

tbarsballe:
Java 9 contains a new native math library that is faster and more precise, but not backwards compatible with results produces using Java 8 and older.
A tolerance allows us to continue supporting both language versions.

Refer to the mailing list discussion for more details.


aaime:
@tbarsballe looks good to me. Just a question (curiosity of mine), why did you both add a tolerance and changed the values?

tbarsballe:
Since the Java 9 math implementation is ostensibly more precise, I changed the values to match that, then added tolerance so the tests would still pass on Java 8. The value change was not technically necessary, but could be helpful if we eventually drop Java 8 support in the future and remove the tolerance.

aaime:
+1

PR #1785, opened 2018-01-09
NullPointerException clearing coverages in NetCDFReader.dispose

maka-io:
https://osgeo-org.atlassian.net/browse/GEOT-5923

NetCDFReader references a coverage list without checking for null, causing the access cleanup portion of dispose() to be skipped (line 696).
Move "coverages.clear()" into the conditional check that coverages is not null.
```sh
java.lang.NullPointerException
at org.geotools.coverage.io.netcdf.NetCDFReader.dispose(NetCDFReader.java:696)
...
```


bencaradocdavies:
Thanks @maka-io, tidy fix. This one is also eligible for backport to stable and maintenance.

bencaradocdavies:
Backported to 18.x and 17.x.

maka-io:
Hey @bencaradocdavies , do you have an idea of when the backport will be released?

PR #1790, opened 2018-01-15
Make SLDTransformer test compatible with Java <= 8 and Java 9

tbarsballe:
Java 8 and Java 9 do pretty-print slightly differently (The XML is still functionally identical); this makes the SLDTranformer tests function at both language levels.

### DIFF
--- a/modules/library/main/src/test/java/org/geotools/styling/SLDTransformerTest.java
+++ b/modules/library/main/src/test/java/org/geotools/styling/SLDTransformerTest.java
@@ -1114,10 +1114,12 @@ public class SLDTransformerTest {
         intString.add(Locale.CANADA_FRENCH, "titre");
         rule.getDescription().setTitle(intString);
         String xml = transformer.transform(rule);
-        assertTrue(xml.contains("<sld:Title>title"));
-        assertTrue(xml.contains("<sld:Localized lang=\""+Locale.ITALIAN.toString()+"\">titolo</sld:Localized>"));
-        assertTrue(xml.contains("<sld:Localized lang=\""+Locale.FRENCH.toString()+"\">titre</sld:Localized>"));
-        assertTrue(xml.contains("<sld:Localized lang=\""+Locale.CANADA_FRENCH.toString()+"\">titre</sld:Localized>"));
+        Document doc = buildTestDocument(xml);
+
+        assertXpathEvaluatesTo("title", "normalize-space(//sld:Title/text()[1])", doc);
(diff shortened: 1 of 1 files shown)


modules/library/main/src/test/java/org/geotools/styling/SLDTransformerTest.java
```diff
@@ -1392,9 +1396,15 @@ public void testLabelCDataEnd() throws Exception {
...
-                + "                        </sld:Label>"));
+
+        assertTrue(xml.contains("<sld:Label>"));
+        assertTrue(xml.contains("<![CDATA[abc ]]>"));
+        assertTrue(xml.contains("<ogc:PropertyName>myProperty</ogc:PropertyName>"));
+        assertTrue(xml.contains("</sld:Label>"));
+
+        assertTrue(xml.indexOf("<sld:Label>") < xml.indexOf("<![CDATA[abc ]]>"));
+        assertTrue(xml.indexOf("<![CDATA[abc ]]>") < xml.indexOf("<ogc:PropertyName>myProperty</ogc:PropertyName>"));
+        assertTrue(xml.indexOf("<ogc:PropertyName>myProperty</ogc:PropertyName>") < xml.indexOf("</sld:Label>"));
```
jodygarnett: that is one way to do it - nice work :)

modules/library/main/src/test/java/org/geotools/styling/SLDTransformerTest.java
```diff
@@ -1405,10 +1415,28 @@ public void testLabelCDataMid() throws Exception {
         StyledLayerDescriptor sld = buildSLDAroundSymbolizer(ts);
         
         String xml = transformer.transform(sld);
-        assertTrue(xml.contains("<sld:Label><![CDATA[a  bc]]>" + NEWLINE +
-                "                            <ogc:PropertyName>myProperty</ogc:PropertyName>" + NEWLINE
-                + "                        </sld:Label>"));
+        Document doc = buildTestDocument(xml);
+
+        //assertXpathEvaluatesTo("a  bc", "normalize-space(//sld:Label/text()[1])", doc);
```
jodygarnett: can we remove commented out test?

tbarsballe: So, that is xmlunit being wrong. If xmlunit gets fixed, we ought to be able to uncomment it; I'll add a comment to make this more clear.

PR #1791, opened 2018-01-18
GEOT-5925: updated jai-ext to 1.0.19

mbarto:
This is needed to solve: https://github.com/geosolutions-it/jai-ext/issues/175


dromagnoli:
+1 here.

PR #1794, opened 2018-01-22
GEOT-5931: Temporal operators in opengis FilterCapabilities implementation and wrapper

sjudeng:
Update opengis FilterCapabilities implementation and wrapper to support temporal operators. Also fix a bug with `BegunBy` in the IsSupportedFilterVisitors which would cause the operator to never be supported.

*CLA has been sent to OSGeo*

### DIFF
--- a/modules/library/main/src/main/java/org/geotools/filter/Capabilities.java
+++ b/modules/library/main/src/main/java/org/geotools/filter/Capabilities.java
@@ -2,7 +2,7 @@
  *    GeoTools - The Open Source Java GIS Toolkit
  *    http://geotools.org
  * 
- *    (C) 2002-2015, Open Source Geospatial Foundation (OSGeo)
+ *    (C) 2002-2018, Open Source Geospatial Foundation (OSGeo)
  *
  *    This library is free software; you can redistribute it and/or
  *    modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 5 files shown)


modules/library/main/src/main/java/org/geotools/filter/Capabilities.java
```diff
@@ -2,7 +2,7 @@
  *    GeoTools - The Open Source Java GIS Toolkit
  *    http://geotools.org
  * 
- *    (C) 2002-2015, Open Source Geospatial Foundation (OSGeo)
+ *    (C) 2002-2018, Open Source Geospatial Foundation (OSGeo)
```
aaime: Hi, for next time, we have a resolution stating that there is no need to update the copyright header dates (we checked, has no legal value). The only requirement is to add headers on new files, or old ones that are missing them.

aaime:
The change looks good, the CLA has not been added in the repository yet, I've inquired in OSGeo.

aaime:
So there is a new CLA, but I cannot match your nickname (it's all I have) with it. Can you share name and surname please? :-)

aaime:
Finally found the CLA, sorry for the wait! Going to do a full GT/GS build locally

aaime:
Merged, thanks for you patience, next time we already have the CLA ready :)

PR #1795, opened 2018-01-22
GEOT-5930: handling Bands selection parameter in blank response creation

dromagnoli:


modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/RasterLayerResponse.java
```diff
@@ -949,20 +970,27 @@ private MosaicOutput createBlankResponse() {
...
                 cm = new IndexColorModel(1, 2, arr, arr, arr);
             }
             il.setColorModel(cm);
             il.setSampleModel(cm.createCompatibleSampleModel(tileSize.width, tileSize.height));
 
-            final double[] bkgValues = new double[values.length];
-            for (int i = 0; i < values.length; i++) {
-                bkgValues[i] = values[i].doubleValue();
-            }
+			final double[] bkgValues;
```
aaime: Tabs!

modules/plugin/imagemosaic/src/test/java/org/geotools/gce/imagemosaic/ImageMosaicReaderTest.java
```diff
@@ -2039,7 +2040,68 @@ public void testRequestInHole() throws Exception {
     }
 
 
-    
+	@Test
+	public void testBlankResponseWithBandSelection() throws FactoryException, IOException {
+		File source = URLs.urlToFile(rgbURL);
```
aaime: More tabs

dromagnoli:
@aaime tabs removed. Thanks for reporting them.

PR #1796, opened 2018-01-22
[GEOT-5932]:GridCoverageRenderer set parent coverage property

dromagnoli:


modules/library/render/src/main/java/org/geotools/renderer/lite/gridcoverage2d/GridCoverageRenderer.java
```diff
@@ -748,7 +743,18 @@ private GridCoverage2D renderCoverage(final GridCoverage2D gridCoverage,
...
     }
 
-
+	private RenderedImage getImageFromParentCoverage(GridCoverage2D parentCoverage) {
+		if (parentCoverage == null) {
+			return null;
+		}
+		RenderedImage ri = parentCoverage.getRenderedImage();
+		if (ri != null) {
+			PlanarImage pi = PlanarImage.wrapRenderedImage(ri);
```
aaime: This already checks if ri is a planar and avoids double wrapping, right?

modules/library/render/src/main/java/org/geotools/renderer/lite/gridcoverage2d/GridCoverageRenderer.java
```diff
@@ -748,7 +743,18 @@ private GridCoverage2D renderCoverage(final GridCoverage2D gridCoverage,
...
+			pi.setProperty("ParentCoverage", parentCoverage);
```
aaime: Best to put this name in a public constant for clients and tests to reuse

modules/library/render/src/main/java/org/geotools/renderer/lite/gridcoverage2d/GridCoverageRenderer.java
dromagnoli: Yep. wrapRenderedImage avoid double wrapping by simply returning the input object if already a PlanarImage.

modules/library/render/src/main/java/org/geotools/renderer/lite/gridcoverage2d/GridCoverageRenderer.java
dromagnoli: Right. I forgot to do that. Fixing...

dromagnoli:
@aaime Feedbacks addressed (putting the property as a constant for re-use)

PR #1798, opened 2018-01-24
Java 9 Compatibility - Preliminary steps

tbarsballe:
Main changes: 
* Add `-XX:+IgnoreUnrecognizedVMOptions --add-modules=java.xml.bind --add-modules=java.activation` to the {jvmArgs} parameter used by the surfire-test argument in the main pom, to support running tests against Java 8 and Java 9 (This is a short-term workaround)
* Update various test cases to support XML pretty-print changes introduced in java 9
* Update a few image comparison tests with a higher tolerence; verified the image looks the same.
* Modify gt-wfs-ng to use the Java 9 compatible factory registry

I also added some basic docs to the user guide outlining what is currently necessary to run GeoTools on Java 9.


modules/plugin/imagemosaic/src/test/java/org/geotools/gce/imagemosaic/HeterogenousCRSTest.java
```diff
@@ -98,6 +94,7 @@ public static void initCRS() {
         System.setProperty("org.geotools.referencing.forceXY", "true");
     }
 
+    @Ignore
```
tbarsballe: @dvntucker - I have marked this test as ignored for now - you had mentioned it was rather brittle. Now that I have everything else working on Java 9, did you want to take a look at it when you get the chance?

mipastgt:
I'd just like to mention that the --add-modules=java.xml.bind --add-modules=java.activation will fail as soon as someone tries to run with an EA build of Java 11. This could already happen right now if someone builds his own JDK. So maybe it is better to just specify the maven dependencies for the replacement libraries in the POM right now.

aaime:
@mipastgt check the mailing list, @tbarsballe tried to add the dependencies and then something else broke. Maybe you can give it a try, fix whatever else breaks, and provide a new patch on top of this one?

tbarsballe:
@mipastgt - to summarize from the mailing list:

[JAXB](https://github.com/javaee/jaxb-v2) and [Activation](https://github.com/javaee/activation) are available as standalone dependencies, but using them instead causes the following error, which I did not get when adding them as modules:

```
Error: java.lang.IllegalAccessException: class org.geotools.gce.imagemosaic.catalog.AbstractGTDataStoreGranuleCatalog cannot access class com.sun.imageio.plugins.tiff.TIFFImageReaderSpi (in module java.desktop) because module java.desktop does not export com.sun.imageio.plugins.tiff to unnamed module
```

(Normally, imagemosaic should be using the imagei0-exts TiffImageReaderSpi, not the sun native one)

So it seems there are some differences between the modules included in Java 9 and what is available from the standalone dependencies.

This PR is at best a short-term fix anyways, until we move over to using the Java Module system properly.

aaime:
@tbarsballe any reason why this one did not get merged yet?

tbarsballe:
Because it doesn't seem to have had a proper review yet?
While a bunch of the changes has been discussed over the mailing list, I'm not sure that anyone has actually looked at the diff.

aaime:
I see... business wise we have no stake in Java 9 yet, but I'll try to have a look in spare time when I can....

docs/user/build/install/jdk.rst
```diff
@@ -48,6 +48,24 @@ Java 5    GeoTools 2.5.x   GeoTools 8.x     compiler=1.5
...
 
+Using Java 9 with GeoTools
+''''''''''''''''''''''''''
+
+Java 9 introduces a number of changes to the JVM, most notably the module system (Project Jigsaw). Refer to `The State of the Module System <http://openjdk.java.net/projects/jigsaw/spec/sotms/>`_ for more details.
+While GeoTools has not yet been updated to use the module system, GeoTools 19.x can run on Java 9.
+
+If your project depends on GeoTools 19.x, and you want to use it with Java 9, you will need to add the following flags to your JVM runtime arguments::
+
+    --add-modules=java.xml.bind --add-modules=java.activation
```
aaime: Wondering, this makes it look like the issue is GT wide, but I thought it was a problem specific to image mosaic and modules depending on it?

aaime:
Wondering if it's necessary/better to change the reference images that have been altered? Can you elaborate?

mipastgt:
This is an issue everywhere where you try to read or write some data via JAXB for example. I don't know which modules inside GT do that. For a discussion of this issue see: [https://stackoverflow.com/questions/43574426/how-to-resolve-java-lang-noclassdeffounderror-javax-xml-bind-jaxbexception-in-j](how-to-resolve-java-lang-noclassdeffounderror-javax-xml-bind-jaxbexception-in-j)

tbarsballe:
> Wondering if it's necessary/better to change the reference images that have been altered? Can you elaborate?

When testing this, I did update the reference images to that which was generated using Java 9, but there was still enough of a difference that tolerances had to be increased for compatibility with Java 8.

Changing the actual composition of the reference images to be better cross-compatible could work, though I am concerned that figuring out such an image would be mostly trial-and-error, as I am not actually certain what changes are causing the need for increased tolerance. Text / alignment seems like it may be affected, but there are a number of textless images that are as well.

docs/user/build/install/jdk.rst
tbarsballe: This is an issue wherever JAXB and Activation are used. The first place I encountered it was in imagemosaic, so anything that comes before it in the build order should be unaffected. However, since I was mostly addressing issues as they came up in build order, I'm not sure if there are any latter modules that share this issue (For example, it is possible that gt-render also requires this flag).

So, it probably just affects imagemosaic and dependants, but I don't know enough to say for certain; consequently I decided to err on the side of caution.

modules/library/render/src/test/java/org/geotools/renderer/lite/FillTest.java
```diff
@@ -108,7 +108,7 @@ public void testSolidFill() throws Exception {
 
     @Test
     public void testCrossFill() throws Exception {
-        runSingleLayerTest("fillCross.sld");
+        runSingleLayerTest("fillCross.sld", 250);
```
jodygarnett: So these are the minor visual differences then.

modules/library/render/src/test/java/org/geotools/renderer/lite/LabelSpacingTest.java
```diff
@@ -61,6 +61,6 @@ public void testLabelsIncreaseWordSpacing() throws Exception {
         // let's see if the result image match our expectations
         File reference = new File(
                 "./src/test/resources/org/geotools/renderer/lite/test-data/lineLabelsIncreaseWordSpacing.png");
-        ImageAssert.assertEquals(reference, image, 3000);
+        ImageAssert.assertEquals(reference, image, 3500);
```
jodygarnett: This difference seems considerable looking at the images below, it is like we are respecting kerning in Java 9. Okay it is only a change on text that follows a line, so it it a result of different math?

Should we note this as an improvement for those upgrading to Java 9....

jodygarnett:
We have a conflict with "HeterogenousCRSTest.java" but then we should be good to merge.

modules/library/render/src/test/java/org/geotools/renderer/lite/LabelSpacingTest.java
tbarsballe: Yeah, some of the labels look like they didn't change at all, but one or two seem to have gotten shifted. I assume it is related to the math difference, but I can't corroborate it.

Given the already high tolerance on the test, I wouldn't be surprised if this is already somewhat different between different OSes and architectures - if we really wanted to be sure how it changed, we would have to test on multiple different OSes running Java 9 vs Java 8 (and possible OpenJDK vs Oracle)

tbarsballe:
Conflicts fixed; will merge after CI reruns.

PR #1800, opened 2018-01-26
[GEOT-5738] URLs.urlToFile does not percent-decode URLs with no slash after "file:"

johngschulz:
https://osgeo-org.atlassian.net/browse/GEOT-5738

### DIFF
--- a/modules/library/metadata/src/main/java/org/geotools/util/URLs.java
+++ b/modules/library/metadata/src/main/java/org/geotools/util/URLs.java
@@ -165,9 +165,9 @@ public class URLs {
             string = string.replace("+", "%2B");
         }
         try {
-            string = URLDecoder.decode(string, "UTF-8");
+            string = URLDecoder.decode(string, java.nio.charset.StandardCharsets.UTF_8.name());
         } catch (UnsupportedEncodingException e) {
-            throw new RuntimeException("Could not decode the URL to UTF-8 format", e);
+            throw new IllegalArgumentException("UTF-8 encoding not supported", e);
         }
         String path3;
         String simplePrefix = "file:/";
@@ -186,7 +186,12 @@ public class URLs {
             path3 = string.substring(simplePrefix.length() - 1);
         } else {
             String auth = url.getAuthority();
-            String path2 = url.getPath().replace("%20", " ");
+            String path2 = url.getPath();
+            try {
+                path2 = URLDecoder.decode(path2, java.nio.charset.StandardCharsets.UTF_8.name());
+            } catch (UnsupportedEncodingException e) {
+                throw new IllegalArgumentException("UTF-8 encoding not supported", e);
+            }
             if (auth != null && !auth.equals("")) {
                 path3 = "//" + auth + path2;
             } else {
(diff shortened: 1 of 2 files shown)


modules/library/metadata/src/main/java/org/geotools/util/URLs.java
```diff
@@ -186,7 +186,12 @@ public static File urlToFile(URL url) {
             path3 = string.substring(simplePrefix.length() - 1);
         } else {
             String auth = url.getAuthority();
-            String path2 = url.getPath().replace("%20", " ");
+            String path2 = url.getPath();
+            try {
+                path2 = URLDecoder.decode(path2, "UTF-8");
+            } catch (UnsupportedEncodingException e) {
+                throw new RuntimeException("Could not decode the URL to UTF-8 format", e);
```
tbarsballe: Throwing a runtime exception here seems a bit severe.
However, I see that that is what the method is already doing for a similar failure earlier, and that by the WC3 spec URL encoding is assumed to be UTF-8.

I am still a bit concerned this could cause problems, but it does seem to be OK.

modules/library/metadata/src/test/java/org/geotools/util/URLsTest.java
```diff
@@ -112,6 +112,7 @@ public void testUrlToFile() throws Exception {
         URL url = new URL("file", "////oehhwsfs09", "/some/path/on/the/server/filename.nds");
         File windowsShareFile = URLs.urlToFile(url);
         assertNotNull(windowsShareFile);
+        assertEquals("file café", URLs.urlToFile(new URL("file:file%20caf%C3%A9")).toString());
```
tbarsballe: Test case is a bit sparse, but adequate. For completeness, testing all of the following might be better, bit it is OK as-is:

`file:file%20caf%C3%A9"`
`file:/file%20caf%C3%A9"`
`file://file%20caf%C3%A9"`

bencaradocdavies: @johngschulz please add the extra tests suggested by @tbarsballe. We will thank you in the future.

johngschulz: @bencaradocdavies I've added the additional test, when converting urlToFile using `file:/` the single slash remains in place.  I am unsure if this is an expected behavior or not.

johngschulz:
If I understood correctly - I've changed the exception to an IllegalArgumentException and the associated message, changed the "UTF-8" string to java.nio.charset.StandardCharsets.UTF_8.name(), and added the additional tests.

I've added a note about how the single slash is handled. Let me know if the changes look appropriate. Thanks

bencaradocdavies:
@johngschulz thanks very much. Those changes look good. For the single-slash case your test is just documenting the existing behaviour. All these cases are to various degrees dodgy non-conforming uses. I cherry-picked your change locally (so I could remove one extra blank line) and ran a full GeoTools build + full GeoServer build + app-schema online test to make sure integration works. All good, so pushed to master.

johngschulz:
Thanks cheers!

PR #1803, opened 2018-02-01
Improved support for ArcGIS WFS data source

lmoran:
The addition of a new output format allow the execution of POST requests again an ArcGIS Server WFS data source. (GET does not work yet, since ArcGIS Server expects spaces to be encoded as "%20", but class URIs usea the Java URLEncoder, which converts spaces into "+".)


aaime:
Hi @lmoran you're the "maintainer" of the module, so can also go and commit directly. Do you have commit access? If not, please ask on the geotools devel list

PR #1809, opened 2018-02-12
[GEOT-5943] Expose GDAL ASRP/USRP format

aaime:


modules/plugin/imageio-ext-gdal/src/main/java/org/geotools/coverageio/gdal/srp/SRPFormatFactory.java
```diff
@@ -0,0 +1,80 @@
...
+			.getLogger(SRPFormatFactory.class.getPackage().getName());
+
+	/**
+	 * Tells me if the coverage plugin to access Erdas imagine is available or
+	 * not.
+	 * 
+	 * @return <code>true</code> if the plugin is available,
+	 *         <code>false</code> otherwise.
+	 */
+	public boolean isAvailable() {
```
dromagnoli: this file looks like having TAB indentations

modules/plugin/imageio-ext-gdal/src/main/java/org/geotools/coverageio/gdal/jp2ecw/JP2ECWFormat.java
```diff
@@ -21,6 +21,7 @@
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
+import jdk.nashorn.internal.ir.annotations.Ignore;
```
dromagnoli: weird import here. I think you wanted the junit ignore instead

aaime:
Feedback handled

bradh:
Does this need a (follow-up?) docs update?

aaime:
I suppose we could add one more bullet point in this page:
http://docs.geotools.org/stable/userguide/library/coverage/imageio.html
(I believe it's already quite outdated though)

aaime:
List updated, merging

PR #1810, opened 2018-02-14
[GEOT-5945] Expose granule crs and resolution as a recognizable dimension

aaime:


aaime:
Will follow up with docs on the GeoServer side

PR #1811, opened 2018-02-22
Changing release process - create new branch during beta instead of RC

tbarsballe:
For an upcoming improvement proposal.

See also: https://github.com/geoserver/geoserver/pull/2763


aaime:
+1

tbarsballe:
Updated, based on mailing list discussions. The beta has been replaced with the release candidate,
 which now occurs a month before the release (functionally this the same as the initially proposed change, and is just a matter of different terminology).

PR #1812, opened 2018-02-23
[GEOS-8477] Support null values in WMSComplexTypes$_LegendURLType

tbarsballe:
Fix for https://osgeo-org.atlassian.net/browse/GEOS-8477

(Bug was reported against GeoServer, but the actual fix is in GeoTools)

Modified _LegendURLType to support returning null values.
This is consistent with the behaviour of other Types in WMSComplexTypes, such as _OnlineResourceType.

Also tested manually in geoserver, to ensure the WMS store could be created.

### DIFF
--- a/modules/extension/wms/src/main/java/org/geotools/data/wms/xml/WMSComplexTypes.java
+++ b/modules/extension/wms/src/main/java/org/geotools/data/wms/xml/WMSComplexTypes.java
@@ -4349,7 +4349,7 @@ public class WMSComplexTypes {
 				Attributes attrs, Map hints) throws SAXException,
 				OperationNotSupportedException {
                     
-                    String legendURL = value[1].getValue().toString();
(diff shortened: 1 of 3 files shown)


modules/extension/wms/src/main/java/org/geotools/data/wms/xml/WMSComplexTypes.java
```diff
@@ -4349,7 +4349,7 @@ public Object getValue(Element element, ElementValue[] value,
 				Attributes attrs, Map hints) throws SAXException,
 				OperationNotSupportedException {
                     
-                    String legendURL = value[1].getValue().toString();
+                    String legendURL = value[1].getValue() == null ? null : value[1].getValue().toString();
```
jodygarnett: classic null check?

tbarsballe: Yep, pretty much.

PR #1816, opened 2018-03-05
Disable external entity expansion when parsing service exceptions, GEOT-5953

smithkm:


smithkm:
@aaime reviewed via email

PR #1820, opened 2018-03-07
GEOT-5952: implemented HTTP connection pooling option for WFS datastore

mbarto:


modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/WFSDataStoreFactory.java
```diff
@@ -129,7 +129,19 @@ public WFSDataStore createDataStore(final Map<String, Serializable> params)
 
         return dataStore;
     }
-    
+
+    public HTTPClient getHttpClient(final Map<String, Serializable> params) throws IOException {
```
aaime: Public methods should have javadocs (yes, I know there are many that still do not)

modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/internal/WFSConfig.java
```diff
@@ -279,6 +283,16 @@ public boolean isGmlCompatibleTypenames() {
     public EntityResolver getEntityResolver() {
         return entityResolver;
     }
+
```
aaime: Extra white lines between methods here?

PR #1821, opened 2018-03-08
Add findOperations method to CoordinateOperationFactory to find all t…

dispiste:
…he available operations from a source to a target CRS. It was only possible to get the default operation till now. New method signature:

    Set<CoordinateOperation> findOperations(CoordinateReferenceSystem sourceCRS,
                                        CoordinateReferenceSystem targetCRS)

The goal of the method is to be able to find all the available operations from a source to a target CRS (by contrast with the existing "createOperation" method, which only returns the default operation). This is particularly useful for CRSs having different datums, for which many operations are sometimes defined, producing very different results.

### DIFF
--- a/modules/plugin/epsg-hsql/src/test/java/org/geotools/referencing/factory/epsg/OperationFactoryTest.java
+++ b/modules/plugin/epsg-hsql/src/test/java/org/geotools/referencing/factory/epsg/OperationFactoryTest.java
@@ -17,11 +17,8 @@
 package org.geotools.referencing.factory.epsg;
 
 import java.util.Iterator;
+import java.util.Set;
 import java.util.logging.Level;
-
-import junit.framework.Test;
-import junit.framework.TestCase;
-import junit.framework.TestSuite;
 
 import org.geotools.factory.Hints;
 import org.geotools.geometry.DirectPosition2D;
@@ -31,6 +28,7 @@ import org.geotools.referencing.ReferencingFactoryFinder;
 import org.geotools.referencing.operation.AbstractCoordinateOperation;
 import org.geotools.referencing.operation.AuthorityBackedFactory;
 import org.geotools.referencing.operation.BufferedCoordinateOperationFactory;
+import org.geotools.referencing.operation.TransformTestBase;
 import org.geotools.resources.Arguments;
 import org.geotools.resources.Classes;
 import org.opengis.geometry.DirectPosition;
@@ -42,7 +40,12 @@ import org.opengis.referencing.operation.ConcatenatedOperation;
 import org.opengis.referencing.operation.Conversion;
 import org.opengis.referencing.operation.CoordinateOperation;
 import org.opengis.referencing.operation.CoordinateOperationFactory;
+import org.opengis.referencing.operation.MathTransform;
 import org.opengis.referencing.operation.Transformation;
+
+import junit.framework.Test;
+import junit.framework.TestCase;
+import junit.framework.TestSuite;
 
 
 /**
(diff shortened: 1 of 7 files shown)


modules/plugin/epsg-hsql/src/test/java/org/geotools/referencing/factory/epsg/OperationFactoryTest.java
```diff
@@ -209,4 +212,158 @@ public void testTransformedDirectPosition() throws Exception {
...
+            final MathTransform transform = operation.getMathTransform();
+            TransformTestBase.assertInterfaced(transform);
+        }
+
+    }
+
+    /**
+     * Tests findOperations method for a pair of known CRSs. 4258 (Geographic2D) to 4326 (Geographic2D), with different datum
+     */
+    public void testFindOperations4() throws Exception {
```
aaime: As an additional test for geographic to geographic try something that has a grid shift file, e.g., see this test for reference (it leverages grid shift files that you can find in this module test resources)
https://github.com/geotools/geotools/blob/e3e81c4f24b4deb3acde49948c650e85df74d850/modules/library/referencing/src/test/java/org/geotools/referencing/AbstractCRSTest.java#L475

dispiste:
I have updated the pull request considering the changes proposed by @aaime 
In particular, the following changes have been performed:
- Replaced asHashSet by Collections.singleton() and Collections.emptySet()
- Restored all the public and protected methods to keep existing API, delegating the behaviour to the new methods
- Test methods have been given significant names, some code has been refactored
- A test using NADCON grid has been added
- The findOperationSteps and createPassThroughOperations have been replaced by the original createOperationStep method for CompoundCRSs (see explanation bellow)

Also, regarding why some cases still use createOperationStep instead of findOperationSteps: these cases can only produce one operation in current implementation, so I've not changed them to minimize the set of changes (which anyway is quite big). For instance, a Geographic2D to Geographic2D transformation can be done using different parameters (i.e. different operations may be available), but these are managed at the beginning of the method by querying the database. When no transformations are found in the database, they are handled in an alternative way that produces a single operation. See for instance:
https://github.com/dispiste/geotools/blob/e712cae4f9089f164fbec98762d602d27018d5b0/modules/library/referencing/src/main/java/org/geotools/referencing/operation/DefaultCoordinateOperationFactory.java#L331
which calls:
https://github.com/dispiste/geotools/blob/e712cae4f9089f164fbec98762d602d27018d5b0/modules/library/referencing/src/main/java/org/geotools/referencing/operation/DefaultCoordinateOperationFactory.java#L844

Note that I have also restored the original createOperationStep method for CompoundCRSs, removing findOperationSteps and createPassThroughOperations. The reason is that both the old and the new methods are extremely complex to understand, and I've realized that they can only produce one operation, so there is currently no need to support a set. In theory, several operations could be produced in some cases, but these cases are not currently supported by existing implementation:
- horizontal+temporalCRS to horizontal+temporalCRS when they have different datums => this is unsupported because (AFAIK) the database does not contain any transformation for temporal CRSs.
- horizontal+verticalCRS to horizontal+verticalCRS when they have different datums => this cases is explicitly not handled by current implementation because the coordinates are inter-dependent and can't be transformated by separate concatenated operations

Finally, I've created a separate pull request to fix the creation of transformations between CompoundCRS, which I've found it is (indirectly) broken in the existing implementation:
https://github.com/geotools/geotools/pull/1830

I've submitted it as a separate request so that we are sure compound CRSs work properly before and after this pull request is accepted.

I think this pull request is ready now.


``

aaime:
Hi, yep, looks good. I noticed the commits did not reference any Jira ticket, and could not find one,
so went ahead and created it based on the pull request description:
https://osgeo-org.atlassian.net/browse/GEOT-5965

I am  squash-merginbg the pull request setting the commit title so that it refers the ticket number and title.

aaime:
The commit is now on master. Do you need it on 19.x too?
If so, as a new feature it has a "cool down" period of one month before it can be backported, which means it will be available in an official release in May, along with 19.1 (19.0 will be but Monday). If you want to follow this path I'd advise to open a backport pull request around April 17th.

dispiste:
Thanks Andrea for reviewing and merging this PR. I'll be happy to provide the backport for 19.1.

PR #1824, opened 2018-03-09
[GEOT-5960] Allow setting number of decimals for encoding in GML 3.2 (fast path) and GML 3+ (complex features)

aaime:
https://osgeo-org.atlassian.net/browse/GEOT-5960

This pull request is up for review but not for merge yet, a followup GeoServer pull request will land shortly.
@bencaradocdavies you open the original issue on the GeoServer side, do you want to have a look? 
In the meantime I've assigned review to our good @nmco

### DIFF
--- a/modules/extension/xsd/xsd-core/src/main/java/org/geotools/xml/Configuration.java
+++ b/modules/extension/xsd/xsd-core/src/main/java/org/geotools/xml/Configuration.java
@@ -379,19 +379,8 @@ public abstract class Configuration {
      * @since 8.0
      */
     public <C extends Configuration> C getDependency(Class<C> clazz) {
-        //first try straight dependencies
-        for (Configuration dep : (List<Configuration>)getDependencies()) {
-            if (clazz.isInstance(dep)) {
-                return (C) dep;
-            }
-        }
-        //fall back to all
-        for (Configuration dep : (List<Configuration>)allDependencies()) {
-            if (clazz.isInstance(dep)) {
-                return (C) dep;
-            }
-        }
-        return null;
+        List dependencies = allDependencies();
+        return (C) dependencies.stream().filter(dep -> clazz.isInstance(dep)).findFirst().orElse(null);
     }
 
     /**
--- a/modules/library/main/src/main/java/org/geotools/gml/producer/CoordinateFormatter.java
+++ b/modules/library/main/src/main/java/org/geotools/gml/producer/CoordinateFormatter.java
@@ -21,9 +21,11 @@ import java.text.NumberFormat;
 import java.util.Locale;
 
 /**
- * Formats coordinates with a given number of decimals, using code more efficient than NumberFormat when possible
+ * Formats coordinates with a given number of decimals, using code more efficient than NumberFormat 
+ * when possible.
+ * The class is not thread safe, create a new instance for each thread using it.
  */
-public class CoordinateFormatter {
+public final class CoordinateFormatter {
 
     /**
      * The min value at which the decimal notation is used 
@@ -43,7 +45,7 @@ public class CoordinateFormatter {
     private final FieldPosition ZERO = new FieldPosition(0);
 
     /**
-     * The power of ten used for fast rounding
+     * The power of ten used for fast rounding, computed using the provided number of decimal values
      */
     private final double scale;
 
@@ -78,24 +80,25 @@ public class CoordinateFormatter {
      */
     public StringBuffer format(double x, StringBuffer sb) {
         if ((Math.abs(x) >= DECIMAL_MIN && x < DECIMAL_MAX) || x == 0) {
-            x = truncate(x);
+            x = roundDecimals(x);
             long lx = (long) x;
-            if (lx == x)
+            if (lx == x) {
                 sb.append(lx);
-            else
+            } else {
                 sb.append(x);
+            }
         } else {
             if (forcedDecimal) {
                 coordFormatter.format(x, sb, ZERO);
             } else {
-                sb.append(truncate(x));
+                sb.append(roundDecimals(x));
             }
         }
-        
+
         return sb;
     }
 
-    final double truncate(double x) {
+    final double roundDecimals(double x) {
         return Math.floor(x * scale + 0.5) / scale;
     }
     
--- a/modules/library/main/src/main/java/org/geotools/gml/producer/CoordinateFormatter.java
+++ b/modules/library/main/src/main/java/org/geotools/gml/producer/CoordinateFormatter.java
@@ -80,7 +80,7 @@ public final class CoordinateFormatter {
      */
     public StringBuffer format(double x, StringBuffer sb) {
         if ((Math.abs(x) >= DECIMAL_MIN && x < DECIMAL_MAX) || x == 0) {
-            x = roundDecimals(x);
+            x = truncate(x);
             long lx = (long) x;
             if (lx == x) {
                 sb.append(lx);
@@ -91,17 +91,26 @@ public final class CoordinateFormatter {
             if (forcedDecimal) {
                 coordFormatter.format(x, sb, ZERO);
             } else {
-                sb.append(roundDecimals(x));
+                sb.append(truncate(x));
             }
         }
 
         return sb;
     }
 
-    final double roundDecimals(double x) {
-        return Math.floor(x * scale + 0.5) / scale;
+    final double truncate(double x) {
+        // scale the number multiplying it by the power of 10 of the desired decimals
+        //  e.g. if we want 8 decimals: 3.123456786 * 10E8 = 312345678.6
+        double scaled = x * scale;
+        // add 0.5 to round the decimal part, e.g 312345678.6 + 0.5 = 312345679.1
+        scaled += 0.5;
+        // take only the decimal part, e.g.  312345679
+        scaled = Math.floor(scaled);
+        // remove the scale factor, the number will now have the desired number of decimals
+        return scaled / scale;
     }
-    
+
+
     /**
      * Returns the maximum number of digits allowed in the fraction portion of a
      * number.
--- a/modules/library/main/src/main/java/org/geotools/gml/producer/CoordinateFormatter.java
+++ b/modules/library/main/src/main/java/org/geotools/gml/producer/CoordinateFormatter.java
@@ -1,0 +1,129 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ *
+ *    (C) 2018, Open Source Geospatial Foundation (OSGeo)
+ *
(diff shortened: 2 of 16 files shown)


modules/extension/xsd/xsd-core/src/main/java/org/geotools/xml/Configuration.java
```diff
@@ -379,19 +379,8 @@ public int compare(DepNode o1, DepNode o2) {
...
-        }
-        //fall back to all
-        for (Configuration dep : (List<Configuration>)allDependencies()) {
-            if (clazz.isInstance(dep)) {
-                return (C) dep;
-            }
-        }
-        return null;
+        List dependencies = allDependencies();
+        return (C) dependencies.stream().filter(dep -> clazz.isInstance(dep)).findFirst().orElse(null);
```
bencaradocdavies: Does this change the order of dependency resolution?

The local variable "dependencies" is only used once and could be eliminated.

modules/library/main/src/main/java/org/geotools/gml/producer/CoordinateWriter.java
```diff
@@ -138,31 +112,30 @@ public CoordinateWriter(int numDecimals, boolean useDummyZ,
...
-    public CoordinateWriter(int numDecimals, String tupleDelim, String coordDelim, boolean useZ, double z, int dimension) {        
+    public CoordinateWriter(int numDecimals, String tupleDelim, String coordDelim, boolean useZ, double z, int dimension) {
         if (tupleDelim == null || tupleDelim.length() == 0){
             throw new IllegalArgumentException("Tuple delimeter cannot be null or zero length");
         }
         if ((coordDelim != null) && coordDelim.length() == 0) {
             throw new IllegalArgumentException("Coordinate delimeter cannot be null or zero length");
         }        
         D = dimension;
+
```
bencaradocdavies: Two blank lines in a row?

bencaradocdavies (commented):
@aaime my only interest in this functionality is that I created the GeoServer Jira issue in response to a report on the user mailing list.

I have had look over these changes and they seem good to me. One consequence of these changes, as seen in the unit tests, is that the default representation of floating point numbers that are integers (e.g. 5.0) is that they are encoded as strings that are integer literals (e.g. "5") not the floating point literals encoded by the current implementation (e.g. "5.0"). While this has benefits of compactness, it is a change of behaviour, might impact naive clients with careless handling of literals, and makes the tests look like they are testing integers (which they were anyway; we could change some of the test data to use non-integer coordinates). I am not saying that this change is work, just asking you to comment.

modules/extension/xsd/xsd-core/src/main/java/org/geotools/xml/Configuration.java
aaime: I know... but I often like to have those around as debugging aids... otherwise the thing is computed and returned, maybe ending in another method call and return, and one has to do a few steps to actually see what the results was.

modules/library/main/src/main/java/org/geotools/gml/producer/CoordinateWriter.java
aaime: This I should fix

modules/extension/xsd/xsd-core/src/main/java/org/geotools/xml/Configuration.java
aaime: About the order of dependency resolution, it may, but it's a consistency thing. WFS has two dependencies to GML, this method was returning one, and was being used by the GML output format to configure the decimals, and the method configuring the bindings was calling on "allDepenencies" instead, the result is that one GML configuration had the decimails set, but was not used. With this change all methods follow the same order and pick the same of the two GML dependencies.

modules/library/main/src/main/java/org/geotools/gml/producer/CoordinateFormatter.java
```diff
@@ -0,0 +1,129 @@
...
+            } else {
+                sb.append(truncate(x));
+            }
+        }
+        
+        return sb;
+    }
+
+    final double truncate(double x) {
+        return Math.floor(x * scale + 0.5) / scale;
```
nmco: It took me a few minutes to understand where the ``+0.5`` was coming from. Basically it just to round the decimal number right ? So if we want ``8`` decimals and want to truncate number ``3.123456786``:
```
3.123456786 * 10E8 + 05 = 312345678.6 + 0.5 = 312345679.1
```

Then we use ``Math.floor()`` to get the largest integer that is less than or equal to the argument.

modules/library/main/src/main/java/org/geotools/gml/producer/CoordinateFormatter.java
```diff
@@ -0,0 +1,129 @@
...
+     * @param sb
+     */
+    public StringBuffer format(double x, StringBuffer sb) {
+        if ((Math.abs(x) >= DECIMAL_MIN && x < DECIMAL_MAX) || x == 0) {
+            x = truncate(x);
+            long lx = (long) x;
+            if (lx == x)
+                sb.append(lx);
+            else
+                sb.append(x);
```
nmco: I would substitute:
```
if (lx == x)
  sb.append(lx);
else
  sb.append(x);
```

with:

```
sb.append(lx == x ? lx : x);
```

or add the missing ``{}``.

nmco (commented):
Just some minor comments, apart from that looks good to me.

modules/library/main/src/main/java/org/geotools/gml/producer/CoordinateFormatter.java
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    /**
+     * Formats a number with the configured number of decimals. For better performance best use 
+     * {@link #format(double, StringBuffer)} against a re-used StringBuffer
+     * @param x
+     * @param sb
+     */
+    public String format(double x) {
+        StringBuffer sb = new StringBuffer();
```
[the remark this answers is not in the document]
nmco: Just saw that ``AbstractStringBuilder`` is not public, my bad.

modules/library/main/src/main/java/org/geotools/gml/producer/CoordinateFormatter.java
```diff
@@ -0,0 +1,129 @@
...
+
+    /** To be used for formatting numbers, uses US locale. */
+    private final NumberFormat coordFormatter = NumberFormat.getInstance(Locale.US);
+
+    private final FieldPosition ZERO = new FieldPosition(0);
+
+    /**
+     * The power of ten used for fast rounding
+     */
+    private final double scale;
```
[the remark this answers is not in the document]
aaime: If you check the diff, CoordinateFormatter is really a refactor of pre-existing code, so all that's in there has been just moved around. But I can make a few extra improvements to it, no problems :)

modules/library/main/src/main/java/org/geotools/gml/producer/CoordinateFormatter.java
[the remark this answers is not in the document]
aaime: It's a lot of work for likely no benefit (synch removal is performed by the JVM once it determines there can not be sharing among threads) and it would trickle up in various places (IntelliJ finds 14 references in GeoTools).

modules/library/main/src/main/java/org/geotools/gml/producer/CoordinateFormatter.java
aaime: Renamed method to roundDecimals

modules/library/main/src/main/java/org/geotools/gml/producer/CoordinateFormatter.java
aaime: Braces added

modules/library/main/src/main/java/org/geotools/gml/producer/CoordinateFormatter.java
```diff
@@ -0,0 +1,129 @@
...
+package org.geotools.gml.producer;
+
+import java.text.FieldPosition;
+import java.text.NumberFormat;
+import java.util.Locale;
+
+/**
+ * Formats coordinates with a given number of decimals, using code more efficient than NumberFormat when possible
+ */
+public class CoordinateFormatter {
```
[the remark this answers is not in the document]
aaime: Javadoc added and final too

aaime:
@nmco feedback handled and put in a separate commit for ease of review, could you have a look?

modules/library/main/src/main/java/org/geotools/gml/producer/CoordinateFormatter.java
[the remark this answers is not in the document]
nmco: Interesting, thanks for the feedback.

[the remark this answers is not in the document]
nmco: So I agree no point in removing the usage of ``StringBuffer``.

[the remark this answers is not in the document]
aaime: In case it is of interest and you were not aware of lock elision already, see this: https://shipilev.net/jvm-anatomy-park/19-lock-elision/

modules/library/main/src/main/java/org/geotools/gml/producer/CoordinateFormatter.java
nmco: Actually the method is truncating the number, the desired number of decimals is 8:
```
truncate(3.123456786) = 3.12345679
```

Only this part of the code is rounding the decimals:
```
x * scale + 0.5
```

nmco: I would separate the computation and add a few comments:
```
final double truncate(double x) {
         // scale the number multiplying it by the power of 10 of the desired decimals
         //  e.g. if we want 8 decimals: 3.123456786 * 10E8 = 312345678.6
         double scaled = x * scale;
         // add 0.5 to round the decimal part, e.g 312345678.6 + 0.5 = 312345679.1
         scaled += 0.5;
         // take only the decimal part, e.g.  312345679
         scaled = Math.floor(scaled);
         // remove the scale factor, the number will now have the desired number of decimals
         return scaled / scale;
}
```

aaime: Ha ha, ok, that's one seriously commented bit of code. Copying and pasting ;-)

nmco: :yum:

nmco:
Thanks for the changes Andrea, looks good to me !

PR #1657, opened 2017-07-21
GEOT-5588 according to gdal spec iw the integers width is 0 …

biancazzurri:
according to gdal spec iw the integers width is 0 it means the width is not specified.

https://osgeo-org.atlassian.net/browse/GEOT-5588

PR #1825, opened 2018-03-11
5588: Fix OGR integer field widths

jespersm:
I wasn't sure how to properly extend PR #1657 from @biancazzurri so this is a new PR - but my fix actually does more, as it also maps three digit integer fields to shorts, not bytes, so that value  255 < x <= 999 don't get SILENTLY truncated to 8 bits. Same story for bigger integer sizes.
The test also reflects this, too.


jespersm:
I updated the pull request to prevent BigDecimal from being used in the case of integers.

Only now I realized that the OGR tests aren't run on Travis due to GDAL not being available. FYI they work on my machine (macOS Sierra), tested with gdal 1.11.4 and 2.2.3.

aaime:
Looks good, merging

PR #1827, opened 2018-03-13
Enable GDAL on Travis

jespersm:
The current Travis CI setup for GeoTools doesn't actually run the OGR tests for neither Bridj nor JNI, they're are more or less silently skipped. This is because we're not installing the actual gdal shared object into the Travis container.

This small patch adresses this _partly_: It makes the Bridj tests run, but not the JNI tests:

> Tests run: 28, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 4.444 sec - in org.geotools.data.ogr.bridj.BridjOGRDataStoreTest

I tried finding a relevant JIRA issue, but the closest I found was GEOT-4733. Hope you can use it!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,4 +22,4 @@ addons:
   apt:
     packages:
       - python-sphinx
-      - gdal-bin
+      - libgdal1h
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,3 +22,4 @@ addons:
   apt:
     packages:
       - python-sphinx
+      - gdal-bin


.travis.yml
```diff
@@ -22,3 +22,4 @@ addons:
   apt:
     packages:
       - python-sphinx
+      - gdal-bin
```
bencaradocdavies: Thanks @jespersm. Well spotted. Two things:
1. GDAL data stores, such as those used in the test, should not need all of `gdal-bin`, just its dependency `libgdal1h`. Please amend this pull request to change `gdal-bin` to `libgdal1h`.
1. Please also change the commit and PR title to "Enable GDAL on Travis". Try not. Do, or do not. There is no try.  ;-)

jespersm: Done!

bencaradocdavies:
Squash merged locally to preserve your authorship and applied to master, 19.x, 18.x. Thanks again, @jespersm.

bencaradocdavies:
Fixing GDAL JNI seems to need an upgrade of GDAL, installation of `libgdal-java`, and changes in GeoTools to accommodate the changed API.

PR #1828, opened 2018-03-14
[GEOT-5961]: Preserve alpha channel in RasterSymbolizer's ChannelSelect

dromagnoli:


modules/library/render/src/main/java/org/geotools/renderer/lite/gridcoverage2d/BandMergeNode.java
```diff
@@ -172,8 +177,11 @@ protected GridCoverage execute() {
 
 					// color model
 					final ComponentColorModel cm = new ComponentColorModel(
-							ColorSpace.getInstance(ColorSpace.CS_LINEAR_RGB),
-							false, false, Transparency.OPAQUE,
+							(size == 1 && hasAlpha) ?
```
aaime: Ouch, multiple inline conditional, that hurt my eyes. Could you use some temp variables instead? :-)

modules/library/render/src/main/java/org/geotools/renderer/lite/gridcoverage2d/BandMergeNode.java
```diff
@@ -253,4 +267,8 @@ protected GridCoverage execute() {
 
 	}
 
+    public void addAlpha(RenderedImage alpha) {
```
aaime: Javadoc missing for public method

modules/library/render/src/test/java/org/geotools/renderer/lite/gridcoverage2d/TransparencyStyledTest.java
```diff
@@ -0,0 +1,355 @@
...
+        System.clearProperty("org.geotools.referencing.forceXY");
+        CRS.reset("all");
+    }
+
+    @BeforeClass
+    public static void init() {
+
+        // make sure CRS ordering is correct
+        CRS.reset("all");
+        System.setProperty("org.geotools.referencing.forceXY", "true");
```
aaime: I know there is a bad habit of using this one "as string" and I'm also at fault for this, but would be nicer to use GeoTools.FORCE_LONGITUDE_FIRST_AXIS_ORDER

modules/library/render/src/main/java/org/geotools/renderer/lite/gridcoverage2d/BandMergeNode.java
dromagnoli: wiil do.

modules/library/render/src/main/java/org/geotools/renderer/lite/gridcoverage2d/BandMergeNode.java
dromagnoli: will do.

aaime:
Thank you!

PR #1830, opened 2018-03-16
Fix properties creation in DefaultOperationMethod when math transform…

dispiste:
… parameters are null. Indirectly fixes creation of operations between CompoundCRSs. Add a test case for createOperation using CompoundCRSs


dispiste:
Replaced by a new PR: https://github.com/geotools/geotools/pull/1833

PR #1831, opened 2018-03-16
[GEOS-8630] Allow the selection of a default geometry attribute in App-Schema

nmco:
Associated issue: https://osgeo-org.atlassian.net/browse/GEOT-5964


modules/extension/app-schema/app-schema/src/main/java/org/geotools/data/complex/config/AppSchemaDataAccessConfigurator.java
```diff
@@ -329,12 +343,77 @@ private AttributeDescriptor getTargetDescriptor(TypeMapping dto, CoordinateRefer
...
+            try {
+                geom = accessor.get(featureType, xpath, GeometryDescriptor.class);
+            } catch (Exception e) {
+                LOGGER.finest(
+                        String.format("Exception occurred retrieving geometry descriptor "
+                                + "at x-path \"%s\": %s", xpath, e.getMessage()));
+            }
+        }
+
+        return geom;
```
aaime: What if the XPATH is valid, but not found? I see in the tests that a invalid xpath generates an exception, so believe it's covered, but... where is that validation happening? Here?

modules/extension/complex/src/main/java/org/geotools/data/complex/ComplexFeatureConstants.java
```diff
@@ -52,6 +52,16 @@
...
 
+    /**
+     * Static attribute name used for the auxiliary default geometry attribute, which is created when
+     * &lt;defaultGeometry&gt; is specified in the feature type mapping configuration.
+     * 
+     * The constant represents to the local part of the qualified attribute name; the namespace URI
+     * is generated at runtime and will be the same as the namespace URI of the feature containing
+     * the geometry attribute.
+     */
+    public static final String DEFAULT_GEOMETRY_LOCAL_NAME = "__DEFAULT_GEOMETRY__";
```
aaime: In the rest of GeoTools the default goemetry is represented by the empty string, "", was there a specific reason to use a different value?

modules/extension/app-schema/app-schema/src/main/java/org/geotools/data/complex/config/AppSchemaDataAccessConfigurator.java
nmco: In the property accessor, if the XPATH doesn't exists an exception will be throw:

```
accessor.get(featureType, xpath, GeometryDescriptor.class)
```

modules/extension/complex/src/main/java/org/geotools/data/complex/ComplexFeatureConstants.java
nmco: I was just no aware of that :yum:, updating the code to sue the empty String.

nmco:
Pull request updated according to the review ... thanks for the review Andrea !

PR #1833, opened 2018-03-19
Fix properties creation in DefaultOperationMethod when math transform…

dispiste:
… parameters are null. Indirectly fixes creation of operations between CompoundCRSs. Add a test case for createOperation using CompoundCRSs


aaime:
Looks good, merging!

aaime:
Actually damn, did not check the commit message. Next time please open a Jira ticket and make sure the commit message is in the form "[GEOT-XYZW] Ticket title". You can copy paste it from Jira by doing a "export to XML".

PR #1837, opened 2018-03-21
Geojson complex props - supports GEOT-5158

jespersm:
Here is my proposed change for allowing JSON Objects as a Map.class binding, and JSON Arrays as a List.class binding inside a SimpleFeature.

I've extended the existing schema discovery and type checking to work with Map and List also, and included explicit compatibility with "properties which are geometry objects".

I've run the simple benchmark and couldn't measure any differences on my machine. A properties objects full of geometry might be a tad slower now (but it was slower to begin with due to the existing replay-proxy for certain layouts).

I hope you can use the change, please let me know if it needs extra work.

I did send the contributor agreement yesterday, BTW.


aaime:
I've tried to have a look but don't have enough familiarity with the module to say if the changes are good or not.

I've however tried to build the GeoServer modules depending on gt-geojson and this PR makes those build fail, so it's causing regressions:

````
Failed tests:   testJaggedGeoJSON(org.geoserver.importer.ImporterDataTest): expected:<4> but was:<7>
  testGeoJSONImportDirectory(org.geoserver.importer.ImporterDataTest): expected:<COMPLETE> but was:<PENDING>
  testGeoJSONImport(org.geoserver.importer.ImporterDataTest): expected:<COMPLETE> but was:<PENDING>
````

aaime:
Also mind, this weekend we'll have the GeoTools mass reformat, on all branches.

jespersm:
Ok, I’ll take a look at the GeoServer regressions. Some assumption must be mission from the module’s test, then.

-Jesper

aaime:
Make that "many assumptions", it's an unsupported module, so has to satisfy no requrements :-)

aaime:
Hi, I can hold the reformat for a few more hours, then I'll have to commit, or I'll lose the window we have prepared by cleaning up all remaining pull request (besides this one). 
It's going to generate many conflicts for this pull request (not unrecoverable, but somewhat nasty to get back in a mergeable state). How are you doing?

jespersm:
Sorry, I couldn't do it right away. Please commit away -- I'll update the patch after I find the breakage.

tbarsballe:
This PR now has conflicts with the current master. They will need to be resolved before it can be merged.

aaime:
Any update?

aaime:
Seem to have gone stale, closing it for the time being. Still would be happy to see progress on it, if you feel like working on it just make a new PR :-)

PR #1839, opened 2018-03-22
[GEOT-5973] update MySql jdbc driver to a newer version

nprigour:
MySql driver updated to version 5.1.40
The pull request addresses the improvement proposed in jira issue  [[GEOT-5973]](https://osgeo-org.atlassian.net/browse/GEOT-5973)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -961,7 +961,7 @@
       <dependency>
         <groupId>mysql</groupId>
         <artifactId>mysql-connector-java</artifactId>
-        <version>5.1.17</version>
+        <version>5.1.40</version>
       </dependency>
       <dependency>
         <groupId>org.hsqldb</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -961,7 +961,7 @@
       <dependency>
         <groupId>mysql</groupId>
         <artifactId>mysql-connector-java</artifactId>
-        <version>5.1.40</version>
+        <version>5.1.46</version>
       </dependency>
       <dependency>
         <groupId>org.hsqldb</groupId>


nprigour:
Latest version of MySql driver is 5.1.46 (compiled with Java 8). However decided to upgrade to 5.1.40 (compiled with Java 1.5) in order to enable easier backporting to previous geotools versions (14.x and up)

aaime:
Hi, tested locally looks good. Before we merge, you need to create a Jira ticket and refer to the ticket from the commit message, e.g. "[GEOT-XYZW] Updating MySql JDBC driver to version abc".
About backports, you can do as many pull requests as you like for that, but mind, only 18.x and 19.x are still being released.

mprins:
> However decided to upgrade to 5.1.40 (compiled with Java 1.5) in order to enable easier backporting to previous geotools versions (14.x and up)

this is a moot point as there are other places in the pom file that dictate java 8 and people can easily override the database driver version in their own project anyways if they need to downgrade (ofcourse that applies to upgrade as well)

seeing the log list of bugfixes https://dev.mysql.com/doc/relnotes/connector-j/5.1/en/news-5-1.html since .40 it seems worth reconsidering and upgrading to the latest 5.1.x release

aaime:
@nprigour typically the question you have to answer is about mysql version support. Which versions of MySQL does the latest driver support? Those unsupported are out of any LTS distribution, and generally speaking, at least 4-5 years old? (seems much I know, but database are probably the kind of software that is upgraded the least often, on fear of losing data).

nprigour:
Connector/J 5.1 can be used from version 5.5 and onward (with some certain features being available i the latest releases (check [jconnector](https://dev.mysql.com/doc/connector-j/5.1/en/connector-j-versions.html )
Based also on the following link [java versions supported](https://dev.mysql.com/doc/connector-j/5.1/en/connector-j-versions-java.html) I think we are pretty safe even with the latest version. So if there is no objection I will update the pull request to use the latest version.

aaime:
Go ahead :)

aaime:
Tested locally, worked fine. Merging

aaime:
Seemed like a safe change, so quickly backported on 19.x and 18.x too

PR #1843, opened 2018-03-28
Assorted fixes for group by

aaime:
Along with a new function and a fix in another, both of which will be made SQL encodable in a subsequent pull request (so, related even if not visibly so)

### DIFF
--- a/modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
+++ b/modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
@@ -20,6 +20,7 @@ import java.io.IOException;
 import java.io.UnsupportedEncodingException;
 import java.lang.reflect.Method;
 import java.net.URLDecoder;
+import java.security.acl.Group;
 import java.sql.Connection;
 import java.sql.DatabaseMetaData;
 import java.sql.PreparedStatement;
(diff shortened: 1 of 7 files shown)


modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
```diff
@@ -1557,6 +1559,18 @@ protected Object getAggregateValue(FeatureVisitor visitor, SimpleFeatureType fea
...
+    /**
+     * Checks if the groupby is a supported one, that is, if it's possible to turn to SQL the
+     * various {@link Expression} it's using
+     * @param visitor
+     * @return
+     */
+    private boolean isSupportedGroupBy(GroupByVisitor visitor) {
+        // right now everything must be an attribute
+        return visitor.getExpressions().stream().allMatch(xp -> xp instanceof AttributeDescriptor) 
+            && visitor.getGroupByAttributes().stream().allMatch(xp -> xp instanceof AttributeDescriptor);
```
nmco: Not sure if I understand this one, so this basically means it will not be possible to use the data difference function ?

nmco (approved):
Looks good to me, just a minor question.

modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
aaime: This code just updates the checks to the current status, to avoid the exceptions. A new pull request will follow to allow encoding the function in postgresql, that requires a larger number of changes, filter capabilities, a way to check if an expression is supported by the capabilities (right now can only check filter), postgresql specific encoding and so on.

nmco: Ok makes sense, thank you.

PR #1844, opened 2018-03-29
[GEOT-5979] Allow JDBCDataStore to work against a generic GroupByVisitor

aaime:


mbarto (approved):

aaime:
Found issues with prepared statement dialects, the aggregation expressions must be run in a sub-query and aliased, so that a name can be used in the groupby (single query does not work as groupby is executed before select, or so the documentation say). Working on it.

aaime:
Pull request updated with logic to handle expressions in group by statement on prepared statement dialects, and some extra Oracle tests too (the original one worked for non PS path in PostGIS, added PostGIS prepared statements tests and Oracle ones for variety, to reduce risk of using db specific syntax)

nmco (approved):
Looks good to me, actually simplifies what has been done before to support group by visitors and group by visitors with functions.

I'm only suggesting two minors additions, not necessary:

* [complement the documentation](http://docs.geotools.org/latest/userguide/library/main/collection.html) with this new functionality, there is already a section for group by visitors

* add a test where limits are also used, i.e. pagination

aaime:
Followed your feedback, adding a test actually allowed to locate a different, pre-existing bug in aggregate function optimization (if groupby returned only one element the optimization would have failed and not collected the results).
I've added generic docs on what to expect SQL optimization wise, in the next pull request (for encoding floor and dateDifference) I have one histogram based examples that I can add to the docs.

nmco:
Nice, thank you @aaime !

aaime:
Fixes build in geotools, but uncovers a new bug in app-schema, see discussion at:
https://github.com/geotools/geotools/pull/1846

PR #1845, opened 2018-03-30
[GEOT-5986] SimplifyProcess: switch simplifiers

martin-mfg:
It seems that the two simplifiers in the `SimplifyProcess` class got mixed up. This commit fixes that.

I should note that I have just started working with your project and that therefore I did not know how to verify that `SimplifyProcess` really uses the wrong simplifier. I concluded that the simplifiers are mixed up only by reading the source code. So I *might* be wrong about this.


aaime:
I believe you're right. Normally we ask for a test on each change, but it's a straightforward one and the module is not officially supported, so I'll cut you some slack on this one. :-)

PR #1847, opened 2018-03-30
GEOT-5982: CopyingHandler does not copy repeated namespace declarations

dromagnoli:


dromagnoli:
Merging it (travis failure is unrelated, due to ogr-bridj sometime randomly failing)

bencaradocdavies:
Need to watch our for more instances of the gt-ogr-bridj JVM crash. We have seen these before and needed to upgrade libgdal1h. If they become a problem we will need to investigate.

PR #1851, opened 2018-04-03
WIP: make native styles available from ResourceInfo class

jgrocha:
This PR results from the [ResourceInfo with native style](https://github.com/geotools/geotools/wiki/ResourceInfo-with-native-style) proposal.

This adds 3 new methods to `ResourceInfo` class, as discussed in the proposal.

```java
    public interface ResourceInfo {
       StyledLayerDescriptor getDefaultStyle() {
          return null;
       }
       List<String> getNativeStyles(){
          return Collections.emptyList();
       }
       StyledLayerDescriptor getNativeStyle(String name) {
          return null;
       }
    }
```

The proposal is still under discussion. **Comments are very welcome.**

I've added one initial unit test regarding shapefiles with side car SLD, which is: `modules/plugin/shapefile/src/test/java/org/geotools/data/shapefile/ShapefileWithSideCarStyleFileTest.java`.
I can add similar tests for geopackage and postgis datastores, if this test is ok.

suricate-and-their-babies

This is my first PR to Geotools. Please by gentle :-)

### DIFF
--- a/modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCFeatureSource.java
+++ b/modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCFeatureSource.java
@@ -16,21 +16,7 @@
  */
 package org.geotools.jdbc;
 
-import java.io.IOException;
-import java.sql.Connection;
-import java.sql.DatabaseMetaData;
-import java.sql.PreparedStatement;
-import java.sql.ResultSet;
-import java.sql.ResultSetMetaData;
-import java.sql.SQLException;
-import java.sql.Statement;
-import java.sql.Types;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.List;
-import java.util.Set;
-import java.util.logging.Level;
-import java.util.logging.Logger;
+import com.vividsolutions.jts.geom.Geometry;
 
 import org.geotools.data.DefaultQuery;
 import org.geotools.data.FeatureReader;
@@ -51,11 +37,13 @@ import org.geotools.feature.visitor.MaxVisitor;
 import org.geotools.feature.visitor.MinVisitor;
 import org.geotools.feature.visitor.NearestVisitor;
 import org.geotools.filter.FilterAttributeExtractor;
-import org.geotools.filter.visitor.PostPreProcessFilterSplittingVisitor;
 import org.geotools.filter.visitor.SimplifyingFilterVisitor;
 import org.geotools.geometry.jts.ReferencedEnvelope;
 import org.geotools.referencing.CRS;
+import org.geotools.sld.SLDConfiguration;
+import org.geotools.styling.StyledLayerDescriptor;
 import org.geotools.util.logging.Logging;
+import org.geotools.xml.Parser;
 import org.opengis.feature.Association;
 import org.opengis.feature.FeatureVisitor;
 import org.opengis.feature.simple.SimpleFeature;
@@ -67,7 +55,24 @@ import org.opengis.filter.expression.Expression;
 import org.opengis.filter.expression.PropertyName;
 import org.opengis.referencing.crs.CoordinateReferenceSystem;
 
-import com.vividsolutions.jts.geom.Geometry;
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
+import java.sql.Connection;
+import java.sql.DatabaseMetaData;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.ResultSetMetaData;
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.sql.Types;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Set;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 /**
  * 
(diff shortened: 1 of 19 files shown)


modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCFeatureSource.java
```diff
@@ -908,5 +913,145 @@ private boolean handleNearestVisitor(Query query, FeatureVisitor visitor) throws
...
+                String styleNameColumnName = null;
+                // GeoPackage DataSource class org.sqlite.javax.SQLiteConnectionPoolDataSource
+                // PostGIS    DataSource class org.geotools.data.jdbc.datasource.DBCPDataSource
+                if (getDataStore().getSQLDialect() instanceof PreparedStatementSQLDialect) {
+                    // GeoPackage
+                    sql = "select * from layer_styles where f_table_name = '" + featureName + "'";
+                    styleNameColumnName = "styleName";
+                } else {
+                    // Postgresql
+                    sql = "select * from public.layer_styles where f_table_name = '" + featureName + "'";
```
mprins: This query should be agnostic to the pg schema the the data comes from, not limited to the public schema

jgrocha: @mprins Thank you for your comment. You pointed out a very important issue. How to we store styles on the provider? In this [proposal](https://github.com/geotools/geotools/wiki/ResourceInfo-with-native-style), we are just following the QGIS approach to store the style on a single table named `layer_styles` on the `public` schema.

One of the arguments for this PR is to enhance the interoperability between GeoTools/GeoServer with QGIS.

We can make a better proposal to store native styles for GeoPackage and PostGIS and then improve both GeoTools and QGIS to support it.

aaime:
Feedback on the proposal itself (feedback on the PR might follow): the extension API imho should be modified. Issues that I'm seeing (in no particular order):

- There is a way to get all native styles, in an un-parsed way, allowing the caller to possibly throw different parsers at the problem. I like the idea, in case a system is smart enough to parse multiple languages, but setup this way, how do I know what the default style is? Also, the rest of the API assumes a notion of a default style, but in this call I don't see it.
- The other two methods return StyledLayerDescriptor. A full SLD normally represent an entire map, with multiple layers, remote sources, local inline data and so on, but a resource is a single one instead. If the methods are kept, what about returning Style object instead?
- The other two methods assume GeoTools has a way to read the style from the source and turn it into an object representation. GeoTools has different style parsers but no pluggable API to discover and use them in turn, so this approach would seem to limit the usefulness of these methods to SLD sources (and add direct dependency on the SLD parsing modules in data source ones, e.g., gt-xsd-sld if one wants to parse SLD 1.1, which is not necessarily welcomed).
- The string based method returns a list of un-named styles (e.g., could have been a map instead), the StyledLayerDescriptor approach seems to assume a named style situation, besides the default style.

Mind, both approaches have some limitations, even the string one won't work for formats that store styling information in something other than a document (e.g., could be database tables having rules expressed as record, or some other binary format, for those a object approach would work better).

If we want to keep things really simple, I guess we could just use the string approach (and leave other styling system to sponsor further extensions to the API), and have maybe just one method, the one returning the list of styles, with the default one being the first? (and documented as such?). Just trying to start a conversation here, it's not a request, just an idea.

jgrocha:
Thanks @aaime for push the discussion forward. According to your KISS principle, in terms of the API, we could keep this as simple as:

```java
    /**
     * @return a list of key-value pairs with all available native styles, if any.
     * Keys are style names,
     * Values are the style as an unparsed string.
     */
    public Map<String, String> getNativeStyles() {
        Map<String,String> styleMap = new HashMap<String,String>();
        return styleMap;
    }
```

The first element of the list would be the default style, if any. If there is no default style, we will not know. 

With this simple call, is up to the developers to detect the style format (it can be SLD, SE (preferable!), MapCSS, CartoCSS, QML, etc) and then call the required parser.

If names are not unique, we can have the same style name with different style encodings, so developers can use the one they are able to parse.

Do we need some more information to offer these native styles in GeoServer?

aaime:
@jgrocha the map has no order (unless it's explicitly a LinkedHashMap), but maybe there can be a special key to represent the default one.

You raise a good point about figuring out the actual style language... like for example, in GeoServer, we ask the user or we make the REST caller tell us which language is used because for example the SLD 1.0 would parse SLD 1.1 too, but miss a number of elements (the parser is tolerant).

Thinking out loud, what about returning a List of StyleInfo instead?

````
interface StyleInfo {
String getName();
String getLanguage();
Version getVersion();
String getStyle();
}
````

The list is ordered, the first style would be the default one, and we have some extra information about it besides its body.

File: modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCFeatureSource.java (hunk 1 above)
Comment (aaime) [reply modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCFeatureSource.java]: Agree with @mprins this is potentially a problem, particularly in this place, as it's shared among all database implementations (do Oracle and SQLServer even have a "public" schema?).
Maybe it's best moved to SQLDialect, and implemented only for postgis and geopackage.
For postgis at least, I can see the usage of public schema as "valid", geometry_columns is in there after all. Looking at the implementation however, I see we don't schema qualify the lookup, allowing the search path to drive the search. Maybe this could be done for the postgis implementation of the style lookup too?

File: modules/library/jdbc/src/test/java/org/geotools/data/jdbc/FilterToSQLTest.java [hunk 2]
```diff
@@ -67,7 +66,7 @@
 public class FilterToSQLTest extends TestCase {
     private FilterFactory filterFac = CommonFactoryFinder.getFilterFactory(null);
     private static Logger LOGGER = org.geotools.util.logging.Logging.getLogger("org.geotools.data.jdbc");
-    
+
```
aaime: Seems it's just whitespace change, please avoid reformatting the code in areas that you did not touch (automatic formatting and checking is coming, please be patient, it's linked to my very limited spare time)

modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/ShapefileFeatureSource.java
```diff
@@ -599,4 +609,69 @@ protected boolean handleVisitor(Query query, FeatureVisitor visitor) throws IOEx
...
+    protected StyledLayerDescriptor getNativeStyle(String name) throws IOException {
+        ReadableByteChannel in = null;
+        try {
+            ByteBuffer buffer = ByteBuffer.allocate(4*1024);
+            FileReader reader = new FileReader() {
+                public String id() {
+                    return "Shapefile's getStyle method to read sidecar style file";
+                }
+            };
+            in = this.shpFiles.getReadChannel(SLD, reader);
```
aaime: Calling directly shpFiles.getInputStream for this case it's probably easier? :-)
(and then give the stream to Parser directly)

modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/ShapefileFeatureSource.java
```diff
@@ -599,4 +609,69 @@ protected boolean handleVisitor(Query query, FeatureVisitor visitor) throws IOEx
...
+                String sldstr = "";
+                while (in.read(buffer) > 0) {
+                    buffer.flip();
+                    sldstr += StandardCharsets.UTF_8.decode(buffer).toString();
+                    buffer.clear();
+                }
+                InputStream stream = new ByteArrayInputStream(sldstr.getBytes(StandardCharsets.UTF_8));
+                SLDConfiguration configuration = new SLDConfiguration();
+                Parser parser = new Parser(configuration);
+                StyledLayerDescriptor sld = (StyledLayerDescriptor) parser.parse(stream);
```
aaime: This will mostly go away since we are returning the style as a string right?

jgrocha:
Thanks @aaime for the detailed feedback on the code itself.
Regarding the API, I do like the proposal of an ordered `StyleInfo` list. I'll play with it.

aaime:
Hi @jgrocha the "codebase reformat" proposal is about to get applied, once that happens this pull request will become un-mergeable (ton of conflicts). How is this coming along? Thinking out loud, if you have yet to code the changes, you might want to wait for after the reformat, as you'd have to restart almost from scratch anyways using StyleInfo, or not?

aaime:
I have started a thread on geotools-devel and geoserver-devel to coordinate the reformat, you can follow it to get news about reformat timings.

aaime:
@jgrocha we haven't heard about this work in a long while, was it dropped?

jgrocha:
Hi Andrea, I just close it for the code reformat, but I want to review it and make the pull request again. Regards, Jorge 


Sent from my mobile. Please excuse the brevity, spelling and punctuation. 
-------- Mensagem original --------De: Andrea Aime <notifications@github.com> Data: 21/07/18  17:12  (GMT+00:00) Para: geotools/geotools <geotools@noreply.github.com> Cc: Jorge Gustavo Rocha <jgr@di.uminho.pt>, Mention <mention@noreply.github.com> Assunto: Re: [geotools/geotools] Make native styles available from
  ResourceInfo class (#1851) 
@jgrocha we haven't heard about this work in a long while, was it dropped?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub, or mute the thread.
{"api_version":"1.0","publisher":{"api_key":"05dde50f1d1a384dd78767c55493e4bb","name":"GitHub"},"entity":{"external_key":"github/geotools/geotools","title":"geotools/geotools","subtitle":"GitHub repository","main_image_url":"https://assets-cdn.github.com/images/email/message_cards/header.png","avatar_image_url":"https://assets-cdn.github.com/images/email/message_cards/avatar.png","action":{"name":"Open in GitHub","url":"https://github.com/geotools/geotools"}},"updates":{"snippets":[{"icon":"PERSON","message":"@aaime in #1851: @jgrocha we haven't heard about this work in a long while, was it dropped?"}],"action":{"name":"View Pull Request","url":"https://github.com/geotools/geotools/pull/1851#issuecomment-406806972"}}}
[
{
"@context": "http://schema.org",
"@type": "EmailMessage",
"potentialAction": {
"@type": "ViewAction",
"target": "https://github.com/geotools/geotools/pull/1851#issuecomment-406806972",
"url": "https://github.com/geotools/geotools/pull/1851#issuecomment-406806972",
"name": "View Pull Request"
},
"description": "View this Pull Request on GitHub",
"publisher": {
"@type": "Organization",
"name": "GitHub",
"url": "https://github.com"
}
},
{
"@type": "MessageCard",
"@context": "http://schema.org/extensions",
"hideOriginalBody": "false",
"originator": "AF6C5A86-E920-430C-9C59-A73278B5EFEB",
"title": "Re: [geotools/geotools] Make native styles available from ResourceInfo class (#1851)",
"sections": [
{
"text": "",
"activityTitle": "**Andrea Aime**",
"activityImage": "https://assets-cdn.github.com/images/email/message_cards/avatar.png",
"activitySubtitle": "@aaime",
"facts": [

]
}
],
"potentialAction": [
{
"name": "Add a comment",
"@type": "ActionCard",
"inputs": [
{
"isMultiLine": true,
"@type": "TextInput",
"id": "IssueComment",
"isRequired": false
}
],
"actions": [
{
"name": "Comment",
"@type": "HttpPOST",
"target": "https://api.github.com",
"body": "{\n\"commandName\": \"IssueComment\",\n\"repositoryFullName\": \"geotools/geotools\",\n\"issueId\": 1851,\n\"IssueComment\": \"{{IssueComment.value}}\"\n}"
}
]
},
{
"targets": [
{
"os": "default",
"uri": "https://github.com/geotools/geotools/pull/1851#issuecomment-406806972"
}
],
"@type": "OpenUri",
"name": "View on GitHub"
},
{
"name": "Unsubscribe",
"@type": "HttpPOST",
"target": "https://api.github.com",
"body": "{\n\"commandName\": \"MuteNotification\",\n\"threadId\": 320297615\n}"
}
],
"themeColor": "26292E"
}
]

aaime:
The full reformat has been done on all branches and now every build will also automatically reformat the code (the build servers instead verify the formatting is the expecting one).

PR #1859, opened 2018-04-18
GEOT-5876: Skip tests if HSTORE Extension can't be created

dromagnoli:


aaime:
This is custom test skipping... which I guess it's needed because the tests on the build server are likely using "skip.on.test=false" to avoid silent failures on connection issues. One way or another this would need custom skip behavior, this is as good as any. +1

PR #1861, opened 2018-04-19
GEOT-5988 Experimental scale2 operation

dromagnoli:


dromagnoli:
@aaime Thanks for the comments. I have addressed them.

PR #1864, opened 2018-04-24
Mosaic issues wrapping existing store

aaime:
Found a couple of issues working on the opensearch module in GeoServer, here are proposed fixes.


dromagnoli (approved):
Looks good to me.

PR #1865, opened 2018-04-24
Mapbox Styles expressions implementation

johngschulz:
This request allows layers to be styled using mapbox style expressions. In addition to being able to read and parse expressions in a Mapbox style,  It includes much but not all of the functionality for the expressions found in the Mapbox style specification here: https://www.mapbox.com/mapbox-gl-js/style-spec/#expressions

Below is a list of the mapbox expressions included in this pull request and explanations of why/where things may be left out, with links to their specifications for easy navigation:

- String expressions: https://www.mapbox.com/mapbox-gl-js/style-spec/#expressions-string
- Color expressions: https://www.mapbox.com/mapbox-gl-js/style-spec/#expressions-color 
    - We left out "rgba", and "to-rgba" expressions because they will not render the alpha channel in Geoserver, and would conflict other functionality to set the alpha channel.
- Types expressions: https://www.mapbox.com/mapbox-gl-js/style-spec/#expressions-types
- Decision expressions: https://www.mapbox.com/mapbox-gl-js/style-spec/#expressions-decision
- Zoom expression: https://www.mapbox.com/mapbox-gl-js/style-spec/#expressions-zoom
- Lookup expressions: https://www.mapbox.com/mapbox-gl-js/style-spec/#expressions-lookup
-Feature Data expressions: https://www.mapbox.com/mapbox-gl-js/style-spec/#expressions-feature-data
    - "properties" is not implemented.  The way the rendering pipeline is set up requires an explicit property name to be declared which is used to populated a list of properties.  The "properties" expression requires a list for all properties found in a layer.  We have not found a work around yet.
- Math expressions: https://www.mapbox.com/mapbox-gl-js/style-spec/#expressions-math

There is more work to be done to round out all expressions included in the specification list the below have not yet been addressed:
- Ramps, Scales, Curves expressions: https://www.mapbox.com/mapbox-gl-js/style-spec/#expressions-ramps-scales-curves
- Variable Binding expressions: https://www.mapbox.com/mapbox-gl-js/style-spec/#expressions-variable-binding
- Heatmap expression: https://www.mapbox.com/mapbox-gl-js/style-spec/#expressions-variable-binding
    - Heatmap expressions also require an implementation of the Mapbox style Heatmap layer


modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/expression/MBColor.java
```diff
@@ -0,0 +1,99 @@
...
+
+    /**
+     * Creates a color value from red, green, blue components, which must range between 0 and 255, and an alpha
+     * component which must range between 0 and 1. If any component is out of range, the expression is an error.
+     * Example: ["rgba", number, number, number, number]: color
+     * @return
+     */
+    // Currently unsupported
+    public Expression colorRGBA(){
+        throw new UnsupportedOperationException("RGBA colors are currently unsupported");
```
jodygarnett: May be able to support by generating the colour, and generating the opacity.

We have a few other examples where a single Mapbox style option is split up across several geotools style attributes.

modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/expression/MBColor.java
```diff
@@ -0,0 +1,99 @@
...
+     * @return
+     */
+    // Currently unsupported
+    public Expression colorToRGBA(){
+        throw new UnsupportedOperationException("RGBA colors are currently unsupported.");
+    }
+
+    @Override
+    public Expression getExpression() throws MBFormatException {
+        switch (name) {
```
jodygarnett: case statement is a good indication that a strategy object could be employed

modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/expression/MBDecision.java
```diff
@@ -0,0 +1,310 @@
...
+     */
+    private Expression decisionNotEqual() {
+        // validate the arg list
+        if (json.size() != 3) {
+            throwUnexpectedArgumentCount(NOT_EQUALS, 2);
+        }
+        // get the comparables
+        Expression comparable1 = parse.string(json, 1);
+        Expression comparable2 = parse.string(json, 2);
+        return ff.function("mbNotEqualTo", comparable1, comparable2);
```
jodygarnett: why make a new "mbNotEqualTo" function when you can use "notEqualTo"?

See geotools function list entry here: http://docs.geotools.org/latest/userguide/library/main/function_list.html#notequalto-a-b-notequalto

jodygarnett: Reading your description there, the handling of null != null sounds indeed different. The fallback comparison based on strings may be a bug in the implementation of "notEqualTo".

modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/function/MBFunction_notEqualTo.java
```diff
@@ -0,0 +1,101 @@
...
+        try { // attempt to get value and perform conversion
+            arg0 = (Object) getExpression(0).evaluate(feature);
+        } catch (Exception e) // probably a type error
+        {
+            throw new IllegalArgumentException(
+                    "Filter Function problem for function equalTo argument #0 - expected type Object");
+        }
+
+        try { // attempt to get value and perform conversion
+            arg1 = (Object) getExpression(1).evaluate(feature);
```
jodygarnett: Logic should attempt to match types using expression.evaulate(feature, arg0.getClass() ), and if that is not available consider fallbacks such as evaluating both to String.

Using evaulate(feature, class) is preferable to using the geotools converters facilities by hand as a given expression may have a better idea how to evaluate to a number, color, or string as required.

modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/function/MBFunction_toBool.java
```diff
@@ -0,0 +1,66 @@
...
+ * Take an object as an argument and returns a boolean value.
+ * The result is false when then input is an empty string, 0, false, null,
+ * or NaN; otherwise it is true.
+ *
+ * This function is a helper to accommodate MBType expressions. Expressions in this section are provided for the purpose of
+ * testing for and converting between different data types like strings, numbers, and boolean values.
+ */
+class MBFunction_toBool extends FunctionExpressionImpl {
+
+    public static FunctionName NAME = new FunctionNameImpl("mbToBool");
```
jodygarnett: would prefer making a naming convention here if we really need to introduce casting: toBoolean, toInteger, toFloat, etc....

modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/function/MBFunction_toColor.java
```diff
@@ -0,0 +1,140 @@
...
+ * "rgba(int, int, int, double)"
+ *
+ * And in arrays of 3 and 4 numbers for rgb and rgba colors:
+ * [int, int, int]
+ * [int, int, int, double]
+ *
+ * The integer values should be between 0-255, for red, green, and blue color values.
+ * The double value should be between 0-1, and is converted to a 0-255 alpha value.
+ */
+class MBFunction_toColor extends FunctionExpressionImpl {
```
jodygarnett: This "_" thing is due to the original FilterFunctions being autogenerated, please follow normal java class naming conventions if you are making functions.

modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/function/MBFunction_toColor.java
```diff
@@ -0,0 +1,140 @@
...
+    public Object evaluate(Object feature) {
+        Color c;
+        for (Integer i = 1; i <= this.params.size() - 1; i++) {
+            Object evaluation = this.params.get(i).evaluate(feature);
+            if (evaluation instanceof Color) {
+                return evaluation;
+            }
+            if (evaluation instanceof JSONArray) {
+                JSONArray je = (JSONArray) evaluation;
+                if (je.size() == 3 || je.size() == 4) {
```
jodygarnett: Think we made similar logic for processing of color constants, can some thought be given to avoiding duplication of logic.

modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/function/MBFunction_toNumber.java
```diff
@@ -0,0 +1,77 @@
...
+/**
+ * Converts the input value to a number, if possible. If the input is null or false, the result is 0. If the input
+ * is true, the result is 1. If the input is a string, it is converted to a number as specified by the
+ * "ToNumber Applied to the String Type" algorithm of the ECMAScript Language Specification. If multiple values are
+ * provided, each one is evaluated in order until the first successful conversion is obtained.
+ * If none of the inputs can be converted, the expression is an error.
+ */
+class MBFunction_toNumber extends FunctionExpressionImpl {
+
+    public static FunctionName NAME = new FunctionNameImpl("mbToNumber");
```
jodygarnett: See prior comment on setting up a naming convention.

modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/parse/MBObjectParser.java
```diff
@@ -318,7 +326,7 @@ public String get(JSONArray json, int index) {
 
     /**
      * Quickly access required json tag (as a String).
-     * 
+     *
```
jodygarnett: lots of whitespace changes here, mask if this file has actually been modified

modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/function/ListSizeFunction.java
```diff
@@ -0,0 +1,54 @@
+package org.geotools.mbstyle.function;
```
tbarsballe: This should come after the header

modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/function/MBFunction_any.java
```diff
@@ -0,0 +1,97 @@
...
+ *     <td>["mbAny", true, true, false]</td>
+ *     <td align="center">true</td>
+ *   </tr>
+ *   <tr>
+ *     <td>["mbAny", false, false, false, false]</td>
+ *     <td align="center">false</td>
+ *   </tr>
+ * </table>
+ * </p>
+ */class MBFunction_any extends FunctionExpressionImpl {
```
tbarsballe: Missing newline?

modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/function/MBFunction_notEqualTo.java
```diff
@@ -0,0 +1,101 @@
...
+        } catch (Exception e) // probably a type error
+        {
```
tbarsballe: This should be on line 93 (use trailing open-braces)

modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/function/MBFunction_toColor.java
tbarsballe: I believe some of that logic was in the YSLD extension (specifically `#RGB` colors), so duplicating that can't really be avoided (You could pull it out to a util class, but both work slightly differently I think). Any duplication that can be avoided would be good though.

modules/unsupported/mbstyle/src/test/java/org/geotools/mbstyle/expression/MBMathTest.java
```diff
@@ -0,0 +1,282 @@
...
+package org.geotools.mbstyle.expression;
+
+import static org.junit.Assert.assertEquals;
+
+import org.json.simple.JSONObject;
+import org.junit.Test;
+
+
+/**
+ *
```
tbarsballe: Empty javadoc is not very helpful

tbarsballe:
With respect to @jodygarnett s comments about mb-prefixed functions - any functions which do represent MBStyle-specific functionality should remain prefixed (In particular, the equals/notEquals functions). But more general-purpose functions would be fine with having more general names (and maybe eventually moved to core, if they are considered sufficiently useful).

To me, it does look like most of the added functions are specific to MBStyle (or at least not useful outside of MBStyles), but I don't have a particularly strong opinion either way.

modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/expression/MBDecision.java
emerkle826: @jodygarnett The GeoTools implementations for "notEqualTo" and "equalTo" (and the StaticGeometry methods they call) haven't changed since early 2011. While it may be a bug/oversight, I would assume things relying on those functions expect them to behave as they are, buggy or not.

modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/function/MBFunction_notEqualTo.java
emerkle826: I guess the reason I did it this way is that you can't really tell what Class type the expression will evaluate to in some cases. And I can't use a String fallback because in MapBox world, `"true"` is not equal to `Boolean.TRUE`. Doing a String comparison of them would result in a false positive for equalTo (or false negative for notEqualTo).

Since both these functions use the same logic (one is just the "not" of the other), I was able to keep the logic in one place (https://github.com/geotools/geotools/pull/1865/files#diff-7d33831b31b1087f88bd7b4f497dd00bR34). That utility method handles the class type comparisons and relies on the `equals()` implementation of the evaluated expressions, which seems to be how the expression works in a MapBox style.

modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/parse/MBObjectParser.java
emerkle826: I tried to revert the whitespace changes, that is annoying. I think I got all but one.

modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/function/ListSizeFunction.java
emerkle826: I should have gotten all the format issues, would appreciate one last double check.

emerkle826:
@jodygarnett @tbarsballe I think @johngschulz is addressing a few of the review items here. I tried to address some of the bulk items (function names, class names, formatting, etc). Let me or @johngschulz know if the new function names/class names are still inadequate. Thanks!

aaime:
Mass code reformat incoming during the weekend (as discussed on the lists), it will make this PR much harder to merge. Can you merge it in time?

johngschulz:
@tbarsballe @jodygarnett, I was able to to remove down on the duplication I had in the toColor function. The 3 hex value isn't handled correctly by the converter, and is still in place.  Let us know if we can do to get this in before the reformat, thanks!

tbarsballe:
@aaime - We should be able to get this merged in today.

aaime:
@tbarsballe that's good news, thank you!

tbarsballe:
Okay, this passes, everything looks fine.
@jodygarnett - Are you satisfied with the changes? I would like to get this merged so as not to interfere with the reformat.

PR #1867, opened 2018-04-26
[GEOT-5993]: ImageMosaic's Virtual Native Resolution

dromagnoli:


modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/OverviewsController.java
```diff
@@ -89,41 +90,72 @@ int pickOverviewLevel(final OverviewPolicy policy, final double[] requestedResol
...
+            reqy = requestedResolution[1];
+            if (useVirtual) {
+                virtualRequestedScaleFactorX = virtualNativeResolution[0] / max.resolutionX;
+                virtualRequestedScaleFactorY = virtualNativeResolution[1] / max.resolutionY;
+                virtualRequestedScaleFactor = (virtualRequestedScaleFactorX <= virtualRequestedScaleFactorY) ? 
+                        virtualRequestedScaleFactorX : virtualRequestedScaleFactorY;
+
+            }
+        }
+        if (!useVirtual && requestedResolution == null){
```
aaime: I'd add a comment as to why we exit, or not, here

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/GranuleDescriptor.java
```diff
@@ -1033,6 +1034,10 @@ public GranuleLoadingResult loadRaster(final ImageReadParam imageReadParameters,
             // image sizes.
             //
             // place it in the mosaic using the coords created above;
+            if (virtualNativeResolution != null && !Double.isNaN(virtualNativeResolution[0]) && !Double.isNaN(virtualNativeResolution[1])){
```
aaime: Where do you check that the requested resolution is higher than the virtual one? Like, if we are reading an overview that has lower resolution than the virtual one, no rescaling should be applied, right?

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/OverviewsController.java
dromagnoli: The old code was returning 0 (first image, highres) as soon as requestedResolution is null.
The new code should check if a different imageIndex need to be returned when virtualResolution has been specified so it need to perform some additional checks.

I'm adding comment to the code too.

PR #1870, opened 2018-04-28
Big bad reformat

aaime:
Here we go with the reformat. Let's see if this mammoth pull request causes trouble with github or travis


aaime:
Ha ha ha:
selezione_175

PR #1876, opened 2018-05-05
[GEOT-6007] Build on Windows not working due to failing tests

vjeetje:
Fixed build For Windows users.

Current Jenkins build is failing for a while now:
http://winbuild.geo-solutions.it/jenkins/view/GeoTools/job/GeoTools-Master/

### DIFF
--- a/modules/library/coverage/src/main/java/org/geotools/coverage/grid/io/footprint/SidecarFootprintProvider.java
+++ b/modules/library/coverage/src/main/java/org/geotools/coverage/grid/io/footprint/SidecarFootprintProvider.java
@@ -114,7 +114,7 @@ public class SidecarFootprintProvider implements FootprintGeometryProvider {
             return getFootprint(reference.getAbsolutePath());
         } else {
             Object value = feature.getAttribute(FOOTPRINT_LOCATION_ATTRIBUTE);
-            if (value != null && value instanceof String) {
+            if (value instanceof String && !((String) value).matches("^(?i)https?://.*$")) {
                 String strValue = (String) value;
                 String path = getFullPath(strValue);
                 return getFootprint(path);
@@ -219,7 +219,7 @@ public class SidecarFootprintProvider implements FootprintGeometryProvider {
         if (!file.isAbsolute()) {
             file = new File(reference, strValue);
         }
-        return file.getPath();
+        return file.getCanonicalPath();
     }
 
     @Override
--- a/modules/library/coverage/src/main/java/org/geotools/coverage/grid/io/footprint/SidecarFootprintProvider.java
+++ b/modules/library/coverage/src/main/java/org/geotools/coverage/grid/io/footprint/SidecarFootprintProvider.java
@@ -219,7 +219,7 @@ public class SidecarFootprintProvider implements FootprintGeometryProvider {
         if (!file.isAbsolute()) {
             file = new File(reference, strValue);
         }
-        return file.getCanonicalPath();
+        return file.getPath();
     }
 
     @Override
--- a/modules/library/metadata/src/test/java/org/geotools/util/URLsTest.java
+++ b/modules/library/metadata/src/test/java/org/geotools/util/URLsTest.java
@@ -101,8 +101,8 @@ public class URLsTest {
         assertURL(replaceSlashes("C:\\one"), "file://C:/one");
         assertURL(replaceSlashes("C:\\one\\two"), "file://C:/one/two");
         assertURL(replaceSlashes("C:\\one\\two\\and three"), "file://C:/one/two/and three");
-        assertEquals("sample", URLs.urlToFile(new URL("file:sample?query")).toString());
-        assertEquals("sample", URLs.urlToFile(new URL("file:sample#ref")).toString());
+        assertURL("sample", "file:sample?query");
+        assertURL("sample", "file:sample#ref");
         File file = File.createTempFile("hello", "world");
         handleFile(file.getAbsolutePath());
         handleFile(file.getPath());
@@ -110,8 +110,8 @@ public class URLsTest {
         URL url = new URL("file", "////oehhwsfs09", "/some/path/on/the/server/filename.nds");
         File windowsShareFile = URLs.urlToFile(url);
         assertNotNull(windowsShareFile);
-        assertEquals("file café", URLs.urlToFile(new URL("file:file%20caf%C3%A9")).toString());
-        assertEquals("/file café", URLs.urlToFile(new URL("file:/file%20caf%C3%A9")).toString());
-        assertEquals("file café", URLs.urlToFile(new URL("file://file%20caf%C3%A9")).toString());
+        assertURL("file café", "file:file%20caf%C3%A9");
+        assertURL("/file café", "file:/file%20caf%C3%A9");
+        assertURL("file café", "file://file%20caf%C3%A9");
     }
 }
--- a/modules/plugin/coverage-multidim/netcdf/src/test/java/org/geotools/coverage/io/netcdf/NetCDFMosaicReaderTest.java
+++ b/modules/plugin/coverage-multidim/netcdf/src/test/java/org/geotools/coverage/io/netcdf/NetCDFMosaicReaderTest.java
@@ -122,7 +122,7 @@ public class NetCDFMosaicReaderTest extends Assert {
         File nc1 = TestData.file(this, "polyphemus_20130301_test.nc");
         File mosaic = new File(TestData.file(this, "."), "nc_harvest1");
         if (mosaic.exists()) {
-            FileUtils.deleteDirectory(mosaic);
+            FileUtils.forceDelete(mosaic);
         }
         assertTrue(mosaic.mkdirs());
         FileUtils.copyFileToDirectory(nc1, mosaic);
@@ -232,7 +232,7 @@ public class NetCDFMosaicReaderTest extends Assert {
         File nc1 = TestData.file(this, "polyphemus_20130301_test.nc");
         File mosaic = new File(TestData.file(this, "."), "nc_poly_hetero");
         if (mosaic.exists()) {
-            FileUtils.deleteDirectory(mosaic);
+            FileUtils.forceDelete(mosaic);
         }
         assertTrue(mosaic.mkdirs());
         FileUtils.copyFileToDirectory(nc1, mosaic);
@@ -295,7 +295,7 @@ public class NetCDFMosaicReaderTest extends Assert {
         File nc1 = TestData.file(this, "polyphemus_20130301_NO2_time2.nc");
         File mosaic = new File(TestData.file(this, "."), "nc_time2");
         if (mosaic.exists()) {
-            FileUtils.deleteDirectory(mosaic);
+            FileUtils.forceDelete(mosaic);
         }
         assertTrue(mosaic.mkdirs());
         FileUtils.copyFileToDirectory(nc1, mosaic);
@@ -319,6 +319,7 @@ public class NetCDFMosaicReaderTest extends Assert {
         ImageMosaicFormat format = new ImageMosaicFormat();
         ImageMosaicReader reader = format.getReader(mosaic);
         checkCustomTimeAttribute(nc1, reader);
+        reader.dispose();
     }
 
     @Test
@@ -342,7 +343,7 @@ public class NetCDFMosaicReaderTest extends Assert {
         File nc1 = TestData.file(this, "polyphemus_20130301_NO2_time2.nc");
         File mosaic = new File(TestData.file(this, "."), "nc_time2");
         if (mosaic.exists()) {
-            FileUtils.deleteDirectory(mosaic);
+            FileUtils.forceDelete(mosaic);
         }
         assertTrue(mosaic.mkdirs());
         FileUtils.copyFileToDirectory(nc1, mosaic);
@@ -378,6 +379,7 @@ public class NetCDFMosaicReaderTest extends Assert {
         assertEquals(1, typeNames.size());
         assertTrue(typeNames.contains("NO2"));
         dataStore.dispose();
+        reader.dispose();
     }
 
     public void checkCustomTimeAttribute(File nc1, ImageMosaicReader reader) throws IOException {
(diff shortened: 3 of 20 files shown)


pom.xml
```diff
@@ -1530,7 +1524,7 @@
...
           </excludes>
-          <argLine>-Xmx${test.maxHeapSize} ${jvm.opts} -Dorg.geotools.test.extensive=${extensive.tests} -Dorg.geotools.test.interactive=${interactive.tests} -Dorg.geotools.image.test.skip=${skip.image.tests} -Dorg.geotools.image.test.interactive=${interactive.image} -Djava.awt.headless=${java.awt.headless} -Dsun.java2d.d3d=${sun.java2d.d3d} -Djava.io.tmpdir="${java.io.tmpdir}" -Djava.library.path="${java.library.path}" ${test.args}  -XX:+TieredCompilation -XX:TieredStopAtLevel=1 -XX:+IgnoreUnrecognizedVMOptions --add-modules=java.xml.bind --add-modules=java.activation</argLine>
+          <argLine>-Xmx${test.maxHeapSize} ${jvm.opts} -Dorg.geotools.test.extensive=${extensive.tests} -Dorg.geotools.test.interactive=${interactive.tests} -Dorg.geotools.image.test.skip=${skip.image.tests} -Dorg.geotools.image.test.interactive=${interactive.image} -Djava.awt.headless=${java.awt.headless} -Dsun.java2d.d3d=${sun.java2d.d3d} -Djava.io.tmpdir=${java.io.tmpdir} -Djava.library.path="${java.library.path}" ${test.args}  -XX:+TieredCompilation -XX:TieredStopAtLevel=1 -XX:+IgnoreUnrecognizedVMOptions --add-modules=java.xml.bind --add-modules=java.activation</argLine>
```
aaime: Why the change here?

modules/plugin/coverage-multidim/netcdf/src/test/java/org/geotools/coverage/io/netcdf/NetCDFMosaicReaderTest.java
```diff
@@ -231,7 +231,7 @@ public void testHeterogeneous()
         File nc1 = TestData.file(this, "polyphemus_20130301_test.nc");
         File mosaic = new File(TestData.file(this, "."), "nc_poly_hetero");
         if (mosaic.exists()) {
-            FileUtils.deleteDirectory(mosaic);
+            FileUtils.forceDelete(mosaic);
```
aaime: So all these changes from deleteDirectory to forceDelete, are they there to ensure an exception is thrown in case the deletion does not work? Or... something else?

modules/library/coverage/src/main/java/org/geotools/coverage/grid/io/footprint/SidecarFootprintProvider.java
```diff
@@ -219,7 +219,7 @@ private String getFullPath(String strValue) throws IOException {
         if (!file.isAbsolute()) {
             file = new File(reference, strValue);
         }
-        return file.getCanonicalPath();
+        return file.getPath();
```
aaime: Can you explain this change?

pom.xml
vjeetje: This is to give better support for people who a space in their directory structure towards their java library / temp directory. It gave me a lot of troubles for running the tests in Intellij.

modules/plugin/coverage-multidim/netcdf/src/test/java/org/geotools/coverage/io/netcdf/NetCDFMosaicReaderTest.java
vjeetje: During the running of the tests I noticed they were sometimes struggling with the deleteDirectory. I'm not sure why tbh, I guess some other tests are influencing this one. I found it hard to locate the tests that were influencing this one, so I tried forceDeleting it. That solved the problem for me. It's a recurring theme in the tests that they sometimes don't cleanup properly.

modules/library/coverage/src/main/java/org/geotools/coverage/grid/io/footprint/SidecarFootprintProvider.java
vjeetje: There is a test that struggles on Windows with constructing the full path. It was a mosaic test, I forgot the exact name. On line 220 in that test the reference was a file path to the target folder of the test e.g. c/windows/.../target/ and the strValue was something like http://localhost:9080/mosaic. So the file path became: c/windows/.../target/http://localhost:9080/mosaic which doesn't make any sense. The test data that was given, seemed ok at the right places. The file path it constructed wasn't really a valid path. When java tries to make it canonical (standardise) it, it breaks. I see no reason why someone would want to get the canonical version of the path. The only reason one would want to do that, is to compare file paths.

pom.xml
aaime: @vjeetje I don't understand, with paths with spaces the double quotes should be preserved, not removed  "tmpdir" as you did above?

aaime:
I have tested the build on Linux in a path without spaces, seems to work fine.
@dromagnoli could you give it a try on Windows?
@bencaradocdavies you have a "linux build with spaces in path" scenario right? Any chance you could give this a kick too?

bencaradocdavies:
@aaime my build is in a path with spaces but not my tmpdir or java installation. I just built GeoTools with `-Djava.io.tmpdir="/tmp/geotools with spaces"` as a test and it built fine *without* this pull request. I do not think the quotes around the `argLine`  `tmpdir` expansion should be removed. @vjeetje I do not understand why you want to make these `argLine` changes.

pom.xml
bencaradocdavies: @vjeetje please explain or revert the `argLine` `tmpdir` quote removal.

modules/library/coverage/src/main/java/org/geotools/coverage/grid/io/footprint/SidecarFootprintProvider.java
bencaradocdavies: @vjeetje this must be a bug in the code or a test. Nothing should be calling this method with the string encoding of a URL. @aaime is this a bug in `SidecarFootprintProvider`? Can the `location` be an HTTP URL?

bencaradocdavies:
It would also be nice if the `dispose` calls were in `finally` blocks. It would be even nicer if `ImageReader` implemented `AutoCloseable`, then you could use try-with-resources.

modules/library/metadata/src/test/java/org/geotools/util/URLsTest.java
```diff
@@ -101,17 +101,17 @@ public void testUrlToFile() throws Exception {
         assertURL(replaceSlashes("C:\\one"), "file://C:/one");
         assertURL(replaceSlashes("C:\\one\\two"), "file://C:/one/two");
         assertURL(replaceSlashes("C:\\one\\two\\and three"), "file://C:/one/two/and three");
-        assertEquals("sample", URLs.urlToFile(new URL("file:sample?query")).toString());
-        assertEquals("sample", URLs.urlToFile(new URL("file:sample#ref")).toString());
+        assertURL("sample", "file:sample?query");
```
bencaradocdavies: @vjeetje these `assertEquals` to `assertURL` changes are good. I do not know how we missed these. Thanks!

modules/library/coverage/src/main/java/org/geotools/coverage/grid/io/footprint/SidecarFootprintProvider.java
aaime: @bencaradocdavies before Niels changes to allow mosaic pulling from random URL definitely not, now I'm not so sure.

aaime:
@bencaradocdavies I agree on AutoCloseable being the way to go, but don't think it's on @vjeetje plate to do so, it's an API change that will require a GSIP (switching from "dispose" to "close", deprecate the older, fix all calling points). If you're interested maybe we can do it as a team and split the work (in a separate pull request)?

bencaradocdavies:
@aaime agreed that `AutoCloseable` is out of scope for this PR. There are many interfaces that should be considered; I will email the lists to gauge interest and scope.

modules/library/coverage/src/main/java/org/geotools/coverage/grid/io/footprint/SidecarFootprintProvider.java
vjeetje: @aaime @bencaradocdavies The test file that is containing the url which was parsed is: org/geotools/gce/imagemosaic/test-data/remote_test/remote_test.dbf . It contains the remote url  "http://localhost:8089/global_mosaic_0.png". Is this a stable feature that somehow works on linux?

aaime: That was recently introduced in https://github.com/geotools/geotools/pull/1869 and already caused another regression fixed in https://github.com/geotools/geotools/pull/1910
It does not break the build on Linux, but don't believe it's actually working either, that code should probably have been amended to also work with URLs instead of files. I'm however not sure why it gets into the footprint management path...

pom.xml
vjeetje: @bencaradocdavies This is some weird Intellij interaction with maven-surefire-plugin. I'm not sure about the cause but running a test on Intellij with -Djava.io.tmpdir="${java.io.tmpdir}" -Djava.library.path="${java.library.path}" fails with a parsing problem. Removing the quotes around tmpdir fixes those issues. The build in general works fine though without my changes. I'll revert those changes as I can't find a solution that works in general on all systems with the IDE.  I've added a screenshot showing the work-around for those on Windows using Intellij. Just uncheck the argline checkbox.
argline_intellij

vjeetje: @aaime see my reply at the code change request of @bencaradocdavies

vjeetje:
I've added my changes in a new commit:

- Revert the ArgLine quoting for the maven-surefire-plugin. See change request @bencaradocdavies 
- I've fixed some tests which were broken again. The cause could be that the tests are flaky or some new changes introduced some issues on Windows again.

This reminds me; how can we assure (once this pull request is accepted) that the Windows build will keep succeeding? Can we configure Travis / enforce some rule?

bencaradocdavies:
@vjeetje GeoSolutions had a Windows Jenkins instance but it has been muted(?) for some time because the build was failing. Once the Windows build is working, and we have Windows developers available to fix Windows-only failures, Windows build notifications could be re-enabled. @aaime asked @dromagnoli to try building the PR on Windows, which suggests that we may have someone with a Windows development environment.

I only use Linux but always build in a path with spaces to catch File<->URL conversion problems, which are a common source of failures on Windows and non-ASCII platforms.

aaime:
@vjeetje Ben is correct, we'll need further help to keep it going.
I hate working on windows, but I was the one to push for setting up that build server and get the build going by crushing a few days off in a Christmas break on the accursed OS ;-)
I did so because multiple people complained about the build not working on Windows, so was trying to be a good citizen.
Unfortunately all complainers disappeared once the build was fixed and the build went back to broken state within a relatively short time (a few months I think).
So, the build server is there, we can re-enable hate mails from it, but we'll also need some support from someone that actually cares to keep the Windows build going (as in, cares about that more than about other things to do in this community, for which we have plenty).

vjeetje:
@aaime Maybe checking the Windows build should be part of the code reviewing process.

aaime:
@vjeetje there are two issues:
* We don't have a windows build for pull requests
* Pull requests are not required for direct project committers (the vast majority of are not using windows), which normally will use PRs only if they think a review is needed, or if they feel like having a second opinion from Travis

For the first one, are you interested in doing research and propose an approach? 
Believe there are free build server that could be paired with PR checks, but never looked into it.

For the second one, I personally would not want to go through PRs for every commit, would slow me down too much, but if a majority or the other devs decide it's a good idea, I'll eat crow.

aaime:
@vjeetje both are a no go, mostly because you're using the wrong pronoun ;-)
"We" should setup AppVeyour (on our own spare time, nobody is sponsoring windows support) and "we" should do PR and review for every change (same as above, nobody is authorized to use extra working hours for this).

The main problem is that "we" are already way overbooked, and are cutting dead branches on whatever we cannot maintain with extra effort (we are actually in the process dropping modules down to unsupported to reduce the load).

It really boils down to was "you" can do, as a new contributor, to help with the Windows build. Otherwise I'm skeptical anything of significance will happen.

aaime:
@bencaradocdavies aside from the other discussions, are you happy with the changes? Is the pull request mergeable now?

NielsCharlier:
It seems indeed that I might have missed another URL-to-file conversion, for example
https://github.com/geotools/geotools/blob/master/modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/ImageMosaicConfigHandler.java#L946

None of my test broke on this though, not sure why

aaime:
@NielsCharlier see the one in this pull request, looks like on Linux the conversion is lenient, on Windows is not. Probably just one of the possible causes.

dromagnoli:
@aaime I successfully did the build on Windows... no issues there

modules/library/coverage/src/main/java/org/geotools/coverage/grid/io/footprint/SidecarFootprintProvider.java
bencaradocdavies: I think that the `file.getCanonicalPath()` to `file.getPath()` change in `SidecarFootprintProvider.getFullPath` is the wrong fix because this calling code should not be passing the string representation of an HTTP(S) URL in the first place. I would prefer a guard in the calling code to prevent HTTP(S) URLs from being used until they are supported. For example, change line 117 to:
```
            if (value != null && value instanceof String
                    && !((String) value).matches("https?://.*")) {
```

vjeetje: Done

bencaradocdavies:
Thanks @vjeetje, that looks good. I also note your improvements to the guard regex.

PR #1877, opened 2018-05-07
[GEOT-6008]: Supporting multilevel sidecar files

dromagnoli:


modules/library/coverage/src/main/java/org/geotools/coverage/grid/io/footprint/MultiLevelROIProviderFactory.java
```diff
@@ -71,7 +90,7 @@ public static MultiLevelROIProvider createFootprintProvider(File referenceFile)
      * @throws Exception
      */
     public static MultiLevelROIProvider createFootprintProvider(
-            final File referenceFile, Geometry granuleBounds) {
+            final File referenceFile, Geometry granuleBounds, Hints hints) {
```
aaime: The hints are not really used here, probably a leftover from a failed experiment?

dromagnoli: mmm nope. 
this implementation do not use hints, while the MultiLevelROIProviderMosaicFactory is actually using it

aaime: But it's a static method, there is no notion of override...

dromagnoli:
Indeed my intention was to update the GeoServer doc too with new properties and usage. 
I will do it right now and refer it here

modules/library/coverage/src/main/java/org/geotools/coverage/grid/io/footprint/MultiLevelROIProviderFactory.java
dromagnoli: Ah! Right... nope... no reflection. I think that at the very beginning I was having issues with the MultiLevelROIProviderMosaicFactory subclass methods. 
I'm going to remove it, thanks.

dromagnoli:
PR for GeoServer documentation update, ready:
https://github.com/geoserver/geoserver/pull/2876

PR #1878, opened 2018-05-07
[GEOT-5920] WFSClient fails when a function is used in a query

ianturton:
Infact anything that encodes/decodes a Like filter with a LHS function will fail.


ianturton:
closed due to reformat issues - replaced by https://github.com/geotools/geotools/pull/1883

PR #1883, opened 2018-05-08
[GEOT-5920] WFSClient fails when a function is used in a query

ianturton:
Infact anything that encodes/decodes a Like filter with a LHS function will fail.


modules/extension/xsd/xsd-filter/src/main/java/org/geotools/filter/v1_0/OGCPropertyIsLikeTypeBinding.java
```diff
@@ -83,7 +83,7 @@ public Class getType() {
      * @generated modifiable
      */
     public Object parse(ElementInstance instance, Node node, Object value) throws Exception {
-        PropertyName name = (PropertyName) node.getChildValue(PropertyName.class);
+        Expression name = (Expression) node.getChildValue(Expression.class);
```
bencaradocdavies: Aha! If the `Literal` is first, could it be used for the name as well? See [GEOS-8738](https://osgeo-org.atlassian.net/browse/GEOS-8738).

PR #1885, opened 2018-05-09
[GEOS-8736] Extend App-Schema to make it possible to use an HTTP URL for the mapping file location

nmco:
Associated issue:
https://osgeo-org.atlassian.net/browse/GEOS-8736

This PR adds the necessary integration tests, a HTTP static server is instantiated _on the fly_, the documentation was also updated.


modules/extension/app-schema/app-schema/src/test/java/org/geotools/http/HttpStaticServer.java
```diff
@@ -0,0 +1,187 @@
...
+                        String.format(
+                                "Resource '%s' with path '%s' not found.",
+                                resourceName, resourcePath));
+            }
+            // read the resource from the classpath
+            String resource = IOUtils.toString(input);
+            // substitute host and port place holders
+            String relativePath =
+                    URLs.getParentUrl(this.getClass().getResource(resourcePath)).getPath();
+            resource = resource.replaceAll("\\{relative\\}", relativePath);
```
bencaradocdavies: @nmco these parameters should be documented in the user manual. They might also conflict with [app-schema property interpolation](http://docs.geoserver.org/latest/en/user/data/app-schema/property-interpolation.html#using-properties). For example, a user might want to use `${host}` and `${port}`. Consider using the app-schema `InterpolationProperties` class, the `${propertyname}` syntax, and longer property names like `${app-schema.http-server.relative}`, `${app-schema.http-server.host}`, and `${app-schema.http-server.port}`, or even some leading underscores to protect users from surprises. I am not known for brevity so will defer to your good taste and judgement.  :-P

nmco: @bencaradocdavies this are just placeholders for the tests since the port is randomized :-P

PR #1892, opened 2018-05-14
geopackage mosaics should use tile matrix set bounds [GEOS-8503]

rfecher:
geopackages weren't previously using tile matrix set extents and only the `gpkg_contents`.  According to this, the tile matrix set extents should be used for calculating tile bounds: https://www.geopackage.org/guidance/getting-started.html#gpkg_contents


tbarsballe:
Last CI build failed on a compilation failure in gt-geopkg. Restarting the CI build to see if it was fixed by recent changes, but if the failure persists it will have to be fixed before this can be merged

rfecher:
its back to passing again.  it was passing before but has been sitting here awhile.  any chance it can get reviewed before there's another breaking change?  Its a bug fix in that it previously ignored the tile matrix extent bounds which the spec said should be used rather than the `gpkg_contents`.  Per the GEOS-8503 issue reported it resulted in geopackages that showed up fine in QGIS but didn't work correctly in GeoServer.

aaime:
Review re-assigned to module maintainers

tbarsballe:
@ianturton or @etj - are you able to review the recent changes to this PR? Thanks.

simboss:
I have assigned this to @NielsCharlier since this is geopackage

rfecher:
@NielsCharlier thanks for the review.  Having been approved weeks ago, is it now someone else who will pull this in?

PR #1897, opened 2018-05-18
[GEOT-4984] Add test for TemporalConversion

bradh:
Looks like the problem was resolved by Ian as part of https://github.com/geotools/geotools/commit/d690a9695eaf90f47aada51fd22e6af6e1b1a142#diff-b9e6e947d7221cb94fd57bb37eb1bd0e back in 2015.
This just adds the test case.


bradh:
Interesting that the new test fails on Travis.

PR #1902, opened 2018-05-22
[GEOT-6024] Allow usage of SVGs as mark sources (fillable/strokable shapes, instead of finished symbols)

aaime:


modules/extension/xsd/xsd-sld/src/main/java/org/geotools/sld/bindings/SLDMarkBinding.java
```diff
@@ -114,6 +121,13 @@ public Object parse(ElementInstance instance, Node node, Object value) throws Ex
         Mark mark = styleFactory.createMark();
 
         if (wkName != null) {
+            // allow references to files containing mark definitions
+            if (wkName.startsWith("file://")) {
+                URL url = resourceLocator.locateResource(wkName);
+                if (url != null) {
```
nmco: Would an warning message be appropriate where in the case the SVG is not found ?

modules/library/main/src/main/java/org/geotools/styling/DefaultResourceLocator.java
```diff
@@ -71,6 +72,23 @@ public URL locateResource(String uri) {
         return url;
     }
 
+    /**
+     * Variant to URL.getQuery() that won't balk at having # in the query (e.g., a color in a
+     * parametric SVG reference)
```
nmco: I would put here an URL example, it would help understanding what the method does.

modules/library/main/src/main/java/org/geotools/styling/DefaultResourceLocator.java
```diff
@@ -71,6 +72,23 @@ public URL locateResource(String uri) {
...
+     *
+     * @param url
+     * @return
+     */
+    private String getQueryPart(URL url) {
+        String external = url.toExternalForm();
+        int idx = external.indexOf('?');
+        if (idx == -1 || idx == external.length() - 1) {
+            return null;
+        } else {
```
nmco: Unnecessary ``else`` it can return directly.

modules/library/main/src/main/java/org/geotools/styling/SLDParser.java
```diff
@@ -1713,7 +1713,17 @@ protected Mark parseMark(Node root) {
...
                     LOGGER.finest("setting mark to " + getFirstChildValue(child));
-                mark.setWellKnownName(parseCssParameter(child));
+                Expression wellKnownName = parseCssParameter(child);
+                if (wellKnownName instanceof Literal) {
+                    String expanded = wellKnownName.evaluate(null, String.class);
+                    if (expanded != null && expanded.startsWith("file://")) {
+                        URL url = onlineResourceLocator.locateResource(expanded);
+                        if (url != null) {
+                            wellKnownName = ff.literal(url.toExternalForm());
+                        }
```
nmco: If the resource is not found here it will never be found right ? I would a log warning.

modules/plugin/svg/src/main/java/org/geotools/renderer/style/RenderableSVG.java
```diff
@@ -0,0 +1,105 @@
+/* (c) 2017 Open Source Geospatial Foundation - all rights reserved
```
nmco: Should be 2018, at least I hope so :P

modules/plugin/svg/src/main/java/org/geotools/renderer/style/RenderableSVG.java
```diff
@@ -0,0 +1,105 @@
...
+
+class RenderableSVG {
+    Rectangle2D bounds;
+
+    GraphicsNode node;
+
+    public RenderableSVG(Document doc) {
+        this.node = getGraphicNode(doc);
+        this.bounds = getSvgDocBounds(doc);
+        if (bounds == null) bounds = node.getBounds();
```
nmco: Missing braces, I'm not sure if the original GT code guidelines still apply since we move to google format.

modules/plugin/svg/src/main/java/org/geotools/renderer/style/RenderableSVGCache.java
```diff
@@ -0,0 +1,260 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ *
+ *    (C) 2017, Open Source Geospatial Foundation (OSGeo)
```
nmco: Should be 2018.

modules/plugin/svg/src/main/java/org/geotools/renderer/style/RenderableSVGCache.java
```diff
@@ -0,0 +1,260 @@
...
+            }
+        } else {
+            this.resolver = GeoTools.getEntityResolver(null);
+        }
+    }
+
+    public RenderableSVG getRenderableSVG(Feature feature, Expression url, String format)
+            throws Exception {
+        // check we do support the declared format
+        if (format == null || !formats.contains(format.toLowerCase())) return null;
```
nmco: Missing braces.

modules/plugin/svg/src/main/java/org/geotools/renderer/style/RenderableSVGCache.java
```diff
@@ -0,0 +1,260 @@
...
+
+        // grab the url
+        String svgfile = url.evaluate(feature, String.class);
+        if (svgfile == null) {
+            throw new IllegalArgumentException(
+                    "The specified expression could not be turned into an URL");
+        } else {
+            // just for validation parse the URL
+            if (Converters.convert(svgfile, URL.class) == null) {
```
nmco: Interesting way of validating the URL.

modules/plugin/svg/src/main/java/org/geotools/renderer/style/RenderableSVGCache.java
```diff
@@ -0,0 +1,260 @@
...
+        RenderableSVG svg = glyphCache.get(svgfile);
+        if (svg == null) {
+            String parser = XMLResourceDescriptor.getXMLParserClassName();
+            SAXSVGDocumentFactory f =
+                    new SAXSVGDocumentFactory(parser) {
+                        @Override
+                        public InputSource resolveEntity(String publicId, String systemId)
+                                throws SAXException {
+                            InputSource source = super.resolveEntity(publicId, systemId);
+                            if (source == null) {
```
nmco: So basic we give it a first try with the default resolver, and then only if it fails we try the one provided one as an hint ?

modules/plugin/svg/src/main/java/org/geotools/renderer/style/RenderableSVGCache.java
```diff
@@ -0,0 +1,260 @@
...
+     * @return
+     */
+    Map<String, String> getParametersFromUrl(String url) {
+        // url.getQuery won't work on file addresses
+        int idx = url.indexOf("?");
+        if (idx == -1 || idx == url.length() - 1) {
+            return Collections.emptyMap();
+        }
+        String query = url.substring(idx + 1);
+        return Arrays.stream(query.split("&"))
```
nmco: Nice.

modules/plugin/svg/src/main/java/org/geotools/renderer/style/RenderableSVGCache.java
```diff
@@ -0,0 +1,260 @@
...
+            for (int i = 0; i < childNodes.getLength(); i++) {
+                Node n = childNodes.item(i);
+                if (n != null && n instanceof Element) {
+                    if (hasParameters((Element) n)) {
+                        return true;
+                    }
+                }
+            }
+        }
+        return true;
```
nmco: This should be ``false`` no ?

modules/plugin/svg/src/main/java/org/geotools/renderer/style/SVGMarkFactory.java
```diff
@@ -0,0 +1,82 @@
+/* (c) 2017 Open Source Geospatial Foundation - all rights reserved
```
nmco: Should be 2018.

modules/plugin/svg/src/test/java/org/geotools/renderer/lite/DrawTest.java
```diff
@@ -307,6 +307,47 @@ public void testParametricAllValues() throws Exception {
                 1000);
     }
 
+    @Test
+    public void testSvgMark() throws Exception {
+        StreamingRenderer renderer = setupSinglePointRenderer("hospital.sld");
+
+        BufferedImage image = RendererBaseTest.showRender("hospital", renderer, TIME, bounds);
+        ImageAssert.assertEquals(
+                new File("./src/test/resources/org/geotools/renderer/lite/test-data/hospital.png"),
```
nmco: This paths relative to source make me uncomfortable, I would prefer to see the get as resource stream option ... but well they are perfectly valid and the build is passing :)

nmco:
GitHub doesn't let me add more comments in my review, looks like a temporary failure.

nmco:
That say I almost finished, the PR looks good just some minor comments. I didn't see any documentation update not sue how relevant it will be.

modules/plugin/svg/src/main/java/org/geotools/renderer/style/RenderableSVG.java
aaime: This code was actually written in 2017

modules/extension/xsd/xsd-sld/src/main/java/org/geotools/sld/bindings/SLDMarkBinding.java
aaime: Good idea, adding logging

modules/library/main/src/main/java/org/geotools/styling/DefaultResourceLocator.java
aaime: Unless covered by the style guide, this comments are a matter of personal taste and could be skipped (we won't get agreement on this unless by writing an addendum to the Google Style, but then, how does one enforce it?).

modules/library/main/src/main/java/org/geotools/styling/SLDParser.java
aaime: Yep, doing so

modules/plugin/svg/src/main/java/org/geotools/renderer/style/RenderableSVGCache.java
aaime: Hmm... you have a point. This code has not been modified compared to master:
https://github.com/geotools/geotools/blob/master/modules/plugin/svg/src/main/java/org/geotools/renderer/style/SVGGraphicFactory.java#L165

However, if we disallow entity resolution in the one provided in the hints, it should stay that way, that one should be used first.

modules/plugin/svg/src/main/java/org/geotools/renderer/style/RenderableSVGCache.java
aaime: Indeed

modules/plugin/svg/src/test/java/org/geotools/renderer/lite/DrawTest.java
aaime: This approach is used a lot in testing code. If it bothers you enough, raise the issue on the devel list and help fixing all instances, e.g.:

{code}
~/devel/git-gt (svg_mark) $ git grep "src/test/resources" | wc -l
264
{code}

modules/plugin/svg/src/main/java/org/geotools/renderer/style/RenderableSVGCache.java
aaime: Checked, we actually need the code to stay as is. The "super" resolver only picks from selected sources in the classpath, and the default GeoTools resolver won't allow DTD, making the entire SVG parsing fail.

modules/library/main/src/main/java/org/geotools/styling/DefaultResourceLocator.java
nmco: I agree ... I never thought about this as a styling issue, I didn't know you were doing it on propose, my bad.

modules/plugin/svg/src/main/java/org/geotools/renderer/style/RenderableSVG.java
nmco: Ah! ok understood.

modules/plugin/svg/src/main/java/org/geotools/renderer/style/RenderableSVG.java
nmco: Actually Java code ask for them but apparently doesn't enforces them:
https://google.github.io/styleguide/javaguide.html#s4.1.1-braces-always-used

modules/plugin/svg/src/main/java/org/geotools/renderer/style/RenderableSVGCache.java
nmco: Ah! thanks for the info.

modules/plugin/svg/src/test/java/org/geotools/renderer/lite/DrawTest.java
nmco: First time I notice it, since it appears in 264 it has prove to be correct enough.

aaime:
@nmco added docs, applied feedback on the other bits, can you double check?

nmco:
Looks good to me, thank you.

PR #1918, opened 2018-06-05
[GEOT-5968] Add NativeFilter filter

nmco:
Associated issue:
https://osgeo-org.atlassian.net/browse/GEOT-5968

Associated mailing list:
https://sourceforge.net/p/geotools/mailman/message/36268473/

### DIFF
--- a/modules/library/jdbc/src/test/java/org/geotools/jdbc/JDBCNativeFilterOnlineTest.java
+++ b/modules/library/jdbc/src/test/java/org/geotools/jdbc/JDBCNativeFilterOnlineTest.java
@@ -1,0 +1,67 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ *
+ *    (C) 2018, Open Source Geospatial Foundation (OSGeo)
+ *
+ *    This library is free software; you can redistribute it and/or
+ *    modify it under the terms of the GNU Lesser General Public
+ *    License as published by the Free Software Foundation;
+ *    version 2.1 of the License.
+ *
+ *    This library is distributed in the hope that it will be useful,
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ *    Lesser General Public License for more details.
+ */
+package org.geotools.jdbc;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.CoreMatchers.notNullValue;
+import static org.junit.Assert.assertThat;
+
+import java.util.ArrayList;
+import java.util.List;
+import org.geotools.data.simple.SimpleFeatureIterator;
+import org.geotools.data.store.ContentFeatureSource;
+import org.geotools.factory.CommonFactoryFinder;
+import org.geotools.referencing.crs.DefaultGeographicCRS;
+import org.opengis.feature.simple.SimpleFeature;
+import org.opengis.filter.Filter;
+import org.opengis.filter.FilterFactory2;
+import org.opengis.filter.NativeFilter;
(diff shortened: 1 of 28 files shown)


modules/library/opengis/src/main/java/org/opengis/filter/NativeFilter.java
```diff
@@ -0,0 +1,25 @@
...
+ *    version 2.1 of the License.
+ *
+ *    This library is distributed in the hope that it will be useful,
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ *    Lesser General Public License for more details.
+ */
+package org.opengis.filter;
+
+/** */
```
aaime: No javadoc? (here and in the method)

modules/library/xml/src/main/java/org/geotools/xml/filter/FilterCompliancePreProcessor.java
```diff
@@ -855,4 +856,9 @@ public String toString() {
     public boolean requiresPostProcessing() {
         return requiresPostProcessing;
     }
+
+    @Override
+    public Object visit(NativeFilter nativeFilter, Object extraData) {
+        throw new UnsupportedOperationException("Native filter is not supported");
```
aaime: Seems quite similar to what the default method already offers? But maybe best to keep it, but throw an exception stating that native filters are not encodable in XML (this is what the visitor seem to be doing, prepare filters for XML encoding).

modules/plugin/jdbc/jdbc-oracle/src/main/java/org/geotools/data/oracle/OracleFilterToSQL.java
```diff
@@ -173,6 +174,9 @@ protected FilterCapabilities createFilterCapabilities() {
         caps.addType(EndedBy.class);
         caps.addType(TEquals.class);
 
+        // native filter support
+        caps.addType(NativeFilter.class);
```
aaime: If all the native filter handling in SQL does is to take it and write it out in the SQL, why not add support for it in SQLDialect#BASE_DBMS_CAPABILITIES instead? And selected DBMS can remove it, if needs be (don't think there will be such need)

modules/library/jdbc/src/test/java/org/geotools/jdbc/JDBCNativeFilterOnlineTest.java
```diff
@@ -0,0 +1,65 @@
...
+            features.add(iterator.next());
+        }
+        iterator.close();
+        // check that we retrieved the necessary features
+        assertThat(features.size(), is(1));
+        assertThat(features.get(0).getAttribute(tname("code")), is("#2"));
+    }
+
+    /**
+     * Builds a native filter that consider only the measurement of type temperature and wind with a
```
aaime: I would reword as "measurement of type temperature or type wind, and with a value greater than 15" and add the example SQL too in the javadoc for less ambiguity: "(TYPE = 'temperature' OR TYPE = 'wind') AND value > 15"

modules/library/opengis/src/main/java/org/opengis/filter/NativeFilter.java
nmco: Ooops, javadoc added.

modules/plugin/jdbc/jdbc-oracle/src/main/java/org/geotools/data/oracle/OracleFilterToSQL.java
nmco: Good point, done.

nmco:
Thank you for the detailed review, suggestions applied.

PR #1919, opened 2018-06-05
[GEOT-6034] Allow the usage of a place holder for the :where: clause created by GeoServer in virtual tables

nmco:
Associated issue:
https://osgeo-org.atlassian.net/browse/GEOT-6034

There is nod documentation about virtual tables in GeoTools, the relevant documentation for this use case is added in GeoServer documentation.

### DIFF
--- a/modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
+++ b/modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
@@ -16,6 +16,10 @@
  */
 package org.geotools.jdbc;
 
+import static org.geotools.jdbc.VirtualTable.WHERE_CLAUSE_PLACE_HOLDER;
+import static org.geotools.jdbc.VirtualTable.WHERE_CLAUSE_PLACE_HOLDER_LENGTH;
+import static org.geotools.jdbc.VirtualTable.setKeepWhereClausePlaceHolderHint;
+
 import com.vividsolutions.jts.geom.Envelope;
 import com.vividsolutions.jts.geom.Geometry;
 import com.vividsolutions.jts.geom.Point;
@@ -3344,13 +3348,12 @@ public final class JDBCDataStore extends ContentDataStore implements GmlObjectSt
 
         // from
         sql.append(" FROM ");
-        encodeTableName(featureType.getTypeName(), sql, query.getHints());
+        encodeTableName(featureType.getTypeName(), sql, setKeepWhereClausePlaceHolderHint(query));
 
         // filtering
         Filter filter = query.getFilter();
         if (filter != null && !Filter.INCLUDE.equals(filter)) {
             sql.append(" WHERE ");
-
             // encode filter
             filter(featureType, filter, sql);
         }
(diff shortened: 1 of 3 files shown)


modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
```diff
@@ -3480,7 +3481,14 @@ FilterToSQL filter(
             // that uses attributes that aren't returned in the results
             SimpleFeatureType fullSchema = getSchema(featureType.getTypeName());
             toSQL.setInline(true);
-            sql.append(" ").append(toSQL.encodeToString(filter));
+            String filterSql = toSQL.encodeToString(filter);
+            int whereClauseIndex = sql.indexOf(":where_clause:");
```
aaime: ":where_clause:" and 14 are sprinkled around, maybe best to add a constant for both?

modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
```diff
@@ -3350,9 +3350,10 @@ protected String selectSQL(SimpleFeatureType featureType, Query query)
         Filter filter = query.getFilter();
         if (filter != null && !Filter.INCLUDE.equals(filter)) {
             sql.append(" WHERE ");
-
             // encode filter
             filter(featureType, filter, sql);
+        } else {
+            removeWhereClausePlaceHolder(sql);
```
aaime: Any chance ":where_clause:" ends up being a valid token in the request? Would be best to ensure that it gets removed in the sqlview definition before merging in the query instead of after building a query with bits coming from the user. Many things can happen, the filter encoding which is different by DBMS, custom functions and their encoding, this being a base class for many databases... while ":where_clause:" is pretty specific, the changes of conflicts are non zero. Also, if you remove in the sql_view expansion, there is less need to sprinkle around calls to "removeWherePlaceholder".
Idea, get the filter, pass it to encodeTableName (or pass the full query to encodeTableName) and handle the removal there instead. Mind joins, not sure what goes on there... actually, make that happen in VirtualTable.expandParameters for better effect. Or, to avoid changing the existing API, one option could be to add the removal of the placeholder as a hint.

nmco: Good feedback, thank you. Removing of the ``:where_clause:`` place holder code moved to the expand parameters method, info passed through a specific hint in the hints map. Regarding the join, well :where_clause: place holder are not supported in those situations.

aaime: That should at least throw a warning, or something like that, or the user would be left with an unreadable error (WFS allows to do joins on all feature types, including the ones that have the placeholder)

nmco: It actually throws an exception, and there is a test for it:
* https://github.com/geotools/geotools/pull/1919/files#diff-b6216c37b5108fd2371ef39ba16ced18R402
* https://github.com/geotools/geotools/pull/1919/files#diff-3f08f2af86593cd27baeb630bf1b1869R4659

aaime:
+1

PR #1923, opened 2018-06-08
GEOT-6036: Repository Datastore should be marked as shared

dromagnoli:


modules/plugin/coverage-multidim/netcdf/src/test/java/org/geotools/coverage/io/netcdf/NetCDFMosaicReaderTest.java
```diff
@@ -380,6 +380,115 @@ public void testCustomTimeAttributeRepository() throws IOException {
...
+                new ShpFileStoreFactory(new ShapefileDataStoreFactory(), new HashMap());
+        File indexDirectory = new File("./target/repo_idx");
+        FileUtils.deleteQuietly(indexDirectory);
+        indexDirectory.mkdir();
+        File auxiliaryDataStoreFile = new File(indexDirectory, "test.properties");
+        String theStoreName = "testStore";
+        FileUtils.writeStringToFile(
+                auxiliaryDataStoreFile, NetCDFUtilities.STORE_NAME + "=" + theStoreName);
+
+        DirectoryDataStore dataStore = new DirectoryDataStore(indexDirectory, dialect);
```
aaime: Suggestion, make a subclass of this, so that you can check if dispose has been called

modules/plugin/coverage-multidim/netcdf/src/test/java/org/geotools/coverage/io/netcdf/NetCDFMosaicReaderTest.java
```diff
@@ -380,6 +380,115 @@ public void testCustomTimeAttributeRepository() throws IOException {
...
+                try {
+                    imageReader.dispose();
+                } catch (Exception e) {
+                    // Ignore exception on dispose
+                }
+            }
+        }
+
+        // the index files have actually been created
+        List<String> typeNames = Arrays.asList(dataStore.getTypeNames());
```
aaime: And here check that dispose has not been called yet

PR #1924, opened 2018-06-11
[GEOT-5843] Fix AStarIterator to work with directed graphs too

mlvrc:
A small fix to allow AStarIterator to work with directed graphs.

The getRelated method was copied from DijkstraIterator, which checks whether the graph is directed or undirected.

The AStarIteratorTest is passing and there are no changes to the API.

This fix is related to the previously raised issue [GEOT-5843](https://osgeo-org.atlassian.net/browse/GEOT-5843).


bencaradocdavies:
@mlvrc this PR changes all the line endings in the file to CRLF. Please revise this PR to leave the line ending as LF. Please also check your git settings. See: http://docs.geotools.org/latest/developer/procedures/git.html#line-endings

We would also like you to add a test case for the new behaviour.

mlvrc:
@bencaradocdavies  I have now added test_4 and test_5 to the AStarIteratorTest, to check that the getRelated method works correctly over nodes of undirected and directed graphs, respectively. These are equivalent to the already existing test_7 and test_8 of the DijkstraIteratorTest.

PR #1938, opened 2018-06-18
[GEOT-6040]: Support esri_pe_string NetCDF attribute for CRS parsing

dromagnoli:


modules/plugin/coverage-multidim/netcdf/src/main/java/org/geotools/coverage/io/netcdf/crs/NetCDFProjection.java
```diff
@@ -60,6 +60,139 @@
...
+                    return new WKTCRSParser(attr);
+                } else if ((attr = var.findAttribute(NetCDFUtilities.GRID_MAPPING_NAME)) != null) {
+                    return new GridMappingCRSParser(attr);
+                }
+            }
+            return null;
+        }
+
+        /** A CRSParser implementation based on a WKT String */
+        static class WKTCRSParser extends CRSParser {
```
aaime: This is a bit strange, I undrestand you cannot have a simple enum because of parameters, but at this point, what's the value of the enum? I'd rather inline the code to do the parsing in these two subclasses. Mind, not a binding comment, I just find this construct confusing.

dromagnoli: Well. At the very beginning I was putting 4 enums to deal with different cases. But I managed to collapse 3 in 1... so yeah... I will simplify that.
Thanks

aaime: +1

PR #1939, opened 2018-06-18
[GEOT-6035] SLD Parser, Expression support in SelectedChannelType

fernandor777:


modules/library/api/src/main/java/org/geotools/styling/StyleFactory.java
```diff
@@ -266,11 +266,14 @@ public RasterSymbolizer createRasterSymbolizer(
...
 
+    public SelectedChannelType createSelectedChannelType(
+            Expression name, ContrastEnhancement enhancement);
+
     public SelectedChannelType createSelectedChannelType(
             String name, ContrastEnhancement enhancement);
 
     /** @deprecated Use {@link #createSelectedChannelType(String, ContrastEnhancement)} */
-    public SelectedChannelType createSelectedChannelType(String name, Expression gammaValue);
+    public SelectedChannelType createSelectedChannelType(Expression name, Expression gammaValue);
```
aaime: There should also be a variant createSelectedChannelType(String, Expression), for backwards compatibility, but I don't see it?

PR #1940, opened 2018-06-19
[GEOT-6045] Unnecessary assert prevents EPSG HSQL database in path with spaces

bencaradocdavies:
https://osgeo-org.atlassian.net/browse/GEOT-6045


bencaradocdavies:
@nmco I have added a unit test to confirm that, when `java.io.tmpdir` contains spaces, the factory works and the database files are created in the correct location.

PR #1941, opened 2018-06-19
[GEOT-6046] HttpConfigTest app-schema-cache path inconsistent and does not honour java.io.tmpdir

bencaradocdavies:
https://osgeo-org.atlassian.net/browse/GEOT-6046


nmco:
Ouch, thank you @bencaradocdavies !

PR #1944, opened 2018-06-22
Update to JTS 1.15.1

jnh5y:
Signed-off-by: Jim Hughes <jnh5y@ccri.com>


jnh5y:
@jodygarnett and I are still working on these branches...

jodygarnett:
This is building for me locally, now with all the dependencies in order we should be able to watch travis be successful.

PR #1945, opened 2018-06-25
[GEOT-6047] Demote GTOPO30 plugin to unsupported

nmco:
Associated issue: 
https://osgeo-org.atlassian.net/browse/GEOT-6047


modules/unsupported/pom.xml
```diff
@@ -206,7 +218,8 @@
 				<module>solr</module>
 				<module>css</module>
 				<module>geojsonstore</module>
-                <module>s3-geotiff</module>
+        <module>s3-geotiff</module>
+        <module>gtopo30</module>
```
tbarsballe: I see spaces here, but tabs above, around line 59. While I would prefer spaces, tabs are currently more consistent with the current pom file.

nmco:
Thank you @bencaradocdavies !

PR #1949, opened 2018-06-25
[GEOT-6050] Upgrade HSQLDB to 2.4.1

bencaradocdavies:
https://osgeo-org.atlassian.net/browse/GEOT-6050

### DIFF
--- a/modules/plugin/epsg-hsql/src/main/java/org/geotools/referencing/factory/epsg/ThreadedHsqlEpsgFactory.java
+++ b/modules/plugin/epsg-hsql/src/main/java/org/geotools/referencing/factory/epsg/ThreadedHsqlEpsgFactory.java
@@ -74,7 +74,7 @@ public class ThreadedHsqlEpsgFactory extends ThreadedEpsgFactory {
      * version number if there is some changes related to the EPSG-HSQL plugin rather then the EPSG
      * database itself (for example additional database index).
      */
-    public static final Version VERSION = new Version("8.6.0.0");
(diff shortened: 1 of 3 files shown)


aaime:
I think it looks good, but I have a suggestion.

Let's say we have two GeoTools applications running off the gt-epsg-hsql database, one gets upgraded, the other does not. What happens? Is the old format preserved, or hsql updates the on disk format? 
I believe it would be safer to bump the version number to 8.6.0.1, here: https://github.com/geotools/geotools/blob/dd12933f34f06d30dc09b1382106903c57b42afb/modules/plugin/epsg-hsql/src/main/java/org/geotools/referencing/factory/epsg/ThreadedHsqlEpsgFactory.java#L77
This would give the newer version a different path, with no chance of conflict.

bencaradocdavies:
Sure, updated commit force-pushed. Can we also change the capitalisation of the path from `/tmp/Geotools` to `/tmp/GeoTools`. This has always bugged me. We still need the `VERSION` bump for the reasons you described above on case insensitive filesystems.

bencaradocdavies:
Oh my. I did it to myself. I broke the test I added a few days ago.  :-(

bencaradocdavies:
Once more, with feeling.

aaime:
Yep, looking good and merging

bencaradocdavies:
Thanks @aaime.

bencaradocdavies:
If this change reduces the number of `CoverageViewTest` failures, it should be backported. I will run some test builds.

PR #1951, opened 2018-06-29
[GEOT-5905] Batik does not ship ServiceSPI files in Maven distro

aaime:


aaime:
This upgrade basically cuts away the ability to resolve entities, not even if configuration would allow for it, since batik does not do that anymore internally. It's probably for the best.

nmco:
Looks good to me.

PR #1952, opened 2018-06-30
[GEOT-6055] Upgrade commons-io dependency to 2.6

nmco:
Associated issue: https://osgeo-org.atlassian.net/browse/GEOT-6055


aaime:
+1

PR #1953, opened 2018-06-30
[GEOT-6056] EPSG lookup failures caused by concurrent database shutdown in FactoryUsingHSQL finalizer

bencaradocdavies:
https://osgeo-org.atlassian.net/browse/GEOT-6056

### DIFF
--- a/modules/plugin/epsg-hsql/src/main/java/org/geotools/referencing/factory/epsg/ThreadedHsqlEpsgFactory.java
+++ b/modules/plugin/epsg-hsql/src/main/java/org/geotools/referencing/factory/epsg/ThreadedHsqlEpsgFactory.java
@@ -308,7 +308,18 @@ public class ThreadedHsqlEpsgFactory extends ThreadedEpsgFactory {
                 }
             }
         }
-        FactoryUsingHSQL factory = new FactoryUsingHSQL(hints, getDataSource());
+        FactoryUsingHSQL factory =
+                new FactoryUsingHSQL(hints, getDataSource()) {
+                    @Override
+                    protected void shutdown(boolean active) throws SQLException {
+                        // Disabled because finalizer shutdown causes concurrent EPSG lookup via
+                        // other FactoryUsingHSQL instances using the same database URL and thus
+                        // the same org.hsqldb.Database instance to fail in, for example,
+                        // GeoServer gs-main unit tests.
+                        // Note that createDataSource() opens the database with "shutdown=true"
+                        // so the database will be shutdown when the last session is closed.
+                    }
+                };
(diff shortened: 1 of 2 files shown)


modules/plugin/epsg-hsql/src/main/java/org/geotools/referencing/factory/epsg/ThreadedHsqlEpsgFactory.java
```diff
@@ -308,7 +308,18 @@ protected AbstractAuthorityFactory createBackingStore(final Hints hints) throws
                 }
             }
         }
-        FactoryUsingHSQL factory = new FactoryUsingHSQL(hints, getDataSource());
+        FactoryUsingHSQL factory =
+                new FactoryUsingHSQL(hints, getDataSource()) {
+                    @Override
+                    protected void shutdown(boolean active) throws SQLException {
```
aaime: I get that it fixes a test failure, but doing so seems also to break the ability to properly shut down the factory after use. If one disposes the authority factory, will the connection to the database be released? If not, we have a leak.

aaime: I guess it will, provided that connections are properly disposed of, right?

bencaradocdavies:
@aaime the finalizer shutdown does not break *a* test fixture, it breaks *any* test fixture, randomly, if that fixture is looking up an EPSG code at the moment the finaliser is run. The `shutdown` method does not shut down the factory, it shuts down the database, immediately and regardless of whether it is in concurrent use by another factory. If you are lucky, the shutdown will be complete before the second factory tries to use it and the database will be cleanly reopened. If you are unlucky, the database will be placed into a closing state and any in-progress lookups will fail.

The finalizer calls `dispose` which releases all factory resources, including closing the connection, which will close its session:
https://github.com/geotools/geotools/blob/master/modules/library/referencing/src/main/java/org/geotools/referencing/factory/epsg/DirectEpsgFactory.java#L3461

Because the EPSG database is opened with `shutdown=true` (which you added), the database will be shut down when the last session is closed. This is a much better solution because the `org.hsqldb.DatabaseManager` knows all open sessions and is thus in effect managing a reference count and in a position to know when the  `org.hsqldb.Database` can be closed. `FactoryUsingHSQL` has no information about other users of the  `org.hsqldb.Database` instance and is in no position to close it.

aaime:
+1

bencaradocdavies:
Thanks @aaime. I will backport this change once the builds have completed for master.

PR #1957, opened 2018-07-03
[GEOT-6059] Add dependency check profile.

bradh:


aaime:
Looks good, merging

aaime:
By the way, found a way to run a parallel build that seemingly work (at least it does not explode and appears to produce a legit output):
``mvn verify -Dall -T8 -Pdependencycheck -DskipTests``

PR #1959, opened 2018-07-03
provide jts 1.15 upgrade instructions

jodygarnett:
Introduced headings to seperate out different GeoTools 20 upgrade activities as was done for prior releases with more than one API change

### DIFF
--- a/docs/user/welcome/upgrade.rst
+++ b/docs/user/welcome/upgrade.rst
@@ -40,7 +40,7 @@ The transitive dependency will correctly bring in the required jars::
         <version>${jts.version}</version>
      </dependency>
 
-**Package change to org.locationtech.org**
+**Package change to org.locationtech.jts**
 
 This release changes the package names from ``com.vividsolutions.jts`` to ``org.locationtech.jts``. To update your own code follow the `JTS Upgrade Guide <https://github.com/locationtech/jts/blob/master/MIGRATION.md>`__ instructions.
 
--- a/docs/user/welcome/upgrade.rst
+++ b/docs/user/welcome/upgrade.rst
@@ -29,9 +29,41 @@ But first to upgrade - change your dependency to |release| (or an appropriate st
 GeoTools 20.x
 -------------
 
+Upgrade to JTS-1.15
+^^^^^^^^^^^^^^^^^^^
+
+The transitive dependency will correctly bring in the required jars::
+
+     <dependency>
+        <groupId>org.locationtech.jts</groupId>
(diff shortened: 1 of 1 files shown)


docs/user/welcome/upgrade.rst
```diff
@@ -29,139 +29,167 @@ But first to upgrade - change your dependency to |release| (or an appropriate st
...
+
+The transitive dependency will correctly bring in the required jars::
+
+     <dependency>
+        <groupId>org.locationtech.jts</groupId>
+        <artifactId>jts-core</artifactId>
+        <version>${jts.version}</version>
+     </dependency>
+
+**Package change to org.locationtech.org**
```
jnh5y: org.locationtech.org should be org.locationtech.jts

docs/user/welcome/upgrade.rst
```diff
@@ -29,139 +29,167 @@ But first to upgrade - change your dependency to |release| (or an appropriate st
...
+  
+  This is a checked exception, in areas of the GeoTools library where this was found we now return an IllegalArgument exception.
   
-* Getting Unit instances
+* Search ``converter == UnitConverter.IDENTITY`` replace ``converter.isIdentity()``
+* Search ``javax.measure.unit.NonSI`` replace ``import si.uom.NonSI``
+* Search ``javax.measure.unit.SI`` replace ``import si.uom.SI``
+* Search ``SI.METER`` replace ``SI.METRE``
+* Search ``javax.measure.converter.UnitConverter`` replace ``javax.measure.UnitConverter``
+* Search ``javax.measure.unit.UnitFormat`` replace ``import javax.measure.format.UnitFormat``
```
jnh5y: Does JSR-363 have both si.uom.* and javax.measure.* packages?

jnh5y:
Looks good to me.  I asked @elahrvivaz to review it as well, and it makes sense to him as well.

docs/user/welcome/upgrade.rst
jodygarnett: jar-363 has javax.measure.* packages.

You need to choose your own implementation - we chose si.uom.*

aaime:
The JTS part looks good, I'm confused by the UOM one, seems like it's removing some examples and the like but could be reformatting. As a rule of thumb we suggest to keep large reformats in separate commits. At least some description of what changed there would be useful (the commit message and PR description only talk about JTS).

jodygarnett:
The reformatting was done to break out subheadings for JTS instructions and UOM would not trip on each other

PR #1960, opened 2018-07-05
GEOT-6049 - Update NetcdfUtilities.getArray to support 7 dims

Mrc0113:
Fix to address bug: https://osgeo-org.atlassian.net/browse/GEOT-6049


aaime:
@bradh we don't upgrade copyright dates anymore, only add them to files that miss them.

bradh:
@aaime: I think that is what I said...

aaime:
Doh sorry, misread...

PR #1961, opened 2018-07-05
exclude jts-core transitive dependency

jodygarnett:
This change excludes a broken hatbox transitive dependency on a nonexistent jts-core 1.15.1-SNAPSHOT.


bencaradocdavies:
Thanks @jodygarnett, I tested this locally and can confirm that the exclusion is effective.

PR #1964, opened 2018-07-06
[GEOT-6061] Make EPSG:404000 magic when read from WKT

ianturton:


modules/library/referencing/src/test/java/org/geotools/referencing/wkt/ParserTest.java
```diff
@@ -408,4 +410,33 @@ private void testParsing(final AbstractParser parser, final String filename)
...
+                        + "  UNIT[\"m\", 1.0], \n"
+                        + "  AXIS[\"x\", EAST], \n"
+                        + "  AXIS[\"y\", NORTH], \n"
+                        + "  AUTHORITY[\"EPSG\",\"404000\"]]";
+
+        final Parser parser = new Parser();
+        CoordinateReferenceSystem observed = parser.parseCoordinateReferenceSystem(wkt);
+        final String check = parser.format(observed);
+        CoordinateReferenceSystem expected = CRS.decode("EPSG:404000");
+        System.out.println("Exp: " + expected);
```
aaime: Best to use the LOGGER instead of System.out (or just comment out)

PR #1965, opened 2018-07-06
Set the EntityResolver on the XML Validator.

sikeoka:
This pull requests updates ValidatorHandler to handle XML entities in the same way as ParserHandler.  All of the ValidatorHandler changes were copied from ParserHandler.

This pull request can be backported to 18.x and 19.x.


aaime:
Looks ok to me, merging

aaime:
I've backported on 19.x, apparenlty no issues (no failures on my machine on gt or gs), but on 18.x there is a conflict, and a quick resolution still causes test failures.
The conflict is related to the missing backport of https://github.com/geotools/geotools/pull/1735 (maybe the test failures are too, but I cannot be sure).

PR #1974, opened 2018-07-13
[GEOS-8311] tests to check ImageMosaics with no projection info work

ianturton:


tbarsballe:
I've squashed and merged; general reminder that it is helpful if you squash simple changes like this into one commit with the JIRA id in the commit message.

PR #1978, opened 2018-07-15
[GEOT-6064] Upgrade apache commons-lang to 3.7.

bradh:
The upgrade is mostly mechanical.


aaime:
Oh wow, large though. Do we need a follow-up with GeoServer? Will the build break downstream if this gets merged? Hmm... probably not due to the package nave change, but though I'd ask anyways.

bradh:
I don't think it will break, but I do have the lang3 PR  (which needs some more attention) up for geoserver

PR #1979, opened 2018-07-16
Update arcsde docs.

bradh:
This is so complicated, its almost certainly always going to be wrong. Should I just delete it instead?


aaime:
Looks good, merged

PR #1980, opened 2018-07-16
[GEOT-6072] close KML LinearRings automatically if they are invalid and not closed

m-abboud:
If a KMLs polygon start point and end point are not the same it is considered invalid and JTS throws an exception saying it is not closed. Google earth and other KML viewers will still work with these invalid polygons and geoserver doesn't which is a problem.

So this commit rectifies it by closing unclosed linear rings by pushing a cloned start point to the end of the coordinate list if the polygon is unclosed.

### DIFF
--- a/modules/extension/xsd/xsd-kml/src/main/java/org/geotools/kml/bindings/LinearRingTypeBinding.java
+++ b/modules/extension/xsd/xsd-kml/src/main/java/org/geotools/kml/bindings/LinearRingTypeBinding.java
@@ -16,14 +16,15 @@
  */
 package org.geotools.kml.bindings;
 
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
 import javax.xml.namespace.QName;
 import org.geotools.kml.KML;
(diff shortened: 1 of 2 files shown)


m-abboud:
Appveyor is complaining about the imagemosaic module which I didn't touch:

Tests in error: 
  ImageMosaicReaderTest.testHarvestSpatial:2788 � IO Unable to delete file: C:\p...

bradh:
Will close/open to see if that re-kicks the build.

m-abboud:
that fixed it yay, ty!

aaime:
The fix looks good to me, thanks. Can you open a ticket at http://osgeo-org.atlassian.net/projects/GEOT 
so that we have something for the release notes? Would be great if you could refer the ticket from the commit message, so that we have reference to a longer description, and keep everything in a single commit (you can rebase and squash, then force push, we won't mind).

m-abboud:
done, squashed it down to one commit and added GEOT-6072 to commit message

PR #1984, opened 2018-07-20
[GEOT-5077] Fix TOWGS84-property of EPSG:28992 in epsg.properties

gerritdrost:
I'm trying to translate EPSG:28992-coordinates to WGS:84 using geotools, and noticed all translated coordinates were off a bit. While googling I stumbled across [GEOT-5077](https://osgeo-org.atlassian.net/browse/GEOT-5077). The TOWGS84-parameters proposed there work like a charm. 

To test them, I looked up a few reference points from the Dutch national mapping agency at https://rdinfo.kadaster.nl/. They all translate correctly using the proposed values.

The code I used to test the original/altered epsg.properties resource file:

```java
import com.vividsolutions.jts.geom.Coordinate;
import org.geotools.geometry.jts.JTS;
import org.geotools.referencing.crs.DefaultGeographicCRS;
import org.opengis.referencing.FactoryException;
import org.opengis.referencing.operation.MathTransform;
import org.opengis.referencing.operation.TransformException;

public class Test {
    public static void main(String[] args) throws FactoryException, TransformException {
        // EPSG:28992 -> WGS:84
        MathTransform transform = CRS.findMathTransform(
            CRS.decode("EPSG:28992"),
            DefaultGeographicCRS.WGS84
        );

        // The tower of the Zuiderkerk in Amsterdam
        //   https://rdinfo.kadaster.nl/overzicht_rd_medewerker/rd_punt.html?punt=250317
        Coordinate src = new Coordinate(121784.6113, 487036.9695);
        Coordinate dest = new Coordinate();

        JTS.transform(src, dest, transform);
        System.out.println(dest);
    }
}
```


aaime:
Can you please add that as a JUnit test in that module, along with some javadoc explaining it?
When we update the EPSG database the epsg.properties gets re-dumped out of the official database and replaced, if the values are still off it's better to know right away with a clear test failure pointing at that.

gerritdrost:
I know little about geo in general, let alone coordinate systems, so please forgive me if I say something stupid.

I can try to add a unit test for this. I suspect this would need to be added in org.geotools.referencing.crs.CRSTest, in the module I also changed the resource-file in. Is that correct? How should the unit test look? Will I just translate a few of the reference points from https://rdinfo.kadaster.nl/ and check if the translated coordinate matches up?

Is the database you mentioned maintained by the GeoTools-community, or is it an import/dump from an external dataset like the one at http://epsg.org? The original poster of [GEOT-5077](https://osgeo-org.atlassian.net/browse/GEOT-5077) suggests those values are recommended by Dutch authorities, but I cannot find any confirmation of that. I just know they work, and the old ones don't. I did find http://epsg-registry.org/?display=entity&urn=urn:ogc:def:crs:EPSG::28992, but it does not mention TOWGS84-values.

aaime:
About the uni test, yes, you're on the right track, I would do exactly what you have suggested.

The database I mention is a dump of the official EPSG database at epsg.org, we updated it every few years as funding shows up to do so (so, cannot say when it will happen). The latest version of it may or may not contain the values you want, and it normally contains many possible choices, but for WKT the dump tool can choose only one set of parameters (I believe it picks the one that has the higher precision, and among those, the one covering the largest area, but could be wrong).

gerritdrost:
Added the unit tests. Is this what you had in mind?

tbarsballe:
That unit test looks good to me, merging.

Does this need to be backported to an earlier GeoTools version?

gerritdrost:
Nope, we're fine like this. Thanks for merging!

PR #1986, opened 2018-07-22
[GEOT-6066] Remove junit and xmlunit from product

bradh:
https://osgeo-org.atlassian.net/browse/GEOT-6066

Breaks GeoServer. Merge https://github.com/geoserver/geoserver/pull/3019 to fix.


modules/extension/xsd/xsd-core/pom.xml
```diff
@@ -139,4 +137,20 @@
       <scope>test</scope>
     </dependency>
   </dependencies>
+  <build>
+    <plugins>
+      <plugin>
```
bencaradocdavies: @bradh please add a source plugin configuration for the test-jar next to that for the test jar. This will cause Maven to build a test-sources.jar that will be deployed and can be used by downstream consumers of the test jar, allowing them to browse the test source and its javadoc in their IDE and set breakpoints in the test classes. The plugin configuration should look like this (identical to the jar plugin except for the artifactId):
```
     <plugin>
        <groupId>org.apache.maven.plugins</groupId>
        <artifactId>maven-source-plugin</artifactId>
        <version>3.1.0</version>
        <executions>
          <execution>
            <goals>
              <goal>test-jar</goal>
            </goals>
          </execution>
        </executions>
      </plugin>
```

bencaradocdavies:
Both Maven plugins should be under dependency management, but that is a job for another day.

bradh:
@bencaradocdavies Have just submitted the request for the change to geotools-devel.

bencaradocdavies:
Thanks, @bradh.

PR #1987, opened 2018-07-24
GEOT-6071, CSV strategies should save the featureType in createSchema

gwalbran:
Other wise the feature type is regenerated with default Integer attribute
types.


tbarsballe:
Looks good, and has adequate test cases. Merging.

PR #1990, opened 2018-07-30
GEOT-5971: fix wfs-ng axis order handling and enable usage with wfs 2…

mbarto:
….0 sources

### DIFF
--- a/modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/WFSFeatureSource.java
+++ b/modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/WFSFeatureSource.java
@@ -186,20 +186,40 @@ class WFSFeatureSource extends ContentFeatureSource {
      *
      * @param query
      */
-    private void invertAxisInFilterIfNeeded(Query query) {
-        boolean invertXY =
-                WFSConfig.invertAxisNeeded(
-                        client.getAxisOrderFilter(), query.getCoordinateSystem());
-        if (invertXY) {
-            Filter filter = query.getFilter();
-
-            FilterFactory2 ff = CommonFactoryFinder.getFilterFactory2(null);
-            InvertAxisFilterVisitor visitor =
-                    new InvertAxisFilterVisitor(ff, new GeometryFactory());
-            filter = (Filter) filter.accept(visitor, null);
-
-            query.setFilter(filter);
+    private void invertAxisInFilterIfNeeded(Query query, SimpleFeatureType featureType) {
+        CoordinateReferenceSystem crs = query.getCoordinateSystem();
+        if (crs == null) {
+            crs = featureType.getCoordinateReferenceSystem();
         }
+        boolean invertXY = WFSConfig.invertAxisNeeded(client.getAxisOrderFilter(), crs);
+        if (invertXY) {
+            invertAxisInFilter(query);
+        }
+    }
+
+    private void invertAxisInFilter(Query query) {
+        Filter filter = query.getFilter();
+
+        FilterFactory2 ff = CommonFactoryFinder.getFilterFactory2(null);
+        InvertAxisFilterVisitor visitor = new InvertAxisFilterVisitor(ff, new GeometryFactory());
+        filter = (Filter) filter.accept(visitor, null);
+
+        query.setFilter(filter);
+    }
+
+    /**
+     * Repleace FastBBOX filters with "real" BBOX filter, so that CRS is correctly included in the
+     * filter we send to the cascaded WFS.
+     *
+     * @param query
+     */
+    private void removeFastBBOXFilterIfNeeded(Query query) {
+        // HACK: we do a double inversion of axis so that we
+        // get the original filter and
+        // FastBBOX accept method does not convert BBOX back to a
+        // FastBBOX again
+        invertAxisInFilter(query);
+        invertAxisInFilter(query);
     }
 
     protected GetFeatureRequest createGetFeature(Query query, ResultType resultType)
@@ -214,9 +234,8 @@ class WFSFeatureSource extends ContentFeatureSource {
 
         request.setTypeName(remoteTypeName);
         request.setFullType(remoteSimpleFeatureType);
-
-        invertAxisInFilterIfNeeded(query);
-
+        removeFastBBOXFilterIfNeeded(query);
+        invertAxisInFilterIfNeeded(query, remoteSimpleFeatureType);
         request.setFilter(query.getFilter());
         request.setResultType(resultType);
         request.setHints(query.getHints());
--- a/modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/internal/parsers/GetFeatureResponseParserFactory.java
+++ b/modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/internal/parsers/GetFeatureResponseParserFactory.java
@@ -95,9 +95,8 @@ public class GetFeatureResponseParserFactory extends AbstractGetFeatureResponseP
         } else if (request.getStrategy().getVersion().equals(Versions.v1_0_0.toString())) {
             config = new org.geotools.wfs.v1_0.WFSConfiguration();
         }
-        return new PullParserFeatureReader(config, in, queryType);
-
-        // TODO: what about axis order issue solved in old parser?
+        return new PullParserFeatureReader(
+                config, in, queryType, request.getStrategy().getConfig().getAxisOrder());
     }
 
     @Override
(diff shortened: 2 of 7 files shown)


modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/internal/parsers/GetFeatureResponseParserFactory.java
```diff
@@ -95,9 +95,8 @@ protected GetFeatureParser parser(GetFeatureRequest request, InputStream in)
         } else if (request.getStrategy().getVersion().equals(Versions.v1_0_0.toString())) {
             config = new org.geotools.wfs.v1_0.WFSConfiguration();
         }
-        return new PullParserFeatureReader(config, in, queryType);
-
-        // TODO: what about axis order issue solved in old parser?
```
aaime: LOL yeah, what about it indeed! :-D

modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/internal/parsers/PullParserFeatureReader.java
```diff
@@ -81,9 +95,26 @@ public SimpleFeature parse() throws IOException {
...
+                    } catch (TransformException e) {
+                        throw new IOException(e);
+                    }
+                }
+            }
+        }
         return feature;
     }
 
+    private Object invertGeometryCoordinates(Geometry geometry) throws TransformException {
```
aaime: Works anyways but a bit weird to have it return an Object, would look better returning a Geometry (that's what the transformer.tranfsorm method does). Not a critical issue, fix if you like.

modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/WFSFeatureSource.java
```diff
@@ -214,9 +234,8 @@ protected GetFeatureRequest createGetFeature(Query query, ResultType resultType)
 
         request.setTypeName(remoteTypeName);
         request.setFullType(remoteSimpleFeatureType);
-
-        invertAxisInFilterIfNeeded(query);
-
+        removeFastBBOXFilterIfNeeded(query);
```
aaime: A non hacky alternative might be (untested):

```
if (query.getFilter() != null) {
            DuplicatingFilterVisitor duplicator = new DuplicatingFilterVisitor();
            Filter copy = (Filter) query.getFilter().accept(duplicator, null);
            query.setFilter(copy);
        }
```

Tests are still passing using this method instead of the double inversion.

modules/unsupported/wfs-ng/src/test/java/org/geotools/data/wfs/internal/v1_1/AxisOrderTest.java
```diff
@@ -175,11 +175,13 @@ public void testGetFeatureWithEastNorthAxisOrderFilter() throws Exception {
         source.getFeatures(query).features();
         BBOX filter = (BBOX) wfs.getRequest().getFilter();
 
-        // filter coordinates are NOT inverted
-        assertEquals(4623055.0, filter.getMinX(), 0.0);
-        assertEquals(815134.0, filter.getMinY(), 0.0);
-        assertEquals(4629904.0, filter.getMaxX(), 0.0);
-        assertEquals(820740.0, filter.getMaxY(), 0.0);
+        // filter coordinates ARE inverted (EPSG:3857 is EAST/NORTH, so if we ask for NORTH/EAST,
```
aaime: Uh ok, this one is interesting. So the BBOX is built as follows:
ff.bbox("the_geom", 4623055.0, 815134.0, 4629904.0, 820740.0, "EPSG:3857")
and indeed 3857 is projected and not flipped... so this is indeed a fix, the axis order should be preserved, no flipping.

modules/unsupported/wfs-ng/src/test/java/org/geotools/data/wfs/internal/parsers/AbstractGetFeatureParserTest.java
```diff
@@ -486,12 +537,25 @@ public void testParseCubeWerx_RoadSeg() throws Exception {
...
-                getParser(featureName, schemaLocation, featureType, CUBEWERX_ROADSEG.DATA);
+                getParser(featureName, schemaLocation, featureType, CUBEWERX_ROADSEG.DATA, null);
 
         int nof = parser.getNumberOfFeatures();
         assertEquals(-1, nof);
 
-        FeatureVisitor assertor = new FeatureAssertor(featureType);
+        final FeatureAssertor assertor = new FeatureAssertor(featureType);
+        List<Geometry> geometries = new ArrayList<Geometry>();
+        geometries.add(
```
aaime: So here geometries are read in their native order (same as parsed GML file), and the CRS returned is EPSG:4269 with flipped axis order (lat and then lon). Right? Is there also a test for the new axis flipping code in the pull parser?

modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/WFSFeatureSource.java
aaime: Oh, another possible approach would be to implement getSRS in fastBBOX like it's done in BBOX3DImpl: return CRS.toSRS(envelope.getCoordinateReferenceSystem());
Not sure if that will work in all cases, just an idea.

modules/unsupported/wfs-ng/src/test/java/org/geotools/data/wfs/internal/v1_1/AxisOrderTest.java
mbarto: So ... the new tests are ok, right?

aaime: Yep, they look ok, I'm weireded out by the old version of the test verifying that a projected system caused the filter to have inverted axis...

modules/unsupported/wfs-ng/src/test/java/org/geotools/data/wfs/internal/parsers/AbstractGetFeatureParserTest.java
mbarto: PullParserTest extends this one, so this is actually testing (also) PullParser. Does this answer your question?

aaime: I think it does not, but not 100% sure. Unless I'm mistaken the ordinates in the GML file are lat/lon, and the tested ordinates are also lat/lon, so where is the new flipping code path being tested?

modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/internal/parsers/XmlSimpleFeatureParser.java
```diff
@@ -714,14 +714,15 @@ private CoordinateReferenceSystem crs(CoordinateReferenceSystem defaultValue)
...
             return defaultValue;
         }
-        boolean forceXY = false;
+        // boolean forceXY = false;
+        // forceXY=true breaks axis flipping handling 
         if (srsName.startsWith("http://") && srsName.indexOf('#') != -1) {
-            forceXY = true;
+            // forceXY = true;
             srsName = "EPSG:" + srsName.substring(1 + srsName.lastIndexOf('#'));
         } else if (srsName.startsWith("EPSG:")) {
```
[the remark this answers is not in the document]
mbarto: With this code wfs 1.0 is still producing lon - lat coordinates, so I suppose it's not needed for this purpose.

modules/unsupported/wfs-ng/src/test/java/org/geotools/data/wfs/internal/parsers/AbstractGetFeatureParserTest.java
mbarto: I added two more tests, for the northeast and eastnorth cases, so we should cover all combinations with those.

tbarsballe:
@aaime - Are you happy with the changes? Can this be merged or does it still need more work?

modules/unsupported/wfs-ng/src/test/java/org/geotools/data/wfs/internal/v1_1/AxisOrderTest.java
jratike80: I do not remember right now how the correct axis order is checked and where it is documented. For example EPSG:2393 (N-E), EPSG:3067 (E-N), and EPSG:3047 (N-E).

PR #1993, opened 2018-08-08
Fix ESRI NamedIdentifier for Cassini and Equidistant Cylindrical proj…

dispiste:
…ectionsj. Fixes https://osgeo-org.atlassian.net/browse/GEOT-6081

### DIFF
--- a/modules/library/referencing/src/main/java/org/geotools/referencing/operation/projection/CassiniSoldner.java
+++ b/modules/library/referencing/src/main/java/org/geotools/referencing/operation/projection/CassiniSoldner.java
@@ -193,7 +193,7 @@ public class CassiniSoldner extends MapProjection {
                             new NamedIdentifier(Citations.EPSG, "Cassini-Soldner"),
                             new NamedIdentifier(Citations.EPSG, "9806"),
                             new NamedIdentifier(Citations.GEOTIFF, "CT_CassiniSoldner"),
-                            new NamedIdentifier(Citations.ESRI, "Cassini_Soldner"),
+                            new NamedIdentifier(Citations.ESRI, "Cassini"),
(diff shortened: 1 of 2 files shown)


bradh:
There is a formatting failure with your changes, see https://travis-ci.org/geotools/geotools/builds/413725946#L2637

dispiste:
Thanks @bradh, the formatting issues are solved now.

aaime:
The fixes seem ok. A test is required to commit changes :-) (unless one can show it's impossible to write, or too hard)

dispiste:
I have added a test, but unfortunately the test fails because OGC NamedIdentifier declares a name for EquidistantCylindrical which is the same as ESRI declares for EquidistantCylindricalSpherical.

The WKT parser is not able to choose the right Projection (since 2 of them have the same name), so even if the Citations are correct they are not very useful for the EquidistantCylindrical and EquidistantCylindricalSpherical projections.

I wonder what should we do in this case. We could only provide the ESRI citation for the EquidistantCylindrical, which was already provided but was incorrect. In this way, EquidistantCylindricalSpherical would lack a citiation (as it was defined before this PR).

I have also some test code to take the authority into account when choosing the projection. This will help when the authority is provided in the WKT (e.g. AUTHORITY["ESRI","4088"], but it will make no difference for the rest of the WKTs. And ESRI wkts usually lack the authority. Maybe it would be interesting to be able to define a Hint to assume ESRI (or any other) authority when no authority is provided.

aaime:
Can you share what's the use case for this fix? What were you originally trying to do, that did not work?

aaime:
Not sure if you wanted this to be backported, if so, could you pick the single commit I've merged on master and use it to build a backport pull request?

PR #1995, opened 2018-08-10
[GEOT-6082] Add support in App-Schema for JDBC and Solr mutivalued properties

nmco:
Associated issue:
https://osgeo-org.atlassian.net/browse/GEOT-6082

Integrations test for SQL and documentation:
https://github.com/geoserver/geoserver/pull/3040

### DIFF
--- a/modules/extension/app-schema/app-schema/src/main/java/org/geotools/jdbc/WrappedFilterToSql.java
+++ b/modules/extension/app-schema/app-schema/src/main/java/org/geotools/jdbc/WrappedFilterToSql.java
@@ -1,0 +1,80 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ *
+ *    (C) 2018, Open Source Geospatial Foundation (OSGeo)
+ *
+ *    This library is free software; you can redistribute it and/or
+ *    modify it under the terms of the GNU Lesser General Public
+ *    License as published by the Free Software Foundation;
+ *    version 2.1 of the License.
+ *
+ *    This library is distributed in the hope that it will be useful,
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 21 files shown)


aaime:
I've tried to have a look, but I don't understand anything about the module and quickly drowned into my ignorance....

nmco:
@aaime indeed this is some advanced App-Schema usage. The feature chaining syntax support works well for features but not that well for simple values (strings, numbers, etc ...), and if we bring on non relational storages like Solr or MongoDB is hard to handle those mappings in an efficient \ user friendly way.

The approach here was:

1. allow data stores to contribute custom syntax
2. implement that custom syntax to Apache Solr, multivalued fields
3. implement that custom syntax for SQL based data stores

modules/extension/app-schema/app-schema/src/main/java/org/geotools/jdbc/WrappedFilterToSql.java
```diff
@@ -0,0 +1,80 @@
...
+import java.io.StringWriter;
+import java.io.Writer;
+import org.geotools.data.complex.FeatureTypeMapping;
+import org.geotools.data.jdbc.FilterToSQL;
+import org.geotools.data.jdbc.FilterToSQLException;
+import org.opengis.filter.Filter;
+import org.opengis.filter.expression.Expression;
+
+/** Wraps a filter to SQL encoder allowing us to write directly to the output stream. */
+public final class WrappedFilterToSql {
```
mbarto: Does it need to be final?

mbarto:
@nmco it generally looks good to me (from my limited knowledge of app-schema, I didn't spot any evident issue). Is it possible to add more unit tests?

nmco:
The integration tests for this are in GeoServer pull requests:
https://github.com/geoserver/geoserver/pull/3040

Perhaps you missed them Mauro ?

mbarto:
You are right. This is the way it goes with app schema. No blockers from me

PR #2000, opened 2018-08-13
Update datastore tutorial maven archetype command

bradh:
`archetype:create` does not work with recent maven plugins.


aaime:
PR #2000, congratulations, you won a teddy bear!

PR #2009, opened 2018-08-14
ImageMosaic GetCapabilities performance

ivanya82:
We are using GeoServer to display a large number of GeoTIFFs using ImageMosaic data stores. The data has several dimensions (time and custom). We are adding new data to these stores periodically and the goal is to be able to display years of data (millions of granules). We have encountered performance problems with the WMS GetCapabilities operation. It is most noticeable when using Oracle database (or PostGIS with parameter WrapStore=true). We implemented changes which improved the performance for our use case significantly. We would like to publish these changes and hopefully merge them into the official repo.

Details:
our performance problems are caused by the queries for values and defaults for dimensions. The currently used queries vary depending on the selected presentation for the dimension:

1. query for values:
  "List":
    * PostGIS (with WrapStore=false): `SELECT distinct("<dim_name>") FROM "<table_name>"`
    * Oracle: `SELECT FID, <DIM_NAME> FROM <table_name>`
    * PostGIS (with WrapStore=true): `SELECT "fid","<dim_name>" FROM "<table_name>"`
    First version is clearly the fastest, because it loads only distinct values.

"Interval and resolution":
  The queries are the same as for "List", but in this case we only really need the min/max values, which could be calculated much faster in the database.

2. query for defaults (when using smallest/biggest domain value):
  * PostGIS (with WrapStore=false): `SELECT <min | max> ("<dim_name>") FROM "<table_name>"`
  * Oracle: `SELECT FID, <DIM_NAME> FROM <table_name>`
  * PostGIS (with WrapStore=true): `SELECT "fid","<dim_name>" FROM "<table_name>"`
  Again the first PostGIS variant is clearly the fastest. Others load all the values to the application where the min/max values are calculated.

Our proposed solution creates optimized queries depending on which presentation and default setting is used for the dimension. The needed additional information from the settings is sent to the lower level functions using new RenderingHints. The goal of the optimized queries is to load only the needed values from the database (usually what the current PostGIS query does). Unit tests were also added.

We would like to get some feedback on this.


aaime:
So with this change you still get the performance improvement you were looking for, right?

ivanya82:
With this change we get the performance improvement only for obtaining minimum or maximum default values of the dimensions. But we still need to load only the needed values from the database if "Interval and resolution" presentation for the dimension is selected.

To solve the "Interval and resolution" presentation select problem we plan to make changes in the imagemosaic and metadata modules of the GeoTools and in the main module of the GeoServer. Should we add these changes to this branch or should we create another branch?

aaime:
Another one, and please discuss publicly on the list what the issue is and how you intend to implement the changes. In GeoTools/GeoServer it's important to talk before coding (or if you really need to prototype, be prepared to rewrite from scratch, like it happened in this example)

modules/extension/transform/src/main/java/org/geotools/data/transform/TransformFeatureCollection.java
```diff
@@ -206,4 +208,25 @@ public SimpleFeatureCollection subCollection(Filter filter) {
...
+    public void accepts(
+            org.opengis.feature.FeatureVisitor visitor, org.opengis.util.ProgressListener progress)
+            throws IOException {
+        Name typeName = transformer.getSource().getName();
+
+        if (typeName != null) {
+            try {
+                SimpleFeatureSource fs =
+                        (SimpleFeatureSource) source.getDataStore().getFeatureSource(typeName);
+                if (fs instanceof ContentFeatureSource) {
```
aaime: No actual need to check if it's a content feature source, SimpleFeatureSource has the method you need too

modules/extension/transform/src/main/java/org/geotools/data/transform/TransformFeatureCollection.java
```diff
@@ -206,4 +208,25 @@ public SimpleFeatureCollection subCollection(Filter filter) {
...
+                    ((ContentFeatureSource) fs).accepts(query, visitor, progress);
```
aaime: Here is the problem, you cannot, in general, pass down the visitor blindly, but need to check how the property involved was transformed, if at all, and then decide accordingly.
The logic can be easy or complex depending on how far you want to go. My suggestion would be to:
- Recognize CountVisitor and let it go as is
- Recognize FeatureAttributeVisitor implementations, extract the expressions, and see what attributes they are using by using a FilterAttributeExtractor, and then check against the Transformer and see if a rename or a calculation is happening there. 

If the visitors are not a match for the above classes/interfaces, or if any attribute used in the visitor is transformed you cannot pass down the visit and have to rely on super implementation, otherwise you can pass down without issues.

ivanya82:
Method `isTypeCompatible` allows to control pass down the visitor has been appended. We have a question about detecting attributes used in the visitor which were transformed. Method getOriginalNames() of the Transformer class has been used to detect transformed attributes. Because, as described in the method, if attribute has a general transformation then it will not be put into the result list. We understood it right, or is there another way to detect transformed attribute? Thank you for the advises.

modules/extension/transform/src/main/java/org/geotools/data/transform/TransformFeatureCollection.java
```diff
@@ -206,4 +216,56 @@ public SimpleFeatureCollection subCollection(Filter filter) {
...
+    /**
+     * Checks if the visitor is accessing only properties available in the specified feature type,
+     * checks if the target schema contains transformed attributes or as a special case, if it's a
+     * count visitor accessing no properties at all
+     *
+     * @param visitor
+     * @param featureType
+     * @return
+     */
+    protected boolean isTypeCompatible(FeatureVisitor visitor, SimpleFeatureType featureType) {
```
aaime: Verified better and these tests are not valid, could have seen this easily by also checking the results of the visits in tests, like I just did. Since it's been a while (sorry, summer vacations with family and conferences kept all my spare time occupied) I'm going to fix it directly.

ivanya82:
Hello,
We have tested the replacement: #2039.It optimizes dimension computation.

PR #2013, opened 2018-08-17
GEOT-5905: fix batik NPE issues

ianturton:
For some reason best known to themselves Batik does not ship ServiceSPI files in their Maven distributed jars. This leads to failures when trying to render to svg using Batik. 

This PR adds the spi files to gt-svg and adds the now required batik-codec jar to the module too. The spi files are Apache licensed so I believe this is OK.


aaime:
So the module "gt-svg" per se does not render SVG, it just reads it... or not?
I don't mind having the extra SPI files, but is the dependency actually needed? Could it be made test dep for the sake of tests, and then we give an indication to whoever is using the module that if they also want to render, codec must be added?
Just thinking out loud, maybe I'm missing some bits.

ianturton:
In the documentation we say if you want to render to svg you should depend on gt-svg. So if we leave that in we need the dependency. Or we can make it a test dependency and force users to add their own dependency - in which case we could not tell them to use gt-svg and to just import as much batik as they need, including any workaround for the spi issues that batik has.

PR #2014, opened 2018-08-18
[GEOT-6102] Upgrade GeographicLib to 1.49

bradh:
The reason for the test change is that GeographicLib 1.48 (and a little in 1.47) has
a change in behaviour. From the release notes:
"The default range for longitude and azimuth is now (-180d, 180d],
instead of [-180d, 180d). This was already the case for the C++
library; now the change has been made to the other implementations
(C, Fortran, Java, JavaScript, Python, MATLAB, and Maxima)."


aaime:
Thank you!

bradh:
Added a PR with unit tests over at https://github.com/geotools/geotools/pull/2016

PR #2015, opened 2018-08-18
[GEOT-6103] Upgrade EJML to 0.34

bradh:
The EJML inverse function was changed to return a MatrixDimensionException if the input is not square:
https://github.com/lessthanoptimal/ejml/commit/ca9f8d528dd4a460658d3223b626a494de297f73#diff-16e60a287b7af132b57bfc757a7c8dc6R682
I kept the old catch for IllegalArgumentException just in case that still gets used somewhere.


aaime:
All looks good now. @jodygarnett and @jnh5y  any comment?

aaime:
Let's give it a roll then and see if GeoServer and GWC like too

PR #2016, opened 2018-08-18
[GEOT-4207/4604/6026/6077] Add tests for GeodeticCalculator problems.

bradh:
These are a mix of issues that are resolved by the recent incorporation of 1.49,
and those can't be reproduced with either 1.44 or 1.49.


bradh:
Note that GEOT-6077 should stay open for a bit longer, even after this is incorporated.

PR #2020, opened 2018-08-20
[GEOT-6106] Upgrade postgresql JDBC driver to 42.2.5

bradh:
I did a full build of geoserver against the modified geotools, but I don't have an online test environment, and it really does need that kind of test.

@bencaradocdavies Can you run the tests?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -108,7 +108,7 @@
     <fork.javac>true</fork.javac>
     <javac.maxHeapSize>256M</javac.maxHeapSize>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    <postgresql.jdbc.version>42.1.1</postgresql.jdbc.version>
+    <postgresql.jdbc.version>42.2.5</postgresql.jdbc.version>
     <solrj.version>7.2.1</solrj.version>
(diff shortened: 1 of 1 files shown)


mprins:
You probably want to jump to 42.2.5 since that was released this week

aaime:
Hi @bradh, I'm at a conference away from my development machine, but it's not hard to run online tests, roughly:

- Prepare a postgis database that is empty and a user that can do whatever on it (create/delete tables/indexes/views/sequences/anything else)
- Run tests once from the IDE, or from maven with -Ponline, they will skip, look in your home directory, you should find a .geotools/postgis.properties.example (or something like that)
- Change the filename to postgis.propeties, fill good values, and re-run
Not all tests may not pass, there are some checking brand new database creations, others might need an extension to be installed, but hopefully it should give you a good idea

bradh:
I have updated this to 42.2.5. The online tests for postgis (well jdbc-postgis) ran with no errors or failures.

bradh:
Have now got the right configuration - now `-Ponline` passes.

aaime:
Merging on master, let's have a look at what the online test build says.

PR #2022, opened 2018-08-21
[GEOT-6111] Update test dependencies.

bradh:
This is a mix of simple upgrades, some changes that affected the dependency names and test code changes to reflect API changes in the deps. No change to product. 

I did do a full geoserver release build against this version, no impact as expected.


aaime:
Merged. Thanks!

PR #2023, opened 2018-08-22
[GEOT-6109] Centralize SnakeYAML configuration

sikeoka:
https://osgeo-org.atlassian.net/browse/GEOT-6109


bradh:
While doing this cleanup, would it also be appropriate to move to snakeyaml 1.21?

sikeoka:
I've looked into that but I think that it should be a separate ticket and pull request.  I was planning to do it at some point after this one since it contains a fix for illegal reflective access in Java 9.

aaime:
According to the pom the current maintainer is @smithkm, assigned review accordingly

PR #2025, opened 2018-08-23
GEOT-6112: support for expressions in RasterSymbolizer GammaValue for…

mbarto:
… SLD and CSS

### DIFF
--- a/modules/extension/xsd/xsd-sld/src/test/java/org/geotools/sld/SLDTest.java
+++ b/modules/extension/xsd/xsd-sld/src/test/java/org/geotools/sld/SLDTest.java
@@ -84,6 +84,14 @@ public class SLDTest {
         parser.validate(getClass().getResourceAsStream("linePerpendicularOffset.sld"));
     }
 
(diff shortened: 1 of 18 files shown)


modules/extension/xsd/xsd-sld/src/test/java/org/geotools/sld/SLDTest.java
```diff
@@ -84,6 +84,14 @@ public void testValidatePerpendicularOffset() throws Exception {
         parser.validate(getClass().getResourceAsStream("linePerpendicularOffset.sld"));
     }
 
+    @Test
+    public void testValidateGammaValueExpression() throws Exception {
+        Parser parser = new Parser(new SLDConfiguration());
+
+        // if a validato error occurs it will blow up with an exception
```
aaime: "if a validation error"

mbarto: Copied from another test :-)

PR #2027, opened 2018-08-29
[GEOT-6113]: Fix accurate resolution computation

dromagnoli:


modules/library/coverage/src/main/java/org/geotools/coverage/grid/io/ReadResolutionCalculator.java
```diff
@@ -224,10 +228,28 @@ public ReadResolutionCalculator(
...
+                    new GeneralEnvelope(new double[] {x0, y0}, new double[] {x1, y1});
+            envelope = CRS.transform(destinationToSourceTransform, envelope);
+            GridToEnvelopeMapper mapper =
+                    new GridToEnvelopeMapper(new GridEnvelope2D(0, 0, 1, 1), envelope);
+            AffineTransform transform = mapper.createAffineTransform();
+            fullRes[0] = XAffineTransform.getScaleX0(transform);
+            fullRes[1] = XAffineTransform.getScaleY0(transform);
+        }
+        // fall back on the full resolution when zero length
+        double minDistanceX = zeroLength ? fullRes[0] : minDistance;
```
nmco: This doens't look like a fall-back, maybe I'm missing something, in my understanding the ``if (zeroLength && isReprojected)`` is there to make sure that if the read bounding-box is reprojected or is empty (zero length) then we fallback? on creating a full resolution on pixel bounding-box.

modules/library/coverage/src/test/java/org/geotools/coverage/grid/io/ReadResolutionCalculatorTest.java
```diff
@@ -46,4 +46,34 @@ public void testReadResolutionCalculator() throws Exception {
...
+        final ReferencedEnvelope requestBounds =
+                new ReferencedEnvelope(
+                        -1.3184040176569374E7,
+                        -1.31787660216177E7,
+                        4020998.5368273035,
+                        4026272.691778981,
+                        requestCRS);
+        GridGeometry2D gg = new GridGeometry2D(new GridEnvelope2D(0, 0, 256, 256), requestBounds);
+
+        // We intentionally provide a null native res
```
nmco: So providing a NULL native resolution is what will trigger ``ReadResolutionCalculator`` ?

modules/library/coverage/src/main/java/org/geotools/coverage/grid/io/ReadResolutionCalculator.java
dromagnoli: The previous code was simply doing a Max(fullRes, minDistance) to prevent zeroLength (minDistance = 0).
The problem was that in case of reprojection (even if not zeroLength) fullRes can be > minDistance so fullRes was taken (20 meters instead of 0.0002 degrees) returning back huge pixels.

With the new code:
- zeroLength is checked via a zero comparison: abs(minDistance) > 1E-10 .
- in case of zero length minDistance will not be used in favor of fullRes.
- in case of reprojection, fullRes need to be recomputed to report resolution in proper CRS. However this computation is only performed when zeroLength since it will only be used in that case. (minDistance will be used otherwise)

Does it make sense?

modules/library/coverage/src/test/java/org/geotools/coverage/grid/io/ReadResolutionCalculatorTest.java
dromagnoli: that is one case indeed.

modules/library/coverage/src/main/java/org/geotools/coverage/grid/io/ReadResolutionCalculator.java
nmco: Ok understood, thanks.

modules/library/coverage/src/test/java/org/geotools/coverage/grid/io/ReadResolutionCalculatorTest.java
nmco: Nice, thx for the explanation.

PR #2033, opened 2018-09-01
fix possible NPE issue during MySQLVersion56 check

nprigour:

### DIFF
--- a/modules/plugin/jdbc/jdbc-mysql/src/main/java/org/geotools/data/mysql/MySQLDataStoreFactory.java
+++ b/modules/plugin/jdbc/jdbc-mysql/src/main/java/org/geotools/data/mysql/MySQLDataStoreFactory.java
@@ -155,7 +155,7 @@ public class MySQLDataStoreFactory extends JDBCDataStoreFactory {
             e.printStackTrace();
         } finally {
             try {
-                if (con != null || !con.isClosed()) {
+                if (con != null && !con.isClosed()) {
                     con.close();
                 }
             } catch (SQLException e) {


nprigour:
good to be back-ported to version 19.x also

aaime:
Looks like an obvious fix, did you stumble into it by reading code or did it actually happen?
If the latter, please open a ticket and describe how, for the release notes, then refer to the ticket from the commit message.

nprigour:
Indeed it is obvious and to be honest it was introduced by me while updating mysql plugin to work with the latest mysql versions.
Actually I fall on it by change but it may occur while i.e. trying to create a mysql DataStore against a Mysql DB that is down or unreachable.

aaime:
Roger, merging on master only then (based on the lack of tests)

nprigour:
would it be possible to backport this to 19.x?

aaime:
I already made an exception to the test rules by merging on master, under the consideration that is a developer series that will not see a release for some time. If you want backports please add tests :-)

nprigour:
But how is it possible to add a test case for such an obvious error?
I can i.e. add a test that attempts to open a connection to an unreachable (non existent)  Datasource and assert that no exception is thrown from the final block. Would that be enough to accept the fix for backporting?

aaime:
Yes, check that you're not getting a NPE :-)

nprigour:
OK I will try to implement such a test the following days. Should I reopen this or do it in a seperate pull request?

aaime:
Euh, this one got merged, a new PR with the test is probably best.

PR #2035, opened 2018-09-06
Bursa wolf parameters required

hhj1988:
There are (EPSG: 4547 EPSG: 4326) two different coordinate system transformations, there is an exception: org. opengis. referencing. operation. OperationNotFoundException: Bursa wolf parameters required. Now datum is different, can you customize the transformation?
Thank you


bradh:
Are you trying to file a bug report here? The pull request looks like you're trying to merge some commits into the 20.0 branch, which doesn't look right.

PR #2040, opened 2018-09-16
LabelSplitter index-out-of-bounds bug.

wcmatthysen:
- Fixed a bug in the LabelSplitter class where the start index would be
  incorrectly advanced at the end of the buildFontRanges() method. This
  happened when a font could not be found to display the character at
  the current index, the start-index would then be incremented in the
  else. This can lead to an incorrect font-range from being created, or
  at worst a StringIndexOutOfBoundsException.
  The fix for this is to not increment the start-index inside the
  for-loop, but to keep it fixed and allow all fonts to be checked
  whether they can render the current character. Only after the loop has
  iterated through all fonts; and no font has been found; do we
  increment the start-index.
- A new unit test class called LabelSplitterTest was added. These tests
  were written to illustrate the abovementioned bug. With the first test,
  we have a string with some special characters in the middle. If the
  bug-fix is not applied to the LabelSplitter the ranges will be
  completely messed up. The second test has a character at the start,
  that can be renderered by the "Times New Roman" font. However, the
  rest of the characters can only be rendered by the "Lucida" font. If
  the bug-fix is not applied the LabelSplitter will create a second-font
  range that will cause a StringIndexOutOfBoundsException.

### DIFF
--- a/modules/library/render/src/main/java/org/geotools/renderer/label/LabelSplitter.java
+++ b/modules/library/render/src/main/java/org/geotools/renderer/label/LabelSplitter.java
@@ -369,9 +369,10 @@ class LabelSplitter {
                             foundFont = true;
                             result.add(new FontRange(text, base, start, fonts[0]));
                             break;
-                        } else {
-                            start++;
                         }
+                    }
+                    if (!foundFont) {
+                        start++;
                     }
                 }
             }
@@ -389,7 +390,7 @@ class LabelSplitter {
      *
      * @author Andrea Aime - GeoSolutions
      */
-    private static class FontRange {
+    static class FontRange {
         int startChar;
 
         int endChar;
--- a/modules/library/render/src/test/java/org/geotools/renderer/label/LabelSplitterTest.java
+++ b/modules/library/render/src/test/java/org/geotools/renderer/label/LabelSplitterTest.java
@@ -1,0 +1,91 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ *
+ *    (C) 2014 - 2018 , Open Source Geospatial Foundation (OSGeo)
+ *
+ *    This library is free software; you can redistribute it and/or
+ *    modify it under the terms of the GNU Lesser General Public
+ *    License as published by the Free Software Foundation;
+ *    version 2.1 of the License.
+ *
+ *    This library is distributed in the hope that it will be useful,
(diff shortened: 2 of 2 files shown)


wcmatthysen:
I made the fix for the version 19-branch. However, I think this is probably a bug on the other branches as well.

aaime:
Pull requests are only accepted against the master branch (for starters, then backports are welcomed).

The test is also failing:

```
Tests run: 2, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.181 sec <<< FAILURE! - in org.geotools.renderer.label.LabelSplitterTest
testFontRangeWithRenderableCharAtStartAndSpecialCharsAtEnd(org.geotools.renderer.label.LabelSplitterTest)  Time elapsed: 0.005 sec  <<< FAILURE!
java.lang.AssertionError: expected:<2> but was:<1>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:743)
	at org.junit.Assert.assertEquals(Assert.java:118)
	at org.junit.Assert.assertEquals(Assert.java:555)
	at org.junit.Assert.assertEquals(Assert.java:542)
	at org.geotools.renderer.label.LabelSplitterTest.testFontRangeWithRenderableCharAtStartAndSpecialCharsAtEnd(LabelSplitterTest.java:77)
```

wcmatthysen:
Ok, I'll re-open a pull-request against master. I'll also look into why the test is failing. I suspect it is because the machine travis is running on doesn't have the Lucida font or something like that. I'll try and re-work it with a different font.

aaime:
Yes, you cannot assume presence of any font on the machines, the fonts must be included as test resources, which implies they have to freely redistributable.
The code doing labelling in GeoTools uses the included "Vera.ttf" and registers it at the beginning of the test with code looking like:

```
Font.createFont(
                                Font.TRUETYPE_FONT,
                                TestData.getResource(this, "Vera.ttf").openStream())
```

Since you're already playing in the rendering modules, the fonts already available to you are:

````
~/devel/git-gt/modules/library/render (transform_visitor) $ git ls-files "*.ttf"
src/test/resources/org/geotools/renderer/lite/test-data/DroidNaskh-Regular.ttf
src/test/resources/org/geotools/renderer/lite/test-data/DroidSansArmenian.ttf
src/test/resources/org/geotools/renderer/lite/test-data/Vera.ttf
src/test/resources/org/geotools/renderer/lite/test-data/recreate.ttf
````

In case you need any, please make sure to add a font with an open licence

PR #2047, opened 2018-09-18
Additional type mappings for jts geometry types

jodygarnett:
Maps both org.locationtech.jts.geom and com.vividsolutions.jts.geom for backwards compatibility.


jodygarnett:
Chat discussion indicates this fix may help resolve https://osgeo-org.atlassian.net/browse/GEOT-6087

jodygarnett:
Thanks for the tip @aaime that did the trick (and I feel better about the PR fixing GEOT-6087)

tbarsballe:
@jodygarnett Please remember to run the formatter on all your commits before pushing. This had non-compliant files (normally this would get caught by Jenkins, but the GeoTools Jenkins job is currently failing on all builds for unrelated reasons, see mailing list discussion)

PR #2055, opened 2018-09-21
HANA JDBC plugin implementation

stefanuhrig:
Supports direct access to HANA databases.


modules/unsupported/jdbc-ng/jdbc-hana/src/main/java/org/geotools/data/hana/converter/BooleanConverterFactory.java
```diff
@@ -0,0 +1,54 @@
...
+import org.geotools.util.ConverterFactory;
+
+/**
+ * Converter factory for converting Short objects to Boolean objects.
+ *
+ * <p>In HANA, ResultSet#getObject will return a Short object in case of boolean fields.
+ *
+ * @author Stefan Uhrig, SAP SE
+ */
+public class BooleanConverterFactory implements ConverterFactory {
```
aaime: This may not cause troubles, but in general, converters are applied everywhere once loaded, e.g., they will be used anywhere else, not just in the HANA code. Right now I don't think you have much of an alternative, doing something else would require changing the core classes (e.g., calling into the dialects for data type conversion). Take this more of as a warning, if something weird happens at runtime while the HANA plugin is running, this might be a cause.

modules/unsupported/jdbc-ng/jdbc-hana/src/main/java/org/geotools/data/hana/metadata/CsvReader.java
```diff
@@ -0,0 +1,125 @@
...
+import java.io.UnsupportedEncodingException;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * A simple CSV file reader tailored for reading the embedded HANA spatial metadata.
+ *
+ * @author Stefan Uhrig, SAP SE
+ */
+public class CsvReader {
```
aaime: Classes in the metadata package are probably best kept package private, with the exception of MetadataImport, they are not really meant to be reused outside this context.

modules/unsupported/jdbc-ng/jdbc-hana/src/main/java/org/geotools/data/hana/HanaDialect.java
```diff
@@ -0,0 +1,892 @@
...
+        SQL_TYPE_TO_CLASS.put(-10, String.class); // NCLOB
+        SQL_TYPE_TO_CLASS.put(3000, BigDecimal.class); // SMALLDECIMAL
+        SQL_TYPE_TO_CLASS.put(29812, Geometry.class); // ST_GEOMETRY
+    }
+
+    private static final Map<Class<?>, Integer> CLASS_TO_SQL_TYPE =
+            new HashMap<Class<?>, Integer>();
+
+    static {
+        CLASS_TO_SQL_TYPE.put(Geometry.class, 29812);
```
aaime: The 29812 "magic number" is used in several places, I'd use a constant.

modules/unsupported/jdbc-ng/jdbc-hana/src/main/java/org/geotools/data/hana/HanaFilterToSQL.java
```diff
@@ -0,0 +1,553 @@
...
+                out.write(Double.toString(minz));
+            }
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+        return extraData;
+    }
+
+    private BoundingBox clamp(BoundingBox bounds) {
+        // In geographic CRS', HANA will reject any points outside the "normalized" range, which is
```
aaime: This really comes from practical use of WMS/WFS, clients will normally do out of range requests, and admins will often record as WGS84 stuff that's really projected, and expect the result to work anyways.

modules/unsupported/jdbc-ng/jdbc-hana/src/main/java/org/geotools/data/hana/HanaUtil.java
```diff
@@ -0,0 +1,92 @@
...
+ */
+public final class HanaUtil {
+
+    /**
+     * Encodes multiple identifiers, separating them by periods.
+     *
+     * @param ids List of identifiers. All null-identifiers are ignored.
+     * @return Returns the encoded identifiers.
+     */
+    public static String encodeIdentifiers(String... ids) {
```
aaime: Looks like this method is not used?

modules/unsupported/jdbc-ng/jdbc-hana/src/main/java/org/geotools/data/hana/converter/BlobConverterFactory.java
```diff
@@ -0,0 +1,56 @@
...
+import org.geotools.factory.Hints;
+import org.geotools.util.Converter;
+import org.geotools.util.ConverterFactory;
+
+/**
+ * Converter factory for converting all classes implementing java.sql.Blob to byte[].
+ *
+ * @author Stefan Uhrig, SAP SE
+ */
+public class BlobConverterFactory implements ConverterFactory {
```
nmco: Not something that needs to be changed, just an heads up. This converter (and the other two) are targeting native Java types, i.e. nothing specific to SAP HANA, why not contributing them to the ``org.geotools.util`` GeoTools package ? 

Otherwise there is a risk that someone will contribute a converter that does the same conversion (Blob -> byte[]) and the behavior will may be different depending on the converter loaded first. For example, ``org.geotools.data.teradata`` already implements a Blob -> byte[] converter.

modules/unsupported/jdbc-ng/jdbc-hana/src/main/java/org/geotools/data/hana/converter/BlobConverterFactory.java
```diff
@@ -0,0 +1,56 @@
...
+
+    @Override
+    public Converter createConverter(Class<?> source, Class<?> target, Hints hints) {
+        if (byte[].class.equals(target) && Blob.class.isAssignableFrom(source)) {
+            return new Converter() {
+                @Override
+                public <T> T convert(Object source, Class<T> target) throws Exception {
+                    if (source == null) {
+                        throw new IllegalArgumentException();
```
nmco: Suggestion, I would add some basic message saying at least which argument was invalid.

modules/unsupported/jdbc-ng/jdbc-hana/src/main/java/org/geotools/data/hana/converter/BlobConverterFactory.java
```diff
@@ -0,0 +1,56 @@
...
+                    }
+                    Blob blob = (Blob) source;
+                    long length = blob.length();
+                    if (length == 0) {
+                        return target.cast(new byte[0]);
+                    }
+                    if (length > Integer.MAX_VALUE) {
```
nmco: If I understood this correctly, the goal here is too avoid the error:

```
Exception in thread "main" java.lang.OutOfMemoryError:
  Requested array size exceeds VM limit
```

... if the blob size is bigger than the maximum array possible index. Two comments here:

* the array maximum size varies from a VM to another (http://hg.openjdk.java.net/jdk7/jdk7/jdk/file/tip/src/share/classes/java/util/ArrayList.java#l190)

* not sure that simply returning ``NULL`` is the best option, I would at least log some debug message saying: ``Conversion not supported this by this converter, blob to big for this converter to handle.``.

modules/unsupported/jdbc-ng/jdbc-hana/src/main/java/org/geotools/data/hana/converter/BooleanConverterFactory.java
nmco: Oops, just saw this comment now, my apologies for the similar comment above.

stefanuhrig:
Thanks for the feedback!

I'll try to address the issues you've pointed out and will update the pull request.

modules/unsupported/jdbc-ng/jdbc-hana/src/main/java/org/geotools/data/hana/HanaDialect.java
stefanuhrig: I've introduced a constant for the type code.

modules/unsupported/jdbc-ng/jdbc-hana/src/main/java/org/geotools/data/hana/HanaUtil.java
stefanuhrig: You are right. I've removed the method.

modules/unsupported/jdbc-ng/jdbc-hana/src/main/java/org/geotools/data/hana/converter/BlobConverterFactory.java
stefanuhrig: I propose I do that in a separate pull request (for `Blob` and `Clob`) and remove the converter from `org.geotools.data.teradata` then as well.

modules/unsupported/jdbc-ng/jdbc-hana/src/main/java/org/geotools/data/hana/converter/BlobConverterFactory.java
stefanuhrig: I've added a message text pointing out that source must not be null (in all of the three containers).

modules/unsupported/jdbc-ng/jdbc-hana/src/main/java/org/geotools/data/hana/converter/BlobConverterFactory.java
stefanuhrig: Actually I did that because the dimension expression of an array must be of type `int` and cannot be of type `long`. When casting from `long` to `int`, the upper 32 bits are just discarded. So this check is to verify that none of the upper 32 bits is set.

According to the documentation of `org.geotools.util.Converter`, both returning `null` and throwing an exception is fine to signal that the converter is not able to convert. Nevertheless, logging a debug message in case of a failure does not hurt.

I propose I address that in a separate pull request if you agree that a `Blob` to `byte[]` converter and a `Clob` to `String` converter in `org.geotools.util` makes sense.

modules/unsupported/jdbc-ng/jdbc-hana/src/main/java/org/geotools/data/hana/converter/BlobConverterFactory.java
nmco: Sounds good to me, thank you.

aaime:
Checked the CLA was on file, it is. Merged :-)

PR #2056, opened 2018-09-21
[GEOT-6125] Fix an annoying ArrayIndex Issue, but not the path problem

ianturton:
ImageMosaicConfigHandler throws an ArrayIndexOutOfBoundsException when looking up if there is a higher resolution, which this PR fixes.

Unable to demonstrate Windows path problem in this code, so [proceeding](https://osgeo-org.atlassian.net/browse/GEOS-8942) to look at the GeoServer code that creates a panel for the ImagePyramid module.


aaime:
I think it'd be better to copy the "goodpyramid" into a path with spaces under "target" and run from there, rather than adding 100 new files git... or not?

ianturton:
Very good point

PR #2065, opened 2018-09-24
[GEOT-6214] remove validity check and port faster distance from javascript

ianturton:


modules/unsupported/process-geometry/src/main/java/org/geotools/process/geometry/PolyLabeller.java
```diff
@@ -54,9 +54,9 @@ static Geometry getPolylabel(Geometry polygon, double precision) {
             throw new IllegalStateException("Can not label empty geometries");
         }
 
-        if (!polygon.isValid()) {
+       /* if (!polygon.isValid()) {
```
giohappy: if we're sure that validity check is not required anymore why do we keep it commented?

ianturton:
Good point, I forgot about it after I ran the tests.

PR #2081, opened 2018-10-02
[GEOT-6140] Allow the usage in App-Schema of Solr as an external index

nmco:
Associated issue:
https://osgeo-org.atlassian.net/browse/GEOT-6140

More integration tests involving both PostgreSQL and Solr are available in this pull request:
https://github.com/geoserver/geoserver/pull/3143


modules/extension/app-schema/app-schema/src/main/java/org/geotools/data/complex/IndexedMappingFeatureIterator.java
```diff
@@ -0,0 +1,228 @@
...
+        return unrolledFilter;
+    }
+
+    /**
+     * Simplifies id value, cutting "typename." part if exists
+     *
+     * @param feature
+     * @return
+     */
+    protected String simplifyIndentifier(Feature feature) {
```
mbarto: I see this one also in IndexIdIterator, maybe we can factor it out for reuse

mbarto:
Looks very good to me, a lot of unit testing too (I am not able to understand every bit, but I cannot see any evident issue). Just a minor comment (fix if you want)

modules/extension/app-schema/app-schema/src/main/java/org/geotools/data/complex/IndexedMappingFeatureIterator.java
nmco: Ah! Nice catch ! This method was _already_ factored out indeed, I just frogot to remove it from here :yum:.

nmco:
Suggestions applied, thank you for the review @mbarto !

nmco:
I will wait for the build to pass and then merge.

PR #2082, opened 2018-10-03
GEOT-6141 Fix ProjectionHandlers valid area bug

NielsCharlier:


aaime:
Verified transformation is correct with cs2cs, merging

NielsCharlier:
Thank you @aaime

PR #2092, opened 2018-10-05
[GEOT-6145]: ForceComponentColorModel noAlpha

dromagnoli:


modules/library/coverage/src/main/java/org/geotools/image/ImageWorker.java
```diff
@@ -1747,20 +1747,24 @@ public final ImageWorker forceComponentColorModel() {
...
      * @return this {@link ImageWorker}.
      * @see FormatDescriptor
      */
     public final ImageWorker forceComponentColorModel(
-            boolean checkTransparent, boolean optimizeGray) {
+            boolean checkTransparent, boolean optimizeGray, boolean omitAlpha) {
         final ColorModel cm = image.getColorModel();
         if (cm instanceof ComponentColorModel) {
-            // Already an component color model - nothing to do.
+            // Already a component color model - nothing to do.
```
aaime: What if it was a component color model with alpha and omitAlpha was false? The javadoc could be made more explicit, and state that the color model is already component, then the alpha is preserved. Also, "omitAlpha" could be renamed to "omitAlphaOnExpand"

dromagnoli: Thx. Going to update it, accordingly.

dromagnoli: Done

aaime:
Thank you, good to go imho.

PR #2102, opened 2018-10-15
[GEOT-6156] Upgrade Surefire plugin.

bradh:


aaime:
Travis build got stuck with no new output for a while, twice in a row, but now it's fine... let's merge and see if anyone else experiences the same problem (and if so, we'll revert and investigate).

PR #2103, opened 2018-10-16
Removing gml:id on geometry encoding, only for GML 3.1

fernandor777:
Removing gml:id on geometry encoding for WFS 1.1.0 GML 3.1

### DIFF
--- a/modules/extension/xsd/xsd-gml2/src/main/java/org/geotools/gml2/simple/GeometryEncoder.java
+++ b/modules/extension/xsd/xsd-gml2/src/main/java/org/geotools/gml2/simple/GeometryEncoder.java
@@ -33,8 +33,16 @@ import org.xml.sax.helpers.AttributesImpl;
  */
 public abstract class GeometryEncoder<T extends Geometry> extends ObjectEncoder<T> {
 
+    /** Flag for to encode gml:id as attribute if is true. */
+    private boolean encodeGmlId = true;
+
     protected GeometryEncoder(Encoder encoder) {
         super(encoder);
+    }
+
+    protected GeometryEncoder(Encoder encoder, boolean encodeGmlId) {
+        super(encoder);
+        this.encodeGmlId = encodeGmlId;
     }
 
     /**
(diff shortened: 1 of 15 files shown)


modules/extension/xsd/xsd-gml2/src/main/java/org/geotools/gml2/simple/GeometryEncoder.java
```diff
@@ -73,7 +81,7 @@ protected AttributesImpl cloneWithGmlId(AttributesImpl atts, String gmlId) {
         } else {
             result = new AttributesImpl(atts);
         }
-        addGmlId(result, gmlId);
+        if (encodeGmlId) addGmlId(result, gmlId);
```
aaime: Probably best at the beginning, ``if gmlId == null || !encodeGMlId``

modules/extension/xsd/xsd-gml2/src/main/java/org/geotools/gml2/simple/GeometryEncoder.java
```diff
@@ -33,10 +33,18 @@
  */
 public abstract class GeometryEncoder<T extends Geometry> extends ObjectEncoder<T> {
 
+    /** Flag for to encode gml:id as attribute if is true. */
+    private boolean encodeGmlId = true;
```
aaime: This way the GML2 encoder also end up having the encodeGMLId property set to true (but should receive a null gmlid so the encoding does not happen), best to have it set to false and enable it in GML 3.2 instead

fernandor777:
Condition check refactored, thanks.

modules/extension/xsd/xsd-gml2/src/main/java/org/geotools/gml2/simple/GeometryEncoder.java
aaime: We discussed internally, causes too much churn in other places, will be left as is.

PR #2105, opened 2018-10-16
Allow clean removal of image mosaic files and support elements when removing from the granule store

aaime:


modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/GranuleStoreDecorator.java
```diff
@@ -0,0 +1,95 @@
+/*
+ * GeoTools - The Open Source Java GIS Toolkit
+ * http://geotools.org
+ *
+ * (C) 2017, Open Source Geospatial Foundation (OSGeo)
+ *
```
dromagnoli: -> 2018 :)

dromagnoli (approved):
Looks good to me.
I have only spotted a wrong copyright year for a new file.
Finally, it looks like the windows build failed :(

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/GranuleStoreDecorator.java
aaime: Ah yep

aaime:
About the windows failure, the Slice2DIndexManager is indeed leaving RandomAccessFile instances open.. trying to figure out where and how.

aaime:
Was due to un-closed readers and deferred loading. Also added machinery to indicate the issue happened and a flag to debug it (similar to the one available on the JDBCFeatureReader)

aaime:
Phew, finally builds on Windows too... I'll squash all the extra commits and merge.

PR #2114, opened 2018-10-20
Fix mvn site, initial part.

bradh:
At this point, it runs to the point of creating error messages for the javadoc, before we hit https://bugs.openjdk.java.net/browse/JDK-8200432


aaime:
Conflicts on master. However, imho, should be merged before we start working on the JDK11 branch (in the branch too), or it might not get merged at all, during the sprint we'll move classes around to resolve split package issues, that will likely create too many conflicts to recover it.

bradh:
Hmm, merging from master in #2113 seemed like a good idea at the time...

Have resolved conflicts and repushed.

aaime:
Looks good to me, not sure if @jodygarnett wants to have a look? But we need to merge before tomorrow morning, europe time

PR #2115, opened 2018-10-20
[GEOT-5438] Curve support not complete for curved polygons with multiple

fernandor777:
Fix the inner curve multiple holes support on Polygon type for Oracle JDBC data source due to an arithmetic bug in inner rings iteration.

https://osgeo-org.atlassian.net/browse/GEOT-5438


aaime:
Hmm... something wrong happened here, the files seem to have been fully rewritten, likely a case of replacing newline endings.

fernandor777:
Looks like original files use CRLF end line and my Git on windows is configured to replace all end line to LF only.  Rebulding the PR commit, thanks!

aaime:
Seems ok to me

aaime:
+1

PR #2116, opened 2018-10-22
update JavaFX dependencies to build with JDK11

ianturton:
Fix dependencies and generally update them.

### DIFF
--- a/modules/unsupported/javafx/README.md
+++ b/modules/unsupported/javafx/README.md
@@ -5,6 +5,8 @@ A JavaFX parent subclass containing a Canvas. A map from a WMS-Server is drawn t
 
 In development.
 
+Note: You need to [download OpenJDK](https://wiki.openjdk.java.net/display/OpenJFX/Download) to use this module.
+
 ## IP Review
 
 
--- a/modules/unsupported/javafx/README.md
+++ b/modules/unsupported/javafx/README.md
@@ -5,8 +5,6 @@ A JavaFX parent subclass containing a Canvas. A map from a WMS-Server is drawn t
 
 In development.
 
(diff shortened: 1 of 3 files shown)


modules/unsupported/javafx/README.md
```diff
@@ -5,6 +5,8 @@ A JavaFX parent subclass containing a Canvas. A map from a WMS-Server is drawn t
 
 In development.
 
+Note: You need to [download OpenJDK](https://wiki.openjdk.java.net/display/OpenJFX/Download) to use this module.
```
aaime: Why so? Should work with Oracle too given the dependencies in the pom. Afaik it was not working with OpenJDK, not even before (OpenJDK did not package JavaFX)

aaime: Why so? Should work with Oracle too given the dependencies in the pom. Afaik it was not working with OpenJDK, not even before (OpenJDK did not package JavaFX)

ianturton: Well I had to download it, but I guess it should work with Oracle if they still package it. 
I can delete that line if you think it is confusing.

ianturton: Well I had to download it, but I guess it should work with Oracle if they still package it. 
I can delete that line if you think it is confusing.

ianturton: Well I had to download it, but I guess it should work with Oracle if they still package it. 
I can delete that line if you think it is confusing.

mipastgt:
Did you have a look at this link? https://openjfx.io/openjfx-docs/#maven Starting with JavaFX11 you don't have to download anything anymore. You can just add the corresponding Maven/Gradle dependencies.

aaime:
@mipastgt  That's what has been done indeed
@ianturton Looks good to me

PR #2118, opened 2018-10-22
replace sun.awt.OSInfo and add JAXB jar

ianturton:
Replace calls to OSInfo with commons-lang3 equivaltents and added a dependency to JAXB api as it is no longer shipped with JDK-11

### DIFF
--- a/modules/plugin/coverage-multidim/coverage-api/pom.xml
+++ b/modules/plugin/coverage-multidim/coverage-api/pom.xml
@@ -94,6 +94,15 @@
       <artifactId>commons-io</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-lang3</artifactId>
+    </dependency>
+    <dependency>
+	  <groupId>javax.xml.bind</groupId>
+	  <artifactId>jaxb-api</artifactId>
+	  <version>2.2.11</version>
+    </dependency>
+    <dependency>
       <groupId>org.locationtech.jts</groupId>
       <artifactId>jts-core</artifactId>
       <!-- The version number is specified in the parent POM. -->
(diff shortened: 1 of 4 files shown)


modules/plugin/coverage-multidim/coverage-api/pom.xml
```diff
@@ -93,6 +93,15 @@
       <groupId>commons-io</groupId>
       <artifactId>commons-io</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-lang3</artifactId>
+    </dependency>
+    <dependency>
```
aaime: Why is JAXB needed here?

ianturton: Not sure, but it fails to build on my machine without it (as at least a test dependency):

```
Running org.geotools.imageio.GeoSpatialImageReaderTest
[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.003 s <<< FAILURE! - in org.geotools.imageio.GeoSpatialImageReaderTest
[ERROR] testReader(org.geotools.imageio.GeoSpatialImageReaderTest)  Time elapsed: 0.002 s  <<< ERROR!
java.lang.IllegalArgumentException: java.lang.NoClassDefFoundError: javax/xml/bind/JAXBException
	at org.geotools.imageio.GeoSpatialImageReaderTest.testReader(GeoSpatialImageReaderTest.java:102)
Caused by: java.lang.NoClassDefFoundError: javax/xml/bind/JAXBException
	at org.geotools.imageio.GeoSpatialImageReaderTest.testReader(GeoSpatialImageReaderTest.java:102)
Caused by: java.lang.ClassNotFoundException: javax.xml.bind.JAXBException
	at org.geotools.imageio.GeoSpatialImageReaderTest.testReader(GeoSpatialImageReaderTest.java:102)
```

ianturton: there is another commit coming (maybe) that centralizes the version and makes it a test dependency

aaime: So the jaxb dependency should be coming from imagemosaic, which is a dependency of coverage-api. But repeating it does not hurt.

aaime:
Looks good to me

PR #2120, opened 2018-10-22
Fix locale for CommonsConverterFactoryTest

tbarsballe:
The en_CA local only supports "a.m." and "p.m." while the format string assumes AM/PM
Fix to a known locale to avoid this issue.
This only appears to fail on Java 11, not Java 8


aaime:
LOL yeah, it's pretty much what I was expecting :)

PR #2122, opened 2018-10-22
Replace Sun FileImageInputStream usage with imageio-ext FileImageInputStreamExtImplSpi

tbarsballe:
FileImageInputStreamSpi is internal proprietary API and may be removed in a future release

### DIFF
--- a/modules/plugin/imagemosaic/src/test/java/org/geotools/gce/imagemosaic/remote/RemoteImageInputStreamSPI.java
+++ b/modules/plugin/imagemosaic/src/test/java/org/geotools/gce/imagemosaic/remote/RemoteImageInputStreamSPI.java
@@ -16,7 +16,6 @@
  */
 package org.geotools.gce.imagemosaic.remote;
 
-import com.sun.imageio.spi.FileImageInputStreamSpi;
 import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


modules/plugin/imagemosaic/src/test/java/org/geotools/gce/imagemosaic/remote/RemoteImageInputStreamSPI.java
```diff
@@ -16,20 +16,21 @@
...
 import javax.imageio.stream.ImageInputStream;
+
+import it.geosolutions.imageio.stream.input.spi.FileImageInputStreamExtImplSpi;
 import org.geotools.TestData;
 import org.geotools.gce.imagemosaic.RemoteTest;
 
 public class RemoteImageInputStreamSPI extends ImageInputStreamSpi {
 
-    private FileImageInputStreamSpi delegate = new FileImageInputStreamSpi();
+    private FileImageInputStreamExtImplSpi delegate = new FileImageInputStreamExtImplSpi();
```
jodygarnett: This looks like a sensible replacement: https://github.com/geosolutions-it/imageio-ext/blob/master/library/streams/src/main/java/it/geosolutions/imageio/stream/input/spi/FileImageInputStreamExtImplSpi.java

Only concern would be if imageioext would like a more generic class returned to have flexibility, if so we can change it again in the future.


And has been around since 2010

dvntucker:
It should be fine. I wonder if there is any reason not to just use javax.imageio.stream.FileImageInputStream directly for this? The Ext version is also deprecated.

tbarsballe:
This seems to want to extend the SPI for service lookup reasons, and javax doesn't expose an SPI interface for that class.

PR #2125, opened 2018-10-23
add additionalJOption to enable doclint directive properly

moovida:
After trying to fix 100 tons of javadoc, noticed that the plugin now requires additionalJOption to be set to enable the doclint directive.

So just submitting the pom fix, since the javadoc is in an "interesting" state and quick fixes might hurt people's work.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1671,6 +1671,8 @@
           <encoding>UTF-8</encoding>
             <!-- Set an additional parameter for the command line. -->
           <additionalparam>-keywords -Xdoclint:none</additionalparam>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -1671,6 +1671,8 @@
           <encoding>UTF-8</encoding>
             <!-- Set an additional parameter for the command line. -->
           <additionalparam>-keywords -Xdoclint:none</additionalparam>
+          <additionalJOption>-Xdoclint:none</additionalJOption>
```
jodygarnett: Does this mean -Xdoclint:done can be removed from additionalparam above?

PR #2126, opened 2018-10-23
gt-main moduleprovider error fix

moovida:
remove non existing serviceprovider + move xml schema providers into right module


aaime:
Test failure is spurious

PR #2137, opened 2018-10-24
[GEOT-5653] Added YSLD support for Rule LegendGraphic elements.

sikeoka:
https://osgeo-org.atlassian.net/browse/GEOT-5653


sikeoka:
Could you please review this pull request when you have time @smithkm?  Thank you.

modules/extension/ysld/src/main/java/org/geotools/ysld/parse/RuleParser.java
```diff
@@ -49,14 +50,17 @@ public void handle(YamlObject<?> obj, YamlParseContext context) {
             featureStyle.rules().add(rule);
 
             rule.setName(r.str("name"));
-            if (r.has("title")) {
-                rule.setTitle(r.str("title"));
-            }
-            if (r.has("abstract")) {
-                rule.setAbstract(r.str("abstract"));
-            }
```
smithkm: This change would be better off in a separate commit as it is really just housekeeping unrelated to the other change.

sikeoka: These unnecessary code changes have been removed.

sikeoka:
I added more unit tests to the pull request.

sikeoka:
Could you please take another look at this pull request @smithkm? Thank you.

PR #2140, opened 2018-10-24
JDK 11 Compatibility fixes

jodygarnett:
Overview:

* compile target=1.8
* avoid use of jdk internal classes
* version updates to dependencies

After this change GeoTools can be built with Java 8 or Java 11.


jodygarnett:
Looks like a conflict has been introduced, will need to rebase :(

jodygarnett:
builder server passed

PR #2141, opened 2018-10-25
WIP: Further casting of ByteBuffer for compatibility with Java 11 when running on Java 8

tbarsballe:
Casting for all invocations of the following methods:

position(int newPosition)
limit(int newLimit)
flip()
clear()
mark()
reset()
rewind()


tbarsballe:
We will not be going this route, and instead will have two separate builds one with JDK8 and one with JDK11, with two separate local repositories.

PR #2154, opened 2018-10-29
JDK11: Merge gt-main and gt-data to resolve split packages

tbarsballe:
* Merge gt-main and gt-data to resolve split packages
* Refactor poms to account for merged modules (also removed a couple other duplicate dependency declarations)
* Refactor gt-render to move org.geotools.styling classes to org.geotools.render.style


aaime:
The JDK11 is due the build locking up, has been happening a lot on Travis + jdk11, trying to investigate in another thread. Can we merge this and lift the "freeze"? Or is there anything else that need doing?

tbarsballe:
Code-wise, this is fine to merge, and probably best to merge sooner rather than later.

There's still some doc work that will need to be done, although that can be done after this is merged

aaime:
Yep, an extended freeze was not agreed upon in advance, so while I appreciated the PR, I'd rather get it in ASAP and fix docs later, rather than just block everybody else's work.

tbarsballe:
I'll just go ahead and merge then.

PR #2155, opened 2018-10-29
Java 11 quickstart

tbarsballe:
Basic Quickstart for running the Quickstart tutorial app on Java 11 under the module system. I've tested it out and it runs fine without the need for any additional runtime arguments.

Depends upon https://github.com/geotools/geotools/pull/2154 to run.

Also includes a fix to Hints that I discovered when running the tutorial - looks like this was missed during the refactor of ScreenMap

### DIFF
--- a/docs/user/tutorial/quickstart/artifacts/module-info.java
+++ b/docs/user/tutorial/quickstart/artifacts/module-info.java
@@ -1,0 +1,7 @@
+module org.geotools.tutorial.quickstart {
+    requires java.desktop;
+    requires org.geotools.main;
+    requires org.geotools.shapefile;
+    requires org.geotools.swing;
+    requires org.geotools.render;
+}
--- a/docs/user/tutorial/quickstart/intellij.rst
+++ b/docs/user/tutorial/quickstart/intellij.rst
@@ -14,7 +14,7 @@ Pre-Requisites
 --------------
 This guide assumes the following:
 
-* You have the latest JDK installed (8 at the time this article was written) installed. If not the TODO Eclipse Quickstart
+* You have the latest JDK installed (8 at the time this article was written) installed. If not the `Eclipse Quickstart <./eclipse.html>`_
   provides instructions on how to do this.
 * You have IntelliJ installed. This article targets IntelliJ CE 2016; however, previous versions at least as far back as
   13 should work fine. Ultimate versions should also work fine. IntelliJ can be downloaded from JetBrains_ and generally
@@ -202,7 +202,7 @@ shapefile on the screen.
    * We are using a very basic display style here that just shows feature outlines. In the examples that follow we will see how to specify more sophisticated styles.
   
 Things to Try
-=============
+-------------
 
 Each tutorial consists of very detailed steps followed by a series of extra questions. If you get
 stuck at any point please ask your instructor; or sign up to the geotools-users_ email list.
--- a/docs/user/tutorial/quickstart/java11.rst
+++ b/docs/user/tutorial/quickstart/java11.rst
@@ -1,0 +1,152 @@
+:Author: Torben Barsballe
+:Thanks: geotools-user list
+:Version: |release|
+:License: Creative Commons with attribution
+
+**********************
+ Java 11 Quickstart 
+**********************
+
+.. sectionauthor:: Torben Barsballe <tbarsballe@boundlessgeo.com>
+
+Welcome 
+=======
+
+GeoTools 21.0 and newer can be used in a Java 11 modular application. This requires a few additions on top of the regular tutorial application.
+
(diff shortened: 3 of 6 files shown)


docs/user/tutorial/quickstart/intellij.rst
```diff
@@ -14,7 +14,7 @@ Pre-Requisites
 --------------
 This guide assumes the following:
 
-* You have the latest JDK installed (8 at the time this article was written) installed. If not the TODO Eclipse Quickstart
+* You have the latest JDK installed (8 at the time this article was written) installed. If not the `Eclipse Quickstart <./eclipse.html>`_
```
jodygarnett: Java 8 is no longer the latest at the time of writing

docs/user/tutorial/quickstart/java11.rst
```diff
@@ -0,0 +1,143 @@
...
+**********************
+ Java 11 Quickstart 
+**********************
+
+.. sectionauthor:: Torben Barsballe <tbarsballe@boundlessgeo.com>
+
+Welcome 
+=======
+
+GeoTools 21.0 and newer can be used in a Java 11 modular application. This requires a few additions on top of the regular tutorial application.
```
jodygarnett: should we say "jigsaw" modular application, since modules were introduced prior to Java 11 (and will exist for longer than Java 11).

docs/user/tutorial/quickstart/java11.rst
```diff
@@ -0,0 +1,143 @@
...
+
+Java Install
+============
+
+Instead of Java 8, we will want to install Java 11.
+
+#. Download the latest Java 11 JDK:
+
+   * Oracle JDK: http://www.oracle.com/technetwork/java/javase/downloads/
+   * OpenJDK: http://jdk.java.net/11/
```
jodygarnett: Consider putting the Oracle JDK in a information or warning box to explain it is not an open source download:

```
.. info:

  OpenJDK updates every six months, for longer term support consider:
  
  * Adopt OpenJDK: https://adoptopenjdk.net
  
  Oracle distributes a supported JDK 11 for their customers, which is available for development and testing:
  
  * http://www.oracle.com/technetwork/java/javase/downloads/
```

docs/user/tutorial/quickstart/java11.rst
```diff
@@ -0,0 +1,143 @@
...
+   http://www.naturalearthdata.com/ project which is supported by the North American Cartographic
+   Information Society. Head to the link below and download some cultural vectors. You can use the 'Download all 50m cultural themes' at top.
+   
+   * `1:50m Cultural Vectors <http://www.naturalearthdata.com/downloads/50m-cultural-vectors/>`_
+   
+   Unzip the above data into a location you can find easily such as the desktop.
+
+#. You can run the application using Maven on the command line::
+   
+     mvn exec:java -Dexec.mainClass=org.geotools.tutorial.quickstart.Quickstart
```
jodygarnett: That is it? What about module path fun and games?

* https://www.mojohaus.org/exec-maven-plugin/examples/example-exec-for-java-programs.html

docs/user/tutorial/quickstart/artifacts/module-info.java
```diff
@@ -0,0 +1,7 @@
+module org.geotools.tutorial.quickstart {
+    requires java.desktop;
+    requires org.geotools.main;
+    requires org.geotools.shapefile;
+    requires org.geotools.swing;
+    requires org.geotools.render;
+}
```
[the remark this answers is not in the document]
tbarsballe: Nope, you need to list anything you import into one of your classes (I tried shorter initially and got errors)

docs/user/tutorial/quickstart/intellij.rst
tbarsballe: That was mostly just a fix to the TODO; all the quickstarts should probably have updated wordings around java versions

docs/user/tutorial/quickstart/java11.rst
tbarsballe: Maven does it for you as far as I can tell. Some combination of the release version config targeting 11 and declaring the module info results in a state where you don't need to declare anything explicitly.

I tried it out, and it worked fine.

docs/user/tutorial/quickstart/java11.rst
tbarsballe: My only concern is people tend to have no idea what you're talking about when you say "Jigsaw" Maybe this will change as people get more familiar with it, but for now I prefer it as written since most people will just care about what they need to do to get it running on Java 11.

docs/user/tutorial/quickstart/java11.rst
tbarsballe: Good idea, will add

PR #2156, opened 2018-10-30
Trying variations to see where the build gets stuck on JDK11, and hopefully fix it

aaime:
DO NOT MERGE :-)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ env:
   global:
     - MAVEN_OPTS=-Xmx512m
 script:
-  - mvn -B -fae -T1 -Dall clean install -Dfmt.action=check -U
+  - mvn -B -fae -T2 -Dall clean install -Dfmt.action=check -U
 notifications:
   email: false
   on_failure: never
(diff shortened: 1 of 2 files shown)


aaime:
Single threaded build work, but build time is insane. Trying upgrading plugins.

mprins:
You could look into using `travis_wait` with some timeout in minutes, it seems most jdk11 build are timing out with 10min "no output" [1], eg. `travis_wait 20 mvn -B -fae -T4 -Dall clean install -Dfmt.action=check -U` will give you 20 mins instead of (default) 10 mins.

Also not using `-U` should save a lot of network I/O

[1] https://travis-ci.org/geotools/geotools/jobs/447994326#L5719-L5722

aaime:
There is nothing in the build pausing for 10 minutes straight, it's definitely a lockup. I cannot reproduce the issue anymore with upgraded maven plugins, will try a few more times and if that solves it, merge.
GeoTools is currently not depending on snapshots (at least I think it does not) so -U could indeed be removed.

aaime:
Seems to work reliably, I'll squash the changes and merge manually

PR #2157, opened 2018-10-31
GEOT-6166

jericks:
https://osgeo-org.atlassian.net/browse/GEOT-6166

1. The GeoJSON format in older versions of GDAL doesn't support LayerCanCreateField correctly.  Add it to the special exceptions list.
2. The failing spatial query test was due to GDAL (on my computer) not being compiled with GEOS support.  If GEOS is not included then GDAL uses bounding boxes instead of actual geometries.
3. The CSV format test was failing because the first row which contains the headers was being counted because of the property names the test was using.  For some unknown reason the property 'd' was causing GDAL to think the first row had data.  Make the property names slightly more realistic.


tbarsballe:
@jericks - Have you had a chance to look at Andrea's feedback yet? Can this be merged, or does it still need more work.

jericks:
It needs more work.  I need to incorporate Andrea's changes.

aaime:
It seems this one stalled, or did we miss activity on it?

aaime:
Superceded by https://github.com/geotools/geotools/pull/2269

PR #2160, opened 2018-11-01
19.x StreamingRenderer().paint() methods high memory

553226713:
hello!
I justed use StreamingRenderer().paint(graphics, rectangle, referencedEnvelope); methods to export png.
about 10000+.
But I found that the memory was not freed，
I used java1.8.
I'm sure it was caused by this.
thanks.


aaime:
Please don't mis-use pull requests as a mean to get support.
If you want to participate in the project (get and _give_ support) then go to the user list, as described here: http://geotools.org/getinvolved.html
Otherwise you can check commercial support providers too: http://geoserver.org/comm/

PR #2164, opened 2018-11-02
[GEOT-6171] Add a remove indexed color model method in ImageWorker

aaime:


aaime:
Java 11 build still timeout out after 10 minutes of no output, but not sure what else to do about it, does not happen locally, already tried upgrading all maven plugins.

PR #2169, opened 2018-11-09
[GEOT-6177] NEAREST spatial predicate for Postgis

fernandor777:
NEAREST spatial predicate implementation in ECQL, with the usage being something similar to NEAREST(POINT(21707 6558912),10), returning the nearest 10 features to POINT.

Uses postgis `<->` operator with a subquery.

https://osgeo-org.atlassian.net/browse/GEOT-6177

### DIFF
--- a/modules/library/main/src/main/java/org/geotools/data/util/FeatureStreams.java
+++ b/modules/library/main/src/main/java/org/geotools/data/util/FeatureStreams.java
@@ -48,7 +48,7 @@ public final class FeatureStreams {
         return StreamSupport.stream(Spliterators.spliteratorUnknownSize(fi, 0), false)
                 .onClose(
                         () -> {
-                            fc.features().close();
+                            fi.close();
                         });
     }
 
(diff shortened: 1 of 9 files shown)


modules/library/jdbc/src/main/java/org/geotools/jdbc/ISQLDialect.java
```diff
@@ -0,0 +1,961 @@
...
+import org.locationtech.jts.geom.Geometry;
+import org.locationtech.jts.geom.GeometryFactory;
+import org.opengis.feature.FeatureVisitor;
+import org.opengis.feature.simple.SimpleFeatureType;
+import org.opengis.feature.type.AttributeDescriptor;
+import org.opengis.feature.type.GeometryDescriptor;
+import org.opengis.filter.Filter;
+import org.opengis.referencing.crs.CoordinateReferenceSystem;
+
+public interface ISQLDialect {
```
aaime: This is not .NET, interfaces do not start with a capital I.
Also this is a API change, the kind of thing that should be discussed before being done.
Can you elaborate on how this relates to the code sharing change request made yesterday?

fernandor777: The purpose of that is to use both unrelated dialects objects only, anyway is still a work in progress and I'm changing it for a lambda passing instead.  Thanks for your comment and you are right there is a simpler way to do it without breaking the API.

PR #2170, opened 2018-11-13
[GEOT-6178] Invalid select bounds query generated for Oracle database

nmco:
Associated issue:
https://osgeo-org.atlassian.net/browse/GEOT-6178

This issue can only be reproduced with Oracle. The tests where already there and failing, note that some Oracle tests are still failing but its unrelated to this issue cause.

Before this PR:

image

After this PR:

image

### DIFF
--- a/modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
+++ b/modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
@@ -3834,7 +3834,7 @@ public final class JDBCDataStore extends ContentDataStore implements GmlObjectSt
                             "AND " + toSQL.encodeToString(filter));
                     toSQL.setInline(false);
                 } else {
-                    sql.append(toSQL.encodeToString(filter));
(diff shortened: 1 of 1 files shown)


modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
```diff
@@ -3834,7 +3834,7 @@ protected PreparedStatement selectBoundsSQLPS(
                             "AND " + toSQL.encodeToString(filter));
                     toSQL.setInline(false);
                 } else {
-                    sql.append(toSQL.encodeToString(filter));
+                    sql.append(" ").append(toSQL.encodeToString(filter));
```
aaime: So this restores the situation before the placeholder changes, for the record:

````
            try {
                toSQL = createPreparedFilterToSQL(featureType);
                sql.append(" ").append(toSQL.encodeToString(filter));
            } catch (FilterToSQLException e) {
            }
````

PR #2173, opened 2018-11-15
[GEOT-6179] Delegate filter execution matching multiple same-xpath feature chained attribute names to database

fernandor777:
Currently app-schema plugin detect two posible attributes with same xpath and delegate filter execution to JVM runtime. This fix delegates filter execution to database using all available join-fields.

https://osgeo-org.atlassian.net/browse/GEOT-6179


tbarsballe:
This needs a test case as per [CONTRIBUTING.md](https://github.com/geotools/geotools/blob/master/CONTRIBUTING.md)

Everything else seems fine as far as I can tell, although I am not especially familiar with app-schema.

nmco:
Looks good to me too, f I'm not mistake, as usual for App-Schema, tests have been implemented in GeoServer:
https://github.com/geoserver/geoserver/pull/3239

Right @Fernando?

fernandor777:
Indeed app-schema integration test on:
https://github.com/geoserver/geoserver/pull/3239

PR #2176, opened 2018-11-20
[GEOT-6181] Implement a Equal Area function for vector data classification

aaime:


modules/library/main/src/main/java/org/geotools/data/util/ScreenMap.java
```diff
@@ -134,17 +135,24 @@ public boolean checkAndSet(int x, int y) {
...
                 return false;
             }
         } catch (Exception e) {
             return true;
         }
     }
 
+    private int[] getPixels() {
+        if (pixels == null) {
+            pixels = new int[arraySize];
```
ianturton: might want to check arraySize isn't too big here and that there is no OOM issue

modules/library/main/src/main/java/org/geotools/geometry/jts/GeometryClipper.java
```diff
@@ -163,6 +163,10 @@ public Geometry clip(Geometry g, boolean ensureValid) {
         if (g instanceof LineString) {
             return clipLineString((LineString) g);
         } else if (g instanceof Polygon) {
+            Polygon polygon = (Polygon) g;
+            //            if (polygon.getNumInteriorRing() == 0) {
```
ianturton: left over comments

modules/library/main/src/test/java/org/geotools/feature/visitor/EqualAreaListVisitorTest.java
```diff
@@ -0,0 +1,193 @@
...
+    }
+
+    public static double getTotalArea(ListFeatureCollection fc, Expression areaFunction) {
+        return fc.stream().mapToDouble(f -> areaFunction.evaluate(f, Double.class)).sum();
+    }
+
+    /** Returns min and max value applying the expression over the feature collection */
+    public static NumberRange<Double> getMinMax(
+            ListFeatureCollection fc, Expression attributeExpression) {
+        double min =
```
ianturton: while streams are cool, does this really win over a single loop generating min and max together?

modules/library/main/src/main/java/org/geotools/data/util/ScreenMap.java
aaime: Oh sorry, these changes got in by mistake, generating vector tiles allocates a 4000x4000 screenarea now, often for nothing, so I was trying to delay it, but this experimental code should be cleaned up. I have to remove it from this PR, it's unrelated

modules/library/main/src/test/java/org/geotools/feature/visitor/EqualAreaListVisitorTest.java
aaime: It does not, but it's not performance critical and I find it more readable

aaime: That said, I can turn it back to a loop, but it's just test code :-p

modules/library/main/src/main/java/org/geotools/filter/function/AbstractQuantityClassificationFunction.java
```diff
@@ -0,0 +1,144 @@
...
+     * Returns the list visitor for the specific implementation of "quantity" (e.g., count, area,
+     * ...). The visitor must return a "bins" structure matching {code}List<Comparable>[]{code},
+     * where each array entry is a bin, and values inside the bin are sorted from lowest to highest
+     *
+     * @return
+     */
+    protected abstract FeatureCalc getListVisitor();
+
+    private Object calculate(SimpleFeatureCollection featureCollection) {
+        FeatureCalc quantileVisit = getListVisitor();
```
mbarto: Maybe you mean quantityVisit here?

modules/library/main/src/main/java/org/geotools/geometry/jts/GeometryClipper.java
```diff
@@ -163,6 +163,10 @@ public Geometry clip(Geometry g, boolean ensureValid) {
         if (g instanceof LineString) {
             return clipLineString((LineString) g);
         } else if (g instanceof Polygon) {
+            Polygon polygon = (Polygon) g;
```
mbarto: Is this used anywhere?

modules/library/main/src/main/java/org/geotools/data/util/ScreenMap.java
aaime: Actually, found a better way to collect min and max in a single pass, will update the PR.

modules/library/main/src/test/java/org/geotools/feature/visitor/EqualAreaListVisitorTest.java
aaime: Actually, found a better way to collect min and max in a single pass, will update the PR.

modules/library/main/src/main/java/org/geotools/filter/function/AbstractQuantityClassificationFunction.java
aaime: Yes, leftover from the extraction of a common superclass from the old Quantile function

modules/library/main/src/main/java/org/geotools/geometry/jts/GeometryClipper.java
aaime: To be removed, it was weekend code while I was trying to improve the MVT generation perf that got left on disk

modules/library/main/src/test/java/org/geotools/feature/visitor/EqualAreaListVisitorTest.java
ianturton: I was just wondering about the speed difference really

aaime: They are slower :-)

aaime:
@ianturton @mbarto code should have been updated based on your feedback

PR #2177, opened 2018-11-20
[GEOT-6182] Add the possibility to specify the target CRS for the ReprojectingFilterVisitor

nmco:
Associated issue:
https://osgeo-org.atlassian.net/browse/GEOT-6182

Currently the _target CRS_ is always inferred from the schema, this gives the possibility for the client code to explicit set the target CRS. This changes are backward compatible.


modules/library/main/src/main/java/org/geotools/filter/spatial/ReprojectingFilterVisitor.java
```diff
@@ -103,7 +111,11 @@ public Object visit(BBOX filter, Object extraData) {
         } else if (filter.getExpression2() instanceof PropertyName) {
             propertyName = (PropertyName) filter.getExpression2();
         }
-        CoordinateReferenceSystem targetCrs = findPropertyCRS(propertyName);
+        CoordinateReferenceSystem targetCrs = this.targetCrs;
```
aaime: Why make it BBOX specific, would it make sense to move this inside findPropertyCRS instead, so that geometry literals and other comparisons get reprojected as well?

nmco: Good point.

nmco:
Changes applied, thanks for the review @aaime !

PR #2182, opened 2018-11-22
[GEOT-6184] MongoDB complex store iterator enters in an infinite loop if no ID is defined

nmco:
Associated issue: https://osgeo-org.atlassian.net/browse/GEOT-6184

Integration tests updated in GeoServer.

### DIFF
--- a/modules/plugin/mongodb/src/main/java/org/geotools/data/mongodb/complex/MongoComplexDataSource.java
+++ b/modules/plugin/mongodb/src/main/java/org/geotools/data/mongodb/complex/MongoComplexDataSource.java
@@ -1,0 +1,78 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ *
+ *    (C) 2018, Open Source Geospatial Foundation (OSGeo)
+ *
+ *    This library is free software; you can redistribute it and/or
+ *    modify it under the terms of the GNU Lesser General Public
+ *    License as published by the Free Software Foundation;
+ *    version 2.1 of the License.
+ *
+ *    This library is distributed in the hope that it will be useful,
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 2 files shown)


modules/plugin/mongodb/src/main/java/org/geotools/data/mongodb/complex/MongoComplexDataSource.java
```diff
@@ -0,0 +1,78 @@
...
+import org.geotools.data.mongodb.MongoDataStore;
+import org.opengis.feature.Feature;
+import org.opengis.feature.type.FeatureType;
+
+/**
+ * Class that builds an App-Schema iterator ready to be used with MongoDB. Data coming from MongoDB
+ * is assumed to be always normalized, when using MongoDB with App-Schema we cannot chain multiple
+ * collections together.
+ */
+public final class MongoComplexDataSource implements CustomSourceDataStore {
```
tbarsballe: This adds a whole new datasource to handle the iterator issue, but I can't see where this is actually being used?

Looking at the upstream GeoServer changes (https://github.com/geoserver/geoserver/pull/3252 ), they seem to be just changes in some test resources to get the iterator to be called, but again I can't see exactly how this is called.

An actual standalone test case for this issue would be helpful.

nmco: Thanks for the feedback.

```
This adds a whole new datasource to handle the iterator issue, but I can't see where this is actually being used?
```

The name is a bit unfortunate, but this doesn't add a new data source ... it just handles the construction of the existing iterator so it can set the ``isFitlered`` option to ``false``.

``but again I can't see exactly how this is called.``

This is using an existing extension point in App-Schema:
https://github.com/geotools/geotools/blob/master/modules/extension/app-schema/app-schema/src/main/java/org/geotools/data/complex/MappingFeatureIteratorFactory.java#L291-L296

``An actual standalone test case for this issue would be helpful.``

The changes make in the existing test makes it a standalone test case, having a random ID without this fix will make it enter in an infinite loop.

PR #2190, opened 2018-11-27
[GEOT-6190] Disable jdbc-spatialite pending Java 11 upgrade

bradh:
Intended to be temporary. There will need to be corresponding GS changes.


jodygarnett:
Q: Is it worth making a profile so it still can be build on demand?

bradh:
I would hope it won't be out for very long.

PR #2195, opened 2018-11-30
[GEOT-6194] Unit tests for the gt-jdbc-hana metadata installer

stefanuhrig:
The gt-jdbc-hana plugin comes with a metadata installer. Currently,
there are no unit tests for the metadata installer. Hence, unit tests
should be added.


aaime:
Hi @stefanuhrig for changes that are only hitting the SAP HANA module you should be able to just go and commit, if you want (I just invited you in the organization so that you can get commit access). It does not mean that you have to, you can still do PRs as a verification that things are building properly outside of your machine, and you should do PRs for anything that involves changes outside of your module.

stefanuhrig:
Hi @aaime,

Thanks, but can you re-check/re-sent the invitation? I received a notification about your comment, but I did not receive an invitation to the organization. I also checked my spam folder, but could not find anything.

PR #2203, opened 2018-12-05
[GEOT-6198] Do not set ROIs in image mosaic when not strictly necessary

aaime:
Depends on https://github.com/geosolutions-it/jai-ext/pull/227 and a jai-ext release

### DIFF
--- a/modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/granulecollector/BaseSubmosaicProducer.java
+++ b/modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/granulecollector/BaseSubmosaicProducer.java
@@ -17,9 +17,9 @@
 
 package org.geotools.gce.imagemosaic.granulecollector;
 
+import it.geosolutions.jaiext.JAIExt;
 import it.geosolutions.jaiext.vectorbin.ROIGeometry;
-import java.awt.Color;
-import java.awt.Rectangle;
+import java.awt.*;
 import java.awt.image.ColorModel;
 import java.awt.image.IndexColorModel;
 import java.awt.image.MultiPixelPackedSampleModel;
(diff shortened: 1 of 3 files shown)


modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/granulecollector/BaseSubmosaicProducer.java
```diff
@@ -83,6 +83,8 @@
 
     protected Color inputTransparentColor;
 
+    // boolean forceFootprints = JAIExt.;
```
dromagnoli: Is it here for "debugging/testing" purposes?

pom.xml
```diff
@@ -97,7 +97,7 @@
     <src.output>${basedir}/target</src.output>
     <imageio.ext.version>1.2.0</imageio.ext.version>
     <jts.version>1.16.0</jts.version>
-    <jaiext.version>1.1.4</jaiext.version>
+    <jaiext.version>1.1-SNAPSHOT</jaiext.version>
```
dromagnoli: I think you set it to SNAPSHOT for development and testing. should it be updated to 1.1.next once Jai-ext get released?

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/granulecollector/BaseSubmosaicProducer.java
aaime: Yep, must be removed

pom.xml
aaime: That's why I said "Depends on geosolutions-it/jai-ext#227 **and a jai-ext release**" :-D

aaime:
Travis builds keep on failing for network related issues, merging anyways and waiting for the official build server verdict.

PR #2204, opened 2018-12-07
[GEOT-6199] ImageMosaic harvesting does not scale up

aaime:


aaime:
Works fine locally and on AppVeyor, Travis is having what seems to be network problems, the build fails after a handful of modules with a lockup (no output)

PR #2205, opened 2018-12-09
[GEOT-6200] Reproject fitler tries to reproject non geometry properties

nmco:
Associated issue:
https://osgeo-org.atlassian.net/browse/GEOT-6200


nmco:
The build keeps getting stuck when retrieving WMTS dependencies, I build locally and no failures. I'm merging anyway and monitoring Jenkins builds.

PR #2209, opened 2018-12-14
[GEOT-6201] Fixed SLD encoding of BBox function

robward-scisys:
Prevented crash when SLD BBox function is encoded in XML.


aaime:
Looks good to me, thank you! 
The AppVeyor build failure seems to be unrelated (we have had build timeouts lately). Merging. 
If you want backports please issue pull requests against the corresponding branches.

PR #2211, opened 2018-12-15
[GEOT-6161] Fails to read MBTiles with JPG file format

leif81:
Fix for https://osgeo-org.atlassian.net/projects/GEOT/issues/GEOT-6161

The spec confirms "jpg" is a valid format here:

https://github.com/mapbox/mbtiles-spec/blob/master/1.3/spec.md


aaime:
Looks legit to me. Needs a test though, you can check the contribution guidelines here:
Hopefully you can mimick other existing tests (I have no familiarity with the module, sorry I cannot be more specific than that).

leif81:
@aaime Thank you for the feedback. I've added a unit test and squished it into the previous commit here. Can you take a look?

aaime:
The change seems sane to me, but I had no involvement with this module yet, I've asked for @smithkm  and @tbarsballe opinion, they are the last ones to do actual changes to it (as opposed to side effects of some other core refactor).

modules/unsupported/mbtiles/src/test/java/org/geotools/mbtiles/MBTilesMetadataTest.java
```diff
@@ -0,0 +1,12 @@
+package org.geotools.mbtiles;
+
+import org.junit.Test;
+
+public class MBTilesMetadataTest {
```
tbarsballe: One note here - the MBTilesFileTest class has a testMBTilesMetaData, so the class name for this test is a little misleading/redundant - it might be better if you just added your new test as its own test case to the MBTilesFileTest instead of creating a new test class.

leif81:
@tbarsballe The only .mbtiles file I have of jpg type is quite big so I haven't included that here.

tbarsballe:
Looks good to me, I've squashed and merged to master

PR #2212, opened 2018-12-17
[GEOT-6201] Fixed SLD encoding of BBox function, backport

robward-scisys:
Prevented crash when SLD BBox function is encoded in XML.


aaime:
Yep thanks, merged!

PR #2219, opened 2019-01-08
GEOT-6207: Better disposal of inner images through ImageUtilities dis…

dromagnoli:
…posal methods

### DIFF
--- a/modules/library/coverage/src/main/java/org/geotools/image/util/ImageUtilities.java
+++ b/modules/library/coverage/src/main/java/org/geotools/image/util/ImageUtilities.java
@@ -1224,8 +1224,17 @@ public final class ImageUtilities {
                     }
                 } else {
                     // Trying disposing the reader
-                    Object imageReader =
-                            inputImage.getProperty(ImageReadDescriptor.PROPERTY_NAME_IMAGE_READER);
+                    Object imageReader = null;
+
+                    try {
+                        imageReader =
+                                inputImage.getProperty(
+                                        ImageReadDescriptor.PROPERTY_NAME_IMAGE_READER);
+                    } catch (NullPointerException npe) {
+                        // for some reason, chained cleanup may have already cleaned (and null) some
+                        // underlying images. let's ignore it
+                    }
+
                     if ((imageReader != null) && (imageReader instanceof ImageReader)) {
                         final ImageReader reader = (ImageReader) imageReader;
                         final ImageInputStream stream = (ImageInputStream) reader.getInput();
@@ -1257,7 +1266,13 @@ public final class ImageUtilities {
      */
     public static void disposeSinglePlanarImage(PlanarImage planarImage) {
         // Looking for an ROI image and disposing it too
-        final Object roi = planarImage.getProperty("ROI");
+        Object roi = null;
+        try {
+            roi = planarImage.getProperty("ROI");
+        } catch (NullPointerException npe) {
+            // for some reason, chained cleanup may have already cleaned (and null) some images
+            // let's ignore it
+        }
         if ((roi != null)
                 && ((ROI.class.equals(roi.getClass()) || (roi instanceof RenderedImage)))) {
             if (roi instanceof ROI) {
@@ -1276,11 +1291,12 @@ public final class ImageUtilities {
         }
 
         try {
-            if (planarImage instanceof RenderedImageAdapter) {
-                cleanField(planarImage, "theImage");
-            }
+            // Check extended class first
             if (planarImage instanceof WritableRenderedImageAdapter) {
                 cleanField(planarImage, "theWritableImage");
+                cleanField(planarImage, "theImage", true);
+            } else if (planarImage instanceof RenderedImageAdapter) {
+                cleanField(planarImage, "theImage");
             }
         } catch (NoSuchFieldException | IllegalAccessException e) {
             // fine, we tried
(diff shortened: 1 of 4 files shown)


dromagnoli:
coverage disposal has been removed from tests since the underlying image get already disposed.
rendering test has been updated to avoid re-using an already disposed image

modules/library/coverage/src/main/java/org/geotools/image/util/ImageUtilities.java
```diff
@@ -1276,11 +1291,12 @@ public static void disposeSinglePlanarImage(PlanarImage planarImage) {
...
 
         try {
-            if (planarImage instanceof RenderedImageAdapter) {
-                cleanField(planarImage, "theImage");
-            }
+            // Check extended class first
             if (planarImage instanceof WritableRenderedImageAdapter) {
                 cleanField(planarImage, "theWritableImage");
+                cleanField(planarImage, "theImage", true);
+            } else if (planarImage instanceof RenderedImageAdapter) {
```
aaime: Before the two ifs were not alternative, now they are. I guess this is correct as the two fields in question are pointing to the same image right?

modules/library/coverage/src/main/java/org/geotools/image/util/ImageUtilities.java
```diff
@@ -1303,10 +1319,30 @@ public static void disposeSinglePlanarImage(PlanarImage planarImage) {
...
+     * @param theObject
+     * @param fieldName
+     * @throws NoSuchFieldException
+     * @throws IllegalAccessException
+     */
+    private static void cleanField(Object theObject, String fieldName, boolean superClass)
+            throws NoSuchFieldException, IllegalAccessException {
+        // getDeclaredField only provides access to current class,
+        // not superclass.
+        Class<? extends Object> theClass = theObject.getClass();
```
aaime: Why not use the getField method instead? It goes up the inheritance chain by itself, no need for flags

modules/library/render/src/test/java/org/geotools/renderer/lite/StreamingRendererTest.java
```diff
@@ -223,7 +223,7 @@ public void testDrawIntepolation() throws Exception {
         layer.getUserData()
                 .put(
                         StreamingRenderer.BYLAYER_INTERPOLATION,
-                        Interpolation.getInstance(Interpolation.INTERP_BICUBIC));
+                        Interpolation.getInstance(Interpolation.INTERP_BICUBIC_2));
```
aaime: What happened here?

modules/library/render/src/test/java/org/geotools/renderer/lite/StreamingRendererTest.java
```diff
@@ -237,19 +237,9 @@ public void testDrawIntepolation() throws Exception {
         // test right interpolation hint is set on Graphics2D
         assertEquals(
                 graphics.getRenderingHint(JAI.KEY_INTERPOLATION),
-                Interpolation.getInstance(Interpolation.INTERP_BICUBIC));
```
aaime: And here?

modules/library/coverage/src/main/java/org/geotools/image/util/ImageUtilities.java
dromagnoli: getField only get public fields, not private ones :(

modules/library/coverage/src/main/java/org/geotools/image/util/ImageUtilities.java
dromagnoli: right.

modules/library/render/src/test/java/org/geotools/renderer/lite/StreamingRendererTest.java
dromagnoli: Just using a "not commonly used" interpolation to make sure hints is actually using it once set (so it's not using a default)

modules/library/coverage/src/main/java/org/geotools/image/util/ImageUtilities.java
```diff
@@ -1276,11 +1291,12 @@ public static void disposeSinglePlanarImage(PlanarImage planarImage) {
...
         try {
-            if (planarImage instanceof RenderedImageAdapter) {
-                cleanField(planarImage, "theImage");
-            }
+            // Check extended class first
             if (planarImage instanceof WritableRenderedImageAdapter) {
                 cleanField(planarImage, "theWritableImage");
+                cleanField(planarImage, "theImage", true);
+            } else if (planarImage instanceof RenderedImageAdapter) {
+                cleanField(planarImage, "theImage");
```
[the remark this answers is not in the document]
dromagnoli: @lislei do you have any chance to setup a dataset (data + style eventually and configuration steps) and requests to replicate that?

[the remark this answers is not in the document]
lislei: What we have is a webstart app based on geotools that produces this nullpointer when painting a wms layer. I could look into giving you access to it if that would help?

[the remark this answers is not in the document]
lislei: My colegue pointed out to me [a self contained example in this post](https://sourceforge.net/p/geotools/mailman/message/36602879/) - see attachement "WmsTest.java". 
This should produce the nullpointer.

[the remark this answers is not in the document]
aaime: Your colleague deserves to pay and use commercial software only. If you are interested in getting any help instead, please open a ticket on https://osgeo-org.atlassian.net/projects/GEOT attaching the example and wording the description like a decent, polite person would do.

PR #2220, opened 2019-01-08
[GEOT-6209] App-schema Postgresql subquery: replace OR with UNION performance improvement

fernandor777:
Is known that OR queries are difficult to optimize for postgresql and are usually slow.

The fix for slow queries with OR involved is using UNION queries, so we need optimize certain complex subqueries. The slow appschema subquery involved is that generated by NestedFilterToSQL class so we need to change the use of OR clause in the root condition for UNION queries.

Since this performance issue affect postgresql, the fix should be isolated for this datasource only, and adding a system property for disable.

With UNION improvement enabled main binary operator on nested filter subquery will rebuild normal OR query like:
       SELECT id, name FROM table WHERE name = "Morty" OR name = "Rick"
       ->
       SELECT id, name FROM table WHERE name = "Morty" UNION SELECT id, name FROM table WHERE name = "Rick"


fernandor777:
Thanks @nmco I did modifications in base to your review.

nmco:
Cool, thanks @fernandor777 !

PR #2224, opened 2019-01-11
[GEOT-6236] (gt-swing) Remove white flash when pan/zooming map

leif81:
The JMapPane drawLayers() method implementation clears the images viewport rectangle and kicks off an asynchronous rendering task.  Calling "repaint()" immediately after this call pretty much guarantees one will get a "white flash" when the user stops  panning.  Furthermore the "onRenderingComplete()" method of AbstractMapPane  contains a call to "repaint()" already so this one is not required.

Credit to @MurrayMcDonald16


ianturton:
please create a jira issue and reference it here.

leif81:
@ianturton ok will do.

leif81:
Sorry for the delay getting this filed in JIRA.

https://osgeo-org.atlassian.net/browse/GEOT-6236

PR #2225, opened 2019-01-11
[GEOT-6209] App-schema Postgresql subquery: replace OR with UNION performance improvement (20.x backport)

fernandor777:
Is known that OR queries are difficult to optimize for postgresql and are usually slow.

The fix for slow queries with OR involved is using UNION queries, so we need optimize certain complex subqueries. The slow appschema subquery involved is that generated by NestedFilterToSQL class so we need to change the use of OR clause in the root condition for UNION queries.

Since this performance issue affect postgresql, the fix should be isolated for this datasource only, and adding a system property for disable.

With UNION improvement enabled main binary operator on nested filter subquery will rebuild normal OR query like:
SELECT id, name FROM table WHERE name = "Morty" OR name = "Rick"
->
SELECT id, name FROM table WHERE name = "Morty" UNION SELECT id, name FROM table WHERE name = "Rick"

Improvement is disabled by default on backport.

### DIFF
--- a/modules/extension/app-schema/app-schema/src/main/java/org/geotools/data/complex/config/AppSchemaDataAccessConfigurator.java
+++ b/modules/extension/app-schema/app-schema/src/main/java/org/geotools/data/complex/config/AppSchemaDataAccessConfigurator.java
@@ -115,6 +115,8 @@ public class AppSchemaDataAccessConfigurator {
 
     public static String PROPERTY_ENCODE_NESTED_FILTERS = "app-schema.encodeNestedFilters";
 
+    public static final String PROPERTY_REPLACE_OR_UNION = "app-schema.orUnionReplace";
+
     /** DOCUMENT ME! */
     private AppSchemaDataAccessDTO config;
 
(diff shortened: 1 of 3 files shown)


modules/extension/app-schema/app-schema/src/main/java/org/geotools/data/complex/config/AppSchemaDataAccessConfigurator.java
```diff
@@ -149,6 +151,13 @@ public static boolean isJoiningSet() {
         return s != null;
     }
 
+    public static boolean isOrUnionReplacementEnabled() {
+        final String orUnionReplacement =
+                AppSchemaDataAccessRegistry.getAppSchemaProperties()
+                        .getProperty(PROPERTY_REPLACE_OR_UNION);
+        return "true".equalsIgnoreCase(orUnionReplacement);
```
nmco: Ok, so the union optimization is disabled by default.

nmco:
Looks like the matching GeoServer:
https://github.com/geoserver/geoserver/pull/3319 PR is this one:

This is indeed between a bug and performance fix, without queries using ``OR`` in relatively sized PostgreSQL databases are unpractical. The fix is relatively isolated and unless users explicitly activate it they will have not to deal with it.

So I say +1 to merge this.

PR #2226, opened 2019-01-11
Add ROUND_HALF_EVEN case to AngleFormat.doRounding as an AssertionError

schwehr:
A version of ErrorProne complains if all test cases are not covered, even if it is obvious to the reader that they are.  I think the idea being that if someone were to add another ROUND_*, you would want to check all the places where it might not be covered.

### DIFF
--- a/modules/library/referencing/src/main/java/org/geotools/measure/AngleFormat.java
+++ b/modules/library/referencing/src/main/java/org/geotools/measure/AngleFormat.java
@@ -386,6 +386,10 @@ public class AngleFormat extends Format {
                 case ROUND_HALF_UP:
                     rounded = Math.round(scaledValue);
                     return rounded / scale;
+
+                    // Include impossible case to allow detection of new entries in RoundingMethod.
+                case ROUND_HALF_EVEN:
+                    throw new AssertionError(rm);
             }
         }
 


aaime:
Why not, but a human might be confused by the code and remove that branch again, I suggest adding a comment explaining why it has to be there.

aaime:
Just to make sure we're not waiting on each other, this PR is not going to get merged until that comment is added ;-)

schwehr:
Working on the frmt failure

aaime:
Easiest thing is to just "mvn install" in the module, it will do the right formatting for you (or even just "mvn fmt:format" in case you really just want to reformat)

schwehr:
Yup.  `mvn fmt:format` is great!

If you think the comment should be different, please counter propose text.  Thx!

aaime:
Comment looks good, I just found the position a bit odd, would have put it a line below, this way it does not seem associated with the "case". Not a big deal, can stay as is as far as I'm concerned, at least now other devs know they should not be messing with it.

schwehr:
@aaime  Can you merge or are you waiting on me for anything?  Sorry I was so slow on adding the comment.

aaime:
Sorry I normally watch PRs only in the weekend, unless I have a job associated to them. So also slow getting back to them.

schwehr:
No worries.  Didn't mean to rush you.

PR #2237, opened 2019-01-21
[GEOT-6215] Enable JAI-EXT by default

aaime:


modules/library/render/src/test/java/org/geotools/renderer/lite/GridCoverageRendererTest.java
```diff
@@ -1259,6 +1246,8 @@ public void testCoverage_0_360() throws Exception {
 
     /** Test rendering of sampleGrib.tif on its native longitude bounds (302,308). */
     @Test
+    // @Ignore // this looks like a bug in mosaic op, there are black pixels likely due to
```
tbarsballe: Given that the @Ignore is both new, and commented out, this comment seems unnecessary/outdated?

aaime: Doh good catch, I fixed the issue and should un-ignore the test :-) Thanks!

PR #2239, opened 2019-01-27
Integrate Spotbugs QA

aaime:


jodygarnett:
Thanks Andrea I really appreciate all these safeties on the codebase.

modules/extension/validation/src/main/java/org/geotools/validation/relate/RelateIntegrity.java
```diff
@@ -154,8 +154,8 @@ private boolean validateMultipleLayers(
                 }
             }
         } finally {
-            fr1.close();
-            fr2.close();
+            if (fr1 != null) fr1.close();
+            if (fr2 != null) fr2.close();
```
jodygarnett: Every time I look at validation code I cry, looks like fr2 is not being closed in the inner loop (or is relying not iterating to the end to close).

aaime: One could question why we're still keeping it there, has been un-used and un-maintained for years now...

PR #2240, opened 2019-01-31
Set EPSG catalog version back to 8.6.0.0

schwehr:
046c0c102470f6b updated the version number, but the catalog was not changed.

unzip -p modules/plugin/epsg-hsql/src/main/resources/org/geotools/referencing/factory/epsg/hsql/EPSG.zip EPSG.properties
#EPSG database on HSQL
#Sat Jan 10 06:45:27 MST 2015
version=2.2.8
readonly=true
epsg.version=8.6.0.0
modified=no


schwehr:
I botched the commit message, so I'll overwrite the patch when I have a bit more time.

aaime:
The last number is .1 on purpose, please don't change or you're going to break upgrades.

schwehr:
@aaime I definitely don't understand.  Please explain.  It's very confusing.

PR #2245, opened 2019-02-05
DefaultCalendarEra.java equals(): Fix instanceof check before cast

schwehr:
The cast in this equals method could throw a ClassCastException if a non-comparable type is passed in, which would violate the equals contract: it should just return false.

http://errorprone.info/bugpattern/EqualsUnsafeCast


aaime:
Yep thanks

PR #2247, opened 2019-02-06
VPF update

jimgambale:
This pull request provides an updated VPF plugin that includes the required ContentDataStore implementation to function as a Geoserver data store.

Also included are revisions to enable integration of geotools with node.js. Refer to this repository for more details regarding node.js integration: https://github.com/jimgambale/geotools-java.


aaime:
@jodygarnett could you give a kick to the appveryor build? probably just a random failure

jimgambale:
Pushed a fix that will kick off the Travis and AppVeyor builds. Should be good this time.

The AppVeyor build failed with the following:

> [ERROR] Failed to execute goal org.apache.maven.plugins:maven-install-plugin:2.4:install (default-install) on project xmlcodegen: Failed to install metadata org.geotools.maven/maven-metadata.xml: Error updating group repository metadata: in epilog non whitespace content is not allowed but got s (position: END_TAG seen ...</metadata>\ns..

This appears to be a recurrence of this issue:
https://sourceforge.net/p/geotools/mailman/message/36036098/

aaime:
I did not check the updates (git won't give me a diff based on the old code) but generally speaking it looks good. I'm not sure about the directory data store changes, would like to see something pluggable instead of a ad-hoc handling for a specific store, but they don't seem harmful either and I'm about to add another round of QA (checkstyle) that might affect the merge. So merging now, and I'll resolve QA issues in the checkstyle branch.

I see you put the module back in the build, can we contact you back in case there are issues with the VPF datastore? You're probably the only one around with any understanding of what it does.

bradh:
Also, what VPF products did you test it with? Just DNCs?

aaime:
@jimgambale it looks like the build works fine on Travis but fails on one of our build server, here:

https://build.geo-solutions.it/jenkins/view/GeoTools/job/GeoTools-Master-OpenJDK8/org.geotools$gt-shapefile/1086/testReport/junit/org.geotools.data.directory/DirectoryDataStoreTest/testFactoryVPFConflict/

Seems there is some condition where the exclusion for VPF in DirectoryDataStore does not work, but I don't have much more information that that. Can you think of anything?

jimgambale:
I will investigate a better solution for the directory data store conflict.

I will be available to handle future issues with the VPF datastore.

jimgambale:
So far I have tested with DNCs obtained here:

https://dnc.nga.mil/DNC/library.php?c=Z13

I would like to test with other VPF data sources. Let me know if you have
suggestions.

jimgambale:
I'm investigating this.

jimgambale:
I resolved the test failure for
org.geotools.data.directory.DirectoryDataStoreTest.testFactoryVPFConflict.
The failure was occurring for mixed case directory paths. This explains why
the test failed on the internal Travis but not the other. The internal
Travis contained a mixed case component in the project path (
GeoTools-Master-OpenJDK8).
I created a new pull request that resolves the issue.

PR #2254, opened 2019-02-08
filter/function/color: Fix spelling: adjstRelative -> adjustRelative

schwehr:


aaime:
Method is indeed mostly used internally, don't think there is anything really user visible here, so ok not having a ticket associated, merging.

jodygarnett:
Thanks @aaime, I agree this was an internal change.

PR #2260, opened 2019-02-12
[GEOT-6237] Support for big String (byte length > 65535) on SimpleFeatureIO

fernandor777:
Writing a String attribute with a big size (String bytes greater than 65535 bytes) on SimpleFeatureIO we got an exception:
```
java.io.UTFDataFormatException: encoded string too long: 71530 bytes
```

Seems like due to DataOutputStream.writeUTF String size limit, SimpleFeatureIO doesn't allow to write/read big String sizes on Simple Features attributes.

This PR adds support for big String sizes using a flag on FeatureType Attribute Descriptor (userMap):
```java
SimpleFeatureType type = typeBuilder.buildFeatureType();
type.getDescriptor(NAME_FIELD)
                    .getUserData()
                    .put(SimpleFeatureIO.BIG_STRING, Boolean.TRUE);
```

https://osgeo-org.atlassian.net/browse/GEOT-6237

### DIFF
--- a/modules/library/main/src/main/java/org/geotools/data/sort/SimpleFeatureIO.java
+++ b/modules/library/main/src/main/java/org/geotools/data/sort/SimpleFeatureIO.java
@@ -24,6 +24,8 @@ import java.io.IOException;
 import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
 import java.io.RandomAccessFile;
+import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Date;
 import java.util.List;
 import org.geotools.feature.simple.SimpleFeatureBuilder;
@@ -41,6 +43,9 @@ import org.opengis.feature.type.AttributeDescriptor;
  * @author Andrea Aime - GeoSolutions
  */
 public class SimpleFeatureIO {
+
+    public static final int MAX_BYTES_LENGTH = 65535;
+    public static final String BIG_STRING = "bigString";
 
     RandomAccessFile raf;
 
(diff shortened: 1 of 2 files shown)


modules/library/main/src/main/java/org/geotools/data/sort/SimpleFeatureIO.java
```diff
@@ -130,6 +156,13 @@ void writeAttribute(AttributeDescriptor ad, Object value) throws IOException {
         }
     }
 
+    private boolean isBigStringAttribute(AttributeDescriptor ad) {
```
aaime: I'd rather calling "isBigString"

modules/library/main/src/main/java/org/geotools/data/sort/SimpleFeatureIO.java
```diff
@@ -268,4 +316,17 @@ public void close(boolean deleteFile) throws IOException {
     public String toString() {
         return "SimpleFeatureIO [schema=" + schema + ", file=" + file + "]";
     }
+
+    private Collection<String> split(String value, int charSize) {
+        final AtomicInteger atomicInteger = new AtomicInteger(0);
+        Collection<String> result =
```
aaime: I'm sure it works in today's JVMs, but it might break later, see:

https://stackoverflow.com/questions/37277967/in-java-8-does-a-sequential-and-ordered-stream-guarantee-performing-operations

I suggest to use a common iteration loop, which guarantees predictable execution order.

modules/library/main/src/main/java/org/geotools/data/sort/SimpleFeatureIO.java
fernandor777: Fixed name

modules/library/main/src/main/java/org/geotools/data/sort/SimpleFeatureIO.java
fernandor777: Fixed, thanks.

fernandor777:
Some fixes added.

aaime:
Looks good to me

PR #2261, opened 2019-02-12
GEOT-6238 shapefile dump with custom filename

NielsCharlier:


modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/ShapefileDumper.java
```diff
@@ -333,6 +351,16 @@ private void writeToShapefile(
      */
     protected void shapefileDumped(String fileName, SimpleFeatureType remappedSchema)
             throws IOException {
+        shapefileDumped(remappedSchema);
```
aaime: I'd add in the javadocs that this method default implementation just calls the other, should helps sub-classers decide which method to override.

NielsCharlier:
Thank you for review, changes added

NielsCharlier:
this ok to merge?

modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/ShapefileDumper.java
```diff
@@ -326,13 +344,28 @@ private void writeToShapefile(
...
+    }
+
+    /**
+     * Allows subsclasses to perform extra actions against a shapefile that was completely written.
+     *
+     * <p>This method will not be called if {@link #shapefileDumped(String, SimpleFeatureType)} is
+     * overwritten.
+     *
+     * @param fileName
```
aaime: filename is not actually a parameter in this method. Once fixed just go ahead and merge.

NielsCharlier:
thanks

modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/ShapefileDumper.java
```diff
@@ -501,11 +537,11 @@ private StoreWriter getStoreWriter(
      *     geometry type suffix
      * @return
      */
-    protected String getShapeName(SimpleFeatureType schema, String geometryType) {
+    protected String getShapeName(String fileName, String geometryType) {
```
aaime: Ah, I did miss this change in the review, and it's breaking the GeoServer build in the WFS module....

aaime: Also, GeoServer really needs a type name, not a random filename

aaime: Will wait a bit to see if by any chance you can respond quickly, and revert the merge by end of day otherwise.

PR #2262, opened 2019-02-12
[GEOT-6240] Expressing a single channel selection, out of red, green and blue, returns a gray image

aaime:


dromagnoli (approved):
Looks good to me.

PR #2266, opened 2019-02-13
zip slip

360CodeSafe:
Hello:
I am a staff member of 360 Code Guard. During the code audit of our open source project, we found that there is a zip_slip vulnerability in “geotools”. The details are as follows:
geotools-master\modules\extension\graph\src\main\java\org\geotools\graph\util\ZipUtil.java
default

At 108 lines, without checking the zip entry name, the zip entry name is directly used to construct the file name of the decompressed file, resulting in the attacker being able to use “.. /” retrospective to overwrite arbitrary files.
The repairing methods are as follows:

while(e.hasMoreElements()) {
ZipEntry zipEntry = (ZipEntry)e.nextElement();
System.out.println(zipEntry.getName());
if(zipEntry.getName().indexOf("..") != -1 && !file.getCanonicalPath().startsWith(unZipAddress))
{ System.out.println("失败！"); }

else
{ ... }

### DIFF
--- a/modules/extension/graph/src/main/java/org/geotools/graph/util/ZipUtil.java
+++ b/modules/extension/graph/src/main/java/org/geotools/graph/util/ZipUtil.java
@@ -100,6 +100,9 @@ public class ZipUtil {
 
         while (entries.hasMoreElements()) {
             ZipEntry entry = (ZipEntry) entries.nextElement();
+			if(entry.getName().indexOf("..")!=-1){
+				System.out.println("非法的zip文件！");
+			}else{
             byte[] buffer = new byte[1024];
             int len;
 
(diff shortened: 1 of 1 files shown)


aaime:
You should know better than to post a vulnerability in a public place like a PR (the Jira ticket is also public), if you don't find a documented place to discuss vulnerabilities, it would be best to ask in a public place, stating that you need to discuss an issue in private, and get a private mail or list where this can be handled.

Also remember this is a volunteer led project, so there might not be anyone with time to look a this within a given deadline.

Final note, the pull request is not matching our published contribution guidelines:
In particular:

- Won't pass the QA portion of the build because it's misformatted (flagged by google format) and contains a system.out (did not get there, but PMD would complain). Probably best to throw an exception in that case.
- Does not have a test checking it's working and won't regress.

PR #2269, opened 2019-02-15
Fixing OGR tests and bringing it back into the build

aaime:
Resumes @jericks work (I squashed a few changes of mine in GEOT-6166 to make it build on OGR 2.2) and added a commit to bring it back into the build, and fix the current QA tool failures.


aaime:
Travis is having problems fetching from github the sources for this build, trying a "close and reopen".

PR #2274, opened 2019-02-20
[GEOT-6251] Add support for ECQL 'function() IN (...)' syntax

elahrvivaz:


elahrvivaz:
I opened two PRs, as it didn't cleanly cherry-pick against both master and 20.x. Second PR is #2275

elahrvivaz:
If possible, I'd like to also back-port it to 19.x and 21.x.

modules/library/cql/src/test/java/org/geotools/filter/text/cql2/FilterToCQLTest.java
```diff
@@ -330,6 +331,12 @@ public void testAttributeEqualsBoolean() throws Exception {
         cqlTest("foo = true");
     }
 
+    @Test
+    public void testFunctionOr() throws Exception {
+        // this is a contrived example, but it's a function that's available in this modules
+        ecqlReparseTest("PropertyExists('name') = true OR PropertyExists('name') = false");
```
ianturton: This test doesn't seem to test this change!

elahrvivaz: It does, although somewhat tangentially... the `ECQL.toCQL` method re-writes it as `PropertyExists('name') IN (true, false)`. I was mainly interested in the fact that it can survive a round trip through ECQL, so I didn't test for the specific format. Should I add that?

elahrvivaz: added an explicit test

aaime: @ianturton are you satisfied with the changes?
@elahrvivaz I see a PR for 20.x, one for master, but none for 21.x? 
As a rule of thumb, we merge first on master, and then backport, so first PR should be there.

elahrvivaz: @aaime I can open additional PRs, I wasn't sure if just cherry-picking would be good enough or if I should open a PR per major version. Initially I opened one against 20.x because the commit against master didn't cleanly cherry-pick back. I'll go ahead and open the additional PRs if you don't object.

aaime: No objection, we need one for master, one for 21.x and one for 20.x... just need @ianturton opinion on it :-)

ianturton: Fine with me

elahrvivaz: I opened one for [21.x](https://github.com/geotools/geotools/pull/2307), [19.x](https://github.com/geotools/geotools/pull/2306), and the original master one is https://github.com/geotools/geotools/pull/2275. Are there any more plans to release 19.x bug fixes?

aaime: Not that I know of, the 19.x series it quite likely dead and buried.

PR #2276, opened 2019-02-22
GEOS-8975 - [Label Placement] Baseline is displaced when the label has line breaks and descenders

imranrajjad:
Labels with multiple lines are anchored using top line instead of the last line

please refer to discussion at https://osgeo-org.atlassian.net/browse/GEOS-8975


modules/library/render/src/main/java/org/geotools/renderer/label/LabelPainter.java
```diff
@@ -184,9 +184,19 @@ public LabelCacheItem getLabel() {
      * @return
      */
     public double getLineHeight() {
+        //        return lines.get(getLineCount()-1).getLineHeight();
```
simboss: remove me

modules/library/render/src/main/java/org/geotools/renderer/label/LabelPainter.java
```diff
@@ -184,9 +184,19 @@ public LabelCacheItem getLabel() {
...
         return lines.get(0).getLineHeight();
     }
 
+    /**
+     * Returns the line height for last line for this label in pixels
+     *
+     * @return
+     */
+    public double getLastLineHeight() {
```
simboss: this is prone to nullpointerexception

aaime:
Travis is also complaining about reformat (first check to be made ,there are others). I'd suggest to run a "mvn clean install -Dqa" to reformat and run all other QA checks before committing.

simboss:
Ciao @imranrajjad thanks for the improvements.

One last thing, you should address feedback from @aaime about reformatting.

modules/library/render/src/test/java/org/geotools/renderer/lite/LabelDisplacementModeTest.java
```diff
@@ -372,6 +374,65 @@ public void testDisplacementVerticalBothConflictDisabledMultiLayer() throws Exce
         ImageAssert.assertEquals(new File(refPath), image, 900);
     }
 
+    // https://osgeo-org.atlassian.net/browse/GEOS-8975
+    // https://osgeo-org.atlassian.net/projects/GEOT/issues/GEOT-6253
+    public void testGEOT6253() throws Exception {
```
aaime: The pointers to the ticket are fine, but give the test a human readable name. Same goes for the sld style file name and feature type name.

modules/library/render/src/test/java/org/geotools/renderer/lite/LabelDisplacementModeTest.java
```diff
@@ -372,6 +374,65 @@ public void testDisplacementVerticalBothConflictDisabledMultiLayer() throws Exce
...
+                                        new Coordinate(2, 5),
+                                        new Coordinate(3, 5),
+                                        new Coordinate(4, 5),
+                                        new Coordinate(5, 5),
+                                        new Coordinate(6, 5),
+                                        new Coordinate(7, 5),
+                                        new Coordinate(8, 5)
+                                    }),
+                            "Espace\nPierrey Betazg", // label with decender and line breaks
+                            // ""Campagne-lÃ¨s-Hesdin""
```
aaime: I'd remove this commented out line.

modules/library/render/src/main/java/org/geotools/renderer/label/LabelCacheImpl.java
```diff
@@ -1211,7 +1211,9 @@ private void setupPointTransform(
                 (textStyle.getAnchorY() * (textBounds.getHeight()))
                         - textStyle.getDisplacementY()
                         - textBounds.getHeight()
-                        + painter.getLineHeight();
+                        + ((painter.lines.size() == 1)
```
aaime: I believe you found the right area to fix, but the change works fine for y anchor equals to 0, which makes for top alignment. But what if bottom alignment was asked and the last line was significantly smaller than the first one instead?
Try the following, change the y anchor to 1, and the string to "Espace\nabcd ef", and the result is this (with your code changes in):

selezione_577

Before the code change it would have been instead:

image


I believe that the code should be sensitive to the anchor value, at least for the cases where a human visible alignment is evident:

- if anchorY is 1 (or very close, use a tolerance), use first line height
- if anchorY is 0  (or very close, use a tolerance), use the last line height
- for any other value, I'd use the average line height instead, just to care for the 0.5 anchor case better, the others are hardly verifiable

imranrajjad: understood

aaime:
Looks good but found out the LabelWrapTest were failing on my machine. Getting labelling tests to pass across different JKDs and operating systems is a balancing act. @imranrajjad I've made an attempt at that in https://github.com/geotools/geotools/pull/2292, which passes on my machine using Oracle JDK 8 and 11, OpenJDK 8 and 11 (they render fonts in significantly different ways). Can you have a look?
Closing this one, 2292 has the same commits squashed, plus my addition in a separate commit for clarity.

imranrajjad:
they also failed at my end, I regenerated the png files that were used for
assertion

PR #2277, opened 2019-02-24
[GEOT-6252] App-schema multiple sortBy support on SQL

fernandor777:
Executing a WFS query with nested filter and multiple sortBy expressions throws an malformed SQL exception on App-schema and Postgresql.

This PR fix SQL nested filter building on multiple sortBy expressions scenario, also adds some method delegating refactoring due to a too big method doing all.

https://osgeo-org.atlassian.net/browse/GEOT-6252


nmco:
Thanks for the refactoring of that big method, makes things clear. That say, that refactoring makes it also hard to check what changed, but since this is a bug fix I would say its fine.

The commit message shoudl be `[GEOT-6252] App-schema multiple sortBy support on SQL`.

As usual for App-Schema the tests where contributed to GeoServer, I guess this is the related PR:
https://github.com/geoserver/geoserver/pull/3377 ?

Feel free to back-port this fix has needed.

PR #2278, opened 2019-02-24
Spotbugs, up rank to 15 and lower threshold from high to default

aaime:
Going ballistic with Spobugs: up rank to 15 and lower threshold from high to default

https://lgtm.com/projects/g/geotools/geotools/context:java still scores GeoTools pretty low (grade D, meaning we're in the 25% worse scoring projects... this PR should give it a bit of a boost (hopefully at least jump to grade C, low 50%)


aaime:
Closing as Travis build is reporting line numbers that do not match the source code, someething might be happening at the git checkout level, will do another PR in a new branch

PR #2281, opened 2019-02-26
[GEOT-6254] Add parameter to 'prime' OGR data sources

aaime:


modules/unsupported/ogr/ogr-core/src/main/java/org/geotools/data/ogr/OGRDataStoreFactory.java
```diff
@@ -112,6 +112,17 @@
                     false,
                     DEFAULT_EVICTOR_TESTS_PER_RUN);
 
+    public static final boolean DEFAULT_PRIME_DATASOURCE = false;
+
+    /** Number of connections checked during a single evictor run * */
```
dromagnoli: does this comment comes from a copy and paste? :)

aaime: Doh!!!

PR #2288, opened 2019-02-28
[GEOT-6088] - Subsampling enabled irrespective of Overviews/Overview Policy

imranrajjad:
Calculation of Subsampling factor only applied to grids with overviews, now they will be calculated irrespective of overview policy or presence of overviews inside a grid coverage.

https://osgeo-org.atlassian.net/browse/GEOT-6088

### DIFF
--- a/modules/library/coverage/src/main/java/org/geotools/coverage/grid/io/AbstractGridCoverage2DReader.java
+++ b/modules/library/coverage/src/main/java/org/geotools/coverage/grid/io/AbstractGridCoverage2DReader.java
@@ -360,10 +360,21 @@ public abstract class AbstractGridCoverage2DReader implements GridCoverage2DRead
 
         // //
         //
-        // default values for subsampling
+        // Resolution requested. I am here computing the resolution required by
+        // the user.
         //
-        // //
-        readP.setSourceSubsampling(1, 1, 0, 0);
+        double[] requestedRes =
+                getResolution(
+                        requestedEnvelope,
+                        requestedDim,
+                        getCoordinateReferenceSystem(coverageName));
+
+        // /////////////////////////////////////////////////////////////////////
+        // DECIMATION ON READING
+        // /////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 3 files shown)


modules/library/coverage/src/main/java/org/geotools/coverage/grid/io/AbstractGridCoverage2DReader.java
```diff
@@ -360,10 +360,21 @@ protected Integer setReadParams(
...
+        double[] requestedRes =
+                getResolution(
+                        requestedEnvelope,
+                        requestedDim,
+                        getCoordinateReferenceSystem(coverageName));
+
+        // /////////////////////////////////////////////////////////////////////
+        // DECIMATION ON READING
+        // /////////////////////////////////////////////////////////////////////
+        decimationOnReadingControl(coverageName, imageChoice, readP, requestedRes);
```
aaime: I'd move the call to decimationOnReadingControl below the check for requested res being being null (can see tha the method is null safe, but it's a pointless extra call if requestedRes is null)

modules/library/coverage/src/test/java/org/geotools/coverage/grid/GridCoverageTest.java
```diff
@@ -72,4 +76,40 @@ public void testSerialization() throws IOException, ClassNotFoundException {
             assertRasterEquals(coverage, serial);
         }
     }
+
+    @Test
+    public void testSubSampling() throws IllegalArgumentException, IOException, TransformException {
```
aaime: I would add a couple more tests:

- Overviews are supported and enabled, check the right one gets picked
- Overviews are supported and enabled, but the last overview is still much bigger than the requested image, allowing for both overview usage and subsampling on top of it, and checking the subsampling factor is "right" for the overview, and not computed on the base image.

aaime:
Manually merged on master after fixing the formatting issue:

image

PR #2290, opened 2019-03-04
Fix a few threadsafety issues in getBounds

pvalsecc:
Discussion about this in the mailing list:
https://sourceforge.net/p/geotools/mailman/message/36586913/

Also fixed a few errors in the javadoc.


aaime:
Right, it was a case of making a partially constructed object available to other threads. I like that the fix mostly removes code too :-D
Is there a ticket for this? If not please create one and rename the commit message to refer to it, e.g. "[GEOT-xxxx] title"

PR #2291, opened 2019-03-05
[GEOT-6210] - Check for Zip Slip Vulnerability

imranrajjad:
added a generic method in `org.geotools.util.Utilities` to verify zip file names for Zip Slip vulnerability as reported at https://osgeo-org.atlassian.net/browse/GEOT-6210. Code that is inline with suggestion at https://lgtm.com/rules/1506728586782/ was not modified

### DIFF
--- a/modules/library/metadata/src/main/java/org/geotools/util/Utilities.java
+++ b/modules/library/metadata/src/main/java/org/geotools/util/Utilities.java
@@ -16,6 +16,8 @@
  */
 package org.geotools.util;
 
+import java.io.File;
+import java.io.IOException;
 import java.io.ObjectStreamException;
 import java.io.Serializable;
 import java.util.AbstractQueue;
@@ -91,6 +93,9 @@ public final class Utilities {
 
     /** The singleton instance to be returned by {@link #emptyQueue}. */
     private static final Queue<?> EMPTY_QUEUE = new EmptyQueue<Object>();
+
+    /** expression to detect Zip Slip vulnerability */
+    private static final String ZIP_SLIP_EXPRESSION = String.format("..%s", File.separator);
 
     /**
      * The class for the {@link #EMPTY_QUEUE} instance. Defined as a named class rather than
@@ -754,5 +759,21 @@ public final class Utilities {
                 element -> element.getClass().getName(),
                 Function.identity(),
                 (first, second) -> second);
+    }
+
+    /**
+     * Asserting if the passed string is not prefixed with '..' to allow an attacker to traverse
+     * outside the referenced directory
+     *
+     * @param fileName to checked for zip sip vulnerability
+     * @throws IOException if passed fileName is found to be vulnerable to zlip slip, e.g fileName
+     *     should not be prefixed with '..'
+     */
+    public static void assertNotZipSlipVulnerable(String fileName) throws IOException {
+        File destinationfile = new File(fileName);
+        String canonicalDestinationFile = destinationfile.getCanonicalPath();
+        if (canonicalDestinationFile.startsWith(ZIP_SLIP_EXPRESSION)) {
+            throw new IOException("Vulnerable File Name");
+        }
     }
 }
--- a/modules/library/xml/src/main/java/org/geotools/xml/SchemaFactory.java
+++ b/modules/library/xml/src/main/java/org/geotools/xml/SchemaFactory.java
@@ -133,7 +133,11 @@ public class SchemaFactory {
                             new BufferedReader(new InputStreamReader(res.openStream(), "UTF-8"));
 
                     while (rd.ready()) {
-                        String factoryClassName = rd.readLine().trim();
+
+                        String factoryClassName = rd.readLine();
+                        // null check as per NP_IMMEDIATE_DEREFERENCE_OF_READLINE
+                        if (factoryClassName == null) break;
+                        factoryClassName = factoryClassName.trim();
 
                         try {
                             Schema s =
--- a/modules/unsupported/gtopo30/src/main/java/org/geotools/gce/gtopo30/GTopo30Writer.java
+++ b/modules/unsupported/gtopo30/src/main/java/org/geotools/gce/gtopo30/GTopo30Writer.java
@@ -71,6 +71,7 @@ import org.geotools.referencing.operation.matrix.XAffineTransform;
 import org.geotools.referencing.operation.transform.LinearTransform1D;
 import org.geotools.util.NumberRange;
 import org.geotools.util.URLs;
+import org.geotools.util.Utilities;
 import org.geotools.util.factory.Hints;
 import org.opengis.coverage.grid.Format;
 import org.opengis.coverage.grid.GridCoverage;
@@ -502,6 +503,8 @@ public final class GTopo30Writer extends AbstractGridCoverageWriter implements G
 
         if (dest instanceof File) {
 
+            Utilities.assertNotZipSlipVulnerable(name);
+
             final PrintWriter out =
                     new PrintWriter(
                             new BufferedOutputStream(
@@ -571,6 +574,8 @@ public final class GTopo30Writer extends AbstractGridCoverageWriter implements G
             out.flush();
             out.close();
         } else {
+            Utilities.assertNotZipSlipVulnerable(gc.getName().toString());
+
             final ZipOutputStream outZ = (ZipOutputStream) dest;
             final ZipEntry e = new ZipEntry(gc.getName().toString() + ".HDR");
             outZ.putNextEntry(e);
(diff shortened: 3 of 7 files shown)


modules/library/metadata/src/main/java/org/geotools/util/Utilities.java
```diff
@@ -92,6 +94,9 @@
     /** The singleton instance to be returned by {@link #emptyQueue}. */
     private static final Queue<?> EMPTY_QUEUE = new EmptyQueue<Object>();
 
+    /** OS Agnostic expression to detect Zip Slip vulnerability */
+    private static final String ZipSlipExpression = String.format("..%s", File.separator);
```
aaime: ZIP_SLIP_EXPRESSION (constants naming conventions)

modules/library/metadata/src/main/java/org/geotools/util/Utilities.java
```diff
@@ -755,4 +760,12 @@ public static void ensureArgumentNonNull(final String name, final Object object)
                 Function.identity(),
                 (first, second) -> second);
     }
+
+    /**
+     * @param fileName
```
aaime: If at all possible, write a description of what any new public and protected  method and class does, the params documentation is of secondary importance (nice to have) but there is not point in having them there, and empty (``@param fileName``) though CheckStyle won't complain (yet). Other tools do, for example lgtm.com

modules/library/metadata/src/main/java/org/geotools/util/Utilities.java
```diff
@@ -755,4 +760,12 @@ public static void ensureArgumentNonNull(final String name, final Object object)
...
+     * @return
+     */
+    public static void isZipSlipVulnerable(String fileName) throws IOException {
```
aaime: Naming is a bit off... this method name makes it sound like something that would return a boolean, instead it throws an exception. Maybe "assertNotZipSlipVulnerable" or "checkZipSlip" instead?

modules/unsupported/gtopo30/src/main/java/org/geotools/gce/gtopo30/GTopo30Writer.java
```diff
@@ -676,10 +682,12 @@ private void writeSRC(GridCoverage2D gc, final String name, Object dest)
 
         RenderedImage image = gc.getRenderedImage();
         if (dest instanceof File) {
+
```
aaime: Still changing/adding newlines and the like it seems? Avoid.

aaime:
Looking good. As a rule of thumb, we want to merge PRs that have a single commit (one ticket, one commit). Can you rebase and squash the commits into one commit, whose commit message is "[GEOT-6210] - Check for Zip Slip Vulnerability"? (and then force push to replace the contents of the PR)

modules/library/metadata/src/main/java/org/geotools/util/Utilities.java
```diff
@@ -755,4 +760,16 @@ public static void ensureArgumentNonNull(final String name, final Object object)
...
+    /**
+     * Asserting if the passed string is not prefixed with '..' to allow an attacker to traverse
+     * outside the referenced directory
+     *
+     * @param fileName to checked for zip sip vulnerability
+     * @throws IOException if passed fileName is found to be vulnerable to zlip slip, e.g fileName
+     *     should not be prefixed with '..'
+     */
+    public static void assertNotZipSlipVulnerable(String fileName) throws IOException {
+        if (fileName.startsWith(ZIP_SLIP_EXPRESSION)) throw new IOException("Vulnerable File Name");
```
bradh: Is this really a sufficient check? What happens if I pass `./../../something.important` ?

imranrajjad: so it needs to look for any ".." inside the file name, right?

aaime: I'd say one needs a "..<fileSeparator>" to trigger it, anywhere in the file name... or not?

mprins: this check is insufficient; the filename should be canonicalized and then the check for starting with ..%s should be done.
It requires creating a File object and then checking the path.

```java
 File destinationfile = new File(fileName);
 String canonicalDestinationFile = destinationfile.getCanonicalPath();
 if (canonicalDestinationFile.startsWith(ZIP_SLIP_EXPRESSION)) {
   throw new IOException("Vulnerable File Name");
 }
```

In reality the fileName should be compared to the directory where the zipfile should be expanded and not just for starting with `..%s`, see the code at: https://snyk.io/research/zip-slip-vulnerability

imranrajjad: I think @aaime has proposed a simple check, ".." means reverse traversal.

aaime: The approach suggested by @mprins looks good to me (check the page he linked and the code he pasted)

imranrajjad: > this check is insufficient; the filename should be canonicalized and then the check for starting with ..%s should be done.
> It requires creating a File object and then checking the path.
>  File destinationfile = new File(fileName);
>  String canonicalDestinationFile = destinationfile.getCanonicalPath();
>  if (canonicalDestinationFile.startsWith(ZIP_SLIP_EXPRESSION)) {
>    throw new IOException("Vulnerable File Name");
>  }
> In reality the fileName should be compared to the directory where the zipfile should be expanded and not just for starting with ..%s, see the code at: https://snyk.io/research/zip-slip-vulnerability

this looks simple and great, I will update accordingly. Thanks

aaime: @mprins can you check again?

modules/library/xml/src/main/java/org/geotools/xml/SchemaFactory.java
```diff
@@ -133,7 +133,11 @@ private Map loadSchemas() {
                             new BufferedReader(new InputStreamReader(res.openStream(), "UTF-8"));
 
                     while (rd.ready()) {
-                        String factoryClassName = rd.readLine().trim();
+
+                        String factoryClassName = rd.readLine();
+                        // null check as per NP_IMMEDIATE_DEREFERENCE_OF_READLINE
+                        if (factoryClassName == null) break;
+                        factoryClassName = factoryClassName.trim();
```
mprins: these changes seem unrelated to GEOT-6210

modules/library/metadata/src/main/java/org/geotools/util/Utilities.java
```diff
@@ -755,4 +760,20 @@ public static void ensureArgumentNonNull(final String name, final Object object)
                 Function.identity(),
                 (first, second) -> second);
     }
+
+    /**
+     * Asserting if the passed string is not prefixed with '..' to allow an attacker to traverse
```
mprins: this should be "Asserting if the passed filename is not prefixed...."

mprins:
I don't see much point in checking when creating zip files - though I don't see any harm ATM; 
The actual vulnerability that we need to address is the inadvertent overwriting of an existing files by unzipping a (possibly crafted) zipfile, eg. zipfile has `../../../../etc/group` which could overwrite the actual `/etc/group` when certain other conditions are met thus providing the possibility to put a trojan in place.

Having said that I still feel that this fix does not fully address the issue because it does not compare the canonical path to the path that the file should (is expected to) be extracted to (`outdir` param of `unzip()` method). It is quite legal (maybe uncommon IRL) to have relative paths in zipfiles even if these are pointing to a parent directory.

eg. to create such a zipfile in `/tmp` having 2 files from `/bin` is quite easy:
```ksh
$ cd /tmp
$ zip foo ../../../bin/ed /bin/[  
  adding: ../../../bin/ed (deflated 53%)
  adding: bin/[ (deflated 54%)
$ unzip -vl foo.zip                                                             
Archive:  foo.zip
 Length   Method    Size  Cmpr    Date    Time   CRC-32   Name
--------  ------  ------- ---- ---------- ----- --------  ----
  222992  Defl:N   103974  53% 10-11-2018 21:18 2edf422c  ../../../bin/ed
  124600  Defl:N    56989  54% 10-11-2018 21:18 41d0bc28  bin/[
```

modules/library/xml/src/main/java/org/geotools/xml/SchemaFactory.java
aaime: They are, but best go in anyways. @imranrajjad had a Spotbugs report on that line which would break the QA build. It is bizzarre that it happens on his machine and not on mine, but we're using different OS, slightly different JDK versions, and there is a chance the eclipse compiler is involved too... In any case, the report is legit, readLine output should be checked for nulls and was not, so is the fix. Best to have the QA build work on as many platforms as possible.

aaime:
@mprins do you want to start over and do a brand new pull request?

bradh:
> I don't see much point in checking when creating zip files - though I don't see any harm ATM;
The only case I can see is potential information disclosure attack (e.g. when using this in geoserver, you might be able to export arbitrary files from the server system).

mprins:
I can/will prepare a PR this week

PR #2293, opened 2019-03-06
read GeoTiff to Array

SenalChen:
Hello, how can I read a tiff to Array like a matrix use ur API. I read your document but it is hard to find the right answer.Pls help me, thank you!


aaime:
Please don't open PRs to ask questions, use subscribe and participate in user list instead:
http://www.geotools.org/getinvolved.html

PR #2294, opened 2019-03-06
GEOS-9130 support xlink in csw

NielsCharlier:
https://osgeo-org.atlassian.net/browse/GEOS-9130


NielsCharlier:
Thanks for your quick response despite not being active any more @bencaradocdavies !
so sorry to bother you with this still, I only decided to try my luck because it is such a tiny change, and you are someone who knows complex features/app-schema really well 
Indeed, geoserver builds fine.
I think your opinion to approve this is still perfectly valid

PR #2296, opened 2019-03-07
[GEOT-6260] Move Globe2.png to the correct package to prevent NPE when creating WMSInfo

mprins:
This seems an oversight in the package/module reorganisation, `Globe2.png` must be in the same package as `WebMapServer`

relevant code block: 
https://github.com/geotools/geotools/blob/18b6d116d64daa2bef93bf197a4442a99035d563/modules/extension/wms/src/main/java/org/geotools/ows/wms/WebMapServer.java#L97-L111

close [GEOT-6260](https://osgeo-org.atlassian.net/browse/GEOT-6260)


aaime:
Looks good to me, merging

PR #2308, opened 2019-03-09
Fix s3-geotiff external configuration

groldan:
It could be that there's no defaultCache configured,
or that the configuration is invalid. Either way, the
net result would be that the CacheManagement class can't
be loaded by the JVM, which wouldn't preclude the
CoverageStore to be advertised as available, and the
real cause of failure to configure it lost.
    
This patch takes care of falling back to defaultCache if
there's an error in parsing the configuration file passed
as a System property (with a prominent log entry), and
creating a defaultConfig that wont fail if the file was
parsed but doesn't define a defaultCache.


dvntucker:
LGTM

groldan:
@dvntucker ok thanks, merging then

PR #2309, opened 2019-03-09
Deprecation removal

aaime:


aaime:
Work here is almost done, I'm moving to GeoWebCache and GeoServer, applying the same approach and fixing the eventual compile errors due to deprecated methods removal.

One thing, there are still a lot of classes and methods deprecated around, which were not touched because the compiler was not catching them: this happens when nothing at all uses them in the GeoTools code base. We might want to get rid of all residual deprecated classes and methods, unless they have been deprecated since the 21.x cut  (e.g., need to keep at least one deprecation cycle)

aaime:
Btw, please **do not merge** yet, doing so will break the GeoServer build!

modules/library/main/src/main/java/org/geotools/feature/FeatureTypes.java
```diff
@@ -427,6 +433,34 @@ public static SimpleFeatureType newFeatureType(AttributeDescriptor[] types, Stri
...
     }
 
+    /**
+     * Whether the feature type has the specified name, or is a descendent from it
+     *
+     * @param featureType typeName with parentage in question
+     * @param name name to match against
+     * @return true if featureType has the same name, or is a descendent of the indicated name
+     */
+    public static boolean matches(FeatureType featureType, Name name) {
```
jodygarnett: Is this new functionality? Or was it hiding elsewhere...

docs/user/welcome/upgrade.rst
```diff
@@ -26,6 +26,62 @@ But first to upgrade - change your dependency to |release| (or an appropriate st
...
 
+GeoTools 22.x
+-------------
+
+Removal of Deprecated API
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This release features an extensive clean up of any deprecated methods and classes. If you find a class or method unavailable after upgrading use the list below and the online GeoTools 21.x API reference to locate it's replacement.
+
+* DefaultQuery removed, use new Query().
```
bradh: Maybe format the new Query (i.e. `new Query()`)

docs/user/welcome/upgrade.rst
```diff
@@ -26,6 +26,62 @@ But first to upgrade - change your dependency to |release| (or an appropriate st
...
+
+* DefaultFeatureLock removed, use new FeatureLock().
```
bradh: Also here: `new FeatureLock()`

docs/user/welcome/upgrade.rst
```diff
@@ -26,6 +26,62 @@ But first to upgrade - change your dependency to |release| (or an appropriate st
...
+     
+  These lists are modified directly using add() and remove() methods.
+
+* Style API title and abstract information is now contained in a descriptor.
+  
+  .. code-block:: java
+     
+     String description = Optional.ofNullable(style.getDescription().getAbstract())
+         .map(Object::toString)
+         .orElese(null);
```
bradh: typo: `orElse(null)`

modules/library/main/src/main/java/org/geotools/filter/CompareFilterImpl.java
```diff
@@ -153,33 +154,23 @@ protected int compare(Comparable leftObj, Comparable rightObj) {
...
-        if (filterType == NULL) {
+        if (this instanceof IsNullImpl) {
             return "[ " + expression1 + " IS NULL ]";
+        } else if (this instanceof IsNilImpl) {
+            return "[ " + expression1 + " IS NIL ]";
         }
         String operator = null;
-
-        if (filterType == COMPARE_EQUALS) {
+        if (this instanceof IsEqualsToImpl) {
```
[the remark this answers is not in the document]
aaime: Feel free to do some improvements post merge, don't think anyone would complain?

modules/library/main/src/main/java/org/geotools/data/DiffFeatureWriter.java
```diff
@@ -124,7 +124,7 @@ public void remove() throws IOException {
             // mark live as removed
             diff.remove(live.getID());
             fireNotification(
-                    FeatureEvent.FEATURES_REMOVED, ReferencedEnvelope.reference(live.getBounds()));
+                    FeatureEvent.Type.REMOVED.type, ReferencedEnvelope.reference(live.getBounds()));
```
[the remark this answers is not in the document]
aaime: I tried, failed.

modules/library/main/src/main/java/org/geotools/feature/FeatureTypes.java
aaime: It was hidden in some other place, I pushed it up to a re-usable method.

mprins:
While this removed AbstractDataStoreFactory it is still listed in the documentation: https://github.com/geotools/geotools/blob/master/docs/user/library/main/internal.rst

PR #2317, opened 2019-03-13
Quote highest-basedir property to fix build in a path with spaces

bencaradocdavies:
`geotoolsBaseDir` can contain spaces and so should be quoted. Without quotes, building in a path with spaces fails because surefire JVM invocation fails.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1786,7 +1786,7 @@
             <exclude>${stress.skip.pattern}</exclude>
             <exclude>${test.exclude.pattern}</exclude>
           </excludes>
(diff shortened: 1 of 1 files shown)


bencaradocdavies:
Oh, crap. Now I have double quoting hell. I will walk back the change to fix only the surefire argline and close the geowebcache and geoserver pull requests. No good deed goes unpunished.

bencaradocdavies:
This one needs backport to 21.x but not 20.x.

aaime:
Thanks!

PR #2324, opened 2019-03-19
[GEOT-6269] Allow  MapViewports to retain scale or bounds on resize

ianturton:
As discussed in [GEOT-6269](https://osgeo-org.atlassian.net/projects/GEOT/issues/GEOT-6269) this enhancement allows a resizing MapPane to keep it's bounds rather than its bounds. The behaviour defaults to the current practice of maintaining the scale and adjusting the bounds.

A call to `setFixedBoundsOnResize` with True will maintain the bounds and adjust the scale.


modules/library/render/src/test/java/org/geotools/map/MapViewportTest.java
```diff
@@ -49,7 +49,8 @@
     private static final Rectangle SCREEN_2_1 = new Rectangle(200, 100);
     // Screen area with aspect ratio 1:2
     private static final Rectangle SCREEN_1_2 = new Rectangle(100, 200);
-
+    // Screen area with aspect ration 2:2 (or 1:1)
```
aaime: :-D

PR #2329, opened 2019-03-20
Upgrading to ErrorProne 2.3.3

aaime:
Did not try to build, failure expected, just want to see where it gets to :-D


bradh:
Not as error prone as you expected!

PR #2332, opened 2019-03-21
Improved JUnit assertions on floats

lislei:
There are some deprecated useage of  junit.framework.*

Fixed all cases of `junit.framework.Assert.assertEquals(float, float)`

See https://junit.org/junit4/javadoc/latest/org/junit/Assert.html#assertArrayEquals(float[],%20float[],%20float)


aaime:
Thanks!

PR #2335, opened 2019-03-22
ClassCastException when running Oracle database tests

lislei:
Some tests are still failing on asserts - that's how we want the (real) errors to show.

### DIFF
--- a/modules/library/jdbc/src/test/java/org/geotools/jdbc/JDBCPrimaryKeyFinderOnlineTest.java
+++ b/modules/library/jdbc/src/test/java/org/geotools/jdbc/JDBCPrimaryKeyFinderOnlineTest.java
@@ -52,8 +52,8 @@ public abstract class JDBCPrimaryKeyFinderOnlineTest extends JDBCTestSupport {
     }
 
     public void testAssignedSinglePKeyView() throws Exception {
-        JDBCFeatureStore fs =
-                (JDBCFeatureStore) dataStore.getFeatureSource(tname("assignedsinglepk"));
+        JDBCFeatureSource fs =
+                (JDBCFeatureSource) dataStore.getFeatureSource(tname("assignedsinglepk"));
 
         assertEquals(1, fs.getPrimaryKey().getColumns().size());
         assertTrue(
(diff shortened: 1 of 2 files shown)


modules/library/jdbc/src/test/java/org/geotools/jdbc/JDBCPrimaryKeyFinderOnlineTest.java
```diff
@@ -39,6 +40,7 @@ protected void connect() throws Exception {
         }
     }
 
+    @Test
```
aaime: I don't think this is correct, the test in question extends from Junit3 TestCase (all OnlineTest do, there is a ton in the coebase):

image

Migrating them to Junit 4 would be nice, at the same time, a ton of work :-D
If you feel like migrating some, it would certainly be appreciated.

aaime: May also be causing the build failues, all builds on Travis and Appveyor are complaining.

lislei: Sorry I missed that. I am certain I've seen that combination in use on some other tests.
Don't have much experience in migrating to JUnit 4, but I do have some knowledge on the good old JUnit3 :)
If I get some extra time to spend I could look into it. 

I get some test errors running the jdbc tests against Oracle. Are those tests part of Travis and if so, what version of Oracle DB?

aaime: With the exclusion of H2 tests, Travis does not normally run database tests. If the H2 tests fail (I believe they do),  it's likely a generalized condition.

lislei:
Yikes, I'm not sure how to proceed. 
Running the original code throws me a class cast exception for the oracle test. 
Changing it sets the H2 tests off with a different class cast exception. 

Seems like the H2 setup returns a `JDBCFeatureStore` instance 
while the oracle setup returns  a `JDBCFeatureSource`.
See `JDBCPrimaryKeyFinderOnlineTest.testAssignedSinglePKeyView` and the statement 
```
//H2PrimaryKeyFinderTest:
      JDBCFeatureStore fs =
              (JDBCFeatureStore) dataStore.getFeatureSource(tname("assignedsinglepk"));

//OraclePrimaryKeyFinderOnlineTest:
      JDBCFeatureSource fs =
              (JDBCFeatureSource) dataStore.getFeatureSource(tname("assignedsinglepk"));
```

aaime:
The test is supposed to return a FeatureStore, if it does not, there might be some issues in your setup, like for example permissions problems. In JDBCTestSupport there is a section of code that you can comment out, it should enable more detailed logging, maybe that will help you figure out what is not working?

aaime:
Closing at it seems related to improper test setup.

PR #2338, opened 2019-03-29
[GEOT-5830] Fids can become corrupted in ShapeFileReader when deletion occurs

ianturton:
See [GEOT-5830](https://osgeo-org.atlassian.net/projects/GEOT/issues/GEOT-5830) for an explanation of the issue.

### DIFF
--- a/modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/ShapefileDataStore.java
+++ b/modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/ShapefileDataStore.java
@@ -133,13 +133,13 @@ public class ShapefileDataStore extends ContentDataStore implements FileDataStor
     }
 
     Name getTypeName() {
-      String typeName="Null";
-      if(shpFiles!=null) {
-        typeName = shpFiles.getTypeName();
-      } else {
-        typeName = "Null";
-      }
-      return new NameImpl(namespaceURI, typeName);
+        String typeName = "Null";
+        if (shpFiles != null) {
+            typeName = shpFiles.getTypeName();
+        } else {
+            typeName = "Null";
+        }
+        return new NameImpl(namespaceURI, typeName);
     }
 
     @Override
--- a/modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/ShapefileDataStore.java
+++ b/modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/ShapefileDataStore.java
@@ -133,7 +133,13 @@ public class ShapefileDataStore extends ContentDataStore implements FileDataStor
     }
 
     Name getTypeName() {
-        return new NameImpl(namespaceURI, shpFiles.getTypeName());
+      String typeName="Null";
+      if(shpFiles!=null) {
+        typeName = shpFiles.getTypeName();
+      } else {
+        typeName = "Null";
+      }
+      return new NameImpl(namespaceURI, typeName);
     }
 
     @Override
--- a/modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/ShapefileDataStore.java
+++ b/modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/ShapefileDataStore.java
@@ -136,8 +136,8 @@ public class ShapefileDataStore extends ContentDataStore implements FileDataStor
         String typeName = "Null";
         if (shpFiles != null) {
             typeName = shpFiles.getTypeName();
-        } else {
-            typeName = "Null";
+            /*} else {
+            typeName = "Null";*/
         }
         return new NameImpl(namespaceURI, typeName);
     }
--- a/modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/ShapefileDataStore.java
+++ b/modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/ShapefileDataStore.java
@@ -136,8 +136,6 @@ public class ShapefileDataStore extends ContentDataStore implements FileDataStor
         String typeName = "Null";
         if (shpFiles != null) {
             typeName = shpFiles.getTypeName();
-            /*} else {
-            typeName = "Null";*/
         }
         return new NameImpl(namespaceURI, typeName);
     }
--- a/modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/ShapefileFeatureSource.java
+++ b/modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/ShapefileFeatureSource.java
@@ -363,7 +363,7 @@ class ShapefileFeatureSource extends ContentFeatureSource {
         // get the .fix file reader, if we have a .fix file
         IndexedFidReader fidReader = null;
         if (getDataStore().isFidIndexed()
-                && filter instanceof Id
+                /* && filter instanceof Id*/
                 && indexManager.hasFidIndex(false)) {
             fidReader = new IndexedFidReader(shpFiles);
         }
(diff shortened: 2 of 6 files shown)


HendrikPeilke:
Hi Ian,

I wrote a message regarding this issue in december 2018: https://sourceforge.net/p/geotools/mailman/message/36489901/

I would rather vote for dropping the FID support for Shapefiles (that is, what we did after trying to get up with a fix, and also the people issuing the bug report mentioned) for the following reasons:
- FIDs can only be held constant in a shapefile over time at huge cost (when lot of deletes and adds are issued), that is probably why the current code has some defragmentation inside
- the current code is broken since GeoTools 10, there can't be too many projects out there really relying on this
- if FID stability is wanted while issuing additions and deletions, the user can choose a format (e.g. database) supporting this

A clear documentation that shapefiles will not support the FID stability is in my opinion needed, if the feature is dropped. Maybe a survey, which projects would suffer from the dropped support, could help on that decision?

Hendrik

aaime:
Dropping stable fids support is currently not an option, despite suggestion
of the contrary, many use shapefile for wfs-t where stable fids are a must.
I'm sorry I have difficulties to follow up on this one, overworked by
having to maintain way too many modules in both geotools and geoserver,
I'll try to prioritize looking at this one but it may still take a week or
two (travelling this week)

Il lun 1 apr 2019, 11:05 Hendrik Peilke <notifications@github.com> ha
scritto:

ianturton:
I think that to maintain FID stability overtime we need to make sure all requests go through the IndexedFIDReader/Writer and not allow the shortcut being used if there is no fid filter in the query. I'm not sure when I'll next get a chance to look at this so feel free to jump in if this is important to you.

aaime:
Hi Ian, I'm travelling for work so don't have time to check the code, but believe you're on the right track. The indexed path should however be followed only if there is a fix file already, to avoid creating one, as it adds overhead for the simple read only usage case (basically trying to get the best of both worlds, stable fids for those that need to edit, fast execution for those that only read the shapefile without touching it).

ianturton:
That sounds like a good plan. I'll try to have a look at the code this
weekend.

Ian

ianturton:
Actually, with Andrea's hint this boils down to a one line change. I think I will leave in the checks for updating the fix file if it is stale, though I'm not sure if they make an difference. 

GeoServer builds cleanly against this change.

modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/fid/IndexedFidWriter.java
```diff
@@ -206,11 +206,12 @@ private void finishLastWrite() throws IOException {
      * Increments the fidIndex by 1. Indicates that a feature was removed from the location. This is
      * intended to ensure that FIDs stay constant over time. Consider the following case of 5
      * features. feature 1 has fid typename.0 feature 2 has fid typename.1 feature 3 has fid
-     * typename.2 feature 4 has fid typename.3 feature 5 has fid typename.4 when feature 3 is
-     * removed/deleted the following usage of the write should take place: next(); (move to feature
-     * 1) next(); (move to feature 2) next(); (move to feature 3) remove();(delete feature 3)
-     * next(); (move to feature 4) // optional write(); (write feature 4) next(); (move to feature
-     * 5) write(); (write(feature 5)
+     * typename.2 feature 4 has fid typename.3 feature 5 has fid typename.4
```
aaime: This is indeed hard to follow and partly outdated. I can see it updating the "removed" counter, not fidIndex. 
Probably a victim of mass reformat, do you have time to turn it into a bullet list, ``<ul><li>...</li></ul>``, structure?
If not, no problem, it's not the objective of the fix.

modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/ShapefileDataStore.java
```diff
@@ -133,7 +133,13 @@ public ShapefileDataStore(URL url) {
     }
 
     Name getTypeName() {
-        return new NameImpl(namespaceURI, shpFiles.getTypeName());
+        String typeName = "Null";
```
aaime: How did this happen?

modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/ShapefileFeatureSource.java
```diff
@@ -359,7 +363,7 @@ protected int getCountInternal(Query query) throws IOException {
         // get the .fix file reader, if we have a .fix file
         IndexedFidReader fidReader = null;
         if (getDataStore().isFidIndexed()
-                && filter instanceof Id
+                /* && filter instanceof Id*/
```
aaime: Let's not leave commented out code sections around.

modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/ShapefileFeatureStore.java
```diff
@@ -70,6 +71,9 @@ public ShapefileFeatureStore(ContentEntry entry, ShpFiles files) {
                     new IndexedShapefileFeatureWriter(
                             ds.indexManager, reader, ds.getCharset(), ds.getTimeZone());
         } else {
+            if (ds.indexManager.isIndexStale(ShpFileType.FIX)) {
```
aaime: This would create the fix file if missing, but if it's missing we have no fid remapping, so why waste time to create a fid index in the first place? At the very least, I'd check if the file exists (maybe isIndexStyle needs a flag to check if the file exists before stating it's outdate).

modules/plugin/shapefile/src/test/java/org/geotools/data/shapefile/fid/FidQueryTest.java
```diff
@@ -182,6 +185,57 @@ public void testDeleteFeature() throws Exception {
...
+        featureStore = (SimpleFeatureStore) ds.getFeatureSource();
+        assertEquals((size - 1), featureStore.getCount(Query.ALL));
+
+        SimpleFeatureCollection features2 = featureStore.getFeatures(fidFilter);
+        assertEquals("wrong number of features", 0, features2.size());
+        try (SimpleFeatureIterator features = features2.features()) {
+            assertFalse("found fid", features.hasNext());
+        }
+
+        // when we fetch the features again the first feature is archsites.1 again!
```
aaime: Is this comment is outdated? archsites.1 does not show up anymore after the fix, or does it?

modules/plugin/shapefile/src/test/java/org/geotools/data/shapefile/fid/FidQueryTest.java
```diff
@@ -182,6 +185,57 @@ public void testDeleteFeature() throws Exception {
...
+        // yet the IndexedShapeReader tries very hard to avoid this!
+        try (SimpleFeatureIterator features = allfeatures.features()) {
+
+            SimpleFeature f = (SimpleFeature) features.next();
+            assertFalse(fidFilter.evaluate(f));
+        }
+        // but not if we use the fid filter
```
aaime: Same here, comment looks like it would have to be updated

modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/ShapefileDataStore.java
ianturton: Not sure, possibly while debugging an earlier effort, reverted.

ianturton: Sometimes IndexManager.hasFidIndex() tries to get a lock using a store (in the IndexManager) where shpFiles is null. 

It seems to be if you ask for a FeatureIterator from a FeatureCollection that points to a DataStore that doesn't exist anymore.  (See line 222-223 in FidQueryTest).

aaime: Roger that

PR #2344, opened 2019-04-15
Added authentication parameter to MS SQL Server JDBC Connection String

mattbsea:
This code change brings the MS JDBC driver version up to 7.2.1.  

It adds an "authentication" property to the JDBC Connection string.  Valid options for the property are ActiveDirectoryIntegrated, ActiveDirectoryPassword, ActiveDirectoryMSI, SqlPassword, and the default NotSpecified.

### DIFF
--- a/modules/plugin/jdbc/jdbc-sqlserver/README.txt
+++ b/modules/plugin/jdbc/jdbc-sqlserver/README.txt
@@ -4,33 +4,18 @@ GeoTools SQL Server DataStore README
 JDBC Driver
 -----------
 
-To use the sql server datastore the jdbc 3.0 driver must be downloaded from 
-Microsoft:
+The current tested driver version is 7.2.1 for SQL Server and SQL Azure. According to Microsoft this
+driver provides access to SQL Azure, SQL Server 2017 from any Java application, application server, 
+or Java-enabled applet.
 
-  http://www.microsoft.com/download/en/details.aspx?displaylang=en&id=19847
-
-The current tested driver version is 3.0 for SQL Server and SQL Azure. According to Microsoft this
-driver provides access to SQL Azure, SQL Server 2008 R2, SQL Server 2008, SQL Server 2005 and 
-SQL Server 2000 from any Java application, application server, or Java-enabled applet.
-
-Make sure to unpack the sqljdbc4.jar file, as that's the one providing support for the Java
-version in this series.
+This plugin will automatically download the MS SQL Server JDBC driver from the central maven repository.
 
 Maven
 -----
 
 If developing with the datastore and the rest of GeoTools: 
 
-  1. install the JDBC driver into the location maven repository:
-
-     mvn install:install-file  -Dfile=<path to sqljdbc4.jar>
-         -DartifactId=sqljdbc4 \ 
-         -DgroupId=com.microsoft \ 
-         -Dversion=3.0 \
-         -Dpackaging=jar \ 
-         -DgeneratePom=true
-
-  2. Build the jdbc-sqlserver module using the *sqlServerDriver* profile:
+  1. Build the jdbc-sqlserver module using the *sqlServerDriver* profile:
 
      mvn clean install -P sqlServerDriver
   
--- a/modules/plugin/jdbc/jdbc-sqlserver/pom.xml
+++ b/modules/plugin/jdbc/jdbc-sqlserver/pom.xml
@@ -90,15 +90,25 @@
 			<dependencies>
 				<dependency>
 					<groupId>com.microsoft.sqlserver</groupId>
-					<artifactId>sqljdbc41</artifactId>
+					<artifactId>mssql-jdbc</artifactId>
 					<version>${sqlServerDriverVersion}</version>
+				</dependency>
+				<dependency>
+					<groupId>com.microsoft.azure</groupId>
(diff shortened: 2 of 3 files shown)


mprins:
I'm curious as to why one would need the `com.microsoft.azure:adal4j` and `com.microsoft.rest:client-runtime` as required dependencies because AFAIK it is still just a JDBC connection with plain SQL and those dependencies are amongst the optional dependencies for the driver (but then there are also the [azure keyvault](https://github.com/Microsoft/mssql-jdbc/blob/9f78a13a27271b3780dcdf86f44997ac47dd3b18/pom.xml#L66-L90) dependencies with an optional scope that are not part of this PR).  
The example project doesn't list those deps either: https://docs.microsoft.com/en-us/azure/sql-database/sql-database-connect-query-java

mattbsea:
adal4j is required for Active Directory authentication.  In my case, I need to use `authentication=ActiveDirectoryPassword`.  The mssql-jdbc driver will attempt to load `adal4j ` to handle the active directory interface which in turn loads the `client-runtime`

The mssql-jdbc project on GitHub describes the dependencies under "Azure Key Vault and Azure Active Directory Authentication Dependencies"

https://github.com/Microsoft/mssql-jdbc

Now, I'm not 100% clear if I need to include them in GeoTools or if I only need them in GeoServer where the library gets called.  But I'm guessing it need to be in GeoTools where the dependency originates.

modules/plugin/jdbc/jdbc-sqlserver/pom.xml
```diff
@@ -90,15 +90,25 @@
 			<dependencies>
 				<dependency>
 					<groupId>com.microsoft.sqlserver</groupId>
-					<artifactId>sqljdbc41</artifactId>
+					<artifactId>mssql-jdbc</artifactId>
 					<version>${sqlServerDriverVersion}</version>
 				</dependency>
+				<dependency>
```
aaime: If these are only needed for particular cases, but not always, and are not found in repos, but have to be manually downloaded and installed, I'd mark them as optional and provide info on how to get them installed on a dev machine,
as indicated here: https://github.com/geotools/geotools/blob/master/modules/plugin/jdbc/jdbc-sqlserver/README.txt

modules/plugin/jdbc/jdbc-sqlserver/README.txt
```diff
@@ -4,33 +4,18 @@ GeoTools SQL Server DataStore README
...
 
-  http://www.microsoft.com/download/en/details.aspx?displaylang=en&id=19847
-
-The current tested driver version is 3.0 for SQL Server and SQL Azure. According to Microsoft this
-driver provides access to SQL Azure, SQL Server 2008 R2, SQL Server 2008, SQL Server 2005 and 
-SQL Server 2000 from any Java application, application server, or Java-enabled applet.
-
-Make sure to unpack the sqljdbc4.jar file, as that's the one providing support for the Java
-version in this series.
+This plugin will automatically download the MS SQL Server JDBC driver from the central maven repository.
```
aaime: So wait, now everything is on the mave repos?

aaime: Well duh, looks like it is: https://mvnrepository.com/artifact/com.microsoft.sqlserver/mssql-jdbc

modules/plugin/jdbc/jdbc-sqlserver/pom.xml
aaime: Never mind, it seems that Microsoft has finally seen the light and started adding stuff in the repos without click-through licenses.

mprins:
> Now, I'm not 100% clear if I need to include them in GeoTools or if I only need them in GeoServer where the library gets called. But I'm guessing it need to be in GeoTools where the dependency originates.

I'd say these would need to be loaded by the final product (geoserver or anything else building on geotools) because it's the final product that is making the specifc connection not the library.  I would opt to keep these as `<scope>optional</scope>` in geotools and `runtime/compile` (ie. no scope specified) in the geoserver extension and thus follow the microsoft recommendation: [Azure Key Vault and Azure Active Directory Authentication Dependencies](https://github.com/Microsoft/mssql-jdbc#azure-key-vault-and-azure-active-directory-authentication-dependencies)

aaime:
@mprins I agree on marking them optional in GeoTools, I don't think any test needs them anyways.
About GeoServer, unsure, users tend to get lost pretty easily with stuff to download and such... ideally, since the drivers are now open, it would be great to just pack everything in the zip, with instructions to remove the uneeded bits, for the (I imagine) few that would care doing that.

mprins:
Yes, having a runtime scope for the geoserver extension should pack them in the zip so users have an easy install.

modules/plugin/jdbc/jdbc-sqlserver/README.txt
jodygarnett: Now if only oracle could play nice :P

tbarsballe:
@mattbsea - Have you signed a [code contribution license](http://docs.geotools.org/latest/developer/procedures/contribution_license.html) with OSGeo? This is required before this PR can be merged - thanks.

aaime:
Getting no feedback from the original author, thus closing. But very happy to reopen if @mattbsea restarts activity on it.

PR #2350, opened 2019-04-25
GEOT-6282 SLD 1.1 doesn't support relative images

NielsCharlier:


modules/extension/xsd/xsd-sld/src/test/java/org/geotools/se/v1_1/SEExampleTest.java
```diff
@@ -221,6 +221,43 @@ public void testParsePointSymbolizer4() throws Exception {
...
+        assertEquals("Example Pointsymbolizer", sym.getDescription().getTitle().toString());
+        assertEquals(
+                "This is just a simple example of a point symbolizer.",
+                sym.getDescription().getAbstract().toString());
+
+        Graphic g = sym.getGraphic();
+        assertEquals(15.0, g.getSize().evaluate(null, Double.class));
+        assertEquals(1, g.getExternalGraphics().length);
+
+        ExternalGraphic eg = g.getExternalGraphics()[0];
```
aaime: This is a call to a deprecated API, please avoid (the build will fail when any deprecated API call is used, after the deprecation branch is merged).

modules/extension/xsd/xsd-sld/src/test/java/org/geotools/se/v1_1/SEExampleTest.java
```diff
@@ -221,6 +221,43 @@ public void testParsePointSymbolizer4() throws Exception {
         assertNull(em.getOnlineResource());
     }
 
+    public void testParsePointSymbolizerRelativeURL() throws Exception {
+        /*<PointSymbolizer version="1.1.0" xsi:schemaLocation="http://www.opengis.net/se http://www.opengis.net/se/1.1.0/Symbolizer.xsd" xmlns="http://www.opengis.net/se" xmlns:ogc="http://www.opengis.net/ogc" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" uom="http://www.opengeospatial.org/se/units/pixel">
```
aaime: This example does not reflect the contents of SLD actually parsed, it's confusing. Either update or remove.

modules/extension/xsd/xsd-sld/src/test/java/org/geotools/se/v1_1/SEExampleTest.java
```diff
@@ -221,6 +221,43 @@ public void testParsePointSymbolizer4() throws Exception {
...
+        assertEquals(
+                "This is just a simple example of a point symbolizer.",
+                sym.getDescription().getAbstract().toString());
+
+        Graphic g = sym.getGraphic();
+        assertEquals(15.0, g.getSize().evaluate(null, Double.class));
+        assertEquals(1, g.getExternalGraphics().length);
+
+        ExternalGraphic eg = g.getExternalGraphics()[0];
+        assertEquals("inlineContent-image.png", eg.getURI().toString());
```
aaime: I was expecting the DefaultResolver to create an absolute path for this image (e..g, ``/path/to/geotols/modules/extension/xsd/xsd-sld/src/test/resources/org/geotools/se/v1_1/inlineContent-image.png``), why didn't it happen? 
Ah, probably because the resource locator registered in the config is not given a sourceURL I suppose... however this way the test is pretty much just testing that the code does not error, but no actual resolution happens.... I suggest to get the resolver, register the directory containing the SLD as the sourceUrl, and check proper resolution is actually happening).

NielsCharlier:
Thanks for the review. Changes added as requested.

aaime:
Looks good to me. Did you try out a downstream GeoServer build?

PR #2348, opened 2019-04-24
FIX GEOT-6284 : IllegalArgumentException on DescribeFeatureTypeResponse

eropartz:
We have an IllegalArgumentException on a DescribeFeatureTypeResponse when type name size is less than 3 :

java.lang.IllegalArgumentException: Prefix string too short
at java.io.File.createTempFile(Unknown Source)
at java.io.File.createTempFile(Unknown Source)
at org.geotools.data.wfs.internal.DescribeFeatureTypeResponse.<init>(DescribeFeatureTypeResponse.java:53)
at org.geotools.data.wfs.internal.parsers.DescribeFeatureTypeResponseFactory.createResponse(DescribeFeatureTypeResponseFactory.java:70)
at org.geotools.data.wfs.internal.WFSRequest.createResponse(WFSRequest.java:215)
at org.geotools.data.wfs.internal.WFSRequest.createResponse(WFSRequest.java:36)
at org.geotools.data.ows.AbstractOpenWebService.internalIssueRequest(AbstractOpenWebService.java:441)
at org.geotools.data.wfs.internal.WFSClient.internalIssueRequest(WFSClient.java:313)
at org.geotools.data.wfs.internal.WFSClient.issueRequest(WFSClient.java:374)
at org.geotools.data.wfs.WFSDataStore.getRemoteFeatureType(WFSDataStore.java:206)
at org.geotools.data.wfs.WFSDataStore.getRemoteSimpleFeatureType(WFSDataStore.java:262)
at org.geotools.data.wfs.WFSFeatureSource.buildFeatureType(WFSFeatureSource.java:364)
at org.geotools.data.wfs.WFSFeatureStore.buildFeatureType(WFSFeatureStore.java:106)
at org.geotools.data.store.ContentFeatureSource.getAbsoluteSchema(ContentFeatureSource.java:327)
at org.geotools.data.store.ContentFeatureSource.getSchema(ContentFeatureSource.java:296)
at org.geotools.data.store.ContentDataStore.getSchema(ContentDataStore.java:291)

Extract from javadoc for function createTempFile" : "The prefix argument must be at least three characters long"

PR #2351, opened 2019-04-26
[GEOT-6284] make sure tempfile prefix is longer than 3 chars

ianturton:
No test as I can't find any evidence we have ever tested DescribeFeatureTypeResponse in GeoTools, GeoServer still works after this change.


eropartz:
Hi,

Thanks for the correction of the issue 6284. I Have already created a PR : #2348 

Regards

ianturton:
If only you'd have taken ownership of the bug, and added a link to the PR in the issue, I would have seen it.

PR #2360, opened 2019-04-30
math filter function test cases for null value added

chosenegbe:
math filter function test cases for null added


chosenegbe:
math filter function test cases for null added

modules/library/main/src/main/java/org/geotools/filter/function/math/FilterFunction_IEEEremainder.java
```diff
@@ -40,25 +42,29 @@ public FilterFunction_IEEEremainder() {
...
-        try { // attempt to get value and perform conversion
-            arg0 = getExpression(0).evaluate(feature, Double.class).doubleValue();
-        } catch (Exception e) {
+        Object arg0 = getExpression(0).evaluate(feature);
+        Object arg1 = getExpression(1).evaluate(feature);
+        if (arg0 == null || arg1 == null) {
+            return null;
+        }
+        Hints hints0 = new Hints();
+        arg0 =
```
aaime: No need to pass down hints as a third argument, use the two argument method instead

aaime:
Looks good. For next time, try to make a pull request that has a single commit whose message is in the form "[GEOT-XYZW] Title of the ticket", it helps tracking information across git and jira.

For anyone else reading @chosenegbe is doing an internship at GeoSolutions, he's thus covered by the company's CLA

aaime:
(Did a squash/merge to fix the commits/message issue)

aaime:
@chosenegbe can you do backport pull requests onto 21.x and 20.x?

PR #2364, opened 2019-05-02
Fix mapping of MultiLineString in DataUtilities

groldan:
DataUtilities.createType(String name, String spec)
was mapping full qualified MultiPoint class names
to MultiLineString instead of to MultiPoint.

This is a forward port of #2200 that was not applied to master at its time


aaime:
Merged thanks. I don't see a Jira ticket mentioned, if missing could you create one, for the release notes?
Commit messages should be in the form "[GEOT-XYZW] Ticket title", helps tracking longer description/discussion in the tracker.

groldan:
Done, sorry will keep that in mind.

Fixes https://osgeo-org.atlassian.net/browse/GEOT-6296 on master

PR #2365, opened 2019-05-02
Fix mapping of MultiLineString in DataUtilities (21.x)

groldan:
DataUtilities.createType(String name, String spec)
was mapping full qualified MultiPoint class names
to MultiLineString instead of to MultiPoint.

This is a forward port of #2200 that was not applied to master at its time


groldan:
Fixes https://osgeo-org.atlassian.net/browse/GEOT-6296 on 21.x

PR #2366, opened 2019-05-02
swing: find resources when started from an OSGI bundle (GEOT-6290)

samthiriot:
see JIRA GEOT-6290: https://osgeo-org.atlassian.net/browse/GEOT-6290

The idea is to accept to load resources from bundle:// urls, which occur when geotools are started from an osgi-packaged application.

Uses the standard methods to solved this situation. 
Uses reflection is order not to add any mandatoru dependency to eclipse nor osgi frameworks. 

Tested under Linux and Windows; is anyway not platform dependent. I don't plan to submit any test case, as building unit tests in an osgi setting would be extremely difficult.


ianturton:
looks like you need to run mvn spotbugs:gui in the swing module to find the bug it's spotting

ianturton:
please mark the ticket with this PR and close if you don't plan to backport later

PR #2367, opened 2019-05-03
GEOT-5857: StreamingRenderer MergeLayerRequest does not merge if imag…

mbarto:
…e is null, sometimes this is not correct (e.g. DstIn composite)


modules/library/render/src/main/java/org/geotools/renderer/lite/StreamingRenderer.java
```diff
@@ -3612,18 +3612,23 @@ void execute() {
...
-                        graphics.drawImage(image, 0, 0, null);
-                        ftsGraphics.dispose();
+                    if (image == null) {
+                        Rectangle size = ((DelayedBackbufferGraphic) ftsGraphics).screenSize;
+                        image =
+                                new BufferedImage(
+                                        size.width, size.height, BufferedImage.TYPE_4BYTE_ABGR);
+                    }
+
+                    if (currentLayer.composite == null) {
```
aaime: I'd avoid the buffered image allocation and render in this case, since it's the most common one.

chosenegbe:
https://github.com/geotools/geotools/pull/2397 ---Backport to 21.x
https://github.com/geotools/geotools/pull/2398 ---Backport to 20.x

PR #2371, opened 2019-05-06
GEOT-6289: Property flag to exclude external files lookup

dromagnoli:


modules/library/coverage/src/main/java/org/geotools/coverage/grid/io/imageio/MaskOverviewProvider.java
```diff
@@ -127,45 +128,48 @@ public MaskOverviewProvider(DatasetLayout layout, URL inputFile, SpiHelper spiPr
...
-                    overviewStreamSpi.createInputStreamInstance(
-                            ovrURL, ImageIO.getUseCache(), ImageIO.getCacheDirectory());
-            if (ovrStream == null) {
-                // No Overview file so we fall back to the original file spis
-                overviewStreamSpi = streamSpi;
-                overviewReaderSpi = readerSpi;
-            } else {
-                overviewReaderSpi = getReaderSpiFromStream(null, ovrStream);
-                hasExternalOverviews = true;
+        if (!ImageIOUtilities.isSkipExternalFilesLookup()) {
```
dromagnoli: Note that this external IF resulted into a different indentation so you will notice several modified lines although, beside this check, the only difference is the indentation.

modules/plugin/geotiff/src/test/java/org/geotools/gce/geotiff/GeoTiffSkipExternalTest.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    @AfterClass
+    public static void teardown() throws ReflectiveOperationException {
+        setStaticField(ImageIOUtilities.class, "SKIP_EXTERNAL_FILES_LOOKUP", false);
+    }
+
+    /** Test that the reader didn't access the external prj */
+    @Test
+    public void prjIsSkippedTest() throws IllegalArgumentException, IOException, FactoryException {
```
mbarto: Are there already analogous tests that verify stuff is read when the flag is false?

dromagnoli: These 3 tests have been copied from standard GeoTiffReaderTests which actually uses external PRJs, external masks and external overviews. They run with that flag not being set (which means false as default).
On these tests, different result occurs (PRJ, Masks and Overviews are found).

mbarto: Ok, it looks good to me, then

PR #2375, opened 2019-05-07
[GEOT-6275-20.x] create default value for com.vividsolutions.jts.geom.Geometry

chosenegbe:


aaime:
Should start on master and with just the commit intended to fir GEOT-6272

PR #2377, opened 2019-05-08
[GEOT-6275] default value for com.vividsolutions.jts.geom.Geometry

chosenegbe:


ianturton:
I would prefer an empty geometry as the default rather than a point at null island. 

Also please don't use .* imports

aaime:
The suggestion to use a point was my "fault". The defaults might end up being used by things like the legend generator, an empty collection would result in nothing being rendered.

PR #2384, opened 2019-05-10
Improved entity validation in external DTDs.

sikeoka:
This pull request changes how the baseURI is used to determine whether or not to allow an entity defined in an external DTD.

### DIFF
--- a/modules/library/metadata/src/main/java/org/geotools/util/PreventLocalEntityResolver.java
+++ b/modules/library/metadata/src/main/java/org/geotools/util/PreventLocalEntityResolver.java
@@ -18,8 +18,10 @@ package org.geotools.util;
 
 import java.io.IOException;
 import java.io.Serializable;
+import java.net.URI;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.regex.Pattern;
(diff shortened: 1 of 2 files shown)


aaime:
The change seems good, but building GeoServer I see a lot of test failures whenever XML validation is used (e.g., in the WCS module, but also CSW, think WPS too).
OGC schemas refer to each other via relative paths, maybe that could cause issues?

sikeoka:
The original fix relied on a URI method that didn't work the way I needed for JAR URIs (which is what the GeoServer unit tests use).  I updated the PR so that JAR URIs should now be working correctly.

aaime:
Tested it out locally, works as expected, merging. Can you do backport PRs?

sikeoka:
I created a PR for the 21.x backport.  I'm not sure about the 20.x backport since the packages got renamed between 20.x and 21.x.

aaime:
Yes, the backport to 20.x might be more than a simple cherry-pick indeed.

sikeoka:
I created a PR for the 20.x backport.

aaime:
Forgot to say, for next time, try to have a single commit whose message points to a jira ticket, in the form "[GEOT-XYZW] title of the ticket"

PR #2386, opened 2019-05-13
[GEOT-6272]- FeatureType without default geometry

chosenegbe:
https://osgeo-org.atlassian.net/browse/GEOT-6272


aaime:
@chosenegbe looks good, merging. Can you please do backports to 21.x and 20.x?

PR #2387, opened 2019-05-13
Revert "GEOT-6238 shapefile dump with custom filename"

aaime:
Reverts geotools/geotools#2261 as it breaks the GeoServer build. Sorry about that, did not notice during review.

Ah, @NielsCharlier , in case you're interested or can share information, I've also started a thread on the deve list to see if we can add a build check for this (e.g., to run a geoserver build off the geotools changes and get a firm result before merging a GeoTools PR)


NielsCharlier:
My bad, I missed that this method was API, I thought of it as an internal convenience method, but the protected status and javadoc made it clear so I should have checked the effects on geoserver.

I think that would be a fantastic idea in any case.

In the meantime I've cooked up a fix for the reverted PR
https://github.com/geotools/geotools/pull/2388

PR #2388, opened 2019-05-13
GEOT-6238 shapefile dump with custom filename (second take)

NielsCharlier:


modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/ShapefileDumper.java
```diff
@@ -326,13 +344,27 @@ private void writeToShapefile(
...
+
+    /**
+     * Allows subsclasses to perform extra actions against a shapefile that was completely written.
+     *
+     * <p>This method will not be called if {@link #shapefileDumped(String, SimpleFeatureType)} is
+     * overwritten.
+     *
+     * @param remappedSchema
+     */
+    protected void shapefileDumped(SimpleFeatureType remappedSchema) throws IOException {
```
aaime: The PR certainly got more convoluted but does not break backwards compatibility.
Just a question about this method, why is it here? The original method is the only one still being called, and this one does not add anything, actually, it removed one parameter that subclasses could just ignore when using the pre-existing method? Makes sense? What did I miss?

Besides this, the rest looks good. Did you try out a GeoServer build?

NielsCharlier: I'm confused myself, I thought this was the original method and I was preserving it for backwards compatibility. I will remove it again. Yes, this time geoserver builds

aaime: Then after removing the method go ahead and merge

PR #2397, opened 2019-05-20
Backport [GEOT-5857] to 21.x StreamingRenderer MergeLayerRequest does not merge if imag…

chosenegbe:
...e is null, sometimes this is not correct (e.g. DstIn composite)


mbarto:
Looks good to me

PR #2403, opened 2019-05-21
[GEOT-6305] FootprintBehavior attaches a ROI to blank images when no input is found, even if the footprint behavior is NONE

aaime:


aaime:
The travis failure is bizzare, seems to be related to me having closed and reopened the PR (to try and see if I could kick circleci into action)

dromagnoli:
@aaime looks good to me

PR #2404, opened 2019-05-24
geot6300: Upgrade jackson-databind to 2.9.9

bradh:
Also upgrades dependency checking tools.

### DIFF
--- a/modules/plugin/coverage-multidim/grib/pom.xml
+++ b/modules/plugin/coverage-multidim/grib/pom.xml
@@ -19,6 +19,60 @@
       <groupId>edu.ucar</groupId>
       <artifactId>cdm</artifactId>
       <version>${netcdf.version}</version>
+      <exclusions>
+        <!-- Overrides - check on upgrade -->
+        <exclusion>
+          <groupId>com.fasterxml.jackson.core</groupId>
+          <artifactId>jackson-core</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>com.fasterxml.jackson.core</groupId>
+          <artifactId>jackson-annotations</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>com.fasterxml.jackson.core</groupId>
+          <artifactId>jackson-databind</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>org.quartz-scheduler</groupId>
+          <artifactId>quartz</artifactId>
+        </exclusion>
+        <exclusion>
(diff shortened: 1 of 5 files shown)


modules/plugin/coverage-multidim/grib/pom.xml
```diff
@@ -19,6 +19,60 @@
       <groupId>edu.ucar</groupId>
       <artifactId>cdm</artifactId>
       <version>${netcdf.version}</version>
+      <exclusions>
+        <!-- Overrides - check on upgrade -->
+        <exclusion>
+          <groupId>com.fasterxml.jackson.core</groupId>
```
aaime: Good catch.
About the jackson dependencies, if I try @chosenegbe pull request and do a dependency:tree I get the following:

```
1667 [INFO] |  +- com.amazonaws:aws-java-sdk-s3:jar:1.11.236:compile
1667 [INFO] |  |  +- com.amazonaws:aws-java-sdk-kms:jar:1.11.236:compile
1667 [INFO] |  |  +- com.amazonaws:aws-java-sdk-core:jar:1.11.236:compile
1667 [INFO] |  |  |  +- software.amazon.ion:ion-java:jar:1.0.2:compile
1667 [INFO] |  |  |  \- com.fasterxml.jackson.dataformat:jackson-dataformat-cbor:jar:2.6.7:compile
1667 [INFO] |  |  \- com.amazonaws:jmespath-java:jar:1.11.236:compile
1668 [INFO] |  +- com.fasterxml.jackson.core:jackson-core:jar:2.9.9:compile
1668 [INFO] |  +- com.fasterxml.jackson.core:jackson-annotations:jar:2.7.9:compile
1668 [INFO] |  +- com.fasterxml.jackson.core:jackson-databind:jar:2.7.9.1:compile
```

jackson-core is already taken care of by the dependency management section at the top, I'd rather do the same for databind and annotations instead of adding explicit exclusions and overrides in various modules.

About the quartz scheduled and c3p0, I was not expecting to see upgrades there, can you explain why they are in here? (eventually on private mail, if necessary)

pom.xml
```diff
@@ -225,7 +226,7 @@
           <plugin>
             <groupId>org.owasp</groupId>
             <artifactId>dependency-check-maven</artifactId>
-            <version>3.2.1</version>
+            <version>4.0.2</version>
```
aaime: This seems unrelated, as well as the exclusions updates, maybe best if it goes in a different PR?

PR #2414, opened 2019-05-30
[GEOT-6314] Fix swapped "has" and "not has" conditions

jopi79:
https://osgeo-org.atlassian.net/browse/GEOT-6314


jodygarnett:
This is a small enough change that we can accept it

PR #2416, opened 2019-05-30
[GEOT-6310] Update MapServerOnlineTest.java

mewagner:
I have added a testWFSFilter function that runs a query to test whether the filter is passed on to the external wfs datastore that is being referenced.


aaime:
So this PR introduces a failing test to show a problem, but does not contain a fix for it, right?

mewagner:
That is correct.

ianturton:
See https://github.com/geotools/geotools/pull/2430 for further work

PR #2419, opened 2019-06-03
[GEOT-6313] Support raster-to-vector rendering transforms across the dateline

tbarsballe:
Fix for https://osgeo-org.atlassian.net/browse/GEOT-6313

The "CoverageCrop" in RenderingTransformationHelper was cropping out portions of the coverage that would show up on the other side of the dateline - this change makes it so the crop only happens if the CRS does not wrap the world or the rendering area is entirely inside the coverage bounds.

One edge case this does not handle is a raster to vector transform with a style that could cause features outside the map area to still be drawn inside it (e.g. very large points). I was not able to find a way to  handle that while still cropping the coverage, and not cropping the coverage where possible  during rendering transforms has a significant performance impact.


sikeoka:
I am interested in this fix so I tested this locally using ras:Contour and ras:RasterAsPointCollection and it does seem to fix this problem of rendering across the dateline.

PR #2422, opened 2019-06-04
GEOT-6312 no alpha channel in JPEG

NielsCharlier:


dromagnoli:
@NielsCharlier  
Looks good to me. +1 on merging it

NielsCharlier:
Thanks @dromagnoli

dromagnoli:
@NielsCharlier: Are you also going to backport that?

NielsCharlier:
I can do that, if you recommend it

dromagnoli:
Yes, please. 
I think they have very similar code so the same problem can occur on 21.x too.

PR #2426, opened 2019-06-06
[GEOT-6316] Fix MongoDB store isNull filter processing.

fernandor777:
This PR fixes IsNull filter processing on MongoDB datastore, executing null attribute (or nonexistent attribute) on database search engine using the Null equality filter.

See:
https://docs.mongodb.com/manual/tutorial/query-for-null-fields/

Issue:
https://osgeo-org.atlassian.net/browse/GEOT-6316

### DIFF
--- a/modules/plugin/mongodb/src/main/java/org/geotools/data/mongodb/FilterToMongo.java
+++ b/modules/plugin/mongodb/src/main/java/org/geotools/data/mongodb/FilterToMongo.java
@@ -372,14 +372,12 @@ public class FilterToMongo implements FilterVisitor, ExpressionVisitor {
         return output;
     }
 
-    // There is no "NULL" in MongoDB, but I assume that TODO add null support
-    // the non-existence of a column is the same...
     @Override
     public Object visit(PropertyIsNull filter, Object extraData) {
         BasicDBObject output = asDBObject(extraData);
-
-        String prop = convert(filter.accept(this, null), String.class);
-        output.put(prop, new BasicDBObject("$exists", false));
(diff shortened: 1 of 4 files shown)


modules/plugin/mongodb/src/main/java/org/geotools/data/mongodb/FilterToMongo.java
```diff
@@ -372,14 +372,12 @@ public Object visit(PropertyIsLike filter, Object extraData) {
...
     @Override
     public Object visit(PropertyIsNull filter, Object extraData) {
         BasicDBObject output = asDBObject(extraData);
-
-        String prop = convert(filter.accept(this, null), String.class);
-        output.put(prop, new BasicDBObject("$exists", false));
+        String prop = convert(filter.getExpression().evaluate(null), String.class);
+        // mongodb filter { item: null } supports both: null value and nonexistent attribute
+        output.put(prop, null);
         return output;
```
nmco: So ``PropertyIsNull`` shoudl yield ``TRUE`` if the value of the property is ``NULL``. Although, in the case of MongoDB, which is a schema less store, since the property may not exists at all, this is interpreted as the property doesn't exists or is ``ǸULL``, looks correct to me :smiley:.

MongoDB documentation link:
https://docs.mongodb.com/manual/tutorial/query-for-null-fields/#equality-filter

nmco:
Since this a pure fix on a supported plugin, it should ideally be backported, @fernandor777 do you want to give a try tot he new backport mechanism?

nmco:
@fernandor777 I added the labels, let's see how it goes:)

PR #2437, opened 2019-06-11
[GEOT-6318] Allow setting entity expansion limit on GeoTools XML parser.

fernandor777:
Currently GeoTools XML parser (org.geotools.xsd.Parser) doesn't allow to set a custom entity expansion limit on the embedded SAX Parser. Since the default can be too big for some use cases (64000) it could be useful to can configure the limit using the Geotools Parser API, so it can be easily used, as example, by Geoserver WFS and more.

See Entity Expansion limit on documentation:
https://docs.oracle.com/javase/tutorial/jaxp/limits/using.html

This PR adds a entityExpansionLimit attribute to Geotools XML Parser.

Issue:
https://osgeo-org.atlassian.net/browse/GEOT-6318

### DIFF
--- a/modules/extension/xsd/xsd-core/src/main/java/org/geotools/xsd/Parser.java
+++ b/modules/extension/xsd/xsd-core/src/main/java/org/geotools/xsd/Parser.java
@@ -23,6 +23,7 @@ import java.io.InputStream;
 import java.io.Reader;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Optional;
 import javax.xml.namespace.QName;
 import javax.xml.parsers.ParserConfigurationException;
 import javax.xml.parsers.SAXParser;
@@ -65,11 +66,19 @@ public class Parser {
 
     private static final String SAX_PROPERTY_PREFIX = "http://xml.org/sax/properties/";
 
+    private static final String JAXP_PROPERTY_PREFIX = "http://www.oracle.com/xml/jaxp/properties/";
+    private static final String JDK_ENTITY_EXPANSION_LIMIT =
+            JAXP_PROPERTY_PREFIX + "entityExpansionLimit";
+    private static final Integer DEFAULT_ENTITY_EXPANSION_LIMIT = 100;
+
     /** sax handler which maintains the element stack */
     private ParserHandler handler;
 
     /** the sax parser driving the handler */
     private SAXParser parser;
+
+    /** Entity expansion limit configuration, set to null by default */
+    private Integer entityExpansionLimit;
 
     /**
      * Creates a new instance of the parser.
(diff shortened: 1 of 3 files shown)


modules/extension/xsd/xsd-core/src/main/java/org/geotools/xsd/Parser.java
```diff
@@ -65,12 +66,22 @@
 
     private static final String SAX_PROPERTY_PREFIX = "http://xml.org/sax/properties/";
 
+    private static final String JAXP_PROPERTY_PREFIX = "http://www.oracle.com/xml/jaxp/properties/";
```
aaime: Would it make sense to have a default setup that won't make the system go OOM easily?

modules/extension/xsd/xsd-core/src/test/java/org/geotools/xsd/ParserTest.java
```diff
@@ -343,4 +345,48 @@ public InputSource resolveEntity(
         parser.validate(
                 MLSchemaLocationResolver.class.getResourceAsStream("mails-external-entities.xml"));
     }
+
+    /** Tests entity expansion limit configuration on Parser. */
+    @Test
+    public void testEntityExpansionLimit() throws Exception {
```
aaime: Maybe also have a test showing that an expansion below the set limits works

modules/extension/xsd/xsd-core/src/main/java/org/geotools/xsd/Parser.java
fernandor777: Indeed, adding 100 as default value, it is a good default?

modules/extension/xsd/xsd-core/src/test/java/org/geotools/xsd/ParserTest.java
fernandor777: Indeed makes sense, added.

aaime:
@fernandor777 looks like there is a legit PMD failure in the build:


```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-pmd-plugin:3.11.0:check (default) on project gt-xsd-core: You have 1 PMD violation. For more details see: /home/travis/build/geotools/geotools/modules/extension/xsd/xsd-core/target/pmd.xml -> [Help 1]
```

modules/extension/xsd/xsd-core/src/main/java/org/geotools/xsd/Parser.java
aaime: Seems reasonable yes.

fernandor777:
Right, fixed the PMD issue. Thanks.

PR #2440, opened 2019-06-12
GEOT-5859: Remove httpclient dependency.

bradh:
https://www.unidata.ucar.edu/software/thredds/current/netcdf-java/reference/httpservices.html indicated this was replaced with httpcomponent 4.x. inside the UCAR libraries as of 4.3 

The dependency:tree results are consistent with that.


aaime:
Added the labels for the bot to run automatic backports (but did not check if we have the same UCAR library versions in all branches, that might be a factor)

dromagnoli:
I manually did the 20.x backport
https://github.com/geotools/geotools/pull/2466

PR #2441, opened 2019-06-12
GEOT-5859: ArcGIS REST conversion to httpcomponents 4.x

bradh:
@lmorandini This is part of a larger upgrade (where we're trying to move on from commons-httpclient). I have it converted over, but the tests are so mock'd that it really needs something online. Can you take a look when you have a chance, and do whatever testing is needed?

### DIFF
--- a/modules/unsupported/arcgis-rest/pom.xml
+++ b/modules/unsupported/arcgis-rest/pom.xml
@@ -73,12 +73,12 @@
       <version>1.2.1</version>
     </dependency>
     <dependency>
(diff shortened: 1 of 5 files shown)


lmorandini:
Hi @bradh, I will into it shortly.

lmorandini:
Hi @bradh ,  I am afraid it does not work, as the code sections needed to make requests work in ArcGISRestDataStore class have been deleted (conspicuously, it now sends out only GET requests). I will take it from here and amend this PR or create a new one.

PR #2443, opened 2019-06-12
GEOT-6321: Remove broken app-schema webservice module

bradh:
If/when this is brought back, it'll also need at least upgrade to httpcomponent 4.x


aaime:
Ops, did not add the backport labels.... did you want to backport this?

bradh:
Backporting won't make any real difference - we aren't building / releasing it anyway. So lets not.

PR #2448, opened 2019-06-14
GEOT-6319 S3-geotiff: time-out properties

NielsCharlier:
Current the s3-geotiff cache never expires. I want to make it possible to configure an expiry time.


aaime:
Not sure if you want to backport or not, up to you.
Can you mark the ticket as resolved if you don't?

NielsCharlier:
That's fine, thanks

PR #2449, opened 2019-06-18
return bugs

D12126977:
getMaxOccurs method should return maxOccurs, however it returns minOccurs


D12126977:
in the file modules\library\xml\src\main\java\org\geotools\xml\gml\ChoiceAttributeTypeImpl.java,  the method named getMaxOccurs returns minOccurs, however, in fact it should return maxOccurs. I fixed this problem.

ianturton:
can you add a test to demonstrate the fix please, and make a Jira issue and link it to this issue - http://docs.geotools.org/latest/developer/procedures/contribute.html

aaime:
Closing, superceded by https://github.com/geotools/geotools/pull/2596 (added tests)

PR #2430, opened 2019-06-09
[GEOT-6310] wfsng max features and filter interaction

ianturton:
This supersedes https://github.com/geotools/geotools/pull/2416

PR #2453, opened 2019-06-21
[GEOT-6310] wfsng max features and filter interaction

ianturton:
Fixes rebase issue with #2430


ianturton:
Yes, I've made this change for all strategies, as far as I can tell it only happens if you can not pass the filter to the client. For V2.0.0 this seems to be all filters(?). 

ContentDataStore wraps the result in a LimitingFeatureReader and handles everything locally.

PR #2454, opened 2019-06-24
GeoTIFF sources configured with GeoServer 2.14.x might not work in 2.15.x

chosenegbe:


chosenegbe:
test for this is contained in the GeoServer PR -- https://github.com/geoserver/geoserver/pull/3588

aaime:
Test can and should be written also in geotools, the commit should refer to a geootols ticket (please create one and link it with the geoserver one)

PR #2458, opened 2019-06-25
[GEOT-6327] tolerate geotiff_coverage as the coverageName 

chosenegbe:


chosenegbe:
> is the setGridCoverageName method actually needed in class and test?
> I think that the relevant part of the fix is that it will return true if the coveragename is "geotiff_coverage" so the latest line of test already proves that, right?

Resolved

PR #2459, opened 2019-06-28
[GEOT-6330] XS DateTime encoder doesn't support Local DateTime Handling

fernandor777:
This PR adds Local DateTime Handling system property support for XS DateTime data type, in the same way it is already supported on XS Date.

Issue:
https://osgeo-org.atlassian.net/browse/GEOT-6330

### DIFF
--- a/modules/extension/xsd/xsd-core/src/main/java/org/geotools/xs/XSUtils.java
+++ b/modules/extension/xsd/xsd-core/src/main/java/org/geotools/xs/XSUtils.java
@@ -1,0 +1,40 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ *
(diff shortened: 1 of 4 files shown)


aaime:
@ianturton seems ok to me, but I believe you're the authority when it comes to local time? (think it was you to implement it in the first place, maybe I remember it worng).

aaime:
@fernandor777 can you rebase on master, to get the travis configuration upgrade for java 11?

fernandor777:
@aaime of course, rebasing done and pushed.

nmco:
I guess this ready to be merged? Is this a bug fix or improvement? Do you intend to back-port this? Looks like that without this fix that the behavior of time values is inconsistent, do you foreseen any backwards compatibility in back-porting this @fernandor777? 
In any case, if you intend to back-port, please ask on the ML 🛩

nmco:
Merging this.

PR #2460, opened 2019-06-29
Trying a switch to OpenJDK 11 for Travis, hoping it's not affected by Oracle whims like the oracleJDK is

aaime:


aaime:
Well it works, but for some reason the java11 QA build is almost twice as slow as the old ones with Oracle JDK... oh well, better slow and working I suppose... merging.

PR #2462, opened 2019-07-01
GEOT-6331 not swallowing exceptions anymore

marpet:
Added also a test to ensure the behaviour

### DIFF
--- a/modules/library/metadata/src/main/java/org/geotools/util/logging/LoggingImagingListener.java
+++ b/modules/library/metadata/src/main/java/org/geotools/util/logging/LoggingImagingListener.java
@@ -19,6 +19,7 @@ package org.geotools.util.logging;
 
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import javax.media.jai.OperationRegistry;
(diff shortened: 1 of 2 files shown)


modules/library/metadata/src/main/java/org/geotools/util/logging/LoggingImagingListener.java
```diff
@@ -29,10 +30,15 @@ public boolean errorOccurred(
         Logger log = Logging.getLogger(LoggingImagingListener.class);
         if (message.contains("Continuing in pure Java mode")) {
             log.log(Level.FINER, message, thrown);
+            return false; // we are not trying to recover
         } else {
             log.log(Level.INFO, message, thrown);
+            if (thrown instanceof RuntimeException && !(where instanceof OperationRegistry)) {
+                throw (RuntimeException) thrown;
```
jodygarnett: We should be clear that this represents a change in functionality for the GeoTools library, I hope it does not cause too much trouble!

jodygarnett:
While this was merged the bug was still open, so not sure exactly what release of geotools this was addressed in.

PR #2463, opened 2019-07-01
[GEOT-6332] unregister App-schema included mappings datastores on failure.

fernandor777:
This PR fixes an App-Schema Data store factory bug that don't unregisters included data stores on creation failure.  Now all included stores are disposed and unregistered if an exception occurs.

Issue:
https://osgeo-org.atlassian.net/browse/GEOT-6332


aaime:
I have no experience on the app-schema store but it looks generally good to me.

nmco:
Nice fix, thx @fernandor777.

nmco:
@fernandor777 can you keep an eye on the back-ports? I have updated the JIRA ticket.

PR #2469, opened 2019-07-03
[GEOT-6334] Dateline wrapping heuristic cannot handle large earth objects

aaime:


ianturton:
Looks good to me

aaime:
@ianturton thanks for having a look :-D

PR #2470, opened 2019-07-03
[GEOT-5729] Schema cache autoconfiguration fails for empty GeoServer data directory

fernandor777:
This PR fixes the Schema cache autoconfiguration bug relaxing the directory schema checking in a simpler approach, ensuring mainly the workspaces directory and its default.xml file for detecting the geoserver root folder and creating the app-schema-cache folder.

Also a warning logging line was added on not found valid directory event.

Issue:
https://osgeo-org.atlassian.net/browse/GEOT-5729

### DIFF
--- a/modules/library/xml/src/main/java/org/geotools/xml/resolver/SchemaCache.java
+++ b/modules/library/xml/src/main/java/org/geotools/xml/resolver/SchemaCache.java
@@ -60,13 +60,6 @@ public class SchemaCache {
     private static final boolean DEFAULT_KEEP_QUERY = true;
 
     /**
-     * Filenames used to recognise a GeoServer data directory if automatic configuration is enabled.
-     */
-    private static final String[] GEOSERVER_DATA_DIRECTORY_FILENAMES = {
-        "global.xml", "wcs.xml", "wfs.xml", "wms.xml"
-    };
-
-    /**
      * Subdirectories used to recognise a GeoServer data directory if automatic configuration is
      * enabled.
      */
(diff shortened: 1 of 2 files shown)


modules/library/xml/src/main/java/org/geotools/xml/resolver/SchemaCache.java
```diff
@@ -464,18 +460,16 @@ static boolean isSuitableDirectoryToContainCache(File directory) {
...
         for (String subdirectory : GEOSERVER_DATA_DIRECTORY_SUBDIRECTORIES) {
             File dir = new File(directory, subdirectory);
             if (!dir.isDirectory()) {
                 return false;
             }
         }
+        // so there is a workspaces directory, lets check default.xml file
+        File workspacesDir = new File(directory, "workspaces");
+        File defaultXmlFile = new File(workspacesDir, "default.xml");
+        if (!defaultXmlFile.exists()) return false;
```
aaime: A bit confused, if the data dir is empty, truly empty, there will be no workspace either.... but then you cannot configure a store. Maybe it's a "not so empty" case where at least a workspace has been created?

fernandor777: Since former dir structure check was based on some files present and no file is present on clean data directory, the algorithm was simplified to check the required workspaces folder only (since app-schema store requires a workspace at least).  Indeed the issue should say that the data dir is files-empty, but no so empty.

PR #2477, opened 2019-07-05
[GEOT-6336] GML 2 temporal data is not being format like GML 3 encoding.

fernandor777:
This PR enables temporal data formatting on GML 2 encoder, using same strategy as GML 3.

Issue:
https://osgeo-org.atlassian.net/browse/GEOT-6336


mbarto:
Looks good to me

PR #2478, opened 2019-07-08
[GEOT-6329] MongoDB schema inferrer only uses first item on inner arrays

fernandor777:
This PR allows MongoDB schema inferrer to use all inner arrays items.

Issue:
https://osgeo-org.atlassian.net/browse/GEOT-6329


nmco:
Looks good, thx @fernandor777.

PR #2492, opened 2019-07-12
WFS-NG Check

jodygarnett:
Looking at an axis order issue encountered by GeoNetwork when updating to GeoTools 21.x:

* [x] set up fixture template for GeoServer online tests
* [x] check WFSDataStore ability to reproject based on axis order change

Reference:
* https://github.com/geonetwork/core-geonetwork/pull/3886
* https://github.com/geoserver/geoserver/pull/3674

### DIFF
--- a/modules/unsupported/wfs-ng/src/test/java/org/geotools/data/wfs/online/v2_0/GeoServerOnlineTest.java
+++ b/modules/unsupported/wfs-ng/src/test/java/org/geotools/data/wfs/online/v2_0/GeoServerOnlineTest.java
@@ -17,6 +17,7 @@
 package org.geotools.data.wfs.online.v2_0;
 
 import static org.geotools.data.wfs.WFSTestData.GEOS_STATES_11;
+
 import java.util.Collections;
 import org.geotools.data.wfs.WFSDataStoreFactory;
 import org.geotools.data.wfs.online.AbstractWfsDataStoreOnlineTest;
@@ -61,8 +62,8 @@ public class GeoServerOnlineTest extends AbstractWfsDataStoreOnlineTest {
         return ff.intersects(ff.property("the_geom"), ff.literal(polygon));
     }
 
-        @Override
-        public void testDataStoreHandlesAxisFlipping() {
-            // disabled, not implemented for 2.0.0
-        }
+    @Override
+    public void testDataStoreHandlesAxisFlipping() {
+        // disabled, not implemented for 2.0.0
+    }
 }
--- a/modules/unsupported/wfs-ng/src/test/java/org/geotools/data/wfs/online/v2_0/GeoServerOnlineTest.java
+++ b/modules/unsupported/wfs-ng/src/test/java/org/geotools/data/wfs/online/v2_0/GeoServerOnlineTest.java
@@ -17,7 +17,6 @@
 package org.geotools.data.wfs.online.v2_0;
 
 import static org.geotools.data.wfs.WFSTestData.GEOS_STATES_11;
-
 import java.util.Collections;
 import org.geotools.data.wfs.WFSDataStoreFactory;
 import org.geotools.data.wfs.online.AbstractWfsDataStoreOnlineTest;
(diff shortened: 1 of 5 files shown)


jodygarnett:
Making some notes:

* Many objects do not have toString implementations making debugging annoying
* DataStore uses InProcessingLocking manager when it should delegate locking (if available) to the WFS Service
* Uses ConcurrentHashMap for names, remoteFeatureTypes rather than define a WFSContentEntry
* Has remoteStoredQueries,storedQueryDescriptionTypes,storedQueryDescriptionTypesLock rather than a data structure to manage stored queries

jodygarnett:
ogr output:
```
ogrinfo -so -al states.shp 
INFO: Open of `states.shp'
      using driver `ESRI Shapefile' successful.

Layer name: states
Metadata:
  DBF_DATE_LAST_UPDATE=1904-07-25
Geometry: Polygon
Feature Count: 49
Extent: (-124.731422, 24.955967) - (-66.969849, 49.371735)
Layer SRS WKT:
GEOGCS["WGS 84",
    DATUM["WGS_1984",
        SPHEROID["WGS 84",6378137,298.257223563,
            AUTHORITY["EPSG","7030"]],
        AUTHORITY["EPSG","6326"]],
    PRIMEM["Greenwich",0,
        AUTHORITY["EPSG","8901"]],
    UNIT["degree",0.0174532925199433,
        AUTHORITY["EPSG","9122"]],
    AUTHORITY["EPSG","4326"]]
STATE_NAME: String (25.0)
STATE_FIPS: String (2.0)
SUB_REGION: String (7.0)
STATE_ABBR: String (2.0)
LAND_KM: Real (19.9)
WATER_KM: Real (19.9)
PERSONS: Real (19.9)
FAMILIES: Real (19.9)
HOUSHOLD: Real (19.9)
MALE: Real (19.9)
FEMALE: Real (19.9)
WORKERS: Real (19.9)
DRVALONE: Real (19.9)
CARPOOL: Real (19.9)
PUBTRANS: Real (19.9)
EMPLOYED: Real (19.9)
UNEMPLOY: Real (19.9)
SERVICE: Real (19.9)
MANUAL: Real (19.9)
P_MALE: Real (19.9)
P_FEMALE: Real (19.9)
SAMP_POP: Real (19.9)
```

jodygarnett:
Moving this discussion over to https://github.com/geoserver/geoserver/pull/3674

aaime:
So is there anything left to do here?

aaime:
@jodygarnett if there is no more activity here I'd suggest to close it, it's been around for a while with no sign of resolution?

jodygarnett:
Marking as ready for review, will look at travis-ci failure. 

As many of the online tests were written for the old implementation they do not all pass; by merging this in it corrects one serious flaw; and allows some tests to be turned on using a geoserver.properties file.

jodygarnett:
@aaime thanks for the encouragement, this is ready for your review as an improvement. I can now meet my requirement, and have made it possible to run and pass a few more tests along the way.

modules/unsupported/wfs-ng/src/test/java/org/geotools/data/wfs/online/v2_0/GeoServerOnlineTest.java
```diff
@@ -33,7 +33,7 @@
 public class GeoServerOnlineTest extends AbstractWfsDataStoreOnlineTest {
 
     public static final String SERVER_URL =
-            "http://localhost:9090/geoserver/wfs?service=WFS&request=GetCapabilities&version=2.0.0";
```
aaime: Is the test run on a normal maven build (hopefully not)? There are around a few annoying tests that just fail if there is a GeoServer running locally that does not have the layers they expect.

modules/unsupported/wfs-ng/src/test/java/org/geotools/data/wfs/online/v2_0/GeoServerOnlineTest.java
```diff
@@ -33,7 +33,7 @@
...
+            "http://localhost:8080/geoserver/wfs?service=WFS&request=GetCapabilities&version=2.0.0";
```
jodygarnett: We should move this to a test fixture then

modules/unsupported/wfs-ng/src/test/java/org/geotools/data/wfs/online/v2_0/GeoServerOnlineTest.java
aaime: Checked, no worries, they are not run at all

jodygarnett:
I would like to backport this, but I understand a geoserver test case needs to be modified at the same time

dromagnoli:
> I would like to backport this, but I understand a geoserver test case needs to be modified at the same time

If you needed it, @jodygarnett, this is the commit I made for that GeoServer test fix
https://github.com/geoserver/geoserver/commit/e03b7dc122bd597dd49a6c8e7fcb43556ff70ffa

PR #2497, opened 2019-07-16
fix for GEOT-4319, adding support for loading of gridshift files with or without the ClasspathGridShiftLocator.

brbrown25:
https://osgeo-org.atlassian.net/browse/GEOT-4319

### DIFF
--- a/modules/library/referencing/src/main/java/org/geotools/referencing/operation/transform/NADCONTransform.java
+++ b/modules/library/referencing/src/main/java/org/geotools/referencing/operation/transform/NADCONTransform.java
@@ -21,7 +21,10 @@ import java.io.ObjectInputStream;
 import java.io.PrintWriter;
 import java.io.Serializable;
 import java.net.URI;
+import java.net.URISyntaxException;
 import java.net.URL;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 import java.util.prefs.Preferences;
 import org.geotools.metadata.i18n.ErrorKeys;
 import org.geotools.metadata.i18n.Errors;
@@ -155,6 +158,8 @@ public class NADCONTransform extends AbstractMathTransform
     /** The grid driving this transform */
     NADConGridShift grid;
 
+    private static Logger logger = Logger.getLogger(NADCONTransform.class.getName());
+
     /**
      * Constructs a {@code NADCONTransform} from the specified grid shift files.
      *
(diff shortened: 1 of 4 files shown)


brbrown25:
Another possible solution would be to modify the ClasspathGridShiftLocator but I think this approach would have the least amount of impact on existing projects and provide a path towards deprecating the old method if that is desired.

brbrown25:
@aaime Any thoughts on the approach in this pr?

aaime:
Had a look, I'm not inclined to accept the PR as is. Rationale:

- There are a number of javadoc changes that actually change nothing but formatting or order of tags, please avoid that
- The patch creates special code paths for the case of inputs being straight file locations, but I believe it would have been possible to write a FileGridShiftLocator, and register it in META-INF/services, instead. Am I missing something? (haven't seen this code in several years, might as well have been written by someone else).

If you're wondering why a FileGridShiftLocator does not exist, it's because we did not need it. The original author of the patch needed classpath lookups (grid shift files placed inside the jar files I believe), while I needed GeoServer to be able to run lookups out of the data directory instead, so created the extension point and added an implementation in GeoServer too, see https://github.com/geoserver/geoserver/blob/master/src/main/src/main/java/org/vfny/geoserver/crs/GeoserverGridShiftLocator.java

brbrown25:
Andrea,

Thanks for the feedback!

I can back out the doc formatting changes. The reason I changed some of the names was to match what I saw In the code, I didn’t meant to reorder things. 

The reason I made the change for handling was that as it stands it will only locate resources packaged with the jar. I had attempt to use the code as is and passed in references to files that were definitely on the class path and I would get an exception. 

I believe this has to do with visibility in the underlying ClassPathLocator implementation. Since the locate method takes a URI and returns a URL I do think it might make sense to see if calling toUrl doesn’t throw an exception and then fallback to a locator. 

What if I were to back out the formatting changes and just keep that toUrl try catch would that be ok? 

If you’d like I can put together a sample project showing how I’m using it and the exception it’s giving.

Thanks,
-Brandon

Brandon Brown
(240) 506-8335 (m)

aaime:
Hi Brandon, it does not make sense to create special case code when there is an extension point that could be used instead. Unless you can argument it's not possible to roll a FileGridShiftLocator for some reason I'm not seeing?

aaime:
Or maybe you're not really trying to get stuff out of the file system and the problem is of different nature? Sorry I'm really tired, shouldn't be looking at this, but you insisted, so giving it a half shot...

aaime:
Reading again your message, it seems you are indeed not trying to get anything out of the file system, but out of the classpath but from a jar that's not gt-referencing (or something like that). Yes, a sample project would really help...

brbrown25:
It’s all good! Yes definitely not trying to load from file system. I’ll put together a sample project and send a link to it Monday  

Have a great weekend!

-Brandon 

Brandon Brown
(240) 506-8335 (m)

aaime:
I see recent activity, did you ever send the sample project?

brbrown25:
I have not had a chance to put it together yet, somethings came up at work. I recently did an update to pull in latest master and also remove some of the format changes I had made. 

I do plan to put together the sample project this week though!

Have a great weekend,
Brandon. 

Brandon Brown
(240) 506-8335 (m)

aaime:
Any update? ;-)

tbarsballe:
There's been no update here in over 2 months, closing this PR. If you have an update in the future, please reopen / resubmit.

PR #2499, opened 2019-07-19
[GEOT-5958]: ArithmeticException divide by zero on layoutHelper

dromagnoli:


modules/library/render/pom.xml
```diff
@@ -212,8 +212,18 @@
           </execution>
         </executions>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <configuration>
+          <systemPropertyVariables>
+            <!-- settings these properties on Test's code won't work due to order of initialization -->
+            <org.geotools.coverage.jaiext.enabled>true</org.geotools.coverage.jaiext.enabled>
```
aaime: This seems unecessary on master and 21.x?

dromagnoli: Right... Removing it right now

PR #2504, opened 2019-07-22
[GEOT-6347] Layer creation using REST API very slow in a PostGIS database with thousands of tables

imranrajjad:
Actual Reported Issue
https://osgeo-org.atlassian.net/browse/GEOS-7533?oldIssueView=true

Geotools JIRA
https://osgeo-org.atlassian.net/browse/GEOT-6347

I propose to put commonly used sqltypes inside overrides map and have the behavior inside JDBCDataStore controllable through some param, that if all sqlTypeName have been found through dialect, it should not bother the driver for querying for datatypes inside underlying Database.

### DIFF
--- a/modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
+++ b/modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
@@ -238,6 +238,9 @@ public final class JDBCDataStore extends ContentDataStore implements GmlObjectSt
     /** sql type to sql type name overrides */
     protected HashMap<Integer, String> sqlTypeToSqlTypeNameOverrides;
 
+    /** cache of sqltypes found in database */
+    protected ConcurrentHashMap<Integer, String> dBsqlTypesCache;
+
     /** Feature visitor to aggregate function name */
     protected HashMap<Class<? extends FeatureVisitor>, String> aggregateFunctions;
 
@@ -624,6 +627,20 @@ public final class JDBCDataStore extends ContentDataStore implements GmlObjectSt
         }
 
         return sqlTypeToSqlTypeNameOverrides;
+    }
+
+    /**
+     * Returns a map integer constants for database types (from {@link Types}) to database type
+     * names.
+     *
+     * <p>This method will return an empty map when there are no overrides.
+     */
+    public ConcurrentHashMap<Integer, String> getDBsqlTypesCache() {
+        if (dBsqlTypesCache == null) {
+            dBsqlTypesCache = new ConcurrentHashMap<Integer, String>();
+        }
+
+        return dBsqlTypesCache;
     }
 
     /** Returns the supported aggregate functions and the visitors they map to. */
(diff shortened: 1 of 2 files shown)


modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
```diff
@@ -3228,12 +3229,17 @@ private Integer findVarcharColumnLength(AttributeDescriptor att) {
             if (sqlTypeName != null) {
                 sqlTypeNames[i] = sqlTypeName;
             }
-        }
+        }        
+        //Step 2 Geos-7533 make below two lines controllable from hint or param
+        //if all sql type names have been found in dialect dont
+        //go to database GEOS-7533
+        boolean allTypesFound = !ArrayUtils.contains(sqlTypeNames, null);        
+        if(allTypesFound) return sqlTypeNames;
```
nmco: Space between `ìf`` and ``(`` missing, remember to do a full build to apply automatic formatting.

nmco: I would add a log messages (level warning) if the type doesn't exist and we need to go to the database.

modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/PostGISDialect.java
```diff
@@ -1114,6 +1115,8 @@ public void registerSqlTypeNameToClassMappings(Map<String, Class<?>> mappings) {
 
     @Override
     public void registerSqlTypeToSqlTypeNameOverrides(Map<Integer, String> overrides) {
+        //Step 1 Geos-7533 put the types in dialect
+        overrides.put(4, "int4");
```
nmco: Please use ``Types.INTEGER`` instead of ``4``.

modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/PostGISDialect.java
```diff
@@ -1363,7 +1366,7 @@ public void postDropTable(String schemaName, SimpleFeatureType featureType, Conn
             throws SQLException {
         Statement st = cx.createStatement();
         String tableName = featureType.getTypeName();
-
+
```
nmco: Unrelated formatting change.

aaime:
I don't see the PR well on my pho e, but dialects are long lived, do o e Wat would be to cache the resulting type map at the first call in a non static field, if people add extra types they will have to reload GS/rebuild the store, does not seem like a big deal

modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
```diff
@@ -3229,6 +3230,10 @@ private Integer findVarcharColumnLength(AttributeDescriptor att) {
                 sqlTypeNames[i] = sqlTypeName;
             }
         }
+        // Geos-7533 if all sql type names have been found in dialect dont
```
aaime: Please refer to Geotools tickets, not geoserver ones

modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
```diff
@@ -3285,6 +3290,9 @@ private Integer findVarcharColumnLength(AttributeDescriptor att) {
 
                     if (sqlType == sqlTypes[i]) {
                         sqlTypeNames[i] = sqlTypeName;
+                        // Geos-7533
```
aaime: Please refer to Geotools tickets, not geoserver ones

modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
```diff
@@ -3285,6 +3290,9 @@ private Integer findVarcharColumnLength(AttributeDescriptor att) {
...
+                        // cache the sqlType which required going to database for sql types
+                        getSqlTypeToSqlTypeNameOverrides().put(sqlType, sqlTypeName);
```
aaime: Nope, this is not all right. The getSqlTypeToSqlTypeNameOverrides is public with a specific semantic defined in its javadoc, what you are registering here is not an override any longer.
I suggest you create a separate map that acts as a cache instead. Also, if you plan to update it this way, make it a concurrent hash map, otherwise code in production might experience a concurrent modification exception.

imranrajjad: we already have two hashmaps inside the implementation

would it be acceptable that if this method is made a synchronized method, so that if unseen fields are being fetched from database from user1, user2 can wait for the only one time. 

For hash map, if I use the main hashamp instead of the hashmap behind getSqlTypeToSqlTypeNameOverrides() will that do?

nmco:
@imranrajjad what's the status of this?

imranrajjad:
I have addressed @aaime `s requested change

modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/PostGISDialect.java
imranrajjad: hard coding removed, this is addressed

modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/PostGISDialect.java
imranrajjad: fixed

modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
imranrajjad: fixed

modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
imranrajjad: fixed

modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
imranrajjad: changed to relevant GEOT, fixed

modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
imranrajjad: I will add the warning message and push changes 👍

imranrajjad: fixed

nmco:
Thx @imranrajjad.

simboss:
if I am right this is a fix. Can we backport to 2.16 and 2.15?

PR #2505, opened 2019-07-23
[GEOT-6345] MongoDB OR filters are not executed on native search engine

fernandor777:
MongoDB store does not register OR operation into filter capabilities and due to this OR operations are not executes on native search engine leading to wrong queries results with or involved.

This seems related to older MongoDB versions unable to mix $or operations with spatial indexes, but it is already supported since 2.6.0 version.

This PR add native OR support by default and adds a store parameter for deactivate it if needed for older MongoDB versions.

See:
https://github.com/geotools/geotools/blob/master/modules/plugin/mongodb/src/main/java/org/geotools/data/mongodb/MongoDataStore.java#L214-L218

Issue:
https://osgeo-org.atlassian.net/browse/GEOT-6345

### DIFF
--- a/modules/plugin/mongodb/src/main/java/org/geotools/data/mongodb/MongoDataStore.java
+++ b/modules/plugin/mongodb/src/main/java/org/geotools/data/mongodb/MongoDataStore.java
@@ -34,6 +34,9 @@ import java.util.Set;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.stream.StreamSupport;
+import org.bson.BsonDocument;
+import org.bson.BsonString;
+import org.bson.Document;
 import org.geotools.data.FeatureWriter;
 import org.geotools.data.Transaction;
 import org.geotools.data.store.ContentDataStore;
@@ -75,6 +78,8 @@ public class MongoDataStore extends ContentDataStore {
     final MongoClient dataStoreClient;
     final DB dataStoreDB;
 
+    final boolean deactivateOrNativeFilter;
+
     @SuppressWarnings("deprecation")
     FilterCapabilities filterCapabilities;
 
@@ -111,6 +116,8 @@ public class MongoDataStore extends ContentDataStore {
                     "Unknown mongodb database, \"" + dataStoreClientURI.getDatabase() + "\"");
         }
 
+        this.deactivateOrNativeFilter = isMongoVersionLessThan2_6(dataStoreClientURI);
+
         schemaStore = createSchemaStore(schemaStoreURI);
         if (schemaStore == null) {
             dataStoreClient.close(); // This smells bad too...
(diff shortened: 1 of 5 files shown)


modules/plugin/mongodb/src/main/java/org/geotools/data/mongodb/MongoDataStore.java
```diff
@@ -211,13 +224,18 @@ private MongoSchemaStore createSchemaStore(String schemaStoreURI) {
...
-        capabilities.addType(And.class);
-        capabilities.addType(Not.class);
+        if (deactivateOrNativeFilter) {
+            /*
+             * disable FilterCapabilities.LOGICAL_OPENGIS since it contains Or.class (in
+             * additions to And.class and Not.class. MongodB 2.4 doesn't support '$or' with
+             * spatial operations.
+             */
+            capabilities.addType(And.class);
+            capabilities.addType(Not.class);
```
nmco: So the goal is to remove the ``Or.class``, I would have extended ``FilterCapabilities`` with a remove type function, although not sure how that would play out with the ``FUNCTIONS``. 

Anyway, ok by me, no need to change, this is just a note.

modules/plugin/mongodb/src/main/java/org/geotools/data/mongodb/MongoDataStoreFactory.java
```diff
@@ -58,6 +58,13 @@
                     "Objects IDs for schema generation (comma separated)",
                     false,
                     null);
+    public static final Param DEACTIVATE_NATIVE_OR =
+            new Param(
+                    "deactivate_native_or",
+                    Boolean.class,
+                    "Deactivate native '$or' operator",
+                    false,
+                    Boolean.FALSE);
```
nmco: This is a clean approach to backwards compatibility, thinking out loud, how will this play with migrations? Some upgrades from GS 2.14.x and has 5 MongoDB stores, by default $or will be activated right?

nmco:
Just a few notes, well done :smiley:. That say, can you updated the documentation:
https://docs.geotools.org/latest/userguide/library/data/mongodb.html
... and remove the reference to the ``$or`` limitation (``Implementation Notes`` section).

nmco:
For reference, the ML discussion:
http://osgeo-org.1560.x6.nabble.com/CQL-filter-OR-operator-does-not-work-on-attributes-that-are-multiple-on-MongoDB-DataSource-td5408541.html

nmco:
Regarding the build failure, seems unrelated:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-pmd-plugin:3.11.0:check (default) on project gt-render: You have 1 PMD violation. For more details see: /home/travis/build/geotools/geotools/modules/library/render/target/pmd.xml -> [Help 1]
```

modules/plugin/mongodb/src/main/java/org/geotools/data/mongodb/MongoDataStoreFactory.java
fernandor777: Indeed native $or filter will be activated by default and allowing user for deactivating it on store configuration if needed.

nmco:
@fernandor777, would it make sense to detect [MongoDB version](https://stackoverflow.com/questions/19274805/getting-version-of-mongo-instance-with-java-driver) and automatically deactivate this functionality for versions prior to 2.6.0?

nmco:
I would get ride of the activate \ deactivate option on the UI, if we are dealing whit MongoDB 2.6.0 or superior we delegate the $or, otherwise we don't. If we can't detect MongoDB version, then we don't delegate.

The MongoDB version should be detected the first time the store connects to MongoDB, at each restart  or when the store is reloaded, this will vouch for the case the MongoDB instance gets updated. 

What do you think? Does this makes sense?

fernandor777:
@nmco Indeed this approach makes sense and looks better.

nmco:
Unrelated failure:
```
[INFO] Render ............................................. FAILURE [01:05 min]
```

PR #2507, opened 2019-07-23
[GEOT-5911] MongoDB CQL_Filter not working anymore

tuckerpm:
Change toCursor to generate the MongoDB query the same way getCountInternal generates the MongoDB query.

### DIFF
--- a/modules/plugin/mongodb/src/main/java/org/geotools/data/mongodb/MongoFeatureSource.java
+++ b/modules/plugin/mongodb/src/main/java/org/geotools/data/mongodb/MongoFeatureSource.java
@@ -258,7 +258,7 @@ public class MongoFeatureSource extends ContentFeatureSource {
         Filter f = q.getFilter();
         if (!isAll(f)) {
             Filter[] split = splitFilter(f);
-            query = toQuery(split[0]);
+            query = toQuery(f);
             if (!isAll(split[1])) {
                 postFilter.add(split[1]);
             }


tuckerpm:
Odd, the build failure seems unrelated to my change?

nmco:
At a first look this fix looks valid only for a particular case, when all the filters can be translated to MongoDB query language, let's have a look at the code:

```java
if (!isAll(f)) {
  Filter[] split = splitFilter(f);
  query = toQuery(split[0]);
  if (!isAll(split[1])) {
    postFilter.add(split[1]);
  }
}
``` 

This lines of code are basically splitting a non empty or ``NULL`` filter in parts that can be delegated to MongoDB and parts that need to be executed in memory.  Your suggested fix is basically implying that all OGC filters can be converted to MongoDB query language and consequently delegated to MongoDB, that's not always the case, for example advanced spatial operators.

The first comment on the issue associated whit this PR sugest the cause and possible fix:

nmco:
Question for you @tuckerpm, did you execute MongoBD online tests? Anyway, could you share with you use case \ filter? But IMHO I would definitively have a look at the suggested fixed in the issue. @fernandor777 what do you think?

tuckerpm:
@nmco thank you for the response.

If you step through the code, a count request is issued prior to requesting the records.  This count request works the same as my proposed change.  If it were a problem to execute this way, the count request would fail before the request for records is executed.  I could not think of any reason the count request should work different from the actual request.  If the count request needs to be corrected, I suggest that be done in a feature request as opposed to a minor bug fix.  The most recent release artifact ignores all applied cql filters logging correct counts, but returning the full record set.

I was not able to find existing unit tests that prove this functionality works.  I had to apply this patch in our environment in order to use any filters against our MongoDB store.

I tested the proposed patch in the linked issue first and it does not solve the problem.

Thanks again,
Patrick Tucker

tuckerpm:
Sorry, I tested the code with the following types of CQL Filters:
- "field" = 'value'
- "field" = 'value' and INTERSECT("geometry" , POLYGON((...)))
- "field" = 'value' and WITHIN("geometry", POLYGON((...)))

fernandor777:
The proposed PR breaks integration tests, also is not so good to bypass the filter capabilities split process since it's a standard in Geotools stores, but the bug is real and easily replicated with an equal filter.

Indeed the getCount query don't use the splitted filter, that is a bug.  On the other side filter filter splitter don't support common Attribute equality and this should be fixed too (and its tests needs some fixing as far I investigated).

nmco:
Thanks for the feedback @fernandor777, what do you mean by ``On the other side filter filter splitter don't support common Attribute equality``, can you kindly share an example?

nmco:
By integration tests I guess you mean the ones hitting an actual MongoDB instance?

fernandor777:
Hi @nmco , I mean this piece of code never matches with normal Equal filter expression (FilterFactory.equal):

https://github.com/geotools/geotools/blob/37669f4e4a67e5fe1b1c4fbfd4264e88a4de5240/modules/plugin/mongodb/src/main/java/org/geotools/data/mongodb/MongoFeatureSource.java#L339-L351

And yes, tests failing are the MongoDB instance involved.

tuckerpm:
@fernandor777 do you guys have a MongoDB installation that I can point at to execute the unit tests against?

Is there another class that implements the functionality in visitBinaryComparisonOperator correctly?

tbarsballe:
@tuckerpm We do not currently have an open MongoDB instance to execute online integration tests against. You can install your own version of MongoDB locally, and test it with GeoTools by creating a `~/.geotools/mongodb.properties` with your connection parameters.
See http://docs.geotools.org/latest/developer/conventions/test/online.html for more details.

aaime:
Closing this one due to lack of activity

PR #2516, opened 2019-08-06
GEOT-6353: coordinate axis unit improvement

dromagnoli:


modules/plugin/coverage-multidim/netcdf/src/main/java/org/geotools/imageio/netcdf/utilities/NetCDFCRSUtilities.java
```diff
@@ -138,6 +146,9 @@
         VERTICAL_AXIS_NAMES.add("z");
         VERTICAL_AXIS_NAMES.add("depth");
         VERTICAL_AXIS_NAMES.add("pressure");
+
+        // Default is false, resulting into no automatic conversion anymore
+        CONVERT_AXIS_KM = Boolean.parseBoolean(System.getProperty(CONVERT_AXIS_KM_KEY, "false"));
```
dromagnoli: Note that this system property has been introduced to not break backward compatibility when backporting.
Backport will use "true" as default value, to have it working as it worked since today (doing the automatic conversion).
Only master has this set to false.

modules/plugin/coverage-multidim/netcdf/src/main/java/org/geotools/coverage/io/netcdf/crs/NetCDFCoordinateReferenceSystemType.java
```diff
@@ -198,6 +186,55 @@ public static NetCDFCoordinateReferenceSystemType parseCRS(CoordinateReferenceSy
         return crsType;
     }
 
+    public static boolean hasConcatenatedTransform(ProjectedCRS projectedCRS) {
```
aaime: Public and static? Does not seem to require being public. If it needs to, please add javadoc.

modules/plugin/coverage-multidim/netcdf/src/main/java/org/geotools/coverage/io/netcdf/crs/NetCDFCoordinateReferenceSystemType.java
```diff
@@ -161,31 +164,16 @@ public static NetCDFCoordinateReferenceSystemType parseCRS(CoordinateReferenceSy
...
-            } else if (transform instanceof AlbersEqualArea) {
-                crsType = ALBERS_EQUAL_AREA;
-            } else if (transform instanceof RotatedPole) {
-                crsType = ROTATED_POLE;
+            // NetCDF supports a reduced number of CRS Type (WGS84 + some projection).
+            // Some ProjectedCRS may have unit of measure resulting in concatenated transforms.
+            if (hasConcatenatedTransform(projectedCRS)) {
+                Projection projection = projectedCRS.getConversionFromBase();
+                MathTransform mt = projection.getMathTransform();
+                crsType = getProjectionType(((ConcatenatedTransform) mt).transform1);
```
aaime: How do you know, in general, that the concatenated transform you need to look at is the first? In theory there is a chain of N (multple concatenated transforms) and the projection steps. But I agree that's the common case, and maybe the only case currently implemented in the code base, but in general there will be many... maybe add a check that there are only two, and if one of them is a concatenated itself, throw a meaningful exception? (something like "we dont' support this case").

modules/plugin/coverage-multidim/netcdf/src/main/java/org/geotools/coverage/io/netcdf/crs/NetCDFCoordinateReferenceSystemType.java
```diff
@@ -215,6 +252,32 @@ public static NetCDFCoordinateReferenceSystemType parseCRS(CoordinateReferenceSy
...
+    public NetCDFCoordinate[] getCoordinates(CoordinateReferenceSystem crs) {
+        NetCDFCoordinate[] axisCoordinates = getCoordinates();
+        if (this != NetCDFCoordinateReferenceSystemType.WGS84) {
+            // NetCDF only support WGS84 and a few projections
+            // There might be the case that the projection contains a concatenatedTransform
+            // due to a specific unit in the coordinate axis (as an instance, km or foot instead
+            // of classic meter).
+            // Extract the unit of measure from the actual coordinate axis in that case.
+            ProjectedCRS projectedCrs = (ProjectedCRS) crs;
+            if (NetCDFCoordinateReferenceSystemType.hasConcatenatedTransform(projectedCrs)) {
```
aaime: You're still in the same class, no need to qualify :-)

modules/plugin/coverage-multidim/netcdf/src/main/java/org/geotools/coverage/io/netcdf/crs/NetCDFProjection.java
```diff
@@ -175,8 +176,11 @@ public CoordinateReferenceSystem parseCoordinateReferenceSystem(Variable var)
                     handleParam(paramsMapping, netcdfParameters, ogcParameterKey, var);
                 }
 
+                Map<String, Object> map = new HashMap<String, Object>();
+                map.put(NetCDFUtilities.NAME, projectionName);
```
aaime: So the projection name could be overridden by the provided crsProperties?
If that does not make sense, it's easier to just pass the crsProperties to the hash map constructor and then put then name on top of it... or not?

modules/plugin/coverage-multidim/netcdf/src/main/java/org/geotools/imageio/netcdf/NetCDFGeoreferenceManager.java
```diff
@@ -110,6 +116,13 @@ protected void init() throws FactoryException, IOException {
                 throw new IllegalStateException("Unable to create envelope for this dataset");
             }
 
+            // The previous code was able to automatically convert km coordinates to meter.
+            // Let's check if we still want this automatically conversion occur or we
+            // want the actual axisUnit to be used.
+            if (!NetCDFCRSUtilities.isConvertAxisKm() && axisUnit != null) {
```
aaime: You said in other parts of the code that the conversion could be due to using feet... so why make it specific to KM?

modules/plugin/coverage-multidim/netcdf/src/main/java/org/geotools/imageio/netcdf/utilities/NetCDFCRSUtilities.java
```diff
@@ -123,6 +123,14 @@
 
     private static final Map<AxisType, String> OPPOSITES = new HashMap<AxisType, String>(16);
 
+    /**
+     * this flag states if an automatic conversion from km to m should happen with axis/coordinates
+     */
+    public static final String CONVERT_AXIS_KM_KEY =
```
aaime: You said in other parts of the code that the conversion could be due to using feet... so why make it specific to KM?

modules/plugin/coverage-multidim/netcdf/src/main/java/org/geotools/imageio/netcdf/utilities/NetCDFCRSUtilities.java
aaime: You said in other parts of the code that the conversion could be due to using feet... so why make it specific to KM? Could be a CONVERT_AXIS_UNIT instead.

modules/plugin/coverage-multidim/netcdf/src/main/java/org/geotools/imageio/netcdf/utilities/NetCDFCRSUtilities.java
```diff
@@ -509,4 +520,8 @@ static CoordinateSystemAxis getAxis(
             }
         }
     }
+
+    public static boolean isConvertAxisKm() {
```
aaime: Javadoc.

modules/plugin/coverage-multidim/netcdf/src/test/java/org/geotools/coverage/io/netcdf/NetCDFCRSTest.java
```diff
@@ -419,6 +446,43 @@ public void testUnsupportedGridMappingName() throws IOException {
         reader.dispose();
     }
 
+    @Test
+    public void testPreserveKM() throws Exception {
+        setFinalStaticField("CONVERT_AXIS_KM", false);
```
aaime: Use CONVERT_AXIS_KM_KEY

modules/plugin/coverage-multidim/netcdf/src/test/java/org/geotools/coverage/io/netcdf/NetCDFCRSTest.java
```diff
@@ -419,6 +446,43 @@ public void testUnsupportedGridMappingName() throws IOException {
...
+        assertTrue(
+                "km".equalsIgnoreCase(crs.getCoordinateSystem().getAxis(0).getUnit().toString()));
+        GeneralEnvelope originalEnvelope =
+                reader.getOriginalEnvelope(reader.getGridCoverageNames()[0]);
+        assertEquals(COORDINATE_IN_KM, originalEnvelope.getMaximum(0), DELTA);
+    }
+
+    @Test
+    public void testAutoConversionKmToM() throws Exception {
+        setFinalStaticField("CONVERT_AXIS_KM", true);
```
aaime: Use CONVERT_AXIS_KM_KEY

modules/plugin/coverage-multidim/netcdf/src/test/resources/org/geotools/coverage/io/netcdf/test-data/netcdf.projections.properties
```diff
@@ -2,3 +2,5 @@
...
+970916=PROJCS["polar_stereographic",GEOGCS["unknown",DATUM["unknown",SPHEROID["unknown", 6371229.0, 0.0]],PRIMEM["Greenwich", 0.0],UNIT["degree", 0.017453292519943295],AXIS["Geodetic longitude", EAST],AXIS["Geodetic latitude", NORTH]],PROJECTION["Polar_Stereographic"],PARAMETER["central_meridian", -135.0],PARAMETER["latitude_of_origin", 90.0],PARAMETER["scale_factor", 0.9330127239227295],PARAMETER["false_easting", 0.0],PARAMETER["false_northing", 0.0],UNIT["m", 1.0],AXIS["Easting", EAST],AXIS["Northing", NORTH],AUTHORITY["EPSG","970916"]]
+970917=PROJCS["polar_stereographic",GEOGCS["unknown",DATUM["unknown",SPHEROID["unknown", 6371229.0, 0.0]],PRIMEM["Greenwich", 0.0],UNIT["degree", 0.017453292519943295],AXIS["Geodetic longitude", EAST],AXIS["Geodetic latitude", NORTH]],PROJECTION["Polar_Stereographic"],PARAMETER["central_meridian", -135.0],PARAMETER["latitude_of_origin", 90.0],PARAMETER["scale_factor", 0.9330127239227295],PARAMETER["false_easting", 0.0],PARAMETER["false_northing", 0.0],UNIT["km", 1000.0],AXIS["Easting", EAST],AXIS["Northing", NORTH],AUTHORITY["EPSG","970917"]]
```
aaime: Could be interesting to have a case in feet to test the generality of the approach.

modules/plugin/coverage-multidim/netcdf/src/main/java/org/geotools/imageio/netcdf/utilities/NetCDFCRSUtilities.java
dromagnoli: I thought to name like that at the beginning.
However, there is no conversion at the moment for feet and we are going to stop having automatic-conversion... so an user setting that convert_axis_unit would expect to have it convert feet too which is actually not supported.
This is why it's explicitly named _km.

modules/plugin/coverage-multidim/netcdf/src/main/java/org/geotools/coverage/io/netcdf/crs/NetCDFCoordinateReferenceSystemType.java
dromagnoli: Right. Will do. thanks

modules/plugin/coverage-multidim/netcdf/src/main/java/org/geotools/coverage/io/netcdf/crs/NetCDFCoordinateReferenceSystemType.java
dromagnoli: Right. At the beginning I was using it from another class too. Switching it back to private.

modules/plugin/coverage-multidim/netcdf/src/main/java/org/geotools/coverage/io/netcdf/crs/NetCDFCoordinateReferenceSystemType.java
dromagnoli: Right (see above). at the beginning it was in a different place. Fixing. Thx

modules/plugin/coverage-multidim/netcdf/src/main/java/org/geotools/imageio/netcdf/NetCDFGeoreferenceManager.java
dromagnoli: Nope. There wasn't any previous code in the past doing the foot conversion. I suppose is the GT Referencing doing that.

modules/plugin/coverage-multidim/netcdf/src/test/java/org/geotools/coverage/io/netcdf/NetCDFCRSTest.java
dromagnoli: Nope. Here I'm modifying the static field through reflection since the execution order of tests affect the outcome. the Key is the property name to be passed as JAVA_OPT

modules/plugin/coverage-multidim/netcdf/src/test/java/org/geotools/coverage/io/netcdf/NetCDFCRSTest.java
dromagnoli: same as above

modules/plugin/coverage-multidim/netcdf/src/test/resources/org/geotools/coverage/io/netcdf/test-data/netcdf.projections.properties
dromagnoli: I don't have any dataset at hand with a reference with feet or a definition of that.
I might have to build a fake one.

modules/plugin/coverage-multidim/netcdf/src/main/java/org/geotools/coverage/io/netcdf/crs/NetCDFProjection.java
dromagnoli: Right. At the beginning there was the case of a null crsProperties. I ended up using an emptyMap instead so it's now safe to have it in the constructor.
fixing.

modules/plugin/coverage-multidim/netcdf/src/main/java/org/geotools/imageio/netcdf/utilities/NetCDFCRSUtilities.java
aaime: Ok, did not see that there was already a special case for KM that you're trying to turn off. Go for it.

dromagnoli:
Feedbacks addressed within this updated PR:
https://github.com/geotools/geotools/pull/2517

PR #2518, opened 2019-08-07
Move jdbc-hana to supported modules

stefanuhrig:


modules/plugin/jdbc/jdbc-hana/pom.xml
```diff
@@ -17,7 +17,6 @@
     <groupId>org.geotools.jdbc</groupId>
     <artifactId>gt-jdbc</artifactId>
     <version>22-SNAPSHOT</version>
-    <relativePath>../../../plugin/jdbc/pom.xml</relativePath>
```
jodygarnett: This looks wrong?

docs/user/library/jdbc/index.rst
```diff
@@ -39,6 +39,7 @@ DataStore plugins:
    
    db2
    h2
+   hana
```
jodygarnett: Do you need to change the hana documentation page at all?

modules/plugin/jdbc/jdbc-hana/pom.xml
stefanuhrig: It is the same as in the other jdbc-plugins. The only have elements groupId, artifactId and version.

docs/user/library/jdbc/index.rst
stefanuhrig: As far as I can tell no further change is necessary. The documentation page is already there (https://docs.geotools.org/latest/userguide/library/jdbc/hana.html) and all the documentation is still up-to-date. Just this file needs an update to move the plugin from the list of unsupported plugins to the supported ones.

Or am I forgetting something?

modules/plugin/jdbc/jdbc-hana/pom.xml
jodygarnett: Sorry I misread the previous, so what was odd was previously this module declared the "wrong" parent :P

docs/user/library/jdbc/index.rst
jodygarnett: Had a look, I like the utility to load additional EPSG definitions.

The pom contains instructions on locally adding the jdbc driver, yep I think we are good.

PR #2525, opened 2019-08-16
[GEOT-4931] Fix comparison of attribute bindings in DataUtilities

elahrvivaz:
* Current DataUtilities isMatch implementation compares the class of the
attribute descriptor, not the type binding

Signed-off-by: Emilio Lahr-Vivaz <elahrvivaz@ccri.com>

### DIFF
--- a/modules/library/main/src/main/java/org/geotools/data/DataUtilities.java
+++ b/modules/library/main/src/main/java/org/geotools/data/DataUtilities.java
@@ -469,7 +469,8 @@ public class DataUtilities {
             return true;
         }
 
-        if (a.getLocalName().equals(b.getLocalName()) && a.getClass().equals(b.getClass())) {
+        if (a.getLocalName().equals(b.getLocalName())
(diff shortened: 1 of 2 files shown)


aaime:
Makes sense to me, merging.

nprigour:
Hi,
Maybe it is worth to also let DataUtilities provide the so far used functionality by providing two extra methods for DataUtilities.compare() and DataUtilities.isMatch() and differentiate based on an appropriate flag. I suggest this since sometimes it may be useful to have a method allowing the comparison of 2 featureTypes based on their attributes names.

aaime:
@nprigour feel free to add a new comparison method, the current one was clearly broken code wise, the class comparison was incorrect.

nprigour:
Hi @aaime,
Do I need to create first a new JIRA issue or is it sufficient to proceed directly with the creation of a new PR to address this?

aaime:
Can you please create a Jira ticket? We use Jira to create the release notes.

PR #2529, opened 2019-08-22
bump commons-beanutils to 1.9.4

bradh:


aaime:
I believe we had to fork commons-beanutils to add an extra fix in there... I believe it was because in order to get safe one had to manually change the code, and we have dependencies that do use beanutils and did not upgrade back then (or were not enabling the safe mode). @tbarsballe should know more.

PR #2531, opened 2019-08-23
[GEOT-6360] add preffered image format support in WMSLayer and

imranrajjad:
WMSCoverageReader

added constructors that take preffered WMS image format for remote wms
service

https://osgeo-org.atlassian.net/browse/GEOT-6360

### DIFF
--- a/modules/extension/wms/src/main/java/org/geotools/ows/wms/map/WMSCoverageReader.java
+++ b/modules/extension/wms/src/main/java/org/geotools/ows/wms/map/WMSCoverageReader.java
@@ -133,6 +133,32 @@ public class WMSCoverageReader extends AbstractGridCoverage2DReader {
         }
     }
 
+    public WMSCoverageReader(WebMapServer wms, Layer layer, String style, String preferredFormat) {
+        this.wms = wms;
+
+        // init the reader
+        addLayer(layer, style);
+
+        // best guess at the format with a preference for PNG (since it's normally transparent)
+        List<String> formats = wms.getCapabilities().getRequest().getGetMap().getFormats();
+        this.format = preferredFormat;
+        // verify if prefferedFormat is supported
+        if (!formats.contains(preferredFormat)) {
+            // if prefereded format is not supported default to first available
+            LOGGER.warning(preferredFormat + " format is not supported, falling back to defaults");
+            for (String format : formats) {
+                if ("image/png".equals(format)
+                        || "image/png24".equals(format)
+                        || "png".equals(format)
+                        || "png24".equals(format)
+                        || "image/png; mode=24bit".equals(format)) {
+                    this.format = format;
+                    break;
+                }
+            }
+        }
+    }
+
     void addLayer(Layer layer) {
         addLayer(layer, "");
     }
--- a/modules/extension/wms/src/main/java/org/geotools/ows/wms/map/WMSCoverageReader.java
+++ b/modules/extension/wms/src/main/java/org/geotools/ows/wms/map/WMSCoverageReader.java
@@ -120,43 +120,32 @@ public class WMSCoverageReader extends AbstractGridCoverage2DReader {
 
         // best guess at the format with a preference for PNG (since it's normally transparent)
         List<String> formats = wms.getCapabilities().getRequest().getGetMap().getFormats();
-        this.format = formats.iterator().next();
+        this.format = getDefaultFormat(formats);
+    }
+
+    public WMSCoverageReader(WebMapServer wms, Layer layer, String style, String preferredFormat) {
+        this.wms = wms;
+        // init the reader
+        addLayer(layer, style);
+        List<String> formats = wms.getCapabilities().getRequest().getGetMap().getFormats();
+        this.format = preferredFormat;
+        // verify if preferred Format is supported else fallback to default functionality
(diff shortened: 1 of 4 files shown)


nmco:
The QA failure is valid, related with code formatting:
```
[INFO] Web Map Service Module ............................. FAILURE [02:02 min]
```

Please fix this @imranrajjad :smile:

imranrajjad:
@nmco and @aaime, please review and merge

nmco:
The PR description looks like a few notes that we need to figure out how to merge together, a more constructive description would be appreciated. IMHO the issue should describe the problem, and if the issue doesn't contain it already, the PR description shoudl quickly summarize the intended fix or link to any relevant ML discussion. This would be really helpful :smiley:.

modules/extension/wms/src/main/java/org/geotools/ows/wms/map/WMSCoverageReader.java
```diff
@@ -133,6 +133,32 @@ public WMSCoverageReader(WebMapServer wms, Layer layer, String style) {
...
     }
 
+    public WMSCoverageReader(WebMapServer wms, Layer layer, String style, String preferredFormat) {
+        this.wms = wms;
+
+        // init the reader
+        addLayer(layer, style);
+
+        // best guess at the format with a preference for PNG (since it's normally transparent)
+        List<String> formats = wms.getCapabilities().getRequest().getGetMap().getFormats();
```
nmco: The comment looks out of context. You are basically getting the supported formats form the remote and then cheeking that the pass by format is supported.

modules/extension/wms/src/main/java/org/geotools/ows/wms/map/WMSLayer.java
```diff
@@ -88,6 +88,18 @@ public WMSLayer(WebMapServer wms, Layer layer, String style) {
...
     }
 
+    /**
+     * Builds a new WMS layer
+     *
+     * @param wms
+     * @param layer
+     * @param style
+     * @param format
+     */
```
nmco: Yes that's what a constructor usually do, this Javadoc doesn't add any valuable information.

modules/extension/wms/src/test/resources/org/geotools/ows/wms/map/caps130.xml
```diff
@@ -23,6 +23,7 @@
       </GetCapabilities>
       <GetMap>
         <Format>image/png</Format>
```
nmco: What will happen if `ìmage/png` is not supported by the cascaded server?

modules/extension/wms/src/main/java/org/geotools/ows/wms/map/WMSCoverageReader.java
```diff
@@ -133,6 +133,32 @@ public WMSCoverageReader(WebMapServer wms, Layer layer, String style) {
...
+        this.format = preferredFormat;
+        // verify if prefferedFormat is supported
+        if (!formats.contains(preferredFormat)) {
+            // if prefereded format is not supported default to first available
+            LOGGER.warning(preferredFormat + " format is not supported, falling back to defaults");
+            for (String format : formats) {
```
nmco: This code is an exact copy paste of the code just 20 lines above, can we avoid this :fearful: ?

nmco:
A more general comment, I can definitely see this as an improvement, but I can also foreseen future improvements, for example, like passing the CRS the caller is interested in. If we follow the current path of adding separate elements to the constructor ... well we may end up with a significant number of constructors.

Why no creating a parameters aggregator class? MapSpec? RequestParams? Can you have a look at GeoServer WMS implementation for ideas?

modules/extension/wms/src/main/java/org/geotools/ows/wms/map/WMSCoverageReader.java
imranrajjad: sure I will move this code into a utility function

modules/extension/wms/src/main/java/org/geotools/ows/wms/map/WMSLayer.java
imranrajjad: will remove it

modules/extension/wms/src/test/resources/org/geotools/ows/wms/map/caps130.xml
imranrajjad: the current implementation assumes that the cascaded server will provide image/png as a bare mimimum. With the currently available constructor is if `image/png` is not supported then the WMS requests will probably go incomplete, shall I throw an Exception for this?

imranrajjad: The previous implementation expects `ìmage/png` as bare minimum to be available, thats why it always defaults to `image/png`

modules/extension/wms/src/main/java/org/geotools/ows/wms/map/WMSCoverageReader.java
imranrajjad: fixed it

imranrajjad:
@nmco comments addressed

imranrajjad:
@nmco can this be merged?

nmco:
Looks good, thx.

PR #2532, opened 2019-08-23
Updates to the release guide based on 22-RC release

tbarsballe:
* Updated path to `org/geotools/util/factory/GeoTools.java`
* Added step to create release series folder in sourceforge (SF doesn't let us do this via ssh anymore)
* Added sed commands for updating stable/latest links in readme/docs

These changes are really only applicable for RC releases, so will next be relevant for the 23-RC.


aaime:
Looks like there is a legit build failure in the docs:

```
     [exec] Warning, treated as error:
     [exec] /home/travis/build/geotools/geotools/docs/developer/procedures/release.rst:4: WARNING: Duplicate explicit target name: "sourceforge".
```

tbarsballe:
Yep, looks like I'm short an underscore. Should be fixed now.

PR #2541, opened 2019-09-08
Improved the ArcGIS ReST API datastore

lmorandini:


aaime:
20.x is dead, wondering, are these changes already in master too?

PR #2556, opened 2019-09-11
[GEOT-6365] Reset tempValue more consistently to avoid incorrectly retaining whitespace for attribute value

soc:


soc:
The build failure is the same as on the master branch, and doesn't appear to have been caused by this change.

aaime:
Thanks for the fix! Can you add a test to go with it, we normally don't accept fixes without a corresponding test.

soc:
> Can you add a test to go with it, we normally don't accept fixes without a corresponding test.

@aaime Sure! I wasn't sure how to add the attached test case – do you have a preferred naming convention for tests associated with bugs? Or should I simply pick a name like `GMLFilterFeature2Test.java`?

soc:
@aaime Test added! Does this look good to you?

aaime:
@bestrauss can you have a look, fixes and issue caused by your earlier change, you should know that code better than anyone else by now :-D

soc:
@aaime The only error I could find is a JavaDoc-related issue in a file that wasn't changed in this PR:

```
[INFO] --- fmt-maven-plugin:2.4.0:check (default) @ docs ---

[warn] Test source directory '/home/travis/build/geotools/geotools/docs/src/test/java' does not exist, ignoring.

[debug] Using AOSP style

[INFO] Processed 93 files (1 non-complying).

[ERROR] Found 1 non-complying files, failing build

[ERROR] Non complying file: /home/travis/build/geotools/geotools/docs/src/main/java/org/geotools/main/FilterExamples.java
```

soc:
@aaime I rebased an the error seems to be gone. Is this good to go?

soc:
@aaime Any chance to get this merged? It looks like this fix already missed the window for the previous minor release. :-(

soc:
Thank you!

PR #2561, opened 2019-09-16
[Backport 22.x][GEOT-6318] Exception handling on SAX parser entity expansion limit.

fernandor777:
This PR adds exception handling for corner cases having an outdated XercesImpl version on classpath not supporting entity expansion limit property.

### DIFF
--- a/modules/extension/xsd/xsd-core/src/main/java/org/geotools/xsd/Parser.java
+++ b/modules/extension/xsd/xsd-core/src/main/java/org/geotools/xsd/Parser.java
@@ -24,6 +24,7 @@ import java.io.Reader;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Optional;
+import java.util.logging.Logger;
 import javax.xml.namespace.QName;
 import javax.xml.parsers.ParserConfigurationException;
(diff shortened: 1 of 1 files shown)


modules/extension/xsd/xsd-core/src/main/java/org/geotools/xsd/Parser.java
```diff
@@ -503,16 +510,29 @@ protected SAXParser parser(boolean validate) throws ParserConfigurationException
...
+                    JDK_ENTITY_EXPANSION_LIMIT,
+                    entityExpansionLimit != null
+                            ? entityExpansionLimit
+                            : DEFAULT_ENTITY_EXPANSION_LIMIT);
+        } catch (SAXNotRecognizedException ex) {
+            LOGGER.warning(
+                    "Sax parser property '"
+                            + JDK_ENTITY_EXPANSION_LIMIT
+                            + "' not recognized.  "
+                            + "Please check if XercesImpl jar dependency must be updated.");
```
aaime: Could actually be newer and require a different property. Better to just say that the xerces version is incompatible.

aaime:
Better thanks, +1

PR #2568, opened 2019-09-20
[GEOT-6372] Misaligned tile contents when rendering heterogenenous CRS mosaics, with reprojection, APH enabled, and high oversampling (21.x)

aaime:


dromagnoli (approved):
Looks good to me.

PR #2571, opened 2019-09-22
[GEOT-6396] Avoid preloading GML schema on GML object construction

aaime:
@nmco can you have a look?

Removing this eager loading speeds up single test runs in the IDE by over a second (reference, GetMapIntegrationTest, since WMS unfortunately depends on WFS which in turn loads this), and I could not get a single test to fail, not even in app-schema integration. 

So... maybe the original motivation for this code is just outdated?


nmco:
Looks like the original motivation was to fix some cyclic, that say, since no unit tests and no App-Schema integrations tests fails (I didn't tun them), then I would say go ahead and merge.

@aaime I would suggest to create a JIRA ticket for this, just to make it easier to keep track of, I hope we will not need it, but we may need tor revert if someone is hit whit that cyclic dependency again.

aaime:
Ticket created and referenced from commit message

PR #2577, opened 2019-09-24
[GEOT-6378] Remove jdom dependency from main and shapefile modules

groldan:
there's a direct dependency on the jdom library
from main and shapefile, which adds nothing that
can't be done with official XML API's, hence
drawing the dependency unnnecessary.

This patch removes such dependency and replaces the
XML parsing done with it by the StAX API which is
faster and is part of the default Java library.

For instance, the following classes have been adapted:

* org.geotools.data.ows.ServiceExceptionParser
* org.geotools.data.shapefile.shp.xml.ShpXmlFileReader

### DIFF
--- a/modules/extension/app-schema/app-schema/pom.xml
+++ b/modules/extension/app-schema/app-schema/pom.xml
@@ -53,6 +53,10 @@
             <classifier>tests</classifier>
             <scope>test</scope>
         </dependency>
+       <dependency>
+           <groupId>org.jdom</groupId>
+           <artifactId>jdom2</artifactId>
+       </dependency>
         <dependency>
             <groupId>commons-digester</groupId>
             <artifactId>commons-digester</artifactId>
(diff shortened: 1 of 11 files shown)


groldan:
The following classes also depend on JDOM transitively:
```
modules/extension/app-schema/app-schema/src/main/java/org/geotools/data/complex/xml/XmlXpathFilterData.java
modules/extension/app-schema/app-schema/src/main/java/org/geotools/data/complex/xml/XmlResponse.java
modules/extension/app-schema/app-schema/src/main/java/org/geotools/appschema/util/XmlXpathUtilites.java
modules/plugin/coverage-multidim/netcdf/src/main/java/org/geotools/coverage/io/netcdf/tools/CreateIndexer.java
modules/plugin/geotiff/src/main/java/org/geotools/gce/geotiff/GeoTiffWriter.java
modules/library/coverage/src/main/java/org/geotools/coverage/grid/io/imageio/geotiff/GeoTiffIIOMetadataEncoder.java
```

groldan:
* `modules/library/main`: Removed dependency on JDOM from 
* `modules/plugin/shapefile`: Removed dependency on JDOM from 
* `modules/extension/app-schema/app-schema`: Declared explicit dependency on JDOM, instead of depending transitively from main
* `modules/plugin/coverage-multidim/netcdf`: Declared explicit dependency on JDOM, depended transitively from `edu.ucar:cdm` and `main`

aaime:
The replacement code is a lower level replacement, ugly to look at, but it's indeed one less dependency . Merging.

PR #2578, opened 2019-09-25
[GEOT-6397] Upgrading jgridshift to version 1.2

aaime:


mprins:
this actually fixes [GEOT-6379](https://osgeo-org.atlassian.net/projects/GEOT/issues/GEOT-6379) for master (not GEOT-6397)

PR #2580, opened 2019-09-26
[GEOT-6383] Allow GranuleSource users to get a file oriented view, rather than a slice oriented one

aaime:
This comes with a sequence of other features and improvements required to make the file oriented view extraction actually work.


dromagnoli (approved):
Looks good to me.

PR #2581, opened 2019-09-26
gt-imagemosaic/testHarvestHeteroUTM unit test fails

dmitri-sirobokov:
Set AbsolutePath property of the Indexer to true. That will help to pass unit test when the code is buld and run from D-drive.  

Default relative path behaviour does not work with different drives (Temporary directory with test data is copied to C: drive).

forum message:
https://sourceforge.net/p/geotools/mailman/message/36771990/


aaime:
Seems like a good solution, thanks. Merging.

PR #2582, opened 2019-09-27
GEOT-6018: WMTS doesn't correctly handle `image/jpeg` format

dmitri-sirobokov:
Default mime type for WMTS request is chosen from layer capabilities, instead of hard-coded 'png'.

This was previously working for REST request. Now the code is modified to have the same behaviour for KVK requests.


ianturton:
Looks good, once we have a CLA on record we can merge it - https://docs.geotools.org/latest/developer/procedures/contribute.html

dmitri-sirobokov:
CLA was signed and sent on 19 july 2019.

PR #2591, opened 2019-10-03
GEOT-6002, custom HttpClient in WMTSTile requests

dmitri-sirobokov:
Using custom injected HttpClient in WMTSTile class allows for more configurations. For example HttpClient can be externally configured to use credentials for Basic Authentication.


dmitri-sirobokov:
Tested the usage of it in GeoServer. In my case I needed to access WMTS server which requires Basic Authentication.

The solution was to inject the instance of HttpClient into WMTSTile class and customize it in WebMapTileServer class with the same credentials which are used for GetCapabilities request.

There are a number of overloaded constructors which I am not fun of, but they are needed to keep existing code backward compatible. Not sure whether to mark it Depricated or not.

Also it turns out that WebMapTileServer is dependent on two different HttpClient implementations. Would be nice to merge them later into one.

aaime:
Had a quick look, I believe as written it cannot be accepted. Geotools code should depend only on the https://github.com/geotools/geotools/blob/865cc72a90b07841a4c871d1b07fcc8e5defd35b/modules/library/main/src/main/java/org/geotools/data/ows/HTTPClient.java interface, allowing users to decide which implementation to use. 
Both implementation of it allow for basic authentication already.

dmitri-sirobokov:
> Had a quick look, I believe as written it cannot be accepted. Geotools code should depend only on the https://github.com/geotools/geotools/blob/865cc72a90b07841a4c871d1b07fcc8e5defd35b/modules/library/main/src/main/java/org/geotools/data/ows/HTTPClient.java interface, allowing users to decide which implementation to use.
> Both implementation of it allow for basic authentication already.

That was my first try. The old WMTSTile (which was already dependent on other HttpClient) sets custom headers in the 'setupInputStream' method. I  could not figure out an easy way to do it with HTTPClient without refactoring more code. But I can give it a second try.

dmitri-sirobokov:
To be able to set additional headers in requests using HTTPClient interface, we need to add additional method with the following signature:

get(URL url, Map<String, String> headers)

However this is a breaking change, all classes that implement this interface have to add this implementation. Is it acceptable? Any thoughts about this, @aaime ?

aaime:
There are already methods to set basic authentication. Do you want to add different headers?

dmitri-sirobokov:
@aaime. I don't need different headers, but existing implementation does something with custom headers. See `WMTSTile.setupInputStream()` method. I assume this functionality has to be kept.

aaime:
Got it now (sorry, was not involved in the WMTS client developement). No, you should not break the HttpClient interface, but nothing prevents you from extending it in a non breaking way. Just add the method you suggested, with a default implementation (Java 8+) that delegates to the previous method. 
Then have the actual implementations of HTTPClient override it and actually use the extra headers (plus the auth ones).

dmitri-sirobokov:
Modified the code. Common `HttpClient` dependency is replaced by abstract `org.geotools.data.ows.HTTPClient`.

aaime:
Looks good, merging

aaime:
Btw, this is a new feature so it would not be normally backported, but you can ask on the gt-devel list about it, the changes do not look disrutpive to me, but it's not just my call.

dmitri-sirobokov:
Thank you Andrea.

PR #2601, opened 2019-10-08
[GEOT-6397] provide extra comparison method for featureTypes in DataUtilities class

nprigour:
see also #2525 for the reason for this addition. Existing functionality is not affected. Additional comparison behavior is added.

### DIFF
--- a/modules/library/main/src/main/java/org/geotools/data/DataUtilities.java
+++ b/modules/library/main/src/main/java/org/geotools/data/DataUtilities.java
@@ -377,6 +377,48 @@ public class DataUtilities {
      * Compare attribute coverage between two feature types (allowing the identification of
      * subTypes).
      *
+     * <p>Strict compatibility is assumed meaning that both the local name and java binding are
+     * compatible (see {@link #compareInternal(SimpleFeatureType, SimpleFeatureType, boolean)} for
+     * more details.
+     *
+     * <p>Namespace is not considered in this operations. You may still need to reType to get the
+     * correct namespace, or reorder.
+     *
+     * <p>Please note this method will not result in a stable sort if used in a {@link Comparator}
+     * as -1 is used to indicate incompatiblity (rather than simply "before").
+     *
+     * @param typeA FeatureType being compared
+     * @param typeB FeatureType being compared against
+     */
+    public static int compare(SimpleFeatureType typeA, SimpleFeatureType typeB) {
+        return compareInternal(typeA, typeB, true);
+    }
+
+    /**
+     * Compare attribute coverage between two feature types (allowing the identification of
+     * subTypes).
+     *
+     * <p>loose compatibility is assumed based on local name and attribute descriptor class (java
+     * binding may differ) (see {@link #compareInternal(SimpleFeatureType, SimpleFeatureType,
+     * boolean)} for more details.
+     *
+     * <p>Namespace is not considered in this operations. You may still need to reType to get the
+     * correct namespace, or reorder.
+     *
+     * <p>Please note this method will not result in a stable sort if used in a {@link Comparator}
+     * as -1 is used to indicate incompatiblity (rather than simply "before").
+     *
+     * @param typeA FeatureType being compared
+     * @param typeB FeatureType being compared against
+     */
+    public static int compareNames(SimpleFeatureType typeA, SimpleFeatureType typeB) {
+        return compareInternal(typeA, typeB, false);
+    }
+
+    /**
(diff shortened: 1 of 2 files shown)


nprigour:
Reason of AppVeyor failure seems irrelevant to the proposed improvement

aaime:
Yes, it seems the appveyor build just died mid flight?

modules/library/main/src/main/java/org/geotools/data/DataUtilities.java
```diff
@@ -469,9 +531,15 @@ public static boolean isMatch(AttributeDescriptor a, AttributeDescriptor b) {
...
-        if (a.getLocalName().equals(b.getLocalName())
-                && a.getType().getBinding().equals(b.getType().getBinding())) {
-            return true;
+        if (strict) {
+            if (a.getLocalName().equals(b.getLocalName())
+                    && a.getType().getBinding().equals(b.getType().getBinding())) {
+                return true;
+            }
+        } else {
+            if (a.getLocalName().equals(b.getLocalName()) && a.getClass().equals(b.getClass())) {
```
aaime: I'm not sure why we want to compare the descriptor classes, outside of maybe backwards compatibility with what looked like a simple coding mistake (to me)... AttributeDescriptor is an interface, two different implementations for it should effectively behave the same way. If what you're looking for is a comparison that only accounts for name, I'd limit it to it.

aaime: Ah, now I see the test... so the comparison would also fail between a common attribute vs a geometry attribute, AttributeDescriptor vs GeometryDescriptor. Uh... still wondering if this is useful though, do you have a specific use case where this happens? I guess you must have one that prompted this pull request to start with :-D

nprigour: The motivation behind the present PR should by sought to the enhancement introduced recently in udig via https://github.com/locationtech/udig-platform/pull/218 and the ability provided by eclipse framework to define dynamic adapters that adapt/transform between java types essentially without limitations.
The changes introduced by GEOT-4931 and its backporting will render this feature useless and we need to fully rewrite it while by adding the extra proposed comparison method will require only minor modifications (only shifting to use the new compareNames method)

aaime: Ok, and is the comparison of descriptor class necessary for that (sorry, I have no time to read into the udig ticket)? Method says "compareNames" but in reality it's comparing names and descriptor classes

nprigour: Provides for an extra level of 'logical' safety distinguishing between geometry filelds and standard attribute fields. If we have a GeometryDescriptor with name 'name1' in a shapefile and an AttributeDescriptor with the same name in another file this featuresTypes should definately not match and no adaptation is possible.
If method naming is the problem then propose a more convenient name and I will accept it :)

aaime: Right but it should not distinguish between two impelmentations of "normal attributes" imho... if you don't really need it, I'd suggest to remove it and be true to a simple method name "compareNames", as opposed to a "compareNamesAndDescriptorClass". If it's hard to explain, unless there is a very good reason to go there, avoid it.

nprigour: I would prefer to be backward compatible and in general I do not like not to be able to distingush between an AttributeDescriptor,  GeometryDescriptor etc but you may be right in the sense that since the name of the method is compareNames() we must do just that.
Therefore, I would modify line 540 and correct appropriately the javadoc comments.

nprigour:
@aaime  I have proceed with the needed simplifications so that compareNames() does exactly what its name implies.
If you accept and merge it I would request also backporting to 22.x & 21.x (or at least some insight on how to do it)

aaime:
Merged and added backport labels, the backport bot should be creating the PRs automatically. If it fails to create one due to conflicts, you'll have to look at it yourself.

jodygarnett:
We missed documenting this new strict parameter.

nprigour:
Hi @jodygarnett,
Do you mean in the following url: https://docs.geotools.org/latest/userguide/library/main/data.html
Actually I have never touched this documentation. Should we create a separate PR fo this?

jodygarnett:
Yup, I try and make sure docs get updated when API changes. Especially for these facade classes.

PR #2602, opened 2019-10-09
[GEOT-6399] Arithmetic operations not supported with Complex Attributes

imranrajjad:
-added support for xpath numerical attributes for arithmetic operators
-integration tests added in app_schema

JIRA ticket : https://osgeo-org.atlassian.net/browse/GEOT-6399

### DIFF
--- a/modules/extension/app-schema/app-schema/src/test/java/org/geotools/appschema/filter/ArithmeticFilterTest.java
+++ b/modules/extension/app-schema/app-schema/src/test/java/org/geotools/appschema/filter/ArithmeticFilterTest.java
@@ -79,7 +79,7 @@ public class ArithmeticFilterTest extends AppSchemaTestSupport {
         // evaluating first feature with Location accuracy value = 200
 
         // 200 x 2 = 400
-        Filter arthimeticMultiplyFilter =
+        Filter arithmeticMultiplyFilter =
                 ff.equals(
                         ff.multiply(
                                 ff.property(
@@ -88,7 +88,7 @@ public class ArithmeticFilterTest extends AppSchemaTestSupport {
                         ff.literal(400));
 
         // 200 / 2 = 100
-        Filter arthimeticDivideFilter =
+        Filter arithmeticDivideFilter =
                 ff.equals(
                         ff.divide(
                                 ff.property(
@@ -97,7 +97,7 @@ public class ArithmeticFilterTest extends AppSchemaTestSupport {
                         ff.literal(100));
 
         // 200 + 100 = 300
-        Filter arthimeticAdditionFilter =
+        Filter arithmeticAdditionFilter =
                 ff.equals(
                         ff.add(
                                 ff.property(
@@ -106,7 +106,7 @@ public class ArithmeticFilterTest extends AppSchemaTestSupport {
                         ff.literal(300));
 
         // 200 - 100 = 100
-        Filter arthimeticSubtractionFilter =
+        Filter arithmeticSubtractionFilter =
                 ff.equals(
                         ff.subtract(
                                 ff.property(
@@ -126,10 +126,10 @@ public class ArithmeticFilterTest extends AppSchemaTestSupport {
             Property val = (Property) positionalAccuracy.evaluate(f);
             System.out.println(val.getValue());
             // try all filters
-            assertTrue(arthimeticMultiplyFilter.evaluate(f));
-            assertTrue(arthimeticDivideFilter.evaluate(f));
-            assertTrue(arthimeticAdditionFilter.evaluate(f));
-            assertTrue(arthimeticSubtractionFilter.evaluate(f));
+            assertTrue(arithmeticMultiplyFilter.evaluate(f));
+            assertTrue(arithmeticDivideFilter.evaluate(f));
+            assertTrue(arithmeticAdditionFilter.evaluate(f));
+            assertTrue(arithmeticSubtractionFilter.evaluate(f));
 
         } finally {
             iterator.close();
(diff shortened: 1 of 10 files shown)


modules/library/main/src/main/java/org/geotools/filter/Filters.java
```diff
@@ -412,6 +414,17 @@ public static double number(Object value) {
                 throw new IllegalArgumentException("Unable to decode '" + text + "' as a number");
             }
         }
+        // if value is coming from ComplexAttribute through Xpath
+        if (value instanceof ComplexAttribute) {
```
aaime: Why does it even get there, the calling code should evaluate the expression/attribute before getting here.

imranrajjad: @aaime thanks for the review, so I am guessing it should be processed outside number method, where we can ensure whatever gets into this method will always be a number. Is that acceptable.?

imranrajjad:
@aaime addressed requested changes, Expressions will be evaluated before Filters.number is called.

aaime:
Looks better, but it's still ... odd. I had a look around and tried to figure out what was going on.
It seems that in 2011 support for handling complex values was added to the FilterAbstract and all its subclasses, via the eval(Expression, Object) method, which eventually calls onto unpack.

The move you did is not bad, but could be more complete and support the whole expression hierachy instead. 

Option 1:
- Copy eval(Expression, Object) to the Filters class, and delegate the existing method to it (instead 
- Add a delegating call to ExpressionAbstract too
- Use it in subclases to evaluate values, both in math expressions, but also, for example, in FunctionExpressionImpl.  Handling of classification functions is harder, I'd leave it to another ticket.

Option 2: deal with the unpacking madness at the source, in the relevant PropertyAccessor.
If I look at FeaturePropertyAccessor (which I believe it's the source of the issue, but it could as well be something else) I see it's completely ignoring the target class, and likely returning a ComplexAttribute when, say, a Number or a String was requested.
In this option, if the target class is set to anything other than Object, ComplexAttribute or null, I'd try to unpack the values and convert the to the requested target type.

@nmco what do you think?

imranrajjad:
the subtle changes in second revision take inspiration from Option 1. By passing the intended class to evaluate method expression.evaluate(feature,<expected class>), we were able to obtain the numerical value nested inside Complex Feature attribute

nmco:
So option 1 it is, what do you think from this approach @aaime ?

imranrajjad:
@aaime @nmco 

added list supports, the expression will return a list in which each item will have passed through desired arithmetic operation.

if both expressions are lists, arithmetic operator will done on corresponding elements. For lists with un-equal size, the larger list will perform arithmetic operation as long as it can find a corresponding element, when the index has passed size of smaller list, unmodified element from larger list will be copied to result list.

modules/extension/app-schema/app-schema/src/test/java/org/geotools/appschema/filter/ArithmeticFilterTest.java
```diff
@@ -0,0 +1,138 @@
...
+        fSource = (FeatureSource<FeatureType, Feature>) dataAccess.getFeatureSource(FEATURE_TYPE);
+    }
+
+    @Test
+    public void testWithArithmeticOperator() throws IOException {
+
+        // evaluating first feature with Location accuracy value = 200
+
+        // 200 x 2 = 400
+        Filter arthimeticMultiplyFilter =
```
aaime: arthimetic ;-) (see other variables too)

modules/library/main/src/main/java/org/geotools/filter/MathExpressionImpl.java
```diff
@@ -109,4 +121,110 @@ protected Object number(double number) {
...
+                    .stream()
+                    .map(
+                            n ->
+                                    doArithmeticOperation(
+                                            n.doubleValue(), scalar.doubleValue(), operator))
+                    .collect(Collectors.toList());
+        }
+    }
+
+    private Double doArithmeticOperation(Double operand1, Double operand2, Operator operator) {
```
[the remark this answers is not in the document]
imranrajjad: understood, will delegate to subclass

modules/library/main/src/main/java/org/geotools/filter/MathExpressionImpl.java
```diff
@@ -109,4 +121,110 @@ protected Object number(double number) {
...
+
+    protected Object getValue(org.opengis.filter.expression.Expression expression, Object feature) {
+        Object evaluation = expression.evaluate(feature);
+        return evaluation;
+    }
+
+    protected Object handleCollection(Object value1, Object value2, Operator operator) {
+        // case 1, both are collections
+        if (value1 instanceof Collection && value2 instanceof Collection) {
+            Collection<Number> result = new ArrayList<Number>();
```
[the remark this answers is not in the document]
imranrajjad: understood, will remove

modules/extension/app-schema/app-schema/src/test/java/org/geotools/appschema/filter/ArithmeticFilterTest.java
imranrajjad: will fix spelling typos

imranrajjad:
@aaime addressed requested changes, hopefully this should be it :)

modules/extension/app-schema/app-schema/src/test/java/org/geotools/appschema/filter/ArithmeticFilterTest.java
aaime: Spelling errors still seem to be here?

imranrajjad: missed those, will fix

imranrajjad:
22.x backport PR https://github.com/geotools/geotools/pull/2616

PR #2603, opened 2019-10-10
[GEOT-6400] Implement support for text labels in ColorMapEntry for CSS

taba90:
This pr has a related jira ticket: https://osgeo-org.atlassian.net/browse/GEOT-6400

### DIFF
--- a/modules/unsupported/css/src/main/java/org/geotools/styling/css/CssTranslator.java
+++ b/modules/unsupported/css/src/main/java/org/geotools/styling/css/CssTranslator.java
@@ -1424,7 +1424,7 @@ public class CssTranslator {
                             throw new IllegalArgumentException(
                                     "Invalid color map content, it must be a color-map-entry function"
                                             + entry);
-                        } else if (f.parameters.size() < 2 || f.parameters.size() > 3) {
+                        } else if (f.parameters.size() < 2 || f.parameters.size() > 4) {
                             throw new IllegalArgumentException(
                                     "Invalid color map content, it must be a color-map-entry function "
                                             + "with either 2 parameters (color and value) or 3 parameters "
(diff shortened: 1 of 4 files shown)


modules/library/main/src/main/java/org/geotools/styling/ColorMapEntry.java
```diff
@@ -34,10 +34,10 @@
  */
 public interface ColorMapEntry {
     /** Label for this Color Map Entry */
-    String getLabel();
+    Expression getLabel();
```
aaime: Big nope, this is a core API change, it should have its own proposal and would make this PR non backportable. If you need expressions instead of literals, it can be done, but make it a separate commit, after issuing a proposal and getting it voted.

taba90: brought back to string

aaime:
LGMT!

modules/unsupported/css/src/main/java/org/geotools/styling/css/CssTranslator.java
```diff
@@ -1424,7 +1424,7 @@ private void addRasterSymbolizer(CssRule cssRule, RuleBuilder ruleBuilder) {
                             throw new IllegalArgumentException(
                                     "Invalid color map content, it must be a color-map-entry function"
                                             + entry);
-                        } else if (f.parameters.size() < 2 || f.parameters.size() > 3) {
+                        } else if (f.parameters.size() < 2 || f.parameters.size() > 4) {
                             throw new IllegalArgumentException(
                                     "Invalid color map content, it must be a color-map-entry function "
                                             + "with either 2 parameters (color and value) or 3 parameters "
```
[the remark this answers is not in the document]
aaime: Agreed, good catch.

PR #2610, opened 2019-10-14
[GEOT-6370]  Thread safe XSD access API change.

fernandor777:
This PR makes XSD access thread safe , using the XSDParticle API instead DOM for the same result on the original PR.


aaime:
Been running GetFeatureTest.testConcurrentGet and GetFeatureTest.getConcurrentPost with pumped up settings, 32 threads, 20k requests, no failures. This seems indeed to be doing the trick :-)

PR #2615, opened 2019-10-16
[GEOT-6401] Support OGC Instant dates in ECQL

peterhuffer:
Adds support when converting an OGC Filter containing an Instant to ECQL.

~Sorry about the test. I couldn't get it to run locally so I just pushed it up as is for now. But will fix if this is the correct change.~

### DIFF
--- a/docs/user/library/metadata/convert.rst
+++ b/docs/user/library/metadata/convert.rst
@@ -61,4 +61,4 @@ Using the ``Converters`` utility class this becomes::
 The ``Converters`` are the technology behind our great support for the Filter API Expression::
   
    Expression expr = ff.literal("#FF0000")
-   Color color = expr.evaualte( null, Color.class );
+   Color color = expr.evaluate( null, Color.class );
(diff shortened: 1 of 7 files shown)


peterhuffer:
Does this make sense to do? Currently if a Filter looks like 
`[ testAttribute After Instant{position:Position{Wed Oct 09 14:49:51 MST 2019}} ]`
it will be translated into a CQL expression like
`testAttribute AFTER 'Instant{position:Position{Wed Oct 09 14:49:51 MST 2019}}'
which will not be translatable back to a Filter

aaime:
Can you please open a ticket in Jira and refer to it from the commit message? And yes, filters should be able to round-trip, if not to the exact same filter (would be best), at least to one that evaluates in the same way.

peterhuffer:
@aaime I linked the ticket in the commit message.

I also took a recommendation to use the Converters class that was available to handle the Instant and Date cases.

peterhuffer:
Anymore input on this?

aaime:
It's looking good, merging. Seems a simple change too, does it need backport?

peterhuffer:
> It's looking good, merging. Seems a simple change too, does it need backport?

Probably. We currently use 19.x in our library. Do I need to open a separate PR for each backport?

aaime:
@peterhuffer we have a bot that can be instructed to build backports... however 19.x is dead, it won't be released any longer, and we don't commit there. Active branches now are 21.x and 22.x

PR #2617, opened 2019-10-17
[GEOT-6406] Fix NetCDF/GRIB wrong dimension order for time+custom dimensions

tbarsballe:
Fix for https://osgeo-org.atlassian.net/browse/GEOT-6406

As far as I can tell, the dimensions the NetCDF reader gets from the slice2DIndex are already in COARDS order ([additional dims], T, Z, Y, X), and it is incorrectly reversing them.

Test data is harvested from https://ftp.nhc.ncep.noaa.gov/toa/ with some minor post-processing to get around https://osgeo-org.atlassian.net/browse/GEOT-6340

Note that while the test case is in the gt-grib module (and this error was encountered with grib2 data), the actual fix is in gt-netcdf


dromagnoli:
Looks good to me.
Wondering if we may exclude the automatically generated ncx3 and gbx9 files but that's not a big deal :)

tbarsballe:
Good point, I should have done that in the first place. Updated.

tbarsballe:
Ah sorry, I think I missed your earlier timezone question.
I've just run through all the GeoTools test cases for NetCDF & GRIB , hoping there was one that tested such a core functionality as that. Given that this change didn't break any of them, either they are woefully inadequate or everything it fine.

I don't have a T+Z NetCDF dataset on hand, but if you know of a good one you can point me to, I'd be happy to test it out.

I am likewise surprised this has been working just fine for the past while (although I did find a message in the user list describing a similar error from a couple years ago - https://sourceforge.net/p/geoserver/mailman/message/36787489/ )

tbarsballe:
Alright, I tried that dataset out, and discovered that it is not an adequate test as the time and elevation dimensions are both the same size (2 total values each).
The code I changed is only used for range checks (i.e. is the requested dimension within the range of available dimensions)


I've tried out that dataset, and it works fine. However, I'm not sure it is an adequate test of the change.
In all cases, the actual band(s) to fetch is determined using:
```
List<Integer> indexes = request.source.reader.getImageIndex(query)
```
somewhat earlier in execution. This behaviour hasn't changed.

The loop that is modified by this PR is only used to do a range check on the data, making sure each dimension is within the declared bounds. Since the sample data has the same number of time and elevation values (2 each), this range check will succeed regardless of whether they are in the right order.
T=2012-04-01T00:00:00.000Z, 2012-04-01T01:00:00.000Z
Z=10.0, 450.0

So, I'm still not certain if it was just coincidence that this was working before, or what.
I've tried looking around for more sample data, and haven't had much luck finding any that has both time and elevation data with a different  number of slices for each.

I did find this data set: https://www.unidata.ucar.edu/software/netcdf/examples/rhum.2003.nc
This is from the sample COARDS Conventions data sets, so it seems applicable - See https://www.unidata.ucar.edu/software/netcdf/examples/files.html

The dataset has a time dimension and a "LEVEL" dimension, and that seemed to work fine with my fix (testing both the min and max values of each dimension).
I will note that despite GeoServer identifying LEVEL as a custom dimension, the NetCDF reader did appear to be treating it as an elevation dimension, in that it came after time in the ranges list.

Trying the same data set out on a GeoServer version which didn't have this fix applied failed with a null pointer exception. So, at the very least this fix does seem to allow GeoServer to support more NetCDF files than it did before. I'm still not 100% certain whether or not it breaks a full time + elevation file.

dromagnoli:
Hi @tbarsballe.
Thanks for your effort on this.
I think that the modified piece of code is actually preparing the section's ranges to access a specific 2D slice of the 4D NetCDF matrix, by doing a ranges.add(first, last, stride) on each dimension.
https://www.unidata.ucar.edu/software/netcdf-java/v4.3/v4.2/javadoc/ucar/ma2/Section.html
This is why I provided my initial concern about the change that is going to modify the order of setting the ranges.

If you need it, I have created a fake NetCDF with 3 times, 2 elevations and a 2x2 lonlat grid.
[O3-NO2_small.zip](https://github.com/geotools/geotools/files/3754644/O3-NO2_small.zip)

This is the dump of the values (I put some comments on there):
float O3(time=3, z=2, lat=2, lon=2);
 data:
  {
    {   //T0
      {    //Z0
        {1.0, 2.0},
        {3.0, 4.0}
      },
      {    //Z1
        {11.0, 12.0},
        {13.0, 14.0}
      }
    },
    {   //T1
      {    //Z0
        {21.0, 22.0},
        {23.0, 24.0}
      },
      {    //Z1
        {31.0, 32.0},
        {33.0, 34.0}
      }
    },
    {   //T2
      {    //Z0
        {41.0, 42.0},
        {43.0, 44.0}
      },
      {    //Z1 
        {51.0, 52.0},
        {53.0, 54.0}
      }
    }
  }

So that it should be easy to understand even with a getfeatureinfo if we are accessing the proper section. 
Hope this helps.

tbarsballe:
Thanks for the custom NetCDF file. I tried it out both with and without the fix, and it seemed to behave identically both ways.
Not quite certain why it works fine both ways, possibly the files where this problem shows up have something different in their internal format.

dromagnoli:
Well. At this point I would say go on :)

dromagnoli:
I had to revert the fix since a customer's GeoServer instance updated to 2.17.x stopped working right after the update.
This is some sample data which stopped working.
[sampledata.tar.gz](https://github.com/geotools/geotools/files/3778386/sampledata.tar.gz)

tbarsballe:
Good to know, thanks for the update. I'll try and see if I can figure out whats going wrong.
Did you revert 2.16 / 2.15 backports of this as well, or does that still need to be done?

dromagnoli:
Sorry, also pasting here the message I put on the reverting PR:
------------------
I'm sorry @tbarsballe
I had to revert your fix (and backports too).
Looks like a GeoServer of a customer, based on nightly 2.17.x getting updated every day, stopped properly reading NetCDFs the day after the merge of that fix.
Checking the logs, it looks like a section parsing indeed.
I'm attaching the log for reference:

NetCDFWrongSections.txt

We tried to deploy the gt-netcdf jar of the day before the fix and GeoServer go back working properly (So I have reverted it).
I will add more comments on the original PR
-----------

Already reverted the backports too, thx

PR #2630, opened 2019-10-26
GEOT-6411, Have AbstractGridCoverageWriter release both output stream and destination on dispose()

aaime:


dromagnoli:
Looks good to me.

PR #2637, opened 2019-10-28
Revert "[GEOT-6406] Fix NetCDF/GRIB wrong dimension order for time+custom dimensions"

dromagnoli:
Reverts geotools/geotools#2617


dromagnoli:
I'm sorry @tbarsballe 
I had to revert your fix (and backports too).
Looks like a GeoServer of a customer, based on nightly 2.17.x getting updated every day, stopped properly reading NetCDFs the day after the merge of that fix. 
Checking the logs, it looks like a section parsing indeed.
I'm attaching the log for reference:

[NetCDFWrongSections.txt](https://github.com/geotools/geotools/files/3778249/NetCDFWrongSections.txt)

We tried to deploy the gt-netcdf jar of the day before the fix and GeoServer go back working properly (So I have reverted it).
I will add more comments on the original PR

PR #2644, opened 2019-10-29
[GEOT-6394] Mongodb datastore HTTP(s) schema: Fix name extractor from URL

fernandor777:
Fixes an issue with filename extract from path method for MongoDB HTTP(S) schema support.


nmco:
Looks good, thx.

nmco:
Let's wait for the build to pass.

PR #2604, opened 2019-10-10
[GEOT-6293] ContentFeatureStore.getReader(Query) returns incorrect results when canTransact() returns false.

nprigour:
fix related to https://osgeo-org.atlassian.net/projects/GEOT/issues/GEOT-6293 with a test case
 (see also https://github.com/locationtech/udig-platform/issues/348)

Signed-off-by: Nikolaos Pringouris <nprigour@gmail.com>

PR #2645, opened 2019-10-30
[GEOT-6293] ContentFeatureStore.getReader(Query) returns incorrect results when canTransact() returns false

nprigour:
recreation of #2604 PR to address issue with direct casting
 

aaime:
Ok, let's give this one a try

PR #2648, opened 2019-11-05
[GEOT-6409]: TimeParser refactoring and improvement

dromagnoli:


modules/library/metadata/src/main/java/org/geotools/util/DateTimeParser.java
```diff
@@ -0,0 +1,634 @@
...
+ * formatted according ISO-8601 standard.
+ *
+ * @author Cedric Briancon
+ * @author Martin Desruisseaux
+ * @author Simone Giannecchini, GeoSolutions SAS
+ * @author Jonathan Meyer, Applied Information Sciences, jon@gisjedi.com
+ * @author Daniele Romagnoli, GeoSolutions SAS
+ * @version $Id$
+ */
+public class DateTimeParser {
```
dromagnoli: This class is the merge of code and functionalities from GeoServer gs-ows TimeParser (being now reduced to a simple wrapper as part of https://github.com/geoserver/geoserver/pull/3884/files) and ImageMosaic's TimeParser utility class

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/properties/time/TimestampFileNameExtractor.java
```diff
@@ -20,35 +20,54 @@
...
+            new DateTimeParser(
+                    -1, DEFAULT_TIME_PARSER_FLAGS | DateTimeParser.FLAG_SINGLE_DATE_AS_DATERANGE);
 
     private DateFormat customFormat;
 
+    private String format;
+
+    private boolean inclusive = false;
+
     public TimestampFileNameExtractor(
```
aaime: API change here?

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/properties/time/TimestampFileNameExtractor.java
```diff
@@ -58,13 +77,30 @@ public void setProperties(SimpleFeature feature) {
         final List<Date> dates = new ArrayList<Date>();
         for (String match : getMatches()) {
             // try to convert to date
+
             try {
                 if (customFormat != null) {
                     Date parsed = customFormat.parse(match);
+                    if (inclusive) {
+                        // Special case to make t and z upper case for ISO8601 matching.
+                        format = format.replace("t", "T").replace("z", "Z");
```
aaime: Is this part of the refactor, seems some new functionality?

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/properties/time/TimestampFileNameExtractorSPI.java
```diff
@@ -71,38 +79,36 @@ public PropertiesCollector create(final Object o, final List<String> propertyNam
...
-                        value = tmp.substring(idx + 1);
-                    } else if (tmp.contains("," + FULL_PATH_PREFIX)) {
-                        idx = tmp.indexOf("," + FULL_PATH_PREFIX);
-                        regex = tmp.substring(0, idx);
-                        value = tmp.substring(idx + 1);
-                    } else {
-                        regex = tmp;
-                    }
-                }
+                int minIndex = value.length();
```
aaime: I have no clue what's going on in this block of changes :-D

dromagnoli: Eh :)
that block was looking for parameters (format, fullPath, regex itself) in the parser definition using a convoluted set of if-s and string find.
I have been refactored that by simply putting all of these parameters into a Properties object so it's easier to find them no matter how they are sorted in the String.

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/properties/time/TimestampFileNameExtractor.java
dromagnoli: That was part of the improval of the ImageMosaic Time parser to support the "fuzzy Date search" available in GeoServer's TimeParser which returns the end of a timeRange interval on a reduced accuracy time.
such as a reduced accuracy datetime "2019-11" get translated into the time range 2019-11-01T00:00:00.000 to 2019-11-30T23.59.59.999

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/properties/time/TimestampFileNameExtractor.java
dromagnoli: mmm. actually not. that is a package private class

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/properties/time/TimeParser.java
```diff
@@ -1,266 +0,0 @@
...
-import java.util.Map;
-import java.util.TimeZone;
-
-/**
- * Parses the {@code time} parameter of the request. The date, time and period are expected to be
- * formatted according ISO-8601 standard.
- *
- * @author Simone Giannecchini, GeoSolutions SAS
- */
-public class TimeParser {
```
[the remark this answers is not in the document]
dromagnoli: Checking it. For some reason I didn't notice the "public" qualifier. I think I have confused with another Time*** in ImageMosaic which was package private.
I will check it. The only difference was that the GeoServer TimeParser was returning a Collection with this one returning a List. I'll use some kind of encapsulation probably

dromagnoli:
> The ticket only talks about refactor:
> "Refactor ImageMosaic's TimeParser adopting pieces of code from GeoServer's TimeParser"
> 
> But the changes seem to suggest new functionality is being added along with it?

Part of the change was also improving the ImageMosaic parser using GeoServer's TimeParser parsing capabilities in terms of how it supports timeRange on reduced accuracy time specifications

aaime:
> Part of the change was also improving the ImageMosaic parser using GeoServer's TimeParser parsing capabilities in terms of how it supports timeRange on reduced accuracy time specifications

Then best be clear about it in the ticket (that's actually the main user visible change, the refactor is just how it has been achieved).

dromagnoli:
> Then best be clear about it in the ticket (that's actually the main user visible change, the refactor is just how it has been achieved).

Ticket updated. thanks for the feedback

dromagnoli:
@aaime 
All the feedbacks have been addressed in 
https://github.com/geotools/geotools/pull/2658

Closing this one.

PR #2649, opened 2019-11-05
GEOT-6364 literatefunction doesn't accept empty lists

NielsCharlier:


aaime:
Looks good, merging. Seems a but fix, does it need backport? Can't remeber what version of GeoTools "literate" showed up.

NielsCharlier:
@aaime thanks

PR #2650, opened 2019-11-05
GEOT-6371 gt-complex: app-schema and CSW not compatible

NielsCharlier:


aaime:
Lacks a test.

NielsCharlier:
not sure how to test it? problem only occurs when you run geoserver with both gs-csw (geoserver extension) and gt-app-schema (geotools extension). the fix is in geotools

aaime:
The change seems to fix a data sharing issue, two subsequent instantiation of the FeatureTypeRegistry should not step on each other toes, and add/replace types that were added by the other one, or not? If so, that's what the bug is about in GeoTools, and also what should be tested.

NielsCharlier:
@aaime sorry for the late response. a unit test has been added

thanks for reviewing

aaime:
Looks ok to me, merging.

aaime:
Squash-merged, if you need to backport you can cherry-pick a single commit.

PR #2652, opened 2019-11-08
[GEOT-6416] add extra vendor option FontShrinkSizeMin

nprigour:
An extra SLD vendor option 'FontShrinkSizeMin' to provide more dynamic control of text label font size in polygon geometry types. See also discussions in geotools-dev list [geotools-devel](https://sourceforge.net/p/geotools/mailman/message/36794820/)


modules/library/render/src/main/java/org/geotools/renderer/label/LabelCacheImpl.java
```diff
@@ -1556,15 +1560,50 @@ private boolean paintPolygonLabel(
...
+        int shrinkSize =
+                allowShrinking ? labelItem.getFontShrinkSizeMin() : textStyle.getFont().getSize();
+        int textSize = textStyle.getFont().getSize();
+        // if shrinking is allowed then try to paint polygon label. If no success reduce font size
+        // by 1 unit and retry until fontShrinkSize is reached.
+        while (textSize >= shrinkSize) {
+            tx = new AffineTransform(tempTransform);
+            LabelCacheItem labelItem2 = painter.getLabel();
+            TextStyle2DExt textStyle2 = new TextStyle2DExt(labelItem2);
+            if (labelItem2.getMaxDisplacement() > 0) {
```
aaime: What if the style editor actually wanted a given displacement and anchor? I'd suggest to respect the anchoring....

modules/library/render/src/test/java/org/geotools/renderer/lite/LabelFontShrinkSizeMinTest.java
```diff
@@ -0,0 +1,129 @@
...
+                        null);
+
+        MemoryDataStore data = new MemoryDataStore();
+        data.addFeature(f2);
+        data.addFeature(f1);
+        fs = data.getFeatureSource("labelFontShrinkSizeMin");
+    }
+
+    public void testFonstShrinkSize() throws Exception {
+        Style style = RendererBaseTest.loadStyle(this, "fontShrinkSize/fontShrinkSizeMin.sld");
```
aaime: This file name starts with a lower case, the actual file starts with an uppercase, causing a build failure on Linux:

```
[INFO] Running org.geotools.renderer.lite.LabelFontShrinkSizeMinTest
[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.005 s <<< FAILURE! - in org.geotools.renderer.lite.LabelFontShrinkSizeMinTest
[ERROR] testFonstShrinkSize(org.geotools.renderer.lite.LabelFontShrinkSizeMinTest)  Time elapsed: 0.001 s  <<< ERROR!
java.lang.NullPointerException
	at org.geotools.renderer.lite.LabelFontShrinkSizeMinTest.testFonstShrinkSize(LabelFontShrinkSizeMinTest.java:108)
```

modules/library/render/src/main/java/org/geotools/renderer/label/LabelCacheImpl.java
nprigour: actually this is the was the bahavior so far (check also if statement at lines 1557-1562) that is if a max displacement option is set then anchor point seems to be ignored during label checking and assumed by default as point 0.5, 0.5. I did not touch this behavior at all.
Do you want to change this? Moreover should this be done as part of this PR? It seems that this is a different issue that should be addressed separetely.

modules/library/render/src/test/java/org/geotools/renderer/lite/LabelFontShrinkSizeMinTest.java
nprigour: sld filename changed to be align with the one specified within code

modules/library/render/src/main/java/org/geotools/renderer/label/LabelCacheImpl.java
aaime: Oh ok, I see it now

aaime:
Looking good. Can you please add a line of docs at:
https://github.com/geotools/geotools/blob/26953c26c7b6ee807ca85dfd5374ac098e60805f/docs/user/library/render/style.rst#L557

nprigour:
user docs updated as requested.

If merged would it be possible to backport to 22.x and 21.x?

aaime:
Yes it can be backported, with two cautions:
* A new feature normally stays on master a month before being backported, to have devs and some dev branch demo servers act as guinea pigs (in case of issues, it will be reverted and you can try again fixing the issues that have been found)
* The backporting bot is gone, so you'll have to perform manual backport PRs, sorry!

In the meantime, merging it.

PR #2653, opened 2019-11-08
[GEOT-6417]: Restoring deprecated predicate on ECQL

dromagnoli:


modules/library/cql/src/main/jjtree/ECQLGrammar.jjt
```diff
@@ -138,15 +162,15 @@ TOKEN:
  */
 TOKEN [IGNORE_CASE]:  /* keywords */
 {
-   < AND: "and"> |
+   < AND: "and" | "&&"> |
    < OR: "or"> |
```
aaime: No "||" for or?

modules/library/cql/src/test/java/org/geotools/filter/text/ecql/ECQLComparisonPredicateTest.java
```diff
@@ -57,6 +57,19 @@ public ECQLComparisonPredicateTest() {
         super(Language.ECQL);
     }
 
+    /**
+     * Equals predicate sample
+     *
+     * @throws Exception
+     */
+    @Test
+    public void deprecatedPredicate() throws Exception {
```
aaime: Should test all the added operators  and verify for reach that the result is the expected one

modules/library/cql/src/main/jjtree/ECQLGrammar.jjt
dromagnoli: Nope. it wasn't part of the old CQLGrammar either.
Not sure why :)

modules/library/cql/src/test/java/org/geotools/filter/text/ecql/ECQLComparisonPredicateTest.java
dromagnoli: Ok. I was just using the same tests available on the CQLComparisonPredicateTest.
I'm going to add new tests then.

dromagnoli: @aaime Tests added

aaime:
Looks good thanks

PR #2661, opened 2019-11-12
[GEOT-6377]: added new filter function arrayAnyMatch and a PostGIS en…

mbarto:
…coding using its ANY function

### DIFF
--- a/modules/library/main/src/main/java/org/geotools/filter/function/FilterFunction_arrayAnyMatch.java
+++ b/modules/library/main/src/main/java/org/geotools/filter/function/FilterFunction_arrayAnyMatch.java
@@ -1,0 +1,64 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ *
+ *    (C) 2019, Open Source Geospatial Foundation (OSGeo)
+ *
+ *    This library is free software; you can redistribute it and/or
+ *    modify it under the terms of the GNU Lesser General Public
+ *    License as published by the Free Software Foundation;
+ *    version 2.1 of the License.
+ *
+ *    This library is distributed in the hope that it will be useful,
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ *    Lesser General Public License for more details.
+ */
+package org.geotools.filter.function;
+
+import static org.geotools.filter.capability.FunctionNameImpl.parameter;
+
+import java.util.Arrays;
+import org.geotools.filter.FunctionExpressionImpl;
+import org.geotools.filter.capability.FunctionNameImpl;
+import org.opengis.filter.capability.FunctionName;
+
+public class FilterFunction_arrayAnyMatch extends FunctionExpressionImpl {
+    public static FunctionName NAME =
+            new FunctionNameImpl(
+                    "arrayAnyMatch",
+                    parameter(
+                            "arrayAnyMatch",
+                            Boolean.class,
+                            "Any",
+                            "Returns true if any array value matches candidate."),
+                    parameter("array", Object[].class, "Array", "array of values"),
+                    parameter("candidate", Object.class, "Candidate", "value to match with array"));
+
+    public FilterFunction_arrayAnyMatch() {
+        super(NAME);
+    }
+
+    public Object evaluate(Object feature) {
+        Object[] array;
+        Object candidate;
+
+        try { // attempt to get array and perform conversion
+            array = (Object[]) getExpression(0).evaluate(feature);
+        } catch (Exception e) // probably a type error
+        {
+            throw new IllegalArgumentException(
+                    "Filter Function problem for function between argument #0 - expected type Object[]");
+        }
+
+        try { // attempt to get candidate and perform conversion
+            candidate = (Object) getExpression(1).evaluate(feature);
+        } catch (Exception e) // probably a type error
+        {
+            throw new IllegalArgumentException(
+                    "Filter Function problem for function between argument #1 - expected type Object");
+        }
+
+        return Boolean.valueOf(Arrays.asList(array).contains(candidate));
+    }
+}
--- a/modules/library/main/src/test/java/org/geotools/filter/function/FilterFunction_arrayAnyMatchTest.java
+++ b/modules/library/main/src/test/java/org/geotools/filter/function/FilterFunction_arrayAnyMatchTest.java
@@ -1,0 +1,70 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ *
+ *    (C) 2019, Open Source Geospatial Foundation (OSGeo)
+ *
+ *    This library is free software; you can redistribute it and/or
+ *    modify it under the terms of the GNU Lesser General Public
+ *    License as published by the Free Software Foundation;
+ *    version 2.1 of the License.
+ *
+ *    This library is distributed in the hope that it will be useful,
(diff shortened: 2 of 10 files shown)


modules/library/main/src/main/java/org/geotools/filter/function/FilterFunction_arrayAnyMatch.java
```diff
@@ -0,0 +1,64 @@
...
+package org.geotools.filter.function;
+
+import static org.geotools.filter.capability.FunctionNameImpl.parameter;
+
+import java.util.Arrays;
+import org.geotools.filter.FunctionExpressionImpl;
+import org.geotools.filter.capability.FunctionNameImpl;
+import org.opengis.filter.capability.FunctionName;
+
+public class FilterFunction_arrayAnyMatch extends FunctionExpressionImpl {
```
jodygarnett: Please rename to follow java conventions, *AnyMatchArrayFunction*.

The darn "FilterFunction_" prefix was produced by a code generator back at the dawn of time.

modules/library/main/src/main/java/org/geotools/filter/function/FilterFunction_arrayAnyMatch.java
```diff
@@ -0,0 +1,64 @@
...
+    public FilterFunction_arrayAnyMatch() {
+        super(NAME);
+    }
+
+    public Object evaluate(Object feature) {
+        Object[] array;
+        Object candidate;
+
+        try { // attempt to get array and perform conversion
+            array = (Object[]) getExpression(0).evaluate(feature);
```
jodygarnett: if you expect an `Object[]` then ask for it rather than eating exceptions:
```
array = getExpression(0).evalulate( feature, Object[] )
```

This approach reduce the amount if error checking boilerplate and allows folks to introduce conversions, for example allowing your function to handle an array represented as a string.

modules/library/main/src/main/java/org/geotools/filter/function/FilterFunction_arrayAnyMatch.java
```diff
@@ -0,0 +1,64 @@
...
+import org.geotools.filter.FunctionExpressionImpl;
+import org.geotools.filter.capability.FunctionNameImpl;
+import org.opengis.filter.capability.FunctionName;
+
+public class FilterFunction_arrayAnyMatch extends FunctionExpressionImpl {
+    public static FunctionName NAME =
+            new FunctionNameImpl(
+                    "arrayAnyMatch",
+                    parameter(
+                            "arrayAnyMatch",
```
jodygarnett: Feedback: The name "arrayAnyMatch" is still odd to me when the usual function is called `IN`, would "InArray" work for you? This would also reverse the order of the arguments:
```
InArray( candidate, array )
```

modules/library/main/src/test/java/org/geotools/filter/function/FilterFunction_arrayAnyMatchTest.java
```diff
@@ -0,0 +1,70 @@
...
+    @Test
+    public void testDoesNotMatchStrings() throws Exception {
+        FilterFactoryImpl ff = new FilterFactoryImpl();
+        Function func =
+                ff.function(
+                        "arrayAnyMatch",
+                        ff.literal(new String[] {"a", "b", "c", "d"}),
+                        ff.literal("g"));
+        assertFalse((Boolean) func.evaluate(new Object()));
+    }
```
jodygarnett: Test the edge cases to check if type conversion is working:
```
    @Test
     public void testDoesNotMatchStrings() throws Exception {
         FilterFactoryImpl ff = new FilterFactoryImpl();
         Function func =
                 ff.function(
                         "arrayAnyMatch",
                         ff.literal("a"),
                         ff.literal("g"));
         assertFalse((Boolean) func.evaluate(new Object()));
     }
```

Can I ask how this is represented in `ECQL`?

modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/FilterToSqlHelper.java
```diff
@@ -930,6 +934,20 @@ private String getPrimaryKeyColumnsAsCommaSeparatedList(
         return sb.toString();
     }
 
+    public Object visit(FilterFunction_arrayAnyMatch filter, Object extraData) {
+        Expression array = getParameter(filter, 0, true);
```
jodygarnett: Defensive programming needed here to fit in with the Filter flexibility in representation of values.

modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/FilterToSqlHelper.java
```diff
@@ -930,6 +934,20 @@ private String getPrimaryKeyColumnsAsCommaSeparatedList(
...
+        Expression candidate = getParameter(filter, 1, true);
```
jodygarnett: Do you need to check that the type of values match, will SQL be upset if you check if a boolean is in an array of strings?

modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/PostgisPSFilterToSql.java
```diff
@@ -187,6 +189,8 @@ public Object visit(PropertyIsEqualTo filter, Object extraData) {
                                     throw new RuntimeException(e);
                                 }
                             }));
+        } else if (any != null) {
+            return helper.visit(any, extraData);
```
jodygarnett: Feedback: A question that goes beyond this PR review - is the duplication necessary? Can one class extend the other.

modules/library/main/src/main/java/org/geotools/filter/function/FilterFunction_arrayAnyMatch.java
mbarto: I agree going to rename to inArray, and removing boilerplate

modules/library/main/src/test/java/org/geotools/filter/function/FilterFunction_arrayAnyMatchTest.java
mbarto: From my knowledge you cannot express array literals in ECQL, so this is only going to be used on datasource that have array type properties (we use it that way in the postgis driver)

modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/PostgisPSFilterToSql.java
mbarto: Not sure about this. I suppose the helper class is the current attempt to put some code in a common place for both.

modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/FilterToSqlHelper.java
mbarto: Can you explain more precisely what you would do here?

modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/FilterToSqlHelper.java
mbarto: I am going to add checks for type matching and casting when needed

aaime:
The appveyor failure is unrelated (download issues) but I cannot restart it, guess only @jodygarnett can

PR #2668, opened 2019-11-15
Refactor literal strings to constants in AppSchema mapping feature iterator

fernandor777:
This PR refactors some string literals into documented constants.


aaime:
Yep, thanks!

PR #2675, opened 2019-11-25
MBTiles + MVT datastore

aaime:
Comes with a bunch of changes to make the store work and/or be more efficient in rendering.
See also https://osgeo-org.atlassian.net/browse/GEOT-6427

Sample rendering (notice how tile borders are gone):
image


modules/library/main/src/main/java/org/geotools/feature/simple/SimpleFeatureImpl.java
```diff
@@ -374,6 +374,11 @@ public boolean isNillable() {
         return userData;
     }
 
+    @Override
+    public boolean hasUserData() {
+        return userData != null;
```
fernandor777: A question: Omitting the map.isEmpty() check is intended?  Overrided default method does the check but here only a null check is done.

modules/library/metadata/src/main/java/org/geotools/util/factory/FactoryRegistry.java
```diff
@@ -453,25 +453,27 @@ private static void debug(
...
-        if (key != null) {
-            buffer.append(", ").append(key);
-        }
-        buffer.append(')');
-        if (message != null) {
-            buffer.append(": ").append(message);
-        }
-        if (type != null) {
-            buffer.append(' ').append(Classes.getShortName(type)).append('.');
+        if (LOGGER.isLoggable(DEBUG_LEVEL)) {
```
fernandor777: Nice refactor to avoid log data building if log level required is not reached.

modules/unsupported/mbtiles/src/test/java/org/geotools/mbtiles/VectorLayersMetadataTest.java
```diff
@@ -0,0 +1,45 @@
...
+
+import java.util.List;
+import org.apache.commons.io.IOUtils;
+import org.junit.Test;
+
+public class VectorLayersMetadataTest {
+
+    @Test
+    public void testParseMetadata() throws Exception {
+        String json = IOUtils.toString(getClass().getResourceAsStream("vectorLayers.json"));
```
fernandor777: toString(java.io.InputStream) usage is deprecated.  IOUtils.toString(inputStream, StandardCharsets.UTF_8.name()) is recommended.  This is not critical, just seems like is causing a build issue on some travis build.

modules/library/main/src/main/java/org/geotools/feature/simple/SimpleFeatureImpl.java
aaime: Good question, my intention was just to avoid unecessary allocations on each single feature going through the renderer, but indeed, the semantic of the method calls for checking isEmpty too. Fixing it.

modules/library/metadata/src/main/java/org/geotools/util/factory/FactoryRegistry.java
aaime: Eh yeah, was building a very large string for nothing, every single time.

modules/unsupported/mbtiles/src/test/java/org/geotools/mbtiles/VectorLayersMetadataTest.java
aaime: Indeed, fixed it.

PR #2677, opened 2019-11-25
[GEOT-6423] Hints.ensureSystemDefaultsLoaded() incorrect use of AtomicBoolean semantics

groldan:
Make updating `Hints.GLOBAL` actually an atomic operation.

Do use `AtomicBoolean` semantics at `org.geotools.util.factory.Hints.ensureSystemDefaultLoaded()` to update the GLOBAL class variable.

### DIFF
--- a/modules/library/metadata/src/main/java/org/geotools/util/factory/Hints.java
+++ b/modules/library/metadata/src/main/java/org/geotools/util/factory/Hints.java
@@ -1169,12 +1169,12 @@ public class Hints extends RenderingHints {
      *     otherwise.
      */
     private static boolean ensureSystemDefaultLoaded() {
-        if (needScan.get()) {
+        // update GLOBAL atomically only if needScan == true
+        if (needScan.compareAndSet(true, false)) {
             Map<RenderingHints.Key, Object> newGlobal =
                     new ConcurrentHashMap<RenderingHints.Key, Object>(GLOBAL);
             boolean modified = GeoTools.scanForSystemHints(newGlobal);
             GLOBAL = newGlobal;
-            needScan.set(false);
             return modified;
         } else {
             return false;


modules/library/metadata/src/main/java/org/geotools/util/factory/Hints.java
```diff
@@ -1169,12 +1169,12 @@ public static void scanSystemProperties() {
      *     otherwise.
      */
     private static boolean ensureSystemDefaultLoaded() {
-        if (needScan.get()) {
+        // update GLOBAL atomically only if needScan == true
+        if (needScan.compareAndSet(true, false)) {
```
aaime: Won't this cause a race condition though? Idea:
- Thread1 gets in, needScan is true, sets it to false before doing the scan
- Thread2 gets in and thinks no scan is needed, and will proceeed doing whatever is doing without the system hints being actually loaded yet.

aaime:
@groldan you still around? :-D

modules/library/metadata/src/main/java/org/geotools/util/factory/Hints.java
groldan: You're right. I'm not completely certain this needs to be fixed right now though, will need to re-run the profiling sessions I did before to create the other tickets under GEOT-6420 to see if the extra complexity would be justified. By that I mean that concurrent calls to `Hints.ensureSystemDefaultLoaded()` ultimately get blocked by the `synchronized` block at `GeoTools.scanForSystemHints()` and could be alleviated if need be, but I'll wait till I can justify it or close this instead.

groldan:
Do not merge, waiting for evidence this shows up as a concurrency bottleneck before deciding whether a single execution of `GeoTools.scanForSystemHints` needs to be guaranteed under concurrency.

aaime:
The PR in this state definitely cannot be merged.  I was wondering about closing it, since it has been around for a while with no further progress.

groldan:
I think covering all possible concurrency issues with these two methods (`scanSystemProperties()` and its lazy executor `ensureSystemDefaultsLoaded()`) would require something more elaborate[1] and it's probably not worth it given `scanSystemProperties()` (the public one of them both) is only used in test cases both in geotools and geoserver.

But this one (#2784) does make sense IMO. So we can close this PR.

[1] https://github.com/groldan/geotools/commit/ca7298f0f1d3bb609713bc60def316de595e4155

groldan:
WON'T FIX

PR #2678, opened 2019-11-26
 [GEOT-6397] updated user docs related to extra methods added in DataUtilities class

nprigour:
updated user docs related to DataUtilities class update


jodygarnett:
Thanks for adding docs, every little bit helps especially in these core classes.

PR #2682, opened 2019-12-03
[GEOT-6444]: Add units to DistanceBufferOperator

mglazer:


mglazer:
Actually looking at this more, this doesn't actually fix the issue since there is some weird schema generation magic happening here which is failing for the OGC types.

mglazer:
I actually suspect this is because of the TODO left behind in DistanceBufferType:

https://github.com/geotools/geotools/blob/95a4d16fcdb8473fccc58e2645332354e61a214a/modules/extension/xsd/xsd-filter/src/main/java/org/geotools/filter/v1_0/OGCDistanceBufferTypeBinding.java#L84-L110

aaime:
I don't believe it has anything to do with code generation, I'd look at filter encoding, which is done programmaticaly, not using those annotations (back when this was done, JAXB was unable to deal with the complexity of GML/Filter).

Check here for FES: https://github.com/geotools/geotools/blob/5258a921e0daf26fbbb70a689e7437ccceead886/modules/extension/xsd/xsd-fes/src/main/java/org/geotools/filter/v2_0/bindings/DistanceBufferTypeBinding.java#L75
and probably here for the others:
https://github.com/geotools/geotools/blob/5258a921e0daf26fbbb70a689e7437ccceead886/modules/extension/xsd/xsd-filter/src/main/java/org/geotools/filter/v1_0/OGCDistanceTypeBinding.java#L92

PR #2687, opened 2019-12-06
Mbstyles fixes, round 2

aaime:
Another round of fixes, this time with a couple of new vendor options allowing to match the MBStyles behavior. The bigger change is how symbols are rendered, allowing to go from this:

tiles

to this (all the labels were hidden by the point symbolizers acting as label obstacles, instead of taking part in the conflict resolution game... that plus the "off by one" zoom level fix):

Selezione_999(044)

For reference, the MVT rendering on maptiler.com, of the same area (yep, still quite a bit of work to do):

image

Also added an important other one, related to how stops are processed, which was breaking interpolation of values at intermediate points between zoom levels.


modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/function/CSSFunction.java
```diff
@@ -60,19 +60,16 @@ public CSSFunction() {
     }
 
     @Override
-    public <T> T evaluate(Object object, Class<T> context) {
+    public Object evaluate(Object object) {
```
tbarsballe: Is there any merit to instead just adding:
```
public Object evaluate(Object object) {
   return evaluate(Object object, null);
}
```
in order to preserve the generic implementation?

modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/function/ZoomLevelFunction.java
```diff
@@ -60,7 +60,9 @@
                     parameter("scaleDenominator", Number.class),
                     parameter("srid", String.class));
 
-    public static final double EPSG_3857_O_SCALE = 559_082_263.9508929;
+    public static final int ROOT_TILE_PIXELS =
+            Integer.valueOf(System.getProperty("MBSTYLE_ROOT_TILE_PIXELS", "512"));
+    public static final double EPSG_3857_O_SCALE = 559_082_263.9508929 / ROOT_TILE_PIXELS * 256;
```
tbarsballe: Took me a minute to figure out this one - maybe use brackets to be a little more clear - `(559_082_263.9508929 / ROOT_TILE_PIXELS) * 256;`

modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/function/ZoomLevelFunction.java
```diff
@@ -60,7 +60,9 @@
                     parameter("scaleDenominator", Number.class),
                     parameter("srid", String.class));
 
-    public static final double EPSG_3857_O_SCALE = 559_082_263.9508929;
+    public static final int ROOT_TILE_PIXELS =
+            Integer.valueOf(System.getProperty("MBSTYLE_ROOT_TILE_PIXELS", "512"));
```
tbarsballe: Can you add this to the docs somewhere? Config variables that are only in the code (and release announcements) are a pain to track down later on.

modules/library/main/src/main/java/org/geotools/styling/PointSymbolizer.java
```diff
@@ -65,6 +65,14 @@
  * @version $Id$
  */
 public interface PointSymbolizer extends org.opengis.style.PointSymbolizer, Symbolizer {
+
+    /**
+     * Boolean vendor option, defaults to true. If true, in case no specified mark or graphics can
+     * be used, the default square mark will be used instead. If false, the symbol will not be
+     * painted.
+     */
+    public static String FALLBACK_ON_DEFAULT_MARK = "fallbackOnDefaultMark";
```
tbarsballe: Please add to docs

modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/function/CSSFunction.java
aaime: Not for this one I believe... these functions were implemented "upside down", the base class is designed so that one implements the version of evaluate without the context, and the base class takes care of using converters if the version with the context is called. Now I understand if the implementation can do something intelligent with the context, then both need to be implemented, but this particular function has no use for it.

modules/library/main/src/main/java/org/geotools/styling/PointSymbolizer.java
aaime: Done

modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/function/ZoomLevelFunction.java
aaime: Done

modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/function/ZoomLevelFunction.java
aaime: Done

PR #2689, opened 2019-12-08
[GEOT-6451] Flatgeobuf DataStore

jericks:
Create a DataStore for reading and writing flatgeobuf files:  https://github.com/bjornharrtell/flatgeobuf

JIRA #GEOT-6451 https://osgeo-org.atlassian.net/browse/GEOT-6451


aaime:
Legit QA failures:

```
[INFO] PMD Failure: org.geotools.data.flatgeobuf.FlatgeobufFeatureReader:38 Rule:UnusedPrivateField Priority:3 Avoid unused private fields such as 'featureIndex'..
[INFO] PMD Failure: org.geotools.data.flatgeobuf.FlatgeobufFeatureReader:44 Rule:UnusedPrivateField Priority:3 Avoid unused private fields such as 'featureBuilder'..
[INFO] PMD Failure: org.geotools.data.flatgeobuf.FlatgeobufFeatureReader:51 Rule:UnusedLocalVariable Priority:3 Avoid unused local variables such as 'file'..
[INFO] PMD Failure: org.geotools.data.flatgeobuf.FlatgeobufFeatureSource:84 Rule:UnusedLocalVariable Priority:3 Avoid unused local variables such as 'feature'..
[INFO] PMD Failure: org.geotools.data.flatgeobuf.FlatgeobufWriter:59 Rule:EmptyIfStmt Priority:3 Avoid empty if statements.
```

Besides that, as a community module, not much to add, just have a curiosity.
The module seems to look for a local file, at the same time we have a WFS output format in the making in GeoServer:
https://github.com/geoserver/geoserver/pull/3470
So I think, one could speed up data transfer between the WFS data store and a WFS server (a well known pain point) by using the Flatgeobuf format, if both ends support it. Is the code in the store useful to support such a use case? Like, could it be used as a library of sorts?

bjornharrtell:
Happy to see this @jericks. :) Note that there are some shortcomings in the current ref. FlatGeobuf Java impl. (no index read/write support and not full geometry type support) which I now feel more motivated to finish.

jericks:
Both pull requests are using bjornharrtell's flatgeobuf library (https://github.com/bjornharrtell/flatgeobuf/tree/master/src/java) but there is some duplicated code that I would like to remove.

bjornharrtell:
@aaime, agreed FlatGeobuf uld be a nice "under the covers" use case to speed up cascaded WFS.

@jericks, I would be interested to know more about what made you work on this?

And a general observeration, note that the Java impl. currently only supports FlatGeobuf in it's unindexed form - which is good for on the fly streaming read/write stuff. A FlatGeobuf with index has to be written in two passes but then supports the use case as a container for very large static datasets and fast spatial filtering.

aaime:
As a community module this does not really need a review, so I'm going ahead and merge, but feel free to use this PR as a way to communicate, if you so like :-D

jericks:
Thanks for reviewing and merging Andrea!

@bjornharrtell I added a Geobuf DataStore to GeoTools a while ago and thought doing the same for Flatgeobuf it would be an interesting project. Also, I wanted to add Flatgeobuf support to GeoScript Groovy (which is now done: https://github.com/geoscript/geoscript-groovy/commit/78db46eba2f96752f5e5b498f38f28d9cd8d26ed).

PR #2695, opened 2019-12-11
[GEOT-6460]: Better nodata management on bandMerges

dromagnoli:


modules/library/coverage/src/test/java/org/geotools/image/ImageWorkerTest.java
```diff
@@ -1713,6 +1720,40 @@ public void testBandMerge() {
...
+        double noDataValue = -10000d;
+        Range nodata = RangeFactory.create(noDataValue, noDataValue);
+        double[] background = new double[] {noDataValue};
+
+        RenderedImage twoBands = null;
+        PlanarImage pi = PlanarImage.wrapRenderedImage(imageWithNodata2);
+        pi.setProperty("GC_NODATA", new NoDataContainer(nodata));
+
+        ImageWorker worker =
+                new ImageWorker()
```
aaime: So why the mosaic in the test? The code had no changes there or not?

modules/library/coverage/src/test/java/org/geotools/image/ImageWorkerTest.java
```diff
@@ -1713,6 +1720,40 @@ public void testBandMerge() {
...
+                                null,
+                                null,
+                                new double[][] {{-32768}},
+                                null);
+        twoBands = worker.addBand(worker.getRenderedImage(), false).getRenderedImage();
+        RenderedImage singleBand = new ImageWorker(twoBands).retainBands(1).getRenderedImage();
+        RenderedImage threeBands =
+                new ImageWorker(twoBands).addBand(singleBand, false).getRenderedImage();
+
+        // Check that the noData holes are still noData after bandMerge
```
aaime: What happens if, by accident, the different input bands had different nodata?

modules/library/coverage/src/test/java/org/geotools/image/ImageWorkerTest.java
dromagnoli: Just to add something in the mix :) This was one of the chains initially spotting the issue.
Going to check if the problem still happen after a simple read.

modules/library/coverage/src/test/java/org/geotools/image/ImageWorkerTest.java
dromagnoli: I think that at the moment it isn't supported since there is only a single nodata value support, not an array. Do we ever had a case with different nodata in input?

aaime: Not really, just thinking out loud

aaime:
You're getting a doc build failure, Jody just fixed it, can rebase from master if you want a green build.

dromagnoli:
Yep. I have seen the email this morning about the doc build fix. I was pretty sure I already did the rebase. double checking...
PS: I have removed the mosaic usage from previous test too

aaime:
+1

PR #2701, opened 2019-12-12
More MBStyles fixes 

aaime:


modules/unsupported/mbstyle/src/main/java/org/geotools/mbstyle/sprite/SpriteGraphicFactory.java
```diff
@@ -97,6 +101,9 @@
 
     private static final String ICON_NAME_DELIMITER = "#";
 
+    private static final Pattern CHARSET_PATTERN =
+            Pattern.compile("(?i)\\bcharset=\\s*\"?([^\\s;\"]*)");
+
```
tbarsballe: Can you add a test case for this issue? (GEOT-6464)

aaime: Not sure how to do it though... the code is using JDK facilities to open the URL, one would have to stand up an actual HTTP server no?

tbarsballe: I was thinking you could do a unit test on the problematic json itself, since that seems to be where the issue is coming from, although I suppose the current implementation doesn't really support that approach, since the connection & read is all in one place.
You could split out the parsing part into its own function to make it unit testable, but I don't really like making structural changes to src code just to make the tests work.

I do note that a bunch of the visual style tests seem to function just fine with a `file:` url, e.g. [here](https://github.com/geotools/geotools/blob/master/modules/unsupported/mbstyle/src/test/resources/org/geotools/mbstyle/symbolStyleSimpleIconAndTextLinePlacementTest.json#L26)

```
"sprite": "file:src/test/resources/org/geotools/mbstyle/sprite/test-data/sprite",
```
So you might be able to get away with using something like that?

Alternatively, you could probably do something with a MockServer, but thats starting be more effort than its worth.

aaime: I've set it up using a HttpClient interface and mock implementation, seems to work.

PR #2704, opened 2019-12-12
Remove XML Entities when parsing via PullParser

brandon-franklin0:
The XMLInputFactory is currently using the default configuration for handling XML entities when creating the XMLStreamReader. If there is a more private way for me to communicate with a maintainer, please let me know.

### DIFF
--- a/modules/extension/xsd/xsd-core/src/main/java/org/geotools/xsd/PullParser.java
+++ b/modules/extension/xsd/xsd-core/src/main/java/org/geotools/xsd/PullParser.java
@@ -140,6 +140,9 @@ public class PullParser {
         XMLInputFactory factory = XMLInputFactory.newInstance();
         factory.setProperty(XMLInputFactory.IS_NAMESPACE_AWARE, true);
         factory.setProperty(XMLInputFactory.IS_VALIDATING, false);
+        factory.setProperty(XMLInputFactory.SUPPORT_DTD, false);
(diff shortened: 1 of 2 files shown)


aaime:
HI Brandon, the gotcha is, the modules does not really have a dedicated maintainer anymore, and the pull parser is only used, as far as I know, in the WFS datastore (see ``modules/unsupported/wfs-ng``).
That said, the approach we've taken on this matter elsewhere is to make the resolution configurable, rather than being turned off, to still allow entity resolution when one can guarantee it's not harmful.
See:

```
modules/library/metadata/src/main/java/org/geotools/util/NullEntityResolver.java:public class NullEntityResolver implements EntityResolver2, Serializable {
modules/library/metadata/src/main/java/org/geotools/util/PreventLocalEntityResolver.java:public class PreventLocalEntityResolver implements EntityResolver2, Serializable {
modules/library/metadata/src/test/java/org/geotools/util/factory/PlaceholderEntityResolver.java:public class PlaceholderEntityResolver implements EntityResolver {
```

and EntityResolverProvider in the GeoServer code.

aaime:
@brandon-franklin0 will you have time to make entity handling conform to the existing approach (used by Parser for example?). See "setEntityResolver" in the Parser class.

aaime:
Closing due to inactivity (pity really...)

PR #2708, opened 2019-12-15
[GEOT-6453] Bring CSV Datastore upto supported status

ianturton:
Progress:

* [x] Coverage at 89%
* [x] Docs untangled from live code
* [x] Module moved to plugins 
* [x] Docs updated to reflect changes


docs/src/main/java/org/geotools/tutorial/csv3/CSVDataStore.java
```diff
@@ -0,0 +1,122 @@
...
+    private final CSVStrategy csvStrategy;
+
+    private final CSVFileState csvFileState;
+
+    public CSVDataStore(CSVFileState csvFileState, CSVStrategy csvStrategy) {
+        this.csvFileState = csvFileState;
+        this.csvStrategy = csvStrategy;
+    }
+
+    // docs start getTypeName
```
aaime: I read somewhere about un-tangling the docs from the gt-csv code... if it has happened these comments here (and probably other places) are not needed anymore. But not harmful either.

docs/pom.xml
```diff
@@ -218,6 +218,11 @@
         <artifactId>junit</artifactId>
         <scope>provided</scope>
     </dependency>
+    <dependency>
```
aaime: I know other parts of the ecosystem use opencsv, like for example gs-wfs:

```
    <dependency>
      <groupId>net.sf.opencsv</groupId>
      <artifactId>opencsv</artifactId>
      <version>1.8</version>
      <scope>test</scope>
    </dependency>
```
and there are some in app-schema too.

Well it's a test and it's unlikely to cause conflicts (only importer depends on gt-csv right?), but something to keep in mind and maybe verify.

aaime: And oh, the switch will require a corresponding PR in GeoServer to upgrade the gt-csv dependencies, in particular, the package.

docs/src/main/java/org/geotools/tutorial/csv3/CSVDataStore.java
ianturton: They do no harm but taking them out will probably be good

jodygarnett: Agreed, the production CSV instance will gather fixes and optimizations over time that can impact its use as a teaching tool

docs/pom.xml
ianturton: I think GeoServer will be OK for the time being - https://osgeo-org.atlassian.net/browse/GEOS-9439

PR #2709, opened 2019-12-17
Change log level from "warning" to "fine"

KeyBridge:
A heterogenous feature collection is a _potential_ configuration error but certainly NOT a software fault.
If this condition is to be logged at all, it maximum level should be 'FINE'.

Current level creates false warnings in application logs.

### DIFF
--- a/modules/library/main/src/main/java/org/geotools/feature/DefaultFeatureCollection.java
+++ b/modules/library/main/src/main/java/org/geotools/feature/DefaultFeatureCollection.java
@@ -182,7 +182,7 @@ public class DefaultFeatureCollection
         }
         SimpleFeatureType childType = (SimpleFeatureType) getSchema();
         if (!feature.getFeatureType().equals(childType)) {
-            LOGGER.warning("Feature Collection contains a heterogeneous" + " mix of features");
+            LOGGER.fine("Feature Collection contains a heterogeneous mix of features");
         }
         // Check inheritance with FeatureType here?
         contents.put(ID, feature);


aaime:
There is nothing "false" there, GeoTools feature collections (and the library in general) is designed for predictable schemas and a collection has one that features should correspond to. Some parts of the library are tolerant to deviations (e.g. rendering) while others will break in various ways.

That said, I believe a move towards better support for schemaless data is going to be a necessity sooner or later... we just don't have dedicated funding to work on it at the moment.
@jodygarnett what do you think? Is relaxing log statements good enough of a start?

ianturton:
On Wed, 18 Dec 2019 at 10:19, Andrea Aime <notifications@github.com> wrote:

> There is nothing "false" there, GeoTools feature collections (and the
> library in general) is designed for predictable schemas and a collection
> has one that features should correspond to. Some parts of the library are
> tolerant to deviations (e.g. rendering) while others will break in various
> ways.
>
> That said, I believe a move towards better support for schemaless data is
> going to be a necessity sooner or later... we just don't have dedicated
> funding to work on it at the moment.
> @jodygarnett <https://github.com/jodygarnett> what do you think? Is
> relaxing log statements good enough of a start?
>
>
> It's certainly something to be thinking about, to be honest I'm not sure
logging is the most important place to start. But providing nothing is
breaking down the line with out a warning where it breaks I'd be happy with
this change.

Ian

-- 
Ian Turton

jodygarnett:
I think we have an explicit FeatureCollection class for heterogenous feature collections:

* Having mixed feature content in a SimpleFeatureCollection would be bad.
* A Feature Collection with member type AbstractFeature should be able to hold anything
* A Feature Collection with member type AbstractFeature and a choice restriction should be able to limit what is valid to contain (if your use case really cares).

jodygarnett:
I was "waiting" for the app-schema community to make helper classes for working with complex and mixed content in the same fashion as DefaultFeatureCollection provides tools for those working with SimpleFeatures.

I would like:

* DefaultFeatureCollection to strictly be a SimpleFeatureFeatureCollection with as many safety features as we can provide.
* A new class, say AbstractFeatureCollection created as a base for those working with Complex and/or heterogenous content

Question for Andrea: I believe we striped out a lot of the generic "collection/pojo" support from rendering. Would the rendering system be willing to work with a generic FeatureCollection? The data structures are in place for styling to control the process (FetureTypeStyles are restricted by type but I am not sure we consider that on a feature by feature basis).

jodygarnett:
@KeyBridge how disruptive would it be to throw an exception? Using a heterogenous feature collection in this specific data structure is a software fault. The rest of GeoTools depends on this API contract to return a consistent set of features from a SimpleFeatureCollection (as an API contract, not an assumption).

aaime:
@jodygarnett about rendering I honestly don't know, we only work with data with predictable structures (even from app-schema, we have a predictable schema most of the time, not sure how much type inheritance with variants is actually used). It may work or not, depending a lot on how much testing was put in place when the code got relaxed to work with unpredictable structures (I can't say I have seen any integration test doing rendering in those conditions...).

aaime:
@jodygarnett I was thinking another thing... is the feature type required to be exactly the same? Or coudl it be that the feature has more attributes than the base type? The latter seems acceptable even in simple features land.

KeyBridge:
@jodygarnett Throwing an exception would probably be bad unless the condition really is a fault. I'm not aware of any requirement for feature collection homogeneity, at least not from RFC 7946. Is there something else I'm missing? 

Regardless, I recognize some applications may have been coded to expect homogeneity, and certainly don't want to break those, which is why I'm only suggesting changing the LOG level from _warning_ to _fine_.

KeyBridge:
Recognizing that status quo may be best suited for existing applications, our team could help code your suggestion of a new `AbstractFeatureCollection` and a more relaxed `MixedFeatureCollection` instance to accommodate heterogeneous schemas. 

What is the sense of the group on this?

aaime:
@jodygarnett see the javadoc for FeatureCollection.getSchema:
https://github.com/geotools/geotools/blob/e04352a97814d2abf9491c5494e582c776744679/modules/library/main/src/main/java/org/geotools/feature/FeatureCollection.java#L127

Check it all, it has provisions for mixed content collections and DefaultFeatureCollection does not further constraints it in its javadocs. However it uses SimpleFeatureType, that has restrictions of its own:
https://github.com/geotools/geotools/blob/8ed704ba86afe77b48628559fa3739db046feebc/modules/library/opengis/src/main/java/org/opengis/feature/simple/SimpleFeatureType.java#L18

Reading it confirms (in my mind) that the current equality check is faulty, and that the actual check should go for type inheritance instead (a feature is a valid member of the collection is its feature type inherits from the one of the collection, but does not need to be the same).

jodygarnett:
Thanks Andrea, I agree that checking type inheritance works that way, so technically one using AbstractFeatureType should be able to do anything....

In the specific case of this implementation (SimpleFeature and SimpleFeatureType) the whole point is the content has the same column order so that getting a value by int, is just as good as explicit looking up value by a string.

See the heading on "Attribute Indexing" on SimpleFeatureType class [javadocs\(https://docs.geotools.org/stable/javadocs/org/opengis/feature/simple/SimpleFeatureType.html).

@KeyBridge for this class the goal was to have an "optimized" implementation for content like shapefiles; as folks had a lot of code working against a previous GeoTools API which we were replacing.

modules/library/main/src/main/java/org/geotools/feature/DefaultFeatureCollection.java
```diff
@@ -182,7 +182,7 @@ protected boolean add(SimpleFeature feature, boolean fire) {
         }
         SimpleFeatureType childType = (SimpleFeatureType) getSchema();
         if (!feature.getFeatureType().equals(childType)) {
-            LOGGER.warning("Feature Collection contains a heterogeneous" + " mix of features");
+            LOGGER.fine("Feature Collection contains a heterogeneous mix of features");
```
jodygarnett: This test is behaving as intended, and should probably be upgraded to an exception if the warning is not being sufficient

Long term we should change this to an exception and arrange an helper class for FeatureCollection that supports mixed content as I agree we will see more of these over time.

Is there a helper class in app schema that we can make use of right now?

jodygarnett: I was thinking if we could make the class have two "modes", one for when AbstractFeatureType is used and then relax this check as @aaime indicated. 

Sadly the class explicitly asks getSchema() to return SimpleFeatureType so we cannot return AbstractFeatureType here

jodygarnett:
@aaime I agree FeatureCollection is written to expect content, this specific case of SimpleFeatureCollection locks that down to document/constrain a attribute order.

Can we quickly copy DefaultFeatureCollection as an generic implementation of FeatureCollection?

jodygarnett:
> A heterogenous feature collection is a _potential_ configuration error but certainly NOT a software fault.
> If this condition is to be logged at all, it maximum level should be 'FINE'.

@KeyBridge can I ask what data store or parser you were using when you got this? Perhaps we could fix that implementation to work with generic FeatureCollection?

aaime:
After some discussion in person we indeed found out that SimpleFeatureCollection prevents this change, as:

- it states that the schema for all feature is _the same_ (first sentence)
- code should be able to extract a property index from the schema, and then use it to access features, to gain performance, that won't work in case inheritance is involved

So indeed, to confirm, a new class is needed. Should be easy to make though, create an empty base class that uses Feature and FeatureType, move all the code in DefaultFeatureCollection up, generify, have DefaultFeatureCollection extend it casting to SimpleFeature/SimpleFeatureType.

aaime:
Closing as we determined this is not the way to go, a  DefaultFaetureCollection must not have heterogeneous features inside. Please feel free to open a separate PR with a new feature collection that is designed to be heterogeneous (and thus, not implementing SimpleFeatureCollection).

KeyBridge:
Thank you all for the consideration. It is frustrating and disappointing that our simple request was rejected.

From how to exchange turned perhaps the team misunderstood our request. As a reminder: our suggested change was for a logging level to be _reduced_, not eliminated or removed, nor for any other changes to the software.

We believe the current logging strategy is incorrect as it warns of a condition in full compliance with the relevant specification. 

We requested this change as the warnings are neither useful nor correct and create a real operational burden for us. Their impact is to fill our application logs with large amounts of garbage which we must filter through to find actual issues.

PR #2713, opened 2019-12-18
[GEOT-3300] Re-enable Windows share URL test, only on Windows

groldan:
WIP, checking if the test with a windows share URL will work on Windows


aaime:
Thanks!

aaime:
This is actually broken on Wndows too, should be removing the new test.

PR #2719, opened 2019-12-23
Updating to latest PMD version, focus on Closeable

aaime:


ianturton:
No clue as to what's up with Appveyor

aaime:
Build just disappeared after 10 minutes running.. no good reason really, guessing they might be using amazon fleet or something like that...

PR #2733, opened 2020-01-02
Mbstyles, disables partials, enable multi-font search, fix labelling as a consequence

aaime:
The end result of the activity looks like this:

image

image


modules/library/render/src/main/java/org/geotools/renderer/style/FontCache.java
```diff
@@ -252,4 +259,31 @@ public synchronized void resetCache() {
...
+                            .stream()
+                            .filter(f -> f.startsWith(name))
+                            .filter(
+                                    f -> { // leave out alterations, use base fonts
+                                        String lc = f.toLowerCase();
+                                        return !lc.contains(" bold") && !lc.contains(" italic");
+                                    })
+                            .sorted() // leave further altered fonts down the line, base ones first
+                            .collect(Collectors.toList());
+        }
```
tbarsballe: Should this put the result into alternatives before returning? As far as I can tell, nothing ever gets written to alternatives currently.

modules/library/render/src/test/java/org/geotools/renderer/lite/LetterConflictTest.java
```diff
@@ -55,8 +55,8 @@
      *
      * <p>Build with mvn -P image.interactive
      */
-    static final boolean IMAGE_INTERACTIVE =
-            Boolean.getBoolean("org.geotools.image.test.interactive");
+    static final boolean IMAGE_INTERACTIVE = true;
+    // Boolean.getBoolean("org.geotools.image.test.interactive");
```
tbarsballe: This should be restored to its prior state before merging

tbarsballe (changes requested):
A couple things need fixing - see comments inline.

I also think the Javadocs could maybe do with a slightly more detailed explanation of what "Font Alternatives" actually are - the tests give a good example but the javadocs themselves are kinda vague (and the FontAlternativesFunction doesn't explain it at all), but this is relatively minor.

Everything else looks good.

modules/library/render/src/main/java/org/geotools/renderer/style/FontCache.java
aaime: Good catch, thanks!

modules/library/render/src/test/java/org/geotools/renderer/lite/LetterConflictTest.java
aaime: Yep

aaime:
@tbarsballe review feedback (hopefully) addressed.

tbarsballe (approved):

aaime:
Thanks! Merging.

PR #2736, opened 2020-01-05
Fmt maven plugin 2.9.0

aaime:
Trying out latest version of the format plugin. Two visible differences:

- Annotation formatting avoids wrapping up the closing parenthesis on the next line
- stream() is treated as part of the main block and not wrapped

Do we want to to adopt? I'm fine either way.


ianturton (approved):

jodygarnett:
As expressed on email I would like to adopt, do we need to backport across all active versions etc...

aaime:
Yes, if we adopt, it should be uniform on all projects and active branches. We should also try to reduce the PR wip, this commit has a chance of causing conflicts (for anything that uses streams).

aaime:
Hey is there anyone that would like to take up on this one? I'm likely going to drop it myself, the formatting improvements are minor.

PR #2737, opened 2020-01-07
[GEOT-6473]: ContrastEnhancement to byte. Reserve nodata and adjust s…

dromagnoli:
…cale/offset


dromagnoli:
Merging it. (AppVeyor having unrelated failure)

PR #2739, opened 2020-01-10
[GEOT-6488] GeoJSONDatastore seems to ignore axis order requirements

ianturton:
This PR fixes the axis order and reprojects to 4326 if needed.


aaime:
Mind, the axis order is not an absolute. The specification allows for using other CRSs when there is an agreement between server and client. OGC Features API will use this actively to generate GeoJSON in whatever projection the client requested.

ianturton:
That's in the future work pile, currently I just want to match the RFC spec. Please open a ticket for the enhancement and assign it to me :-)

PR #2757, opened 2020-01-17
Add SQL error message to insertion error message

mac9416:
I got `Error inserting features` when I just needed to expand a column width. Adding the underlying SQLException message would have helped me avoid searching the source for the error message, attaching a debugger (a little cumbersome in my environment), and re-running the code to reproduce the issue.

I see there's precedent [here](https://github.com/geotools/geotools/blob/376585cd7517451c9a53846c6cc3c392e8e1fdb3/modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java#L2171), in the same file, for appending the underlying message.

```java
} catch (SQLException e) {
    throw new RuntimeException("Unable to obtain connection: " + e.getMessage(), e);
}
```

## Checklist

For all pull requests:

_I'm not sure if this is a core or extension module?_

### DIFF
--- a/modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
+++ b/modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
@@ -1770,7 +1770,7 @@ public final class JDBCDataStore extends ContentDataStore implements GmlObjectSt
                     insertNonPS(notUseExistings, featureType, cx, key, false);
                 }
             } catch (SQLException e) {
-                String msg = "Error inserting features";
+                String msg = "Error inserting features: " + e.getMessage();
                 throw (IOException) new IOException(msg).initCause(e);
             }
         }


aaime:
In the case of GeoServer, the exceptions pop up to the service level and their message is reported to the client. Are we sure it may not contain sensitive information? Otherwise logging the message would be a more secure way to make the information available.

mac9416:
That's a fair concern. In the case of my application, I'm just going to get the exception's `cause` and log its message. That's probably a viable workaround for most folks.

I can modify this PR to use logging, if you think that would be a potential good solution.

aaime:
Yes please. :-D

aaime:
@mac9416 can you update the PR? Otherwise I'm afraid I'll have to close it for inactivity (we're beyond fully booked, it's unlikely that we'll find time to update it in our spare time).

mac9416:
@aaime thanks for checking back in! I'm going to close it for now, since I've had to move on to unrelated work. If I'm back in the same area of code, I'll re-open the PR.

PR #2784, opened 2020-02-03
[GEOT-6506] Unnecessary synchronization in GeoTools.scanForSystemHints()

groldan:
`GeoTools.scanForSystemHints()` synchronizes on its static
`BINDINGS` variable, which is only modified during the class'
static initialization. The method is side-effect free and
only mutates the argument object, hence there's no need
to synchronize on this field.

## Checklist

For all pull requests:

### DIFF
--- a/modules/library/metadata/src/main/java/org/geotools/util/factory/GeoTools.java
+++ b/modules/library/metadata/src/main/java/org/geotools/util/factory/GeoTools.java
@@ -130,12 +130,6 @@ public final class GeoTools {
     private static final EventListenerList LISTENERS = new EventListenerList();
 
     /**
-     * The bindings between {@linkplain System#getProperties system properties} and a hint key. This
-     * field must be declared before any call to the {@link #bind} method.
-     */
-    private static final Map<String, RenderingHints.Key> BINDINGS = new HashMap<>();
-
-    /**
      * The {@linkplain System#getProperty(String) system property} key for the default value to be
      * assigned to the {@link Hints#CRS_AUTHORITY_EXTRA_DIRECTORY CRS_AUTHORITY_EXTRA_DIRECTORY}
      * hint.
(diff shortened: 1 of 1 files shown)


modules/library/metadata/src/main/java/org/geotools/util/factory/GeoTools.java
```diff
@@ -133,8 +133,7 @@ private static Properties loadProperites(String resource) {
      * The bindings between {@linkplain System#getProperties system properties} and a hint key. This
      * field must be declared before any call to the {@link #bind} method.
      */
-    private static final Map<String, RenderingHints.Key> BINDINGS =
-            new HashMap<String, RenderingHints.Key>();
+    private static final Map<String, RenderingHints.Key> BINDINGS = new HashMap<>();
```
aaime: So... the synchronizations are not needed because BINDINGS is modified only during the class initialization itself (at least, with the current code). Right?

aaime: @groldan can you confirm/elaborate?

aaime:
@groldan we might have to close this for "lack of feedback". Any chance you can have a look at my simple question?

groldan:
don't close, I'll find some time this week to wrap this up

modules/library/metadata/src/main/java/org/geotools/util/factory/GeoTools.java
groldan: Hi @aaime yes, I think the commit message states it pretty clearly, let me know if it's confusing anyhow so I can rephrase:
```
GeoTools.scanForSystemHints() synchronizes on its static
BINDINGS variable, which is only modified during the class'
static initialization. The method is side-effect free and
only mutates the argument object, hence there's no need
to synchronize on this field.
```

groldan: well,  `GeoTools.scanForSystemHints(final Map<RenderingHints.Key, Object> hints)` is not "side effect free" because it mutates the argument object, I `BINDINGS` is never modified once the class is initialized.

groldan:
I think this better clarifies it: https://github.com/geotools/geotools/pull/2784/files#diff-847732244ac36aa80c8edb2c9064233cR259

aaime:
I was really just looking for a confirmation, but the new code is clearer indeed.

PR #2794, opened 2020-02-05
[GEOT-6509] gt-jdbc-hana - MultiPoint WKBs not parsed properly

stefanuhrig:
MultiPoint WKBs are not parsed properly by gt-jdbc-hana's WKB parser.

Instead of parsing each member of the multi-point as an independent
point, the parser interprets the following data as raw coordinates.


stefanuhrig:
The failed Java 11 build seems unrelated, because the gt-jdbc-hana plugin is not build at all in that profile. Hence, I'm merging the bugfix.

PR #2804, opened 2020-02-10
[GEOT-6512] Control remote HTTP requests sent by GeoTools

imranrajjad:
[GEOT-6512](https://osgeo-org.atlassian.net/browse/GEOT-6512) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

JIRA : https://osgeo-org.atlassian.net/browse/GEOT-6512
#### SPI implementation

- org.geotools.data.ows.URLChecker
- org.geotools.data.ows.URLCheckerFactory

#### HttpClient Wrapping

- org.geotools.data.ows.ControlledHttpClient
- org.geotools.data.ows.ControlledHttpClientFactory

#### Implementation overview
-wrap all known HTTPClient instances with ControlledHttp
-evaluating URL connections

#### Modules touched in this PR

--extension/tile-client
--extension/wmsts
--extension/xsd/xsd-core
--unsupported/wfs-ng
--unsupported/wps
--library/xml
--library/coverage

### Wrapping an HttpClient
`ControlledHttpClientFactory.wrap(new SimpleHttpClient())`

### Directly evaluating a URL,URI or String
`URLCheckerFactory.evaluate(url);`

## Checklist

For all pull requests:

### DIFF
--- a/modules/extension/tile-client/src/test/java/org/geotools/tile/impl/osm/OSMTileFactoryTest.java
+++ b/modules/extension/tile-client/src/test/java/org/geotools/tile/impl/osm/OSMTileFactoryTest.java
@@ -19,7 +19,7 @@ package org.geotools.tile.impl.osm;
 import static org.junit.Assert.fail;
 
 import org.geotools.data.ows.MockURLChecker;
-import org.geotools.data.ows.URLCheckerFactory;
+import org.geotools.data.ows.URLCheckers;
 import org.geotools.geometry.jts.ReferencedEnvelope;
 import org.geotools.referencing.crs.DefaultGeographicCRS;
 import org.geotools.tile.Tile;
@@ -30,6 +30,7 @@ import org.geotools.tile.impl.WebMercatorTileFactory;
 import org.geotools.tile.impl.WebMercatorTileService;
 import org.geotools.tile.impl.WebMercatorZoomLevel;
 import org.geotools.tile.impl.bing.BingService;
+import org.hamcrest.CoreMatchers;
 import org.junit.Assert;
 import org.junit.Test;
 
@@ -115,7 +116,7 @@ public class OSMTileFactoryTest extends TileFactoryTest {
         // should throw exception
         MockURLChecker urlChecker = new MockURLChecker();
         urlChecker.setEnabled(true);
-        URLCheckerFactory.addURLChecker(urlChecker);
+        URLCheckers.addURLChecker(urlChecker);
 
         try {
             Tile tile =
@@ -129,9 +130,12 @@ public class OSMTileFactoryTest extends TileFactoryTest {
 
             fail();
         } catch (Exception e) {
-            Assert.assertTrue(e.getMessage().contains("Evaluation Failure:"));
+            Assert.assertThat(
+                    e.getMessage(),
+                    CoreMatchers.is(
+                            "Evaluation Failure: http://tile.openstreetmap.org/5/20/15.png: did not pass security evaluation"));
         } finally {
-            URLCheckerFactory.removeURLChecker(urlChecker);
+            URLCheckers.removeURLChecker(urlChecker);
         }
     }
 
--- a/modules/extension/tile-client/src/test/java/org/geotools/tile/impl/osm/OSMTileFactoryTest.java
+++ b/modules/extension/tile-client/src/test/java/org/geotools/tile/impl/osm/OSMTileFactoryTest.java
@@ -31,6 +31,7 @@ import org.geotools.tile.impl.WebMercatorTileService;
 import org.geotools.tile.impl.WebMercatorZoomLevel;
 import org.geotools.tile.impl.bing.BingService;
 import org.hamcrest.CoreMatchers;
+import org.hamcrest.MatcherAssert;
 import org.junit.Assert;
 import org.junit.Test;
 
@@ -130,7 +131,7 @@ public class OSMTileFactoryTest extends TileFactoryTest {
 
             fail();
         } catch (Exception e) {
-            Assert.assertThat(
+            MatcherAssert.assertThat(
                     e.getMessage(),
                     CoreMatchers.is(
                             "Evaluation Failure: http://tile.openstreetmap.org/5/20/15.png: did not pass security evaluation"));
--- a/modules/extension/tile-client/src/test/java/org/geotools/tile/impl/osm/OSMTileFactoryTest.java
+++ b/modules/extension/tile-client/src/test/java/org/geotools/tile/impl/osm/OSMTileFactoryTest.java
@@ -16,6 +16,10 @@
  */
 package org.geotools.tile.impl.osm;
 
+import static org.junit.Assert.fail;
+
+import org.geotools.data.ows.MockURLChecker;
+import org.geotools.data.ows.URLCheckerFactory;
 import org.geotools.geometry.jts.ReferencedEnvelope;
 import org.geotools.referencing.crs.DefaultGeographicCRS;
 import org.geotools.tile.Tile;
@@ -102,6 +106,33 @@ public class OSMTileFactoryTest extends TileFactoryTest {
         Assert.assertEquals(env.getMinY(), expectedEnv.getMinY(), 0.000001);
         Assert.assertEquals(env.getMaxX(), expectedEnv.getMaxX(), 0.000001);
         Assert.assertEquals(env.getMaxY(), expectedEnv.getMaxY(), 0.000001);
+    }
+
+    @Test
+    public void testSecuredHttpClient() {
+
+        // mock URL checker is not configured for tile.openstreetmap.org
+        // should throw exception
+        MockURLChecker urlChecker = new MockURLChecker();
+        urlChecker.setEnabled(true);
+        URLCheckerFactory.addURLChecker(urlChecker);
+
+        try {
+            Tile tile =
+                    factory.findTileAtCoordinate(
+                            51, 7, new WebMercatorZoomLevel(5), createService());
+
+            TileService service = createService();
+            OSMTile expectedTile = new OSMTile(20, 15, new WebMercatorZoomLevel(5), service);
+            // should throw exepction
+            service.getHttpClient().get(expectedTile.getUrl());
+
+            fail();
+        } catch (Exception e) {
+            Assert.assertTrue(e.getMessage().contains("Evaluation Failure:"));
+        } finally {
+            URLCheckerFactory.removeURLChecker(urlChecker);
+        }
     }
 
     private TileService createService() {
--- a/modules/library/main/src/main/java/org/geotools/data/ows/ControlledHttpClientFactory.java
+++ b/modules/library/main/src/main/java/org/geotools/data/ows/ControlledHttpClientFactory.java
@@ -1,0 +1,25 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ *
+ *    (C) 2020, Open Source Geospatial Foundation (OSGeo)
+ *
+ *    This library is free software; you can redistribute it and/or
+ *    modify it under the terms of the GNU Lesser General Public
+ *    License as published by the Free Software Foundation;
+ *    version 2.1 of the License.
+ *
+ *    This library is distributed in the hope that it will be useful,
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ *    Lesser General Public License for more details.
+ */
+package org.geotools.data.ows;
+
+/** @author ImranR */
+public class ControlledHttpClientFactory {
+
+    public static synchronized HTTPClient wrap(HTTPClient client) {
+        return new ControlledHttpClient(client);
+    }
+}
--- a/modules/library/main/src/main/java/org/geotools/data/ows/ControlledHttpClientFactory.java
+++ b/modules/library/main/src/main/java/org/geotools/data/ows/ControlledHttpClientFactory.java
@@ -16,10 +16,27 @@
  */
 package org.geotools.data.ows;
 
-/** @author ImranR */
+/**
+ * ControlledHttpClientFactory provides static method to enable URL/URI evaluation on any
+ * implementation of org.geotools.data.ows.HTTPClient
+ *
+ * @author Imran Rajjad - Geo Solutions
+ */
 public class ControlledHttpClientFactory {
 
-    public static synchronized HTTPClient wrap(HTTPClient client) {
+    /**
+     * This method accepts an instance org.geotools.data.ows.HTTPClient and wraps it in
+     * ControlledHttpClient to enabled URL evaluation
+     *
+     * <p>If no URLCheckers are configured or there are no enabled URLCheckers, then no wrapping
+     * will occur
+     *
+     * @param client instance of org.geotools.data.ows.HTTPClient
+     * @return ControlledHttpClient
+     */
+    public static HTTPClient wrap(HTTPClient client) {
+        // only wrap if URLCheckers are configured
+        if (URLCheckers.getEnabledURLCheckerList().isEmpty()) return client;
         return new ControlledHttpClient(client);
     }
 }
--- a/modules/library/main/src/main/java/org/geotools/data/ows/URLChecker.java
+++ b/modules/library/main/src/main/java/org/geotools/data/ows/URLChecker.java
@@ -1,0 +1,31 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ *
+ *    (C) 2020, Open Source Geospatial Foundation (OSGeo)
+ *
+ *    This library is free software; you can redistribute it and/or
+ *    modify it under the terms of the GNU Lesser General Public
+ *    License as published by the Free Software Foundation;
+ *    version 2.1 of the License.
+ *
+ *    This library is distributed in the hope that it will be useful,
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ *    Lesser General Public License for more details.
+ */
+package org.geotools.data.ows;
+
+/**
+ * Defines how to implement URL validation
+ *
+ * @author ImranR
+ */
+public interface URLChecker {
+
+    String getName();
+
+    boolean isEnabled();
+
+    boolean evaluate(String url);
+}
--- a/modules/library/main/src/main/java/org/geotools/data/ows/URLChecker.java
+++ b/modules/library/main/src/main/java/org/geotools/data/ows/URLChecker.java
@@ -23,9 +23,17 @@ package org.geotools.data.ows;
  */
 public interface URLChecker {
 
+    /** @return URLChecker name that best describes its purpose (e.g GML Schema evaluator etc) */
     String getName();
 
+    /** @return Boolean flag indicating if this URLChecker should be used */
     boolean isEnabled();
 
+    /**
+     * Provide implementation to evaluate URL/URI passed in string form
+     *
+     * @param url to be evaluated
+     * @return boolean response to the
+     */
     boolean evaluate(String url);
 }
--- a/modules/library/main/src/main/java/org/geotools/data/ows/URLCheckerFactory.java
+++ b/modules/library/main/src/main/java/org/geotools/data/ows/URLCheckerFactory.java
@@ -1,0 +1,119 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ *
+ *    (C) 2020, Open Source Geospatial Foundation (OSGeo)
+ *
+ *    This library is free software; you can redistribute it and/or
+ *    modify it under the terms of the GNU Lesser General Public
+ *    License as published by the Free Software Foundation;
+ *    version 2.1 of the License.
+ *
+ *    This library is distributed in the hope that it will be useful,
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ *    Lesser General Public License for more details.
+ */
+package org.geotools.data.ows;
+
+import java.io.IOException;
+import java.net.URI;
+import java.net.URL;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.ServiceLoader;
+import java.util.stream.Collectors;
+
+/**
+ * @author ImranR
+ *     <p>URL Checker Factory
+ *     <p>Scans for implementations of
+ *     <p>org.geotools.data.ows.URLChecker
+ */
+public class URLCheckerFactory {
+
+    // list contains all known org.geotools.data.ows.URLChecker implementations
+
+    private static List<URLChecker> urlCheckerList;
+
+    static {
+        urlCheckerList = URLCheckerFactory.initEvaluators(URLChecker.class);
+    }
+
+    private static <T> List<T> initEvaluators(Class<T> type) {
+        // allowing scanning through SDI
+        ServiceLoader<T> loader = ServiceLoader.load(type);
+        loader.reload();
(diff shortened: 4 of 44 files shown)


modules/library/main/src/main/java/org/geotools/data/ows/ControlledHttpClientFactory.java
```diff
@@ -0,0 +1,25 @@
...
+ *
+ *    This library is distributed in the hope that it will be useful,
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ *    Lesser General Public License for more details.
+ */
+package org.geotools.data.ows;
+
+/** @author ImranR */
+public class ControlledHttpClientFactory {
```
aaime: The proposal uses different names:

image

Please be consistent on both sides.

modules/library/main/src/main/java/org/geotools/data/ows/URLCheckerFactory.java
```diff
@@ -0,0 +1,119 @@
...
+
+    private static List<URLChecker> urlCheckerList;
+
+    static {
+        urlCheckerList = URLCheckerFactory.initEvaluators(URLChecker.class);
+    }
+
+    private static <T> List<T> initEvaluators(Class<T> type) {
+        // allowing scanning through SDI
+        ServiceLoader<T> loader = ServiceLoader.load(type);
```
aaime: Odd way to lookup extensions, GeoTools typical approach is to use SPI (Service Provider Interface) and leverage the factory registry to do so, please see:

* https://github.com/geotools/geotools/blob/a1e08cc358f38dc22bf4f896abdd2bc7569f6f7f/modules/library/main/src/main/java/org/geotools/data/DataStoreFinder.java#L90
* https://github.com/geotools/geotools/blob/a1e08cc358f38dc22bf4f896abdd2bc7569f6f7f/modules/library/main/src/main/java/org/geotools/factory/CommonFactoryFinder.java#L147

Or please elaborate why and how this is equivalent to the above.

modules/library/main/src/main/java/org/geotools/data/ows/URLCheckerFactory.java
```diff
@@ -0,0 +1,119 @@
...
+import java.net.URI;
+import java.net.URL;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.ServiceLoader;
+import java.util.stream.Collectors;
+
+/**
+ * @author ImranR
```
aaime: Please add javadocs for public methods.

modules/library/main/src/main/java/org/geotools/data/ows/URLCheckerFactory.java
```diff
@@ -0,0 +1,119 @@
...
+        }
+        return urlCheckerList;
+    }
+
+    /** @return unmodifiable Collection of URLChecker implementations */
+    public static List<URLChecker> getUrlCheckerList() {
+        return (Collections.unmodifiableList(urlCheckerList));
+    }
+
+    public static List<URLChecker> getEnabledUrlCheckerList() {
```
aaime: Inconsistent Url/URL naming, please use URL in all method names (for local variables it's ok to use "url").

modules/library/main/src/main/java/org/geotools/data/ows/URLCheckerFactory.java
```diff
@@ -0,0 +1,119 @@
...
+        loader.reload();
+        List<T> urlCheckerList = new ArrayList<>();
+        for (T aLoader : loader) {
+            urlCheckerList.add(aLoader);
+        }
+        return urlCheckerList;
+    }
+
+    /** @return unmodifiable Collection of URLChecker implementations */
+    public static List<URLChecker> getUrlCheckerList() {
```
aaime: Does this method need to be public?

modules/library/main/src/main/java/org/geotools/data/ows/URLCheckerFactory.java
```diff
@@ -0,0 +1,119 @@
...
+    public static synchronized void addURLChecker(URLChecker urlChecker) {
+        urlCheckerList.add(urlChecker);
+    }
+
+    public static synchronized boolean removeURLChecker(URLChecker urlChecker) {
+        return urlCheckerList.remove(urlChecker);
+    }
+
+    /**
+     * A Utility method
```
aaime: This javadoc comment is not useful, please tell users that does it do.

aaime: There are some more below in the same situation, please fix all.

modules/library/main/src/main/java/org/geotools/data/ows/ControlledHttpClientFactory.java
imranrajjad: understood, I will change on proposal

modules/library/main/src/main/java/org/geotools/data/ows/URLCheckerFactory.java
```diff
@@ -0,0 +1,119 @@
...
+        //        throw new IOException(
+        //                ">>>Evaluation Failure: " + url + ": did not pass security evaluation");
+        // get enabled URLChecker only
+        List<URLChecker> enabledURLCheckrList = getEnabledUrlCheckerList();
+        // no evaluators.. dont do anything
+        if (enabledURLCheckrList.isEmpty()) return true;
+        boolean passed = false;
+        // evaluate using all available implementations
+        for (URLChecker urlChecker : enabledURLCheckrList) {
+            passed |= urlChecker.evaluate(url);
```
aaime:
```suggestion
       if (!urlChecker.evaluate(url)) {
             throw new IOException(
                    "Evaluation Failure: " + url + ": did not pass security evaluation");
      }
```

and then there is no need to carry a "passed" around.

modules/library/main/src/main/java/org/geotools/data/ows/URLCheckerFactory.java
```diff
@@ -0,0 +1,119 @@
...
+
+    /**
+     * A Utility method
+     *
+     * @param String(URI/URI) to evaluate using all available URLCheckers
+     * @return boolean
+     * @throws IOException
+     */
+    public static synchronized boolean evaluate(String url) throws IOException {
+        //        throw new IOException(
```
aaime: Please don't leave commented out code around. Check the rest of the PR for other possible cases and clean them all up.

modules/library/main/src/main/java/org/geotools/data/ows/URLCheckerFactory.java
```diff
@@ -0,0 +1,119 @@
...
+        boolean passed = false;
+        // evaluate using all available implementations
+        for (URLChecker urlChecker : enabledURLCheckrList) {
+            passed |= urlChecker.evaluate(url);
+        }
+        // passed all evaluation
+        if (passed) return passed;
+        else
+            throw new IOException(
+                    ">>>Evaluation Failure: " + url + ": did not pass security evaluation");
```
aaime: No need to add the ">>>" in front of the error message.

modules/extension/tile-client/src/test/java/org/geotools/tile/impl/osm/OSMTileFactoryTest.java
```diff
@@ -104,6 +108,33 @@ public void testGetExtentFromTileName() {
...
+                            51, 7, new WebMercatorZoomLevel(5), createService());
+
+            TileService service = createService();
+            OSMTile expectedTile = new OSMTile(20, 15, new WebMercatorZoomLevel(5), service);
+            // should throw exepction
+            service.getHttpClient().get(expectedTile.getUrl());
+
+            fail();
+        } catch (Exception e) {
+            Assert.assertTrue(e.getMessage().contains("Evaluation Failure:"));
```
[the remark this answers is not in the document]
aaime: Several other tests in the same boat, please make sure all gets fixed.

modules/library/main/src/main/java/org/geotools/data/ows/URLCheckerFactory.java
```diff
@@ -0,0 +1,119 @@
...
+import java.util.ServiceLoader;
+import java.util.stream.Collectors;
+
+/**
+ * @author ImranR
+ *     <p>URL Checker Factory
+ *     <p>Scans for implementations of
+ *     <p>org.geotools.data.ows.URLChecker
+ */
+public class URLCheckerFactory {
```
aaime: A "factory" creates objects, this one actually looks them up and applies them instead. Could be named "URLCheckers" for example.

modules/library/xml/src/test/java/org/geotools/xml/GMLParserTest.java
```diff
@@ -355,4 +365,47 @@ public void testProblemFeatures() {
...
+
+        // mock URL checker is not configured for tile.openstreetmap.org
+        // should throw exception
+        MockFileURIChecker urlChecker = new MockFileURIChecker();
+        urlChecker.setEnabled(true);
+        URLCheckerFactory.addURLChecker(urlChecker);
+
+        // since the shema loader handles the error
+        // we will assert the security by listening to log
+        Handler exceptionListner =
```
aaime:
```suggestion
        Handler exceptionListener =
```

modules/library/xml/src/test/java/org/geotools/xml/GMLParserTest.java
```diff
@@ -355,4 +365,47 @@ public void testProblemFeatures() {
...
+                new Handler() {
+
+                    public boolean errorThrown = false;
+
+                    @Override
+                    public void publish(LogRecord record) {
+                        errorThrown |=
+                                record.getMessage().contains("did not pass securityevaluation");
```
aaime: There should be a space between "securityevaluation", likely the construction of the error message is faulty, fix it along with the test checking it.

modules/library/xml/src/test/java/org/geotools/xml/GMLParserTest.java
```diff
@@ -355,4 +365,47 @@ public void testProblemFeatures() {
...
+                    }
+
+                    @Override
+                    public void flush() {}
+
+                    @Override
+                    public void close() throws SecurityException {
+                        URLCheckerFactory.removeURLChecker(urlChecker);
+                        if (!errorThrown) fail();
```
aaime: Explain "fail" how, for example:

```suggestion
                        if (!errorThrown) fail("Was expecting an exception loading external schemas, but found none");
```

modules/library/xml/src/test/java/org/geotools/xml/GMLParserTest.java
```diff
@@ -355,4 +365,47 @@ public void testProblemFeatures() {
             // fine, they were expected
         }
     }
+
+    @Test
+    public void testSecured() throws Exception {
```
aaime: Would be nice to also have a test that passes because the parser only hits allowed resources (e..g, allow local file system, schemas.opengis.net, maybe xml.org). Exact same test code, just having a different mock checker setup.

modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/internal/WFSContentComplexFeatureCollection.java
```diff
@@ -72,7 +74,9 @@ public WFSContentComplexFeatureCollection(
 
         try {
             @SuppressWarnings("PMD.CloseResource") // wrapped and returned
-            InputStream stream = request.getFinalURL().openStream();
+            URL finalUrl = request.getFinalURL();
+            URLCheckerFactory.evaluate(finalUrl);
```
aaime: So there is a data store parameter to control whether tests should be performed, but this test runs always instead? This is inconsistent, but also see other comments about this class.

modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/impl/WFSDataAccessFactory.java
```diff
@@ -589,6 +590,22 @@ public T lookUp(final Map params) throws IOException {
                         new WFSFactoryParam<>(name, Integer.class, title, description, 6);
     }
 
+    /**
+     * Optional {@code String} Flag to enabled/disable usage of secured http client wrapper that
+     * will only make requests to allowed urls and paths
+     */
+    public static final WFSFactoryParam<Boolean> SECURED_HTTP_CLIENT;
```
aaime: Makes some sense to have a parameter... but nothing else has it, e.g., WMS and WMTS cascading do not, coverage readers going through ImageIOExt neither, they all have it always on. At this point I'd suggest to do the same here and see if we need to make it parametric (in a consistent way) later.

modules/library/main/src/main/java/org/geotools/data/ows/ControlledHttpClientFactory.java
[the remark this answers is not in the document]
imranrajjad: will add

modules/library/main/src/main/java/org/geotools/data/ows/ControlledHttpClientFactory.java
```diff
@@ -0,0 +1,25 @@
...
+
+    public static synchronized HTTPClient wrap(HTTPClient client) {
```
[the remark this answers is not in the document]
imranrajjad: understood will add. synchronized is used to handle concurrency

modules/extension/tile-client/src/test/java/org/geotools/tile/impl/osm/OSMTileFactoryTest.java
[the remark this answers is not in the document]
imranrajjad: yes the error message has the URL in question, I will update the assert statement(s)

modules/library/main/src/main/java/org/geotools/data/ows/URLCheckerFactory.java
imranrajjad: I use to verify the count of known implementations and also use it in unit tests

modules/library/main/src/main/java/org/geotools/data/ows/URLCheckerFactory.java
imranrajjad: will fix all

modules/library/main/src/main/java/org/geotools/data/ows/URLCheckerFactory.java
imranrajjad: will remove

modules/library/main/src/main/java/org/geotools/data/ows/URLCheckerFactory.java
imranrajjad: ok will rename as suggested

modules/library/main/src/main/java/org/geotools/data/ows/URLCheckerFactory.java
imranrajjad: will remove

modules/library/xml/src/test/java/org/geotools/xml/GMLParserTest.java
imranrajjad: will fix

modules/library/xml/src/test/java/org/geotools/xml/GMLParserTest.java
imranrajjad: understood, will apply suggested change

modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/impl/WFSDataAccessFactory.java
imranrajjad: The WMS and WMTS implementation is in Geoserver, where the parameter and wrapping of HttpClient is done inside HttpStoreInfo abstrat class. Well for ImageIOExt, looks I will have to add new functions that wrap previous one for backward compatability, but then a question arises, what is the backward compatability behavior? Should the evaluation be turned off by default inside ImageIOExt unless requested?

modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/internal/WFSContentComplexFeatureCollection.java
imranrajjad: will add check

modules/library/main/src/main/java/org/geotools/data/ows/ControlledHttpClientFactory.java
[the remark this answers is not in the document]
aaime: This will kill the whole system concurrency. What data structure do you have to protect from concurrency? Find it and secure that one only, this method does not need to be synchronized.

modules/library/main/src/main/java/org/geotools/data/ows/URLCheckerFactory.java
jodygarnett: FactoryFinder uses ServiceLoader internally now; however weo do ask that a factory finder be used so that results are cached in a consistent fashion. This allows provides a chance for folks in odd OSGi environments to inject instances that serviceLodaer cannot otherwise pickup.

modules/library/main/src/main/java/org/geotools/data/ows/ControlledHttpClientFactory.java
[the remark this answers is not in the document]
imranrajjad: makes sense, will remove synchronized

modules/library/main/src/main/java/org/geotools/data/ows/URLCheckerFactory.java
imranrajjad: understood, will use CommonFactoryFinder

modules/library/main/src/main/java/org/geotools/data/ows/URLCheckerFactory.java
imranrajjad: The suggested change will break the behavior. In practice we will have multiple URL checkers registered, and we want the URL/URI to go through all of them.

modules/library/main/src/test/java/org/geotools/data/ows/MockURLChecker.java
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * A Mock implementation to be used in unit tests ONLY
+ *
+ * <p>Follows simple string comparison for URL evaluation
+ *
+ * @author ImranR
+ */
+public class MockURLChecker implements URLChecker {
+
+    List<String> allowedList = Arrays.asList("schemas.opengis.net"); // url calls
```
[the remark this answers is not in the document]
imranrajjad: customizable? as in?

modules/library/main/src/main/java/org/geotools/data/ows/URLCheckerFactory.java
imranrajjad: will fix

modules/library/xml/src/test/java/org/geotools/xml/GMLParserTest.java
imranrajjad: this is going to be troublesome on different OS

modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/impl/WFSDataAccessFactory.java
imranrajjad: will remove this FLAG, same will be done on GeoServer side then.

imranrajjad:
@jodygarnett @aime thank you for your review, I applied your suggestions, let me know if something else is needed.

modules/library/main/src/main/java/org/geotools/data/ows/URLChecker.java
```diff
@@ -0,0 +1,31 @@
...
+ *    Lesser General Public License for more details.
+ */
+package org.geotools.data.ows;
+
+/**
+ * Defines how to implement URL validation
+ *
+ * @author ImranR
+ */
+public interface URLChecker {
```
aaime, line 24: Please add javadocs for all public methods, this is kind of the most important interface in this work, and has no docs.

modules/library/main/src/test/java/org/geotools/data/ows/MockURLChecker.java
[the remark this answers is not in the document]
aaime: As in being able to change its contents. Arrays.asList produces an unmodifiable wrapper. If tests are in the same package, I guess it's also possible to just change the field entirely.

modules/library/main/src/main/java/org/geotools/data/ows/URLCheckers.java
```diff
@@ -0,0 +1,109 @@
...
+import org.geotools.factory.CommonFactoryFinder;
+import org.geotools.util.factory.FactoryRegistry;
+
+/**
+ * @author ImranR
+ *     <p>URL Checkers
+ *     <p>Scans for implementations of
+ *     <p>org.geotools.data.ows.URLChecker
+ */
+public class URLCheckers {
```
aaime, line 35: Another public class with lots of public methods that are not documented. @imranrajjad please go around the entire PR, add javadocs for all public and protected methods, or make them private if they don't really need to be public.

modules/library/xml/src/test/java/org/geotools/xml/GMLParserTest.java
aaime: Please clarify? You can get the absolute path of the file and replace it in the test XML for example.

aaime (changes requested):
Added a few comments. Also, did you check the QA build results? There is one evident failure:

```

[INFO] --- maven-pmd-plugin:3.12.0:check (default) @ gt-main ---
[INFO] PMD Failure: org.geotools.data.ows.URLCheckers$URLCheckers:41 Rule:UnusedPrivateField Priority:3 Avoid unused private fields such as 'registry'..
[INFO] PMD Failure: org.geotools.data.ows.URLCheckers$URLCheckers:41 Rule:UnusedPrivateField Priority:3 Avoid unused private fields such as 'registry'..
```

but I'm sure there are others, for example, @ianturton has upgraded JUnit recently and the assertThat method you're using is deprecated now, see https://github.com/junit-team/junit4/blob/main/doc/ReleaseNotes4.13.md#pull-request-1150-deprecate-assertassertthat
Just replace Assert.assertThat with MatcherAssert.assertThat there.

modules/library/main/src/main/java/org/geotools/data/ows/URLChecker.java
imranrajjad: will do

modules/library/main/src/main/java/org/geotools/data/ows/URLCheckers.java
imranrajjad: will fix

roarbra:
@imranrajjad I've got a [branch](https://github.com/roarbra/geotools/tree/GEOT-6512) ready for the changes needed to align your work with the new gt-http module. I wasn't able to open a PR into your branch. And I'm not completely sure how to accomplish this.

The changes needed is that your classes should be moved to gt-http in package org.geotools.http. And that the call for ControlledHttpClientFactory.wrap is taken within AbstractHTTPClientFactory.createClient. It really doesn't look like a factory class, so you might consider dropping it.

If you're interested we could perhaps work on it together.

imranrajjad:
@roarbra please try now

imranrajjad:
> @imranrajjad I've got a [branch](https://github.com/roarbra/geotools/tree/GEOT-6512) ready for the changes needed to align your work with the new gt-http module. I wasn't able to open a PR into your branch. And I'm not completely sure how to accomplish this.
> 
> The changes needed is that your classes should be moved to gt-http in package org.geotools.http. And that the call for ControlledHttpClientFactory.wrap is taken within AbstractHTTPClientFactory.createClient. It really doesn't look like a factory class, so you might consider dropping it.
> 
> If you're interested we could perhaps work on it together.

@roarbra 
please try now

PR #2805, opened 2020-02-10
[backport 22.x] Harmonizing GridCoverageRenderer and GridCoverageRendererUtilities 

dromagnoli:


aaime:
There are no API incompatible here right? Just some methods made visible that weren't before.

dromagnoli:
> There are no API incompatible here right? Just some methods made visible that weren't before.
Right. 
just moved some private methods to Utility class and made it visible.

PR #2808, opened 2020-02-16
[GEOT-6514] WMTSService class always returns null for getContactInformation()

singhdan:
1. Addition of methods to translate from the net.opengis.ows11.* classes, which are generated when the GetCapabilities is processed, through to org.geotools.metadata.iso.citation.* classes used by WMTSService.
2. Update to existing WMTSServiceTest to include checks for the new values that are now available
3. Includes some typo corrections found elsewhere in the code base


modules/extension/wmts/src/test/java/org/geotools/ows/wmts/WebMapTileServerTest.java
```diff
@@ -72,9 +72,7 @@ public void chGetEnvelopeTest() throws Exception {
         // <ows:UpperCorner>11.47757 48.230651</ows:UpperCorner>
         // </ows:WGS84BoundingBox>
 
-        assertNotNull(wmts.getEnvelope(layer, CRS.decode("urn:ogc:def:crs:OGC:1.3:CRS84")));
-        assertNotNull(wmts.getEnvelope(layer, CRS.decode("CRS:84")));
-        assertNotNull(wmts.getEnvelope(layer, CRS.decode("EPSG:2056")));
+        assertNotNull(wmts.getEnvelope(layer, CRS.decode("urn:ogc:def:crs:EPSG:2056")));
```
ianturton: What's the rational for removing these additional checks?

modules/library/metadata/pom.xml
```diff
@@ -135,6 +135,12 @@
       <groupId>org.apache.commons</groupId>
       <artifactId>commons-lang3</artifactId>
     </dependency>
+      <dependency>
+          <groupId>org.geotools.ogc</groupId>
+          <artifactId>net.opengis.ows</artifactId>
+          <version>23-SNAPSHOT</version>
```
ianturton: This should be the parameter ${project.version}

pom.xml
```diff
@@ -1830,7 +1830,7 @@
...
+	  <argLine>-Xmx${test.maxHeapSize} ${jvm.opts} -Dfile.encoding=UTF-8 -Dorg.geotools.test.extensive=${extensive.tests} -Dorg.geotools.test.interactive=${interactive.tests} -Dorg.geotools.image.test.skip=${skip.image.tests} -Dorg.geotools.image.test.interactive=${interactive.image} -Djava.awt.headless=${java.awt.headless} -Dsun.java2d.d3d=${sun.java2d.d3d} -Djava.io.tmpdir="${java.io.tmpdir}" -Djava.library.path="${java.library.path}" ${test.args}  -XX:+TieredCompilation -XX:TieredStopAtLevel=1 -XX:+IgnoreUnrecognizedVMOptions -Djava.util.logging.config.file="${geotoolsBaseDir}"/${logging-profile}.properties --illegal-access=debug</argLine>
```
ianturton: is this a necessary change? how does it relate to the rest of the PR?

modules/extension/wmts/src/test/java/org/geotools/ows/wmts/WebMapTileServerTest.java
singhdan: Sorry, accidental, will put them back in

pom.xml
singhdan: This is because surefire doesn't appear to pass through the encoding when it forks tests. So without this, on my Windows machine, if I run the tests I get the following fail:
`expected:<Bundesamt für Landestopografie swisstopo> but was:<Bundesamt fÃ¼r Landestopografie swisstopo>`

I tried using `<encoding>UTF-8</encoding>` but that didn't get passed through to the command line when forking so ended up having to put it straight into `<argLine>`

modules/library/metadata/pom.xml
singhdan: Will do

modules/extension/wmts/src/main/java/org/geotools/ows/wmts/model/WMTSService.java
```diff
@@ -14,20 +14,49 @@
...
  * @author ian
  * @author Matthias Schulze (LDBV at ldbv dot bayern dot de)
  */
 public class WMTSService extends Service {
 
+    public WMTSService(ServiceIdentificationType serviceType, ServiceProviderType serviceProvider) {
+
+        this(serviceType);
+
+        ResponsiblePartyImpl contactInfo =
```
singhdan: This needs a check against an empty serviceProvider. In the WMTS spec, the serviceProvder section is optional. Real life example is view-source:https://geodata.nationaalgeoregister.nl/tiles/service/wmts?request=GetCapabilities&service=WMTS

aaime:
I wanted to merged before doing the mass-checkstyle changes, but there is a conflict:

image

Oh well, the checkstyle changes are related to javadocs only, hopefully it won't impact this PR much (if at all).

PR #2811, opened 2020-02-19
[GEOS-9392] geopkg datetime

davidblasby:
GeoPackage supports the definition of a datetime (using a string marked as date or timestamp). 

See https://osgeo-org.atlassian.net/browse/GEOS-9392

## Checklist

For all pull requests:

- [x] You have sent a Contribution Licence Agreement (CLA) as necessary (covered by geocat CLA)


ianturton:
I see new test data but no new test?

jodygarnett:
> I see new test data but no new test?

Yes I asked @davidblasby  to make a PR so I could write a test case. This PR should be marked "draft". 

That said we would love your review, as it was quite tricky watching the aggregate functions not respect the type mappings.

modules/plugin/geopkg/src/main/java/org/geotools/geopkg/GeoPkgDialect.java
```diff
@@ -82,6 +85,18 @@ public GeoPkgDialect(JDBCDataStore dataStore) {
         geomWriterConfig = new GeoPkgGeomWriter.Configuration();
     }
 
+    // The JDBC aggregate functions doesn't support the geopkg DateTime types.
+    //  This is because they are stored as strings in the database instead of real Date types
+    //   and the JDBC aggregate driver does NOT provide the needed conversions.
+    //
+    // since this doesn't work, we don't support it.
+    @Override
+    public void registerAggregateFunctions(
```
aaime: Ouch, this fixes the issue but severely limits usability of geopackages with time/elevation/custom dimensions, when having fast min/max/unique is a must. Production systems will work only on very small geopackages in this case.

Thinking out loud, if it's really just a matter of conversion of result datatype, one might consider also doing type conversions around here?
https://github.com/geotools/geotools/blob/bae4ea158dd39cf766f7ead19c93a81222a4cce5/modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java#L1499
Or is it sqlite that does the aggregate in an unexpected way? min/max should work even with strings, aren't dates are saved as ISO dates in sqlite, so they should sort lexicographically anyways?

Just a thought.

davidblasby: Yes - there's a line all setup for the conversion here;
https://github.com/geotools/geotools/blob/bae4ea158dd39cf766f7ead19c93a81222a4cce5/modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java#L1502

However, there isn't much support for the conversion, so it's likely going to require interface changes.  

I ran into this issue in the Time dimension GetCapabilities - which runs a UniqueVisitor on the dataset.  There were several problems in the GeoPKG with this - i.e. having to change the column from "time" to "datetime(time)" in the filters (otherwise the DB returns the wrong results).

aaime: Right, my only worry is that caps generation will become really really slow... we did quite a bit of work optimizing it with visitors back in the day exactly because caps gen time was getting out of hand real quick...

jodygarnett: I could not find a way of checking the columns the operation is operating on, like if we can confirm everything is simple (and not subject to mapping) then the visitors can be used. In the specific use of GetCapabilities generation making a summary of TIMESTAMP is a common case.

I had thought about making the CalcResult smarter (giving it a target class that it accepts, and it could be responsible for providing a conversation as values are collected). This would solve the problem "once" and not involve making special cases in all the places where visitors are called.

jodygarnett:
Thanks for test @davidblasby, looks like format check may of failed, please build once on the command line to fix up formatting.

modules/plugin/geopkg/src/main/java/org/geotools/geopkg/GeoPkgDialect.java
aaime: Thinking out loud here... the visitor does not really know what the type is, because it's determined at runtime evaluating expressions (besides the visitors that only work on a specific type). JDBCDataStore on the other hand [has the aggregateExpression](https://github.com/geotools/geotools/blob/d8d38228e57638927b745f31bc0b5623f48e61b9/modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java#L1432), from which one can determine the type of the visitor input by using the ExpressionTypeVisitor. 
Now, that is not enough, many visitors just return the same type as the input, but for example avg and sttdev do not.
It would be easy however to add a default method to FeatureCalc or FeatureAttributeVisitor or to the implementations (getExpression is really not present in FeatureCalc for example, but determined through reflection.. not sure why, since there is FeatureAttributeVisitor that can do the job, probably historical reasons).

```
/**
 * Returns the expected output type given the input type, or null if uknown
Class[] getResultType(Class[] inputs);
*/
```

With that one could do the right type conversion no?

aaime: Mind, these "thinking out loud" mean "let's talk but I'm not asking you to do it". If you are hard pressed to get a fix in, I'd suggest to go head with this, open a new ticket warning users that performance of aggregations on sqlite will become really bad (for the cases where it works, like numbers and strings, which is used for example by SLDService), and hope that someone steps in to follow up. Right now I don't have an installation of gpkg + aggregations that would be affected, someone out there might though.

rduivenvoorde: I just added a geopackage ‘model.gpkg’ with several tables and views in it to 
https://osgeo-org.atlassian.net/browse/GEOS-9392

To test I would use  ‘view3’  (>34000 polygons and a datetime column) with given model.sld.

But to test how it ‘detoriates’ the generation of the capabilities it is also a good one because it contains 3 views and tables.

davidblasby: https://osgeo-org.atlassian.net/browse/GEOT-6520

@jodygarnett  - I think this is ok to merge now.

jodygarnett:
Does this wish to be back ported, it is a fix after all?

modules/plugin/geopkg/src/main/java/org/geotools/geopkg/GeoPkgDialect.java
```diff
@@ -82,6 +84,18 @@ public GeoPkgDialect(JDBCDataStore dataStore) {
         geomWriterConfig = new GeoPkgGeomWriter.Configuration();
     }
 
+    // The JDBC aggregate functions doesn't support the geopkg DateTime types.
+    //  This is because they are stored as strings in the database instead of real Date types
+    //   and the JDBC aggregate driver does NOT provide the needed conversions.
+    //
+    // since this doesn't work, we don't support it.
```
jodygarnett: We tend to use javadoc comments, C style comments are okay inside the code block.

modules/plugin/geopkg/src/main/java/org/geotools/geopkg/GeoPkgDialect.java
jodygarnett: Thanks I will review

modules/plugin/geopkg/src/main/java/org/geotools/geopkg/GeoPkgFilterToSQL.java
```diff
@@ -44,6 +44,39 @@ public GeoPkgFilterToSQL(GeoPkgDialect dialect) {
...
 
+    // cf.  visit(Literal expression,...)
+    // When doing temporal queries (like "time BETWEEN t1 AND t2")
+    //
+    // The encoding of the column name ("time") and the literals must be the same!
+    //
+    //  datetime("Time",'localtime') BETWEEN datetime(?,'localtime') AND datetime(?,'localtime')
+    //
+    // For non-time columns, this just relegates to the superclass
+    // For GeoPKG, the time column is actually stored as a STRING.
```
jodygarnett: Please use javadoc comment

modules/plugin/geopkg/src/test/java/org/geotools/geopkg/GeoPkgDatetimeTest.java
```diff
@@ -0,0 +1,125 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ *
+ *    (C) 2017, Open Source Geospatial Foundation (OSGeo)
```
jodygarnett: new file new date, 2020

modules/plugin/geopkg/src/test/java/org/geotools/geopkg/GeoPkgDatetimeTest.java
```diff
@@ -0,0 +1,125 @@
...
+        HashMap<String, Object> params = new HashMap<>();
+        params.put(GeoPkgDataStoreFactory.DBTYPE.key, GeoPkgDataStoreFactory.DBTYPE.sample);
+        params.put(GeoPkgDataStoreFactory.DATABASE.key, URLs.urlToFile(url));
+
+        gpkg = DataStoreFinder.getDataStore(params);
+        assertNotNull(gpkg);
+    }
+
+    /**
+     * Test for GEOS-9392 - recognition of DATETIME and DATE.
```
jodygarnett: thanks, good practice to note the bug report :)

jodygarnett:
The failure on macOS seems unrelated - GridCoverageTest.testSerialization

jodygarnett:
David did you need this backported?

aaime:
I see this one got merged as is. @davidblasby please open a ticket to report the slowdown in aggregation operations? Users with large dataset need to know they have to stay away from geopackage.
@rduivenvoorde any chance you can test how much slower things got after the update, after the nightly builds incorporate it? It can be done either generating capabilities with a layer having dimensions enabled, or using the gs:aggregate process (mapstore for example uses it a lot to generate unique values and the like, to fill dropdowns).

rduivenvoorde:
@aaime I'll do some tests after the weekend (given I can use a nightly then).

The execution path is not very clear to me, but does this also affect the use of gpkg files WITHOUT the use of dimensions? My (naive) impression was that it only affected speed when you enabled dimension(s). So you could mitigate from that by for example putting those layers in separate 'virtual services'? 
I'd be interested in more proper implementation/idea's too if this can be done in say one developer-day work too.

aaime:
@rduivenvoorde no, one needs to run aggregations, before they were done directly in the database, leveraging indexes if present, now they will load the full set of values for the property in question from the database, and aggregate in memory. The most badly affected are min/max extractions, followed by unique values, and group by of sorts. WPS exposes all these operations via gs:aggregate and gs:unique, a GetCapabilities on dimensions will use them to generate the list of times/elevations/custom dimension values, a GetMap won't use them at all.

davidblasby:
Yesterday, I put in a bug report here - https://osgeo-org.atlassian.net/browse/GEOT-6520

I did a quick test on the sample dataset (34k) - it took 3.1s to do the get capabilities.

aaime:
Thanks @davidblasby, did not see it. Yeah, the 34k records case is a small one, will follow up with something more substantial, in the millions of records (we have some installations with a few hundreds of millions records, but don't think those are a use case for geopackage anyways...)

aaime:
Also a reminder, this is a GeoTools PR, needs a GeoTools ticket, for the changelogs when we release GeoTools, and for those that depend on GeoTools only. If the issue was shown in GeoServer, the two tickets can be linked via a dependency.

aaime:
Followed up with my previous suggestion and put together a quick PR, hopefully solving the performance regression without further side effects: https://github.com/geotools/geotools/pull/2822

aaime:
On backport please remember to include https://github.com/geotools/geotools/commit/7529923af1143cfe25cd475bb4f05c11fca83957

rduivenvoorde:
Hi, 
should the first PR be in current nightlies? I've downloaded https://download.osmdata.xyz/data/address_EPSG4326.zip to check times, but thought to test (attached) view3 of modelgpkg first. 
From that one I can indeed register the dimensions, (Time is a iso-timestring there), but I fail to retrieve that layer in the capabilities.
In the log you see: 
```
02 Mar 12:43:10 WARN [wms.capabilities] - Error writing metadata; skipping layer: view3
java.lang.IllegalArgumentException: Date argument should be either a Date or a DateRange, however this one is neither: 2019-09-01 09:00:00
	at org.geoserver.util.ISO8601Formatter.format(ISO8601Formatter.java:60)
	at org.geoserver.wms.capabilities.DimensionHelper.getTemporalDomainRepresentation(DimensionHelper.java:554)
	at org.geoserver.wms.capabilities.DimensionHelper.handleTimeDimensionVector(DimensionHelper.java:701)
	at org.geoserver.wms.capabilities.DimensionHelper.handleVectorLayerDimensions(DimensionHelper.java:101)
```
So I'm not sure if I'm testing a still unfixed geotools in this nightly, or I hit an issue. 
Or should I just wait untill the second PR also lands in the nightlies (will it?)?
Thanks!

jodygarnett:
Moved issue to geotools for correct release notes

PR #2819, opened 2020-02-24
[GEOT-6518]: Heterogeneous CRS ImageMosaic supporting output in Targe…

dromagnoli:
…tCRS matching one of the internal CRSs

## Checklist

For all pull requests:


modules/library/coverage/src/main/java/org/geotools/coverage/grid/io/GridCoverage2DReader.java
```diff
@@ -117,6 +117,12 @@
      */
     public static final String MULTICRS_READER = "MultiCRSReader";
 
+    /**
+     * When the above MULTICRS_READER property is present and evaluates to "true", this property may
+     * contain the list of internal EPSG Codes of the mulitple CRS, if the reader implement it.
```
aaime:
```suggestion
     * contain the list of internal EPSG Codes of the multiple CRS, if the reader implement it.
```

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/ImageMosaicFormat.java
```diff
@@ -190,6 +190,18 @@
                     new Boolean[] {Boolean.TRUE, Boolean.FALSE},
                     Boolean.FALSE);
 
+    /**
+     * A multi-crs reader supporting alternative CRSs can produce output to one of them when the
+     * requested grid geometry is expressed in that CRS and this flag is set to true. No
+     * reprojections will be performed on granules having native CRS matching the target one.
```
aaime: "When this read parameter is set to true, the reader will produce output in the requested CRS, assuming it matches one of the native CRSs, as reported by the MULTICRS_EPSGCODES metadata entry, and that the requested grid geometry is expressed in said CRS. When set to false (default), then only the native CRS declared by GridCoverage2DReader.getCoordinateReferenceSystem will be produced in output".

(I skipped adding the links)

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/RasterLayerRequest.java
```diff
@@ -278,6 +298,53 @@ public RasterLayerRequest(
...
+                        paddedRange.height + DEFAULT_PADDING * 2);
+
+                GridGeometry2D padded =
+                        new GridGeometry2D(
+                                paddedRange,
+                                requestedGridGeometry.getGridToCRS(),
+                                requestedGridGeometry.getCoordinateReferenceSystem());
+                spatialRequestHelper.setRequestedGridGeometry(padded.toCanonical());
+            } else {
+                // Do not apply any padding if we are going to produce output in requested CRS.
```
aaime: Hum... say we have a bunch of granules and some are in the requested CRS, others are not. Is the padding going to be applied to the ones that are getting reprojected?

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/Utils.java
```diff
@@ -208,6 +212,51 @@
...
+    }
+
+    /**
+     * Return the EPSG code number of a provided coordinate reference system by just returning the
+     * identifier's code (not doing any lookup or scan).
+     *
+     * @param crs the CoordinateReferenceSystem to be checked
+     * @return null if no code is available
+     */
+    public static String getEpsgCodeNumber(CoordinateReferenceSystem crs) {
```
aaime: This looks like a broken version of "CRS.lookupEpsgCode(crs, false)", in that it will fail if the CRS happens to have alternative authorities that are not "EPSG". Setting the second parameter to "false" avoids scans.

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/Utils.java
```diff
@@ -208,6 +212,51 @@
...
+        String multiCrsReader =
+                reader.getMetadataValue(AbstractGridCoverage2DReader.MULTICRS_READER);
+        boolean isMultiCRS = multiCrsReader != null && Boolean.valueOf(multiCrsReader);
+        String supportedEpsgCodes = "";
+        if (isMultiCRS) {
+            supportedEpsgCodes =
+                    reader.getMetadataValue(AbstractGridCoverage2DReader.MULTICRS_EPSGCODES);
+        }
+        if (!StringUtils.isBlank(supportedEpsgCodes) && crs != null) {
+            Integer targetCRSEpsgCode = CRS.lookupEpsgCode(crs, true);
```
aaime: Here is does a full scan... is this intended?

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/Utils.java
```diff
@@ -208,6 +212,51 @@
         MOSAIC_SUPPORT_FILES_FILTER = initMosaicSupportFilesFilter();
     }
 
+    /** Check if the provided reader is a MultiCRS Reader and it can support the specified crs. */
+    public static boolean isSupportedCRS(GridCoverage2DReader reader, CoordinateReferenceSystem crs)
```
aaime: Do these methods need to be public?

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/RasterLayerResponse.java
```diff
@@ -477,6 +479,13 @@ public SubmosaicProducerFactory getGranuleCollectorsFactory() {
 
     private ReferencedEnvelope mosaicBBox;
 
+    /**
+     * With Heterogeneous CRS, all the geometries in the catalog are expressed in a single CRS, even
+     * if the granules have a different native CRS. This variable is the mosaicBBOX to be used in
```
aaime: Euh? :-D

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/RasterManager.java
```diff
@@ -1040,6 +1050,32 @@ public RasterManager(
...
         }
+        if (heterogeneousCRS) {
+            // If the reader is kept open (and the rasterManager doesn't change)
+            // it would be useful to cache the result of a query for
+            // a specific EPSG code on the index, so that a DB access won't be
+            // repeated while the info is in cache.
+
+            alternativeCRSCache =
+                    CacheBuilder.newBuilder()
+                            .maximumSize(50)
```
aaime: Magic number, best to make it a system variable if not a read parameter. 
I'd also suggest to switch the default to something larger, a classic world wide satellite mosaic case is having all UTM zones (60 zones, further split into south and north) and the two polar caps, making for 122 CRSs present at the same time in the mosaic.

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/RasterManager.java
```diff
@@ -1040,6 +1050,32 @@ public RasterManager(
...
+                            .expireAfterWrite(60, TimeUnit.SECONDS)
```
aaime: Magic number, best to make it a system variable if not a read parameter.

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/SpatialRequestHelper.java
```diff
@@ -115,8 +116,23 @@ public void setGeographicCRS2D(CoordinateReferenceSystem geographicCRS2D) {
...
 
+        public ReferencedEnvelope getComputedBBox() {
+            return computedBBox;
+        }
+
+        public void setComputedBBox(ReferencedEnvelope computedBBox) {
+            this.computedBBox = computedBBox;
+        }
+
+        public boolean isNeedsReprojection() {
```
aaime: Painful... "isReprojectionNeeded" reads better.

modules/plugin/imagemosaic/src/test/java/org/geotools/gce/imagemosaic/ImageMosaicReaderTest.java
```diff
@@ -2491,7 +2495,7 @@ public void multipleDimensionsStackedSar() throws Exception {
 
         final String[] metadataNames = reader.getMetadataNames();
         assertNotNull(metadataNames);
-        assertEquals(15, metadataNames.length);
+        assertEquals(16, metadataNames.length);
```
aaime: What happened here? I see it does not match any of the constants.

modules/plugin/imagemosaic/src/test/resources/org/geotools/gce/imagemosaic/test-data/heterogeneous_crs_2/indexer.properties
```diff
@@ -0,0 +1,10 @@
+GranuleAcceptors=org.geotools.gce.imagemosaic.acceptors.HeterogeneousCRSAcceptorFactory
+GranuleHandler=org.geotools.gce.imagemosaic.granulehandler.ReprojectingGranuleHandlerFactory
+HeterogeneousCRS=true
+MosaicCRS=EPSG\:4326
+Schema=*the_geom:Polygon,location:String,crs:String,resX:double,resY:double
+ResolutionXAttribute=resX
```
aaime: Are these resolution attributes needed?

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/Utils.java
dromagnoli: Thanks for the notice. I will fix it.

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/RasterManager.java
```diff
@@ -1660,9 +1718,31 @@ String getMetadataValue(String name) {
...
+                && heterogeneousCRS) {
+
+            // Extract the internal EPSG Codes found on the catalog
+            String crsAttribute = null;
+            try {
+                crsAttribute = getCrsAttribute();
+                if (crsAttribute != null) {
+                    Set<String> crsSet = extractDomain(crsAttribute);
+                    for (String crs : crsSet) {
+                        // Cache the results for future queries
```
aaime: This seems like opportunistic caching, as in "while we computed the list of epsg codes supported, let's add them in the cache for the code by code checks", but I'm just guessing here. Would be best if it was clear.

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/Utils.java
dromagnoli: Yep, It will be used by WPS Raster Download

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/SpatialRequestHelper.java
dromagnoli: Will fix. I was simply using the name of the already existing boolean.

modules/plugin/imagemosaic/src/test/java/org/geotools/gce/imagemosaic/ImageMosaicReaderTest.java
dromagnoli: It was a different isolated test already returning a different number of entries with respect to the others. I didn't add constants for all the possible metadataNames.length across the whole JUNIT Test class. Do you think it's better to put all the possible values into constants?

modules/plugin/imagemosaic/src/test/resources/org/geotools/gce/imagemosaic/test-data/heterogeneous_crs_2/indexer.properties
dromagnoli: Not really in this test case. 
I was using the same dataset from WPS Raster Download where it's used by the GridGeometryProvider.

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/RasterManager.java
dromagnoli: K. Improving that comment

dromagnoli:
> I'm confused... a machinery to avoid un-needed reprojection is being added, but the rendering subsystem takes no advantage of it?
Right. That would be part of a different ticket. The current effort is mainly for WPS Raster download aiming to preserve the original resolution and referencing of the underlying datasets

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/RasterManager.java
dromagnoli: I was originally planning to use a system variable indeed. 
Forgot to do at the end. Done, thx

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/Utils.java
dromagnoli: Nope. fixing.

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/RasterLayerResponse.java
dromagnoli: Whenever CRS you ask in output, the geometries in the catalog are stored into a unique CRS. So the query on the DB need to be made on that CRS even if the requested mosaic BBOX is using a different CRS. I'm going to clarify a bit this comment :)

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/ImageMosaicFormat.java
dromagnoli: Updated accordingly

PR #2822, opened 2020-02-29
[GEOT-6520] GeoPackage store aggregates performance regression

aaime:
Follows up https://github.com/geotools/geotools/pull/2811 to restore in-database aggregation.
I've basically followed the suggestion outlined in the previous PR, reporting it here since the comment was marked as "resolved" and it's thus not visible unless one knows where to search.

------------

Thinking out loud here... the visitor does not really know what the type is, because it's determined at runtime evaluating expressions (besides the visitors that only work on a specific type). JDBCDataStore on the other hand [has the aggregateExpression](https://github.com/geotools/geotools/blob/d8d38228e57638927b745f31bc0b5623f48e61b9/modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java#L1432), from which one can determine the type of the visitor input by using the ExpressionTypeVisitor. 
Now, that is not enough, many visitors just return the same type as the input, but for example avg and sttdev do not.
It would be easy however to add a default method to FeatureCalc or FeatureAttributeVisitor or to the implementations (getExpression is really not present in FeatureCalc for example, but determined through reflection.. not sure why, since there is FeatureAttributeVisitor that can do the job, probably historical reasons).

```
/**
 * Returns the expected output type given the input type, or null if uknown
Class[] getResultType(Class[] inputs);
*/
```

With that one could do the right type conversion no?

------

I've isolated the changes so that only the dialects that actually need the conversion dance use it, others will just go with the good behavior of the database.

## Checklist

For all pull requests:

### DIFF
--- a/modules/library/main/src/main/java/org/geotools/feature/visitor/AverageVisitor.java
+++ b/modules/library/main/src/main/java/org/geotools/feature/visitor/AverageVisitor.java
@@ -18,6 +18,7 @@ package org.geotools.feature.visitor;
 
 import java.util.Arrays;
 import java.util.List;
+import java.util.Optional;
 import org.geotools.data.simple.SimpleFeatureCollection;
 import org.geotools.factory.CommonFactoryFinder;
 import org.geotools.filter.IllegalFilterException;
@@ -84,6 +85,20 @@ public class AverageVisitor implements FeatureCalc, FeatureAttributeVisitor {
     @Override
     public List<Expression> getExpressions() {
         return Arrays.asList(expr);
+    }
+
+    @Override
+    public Optional<List<Class>> getResultType(List<Class> inputTypes) {
+        if (inputTypes == null || inputTypes.size() != 1)
+            throw new IllegalArgumentException(
+                    "Expecting a single type in input, not " + inputTypes);
+
+        Class type = inputTypes.get(0);
+        if (Number.class.isAssignableFrom(type)) {
+            return Optional.of(inputTypes);
+        }
+        throw new IllegalArgumentException(
+                "The input type for sum must be numeric, instead this was found: " + type);
     }
 
     /**
--- a/modules/library/main/src/main/java/org/geotools/feature/visitor/AverageVisitor.java
+++ b/modules/library/main/src/main/java/org/geotools/feature/visitor/AverageVisitor.java
@@ -17,6 +17,7 @@
 package org.geotools.feature.visitor;
 
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
 import org.geotools.data.simple.SimpleFeatureCollection;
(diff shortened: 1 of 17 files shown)


davidblasby:
Thanks for putting this together.  I am very sick right now, but will take a more detailed look later in the week.

Thanks a lot!

jodygarnett:
I like the minimal API change, and like doing this work at the visitor level. 

One thing I considered about handling bindings coming out of a generic result set, what about when two sql types match to the same java type ... do we care? With this visitor approach (focused on java bindings) we do not have to care...

For geopackage our trouble was the result set containing a string, and we had to look at the resultset metadata to see the pragma marking it as a datetime.

modules/library/main/src/main/java/org/geotools/feature/visitor/AverageVisitor.java
```diff
@@ -86,6 +87,20 @@ public void init(SimpleFeatureCollection collection) {
...
 
+    @Override
+    public Optional<List<Class>> getResultType(List<Class> inputTypes) {
+        if (inputTypes == null || inputTypes.size() != 1)
+            throw new IllegalArgumentException(
+                    "Expecting a single type in input, not " + inputTypes);
+
+        Class type = inputTypes.get(0);
+        if (Number.class.isAssignableFrom(type)) {
+            return Optional.of(inputTypes);
```
jodygarnett: Did we want Double here? Like standard deviation below?

modules/plugin/geopkg/src/test/java/org/geotools/geopkg/GeoPkgDatetimeTest.java
```diff
@@ -102,10 +115,53 @@ public void testUnique() throws IOException {
...
+
+    @Test
+    public void testMax() throws IOException {
+        MaxVisitor max = new MaxVisitor("date");
+        SimpleFeatureSource fs = gpkg.getFeatureSource(gpkg.getTypeNames()[0]);
+
+        SimpleFeatureCollection features = fs.getFeatures();
+        features.accepts(max, NULL_LISTENER);
+
+        assertEquals(java.sql.Date.valueOf("2020-02-23"), max.getMax());
```
jodygarnett: Thanks for testing/fixing this!

jodygarnett (approved):

modules/library/main/src/main/java/org/geotools/feature/visitor/AverageVisitor.java
aaime: Well spotted. I've fixed it and added tests, along with some other type related fixes. They are not used by the aggregation fix, but they better be right anyways, could be of use down the road.

aaime:
Added a separate commit for ease of verification, if you merge it please use a "squash merge"

PR #2830, opened 2020-03-06
Additional map projection: Homolosine

ldesousa:

This is a literal implementation of the projection described by John Paul Goode, relying on the existing projection classes for the Mollweide and the Sinusoidal. Thus no actual mathematical transformation is performed in the new class. This implementation applies tow essential steps:
1. Identify the Homolosine lobe to which the points belongs and apply the respective central meridian.
2. Deduct the offset between the transformed northing values of the Sinusoidal and Homolosine at the treshold latitude.

All projection unit tests are currently passing with this commit.

## Checklist
- Will submit CLA to OSGeo in the coming days.

For all pull requests:


jodygarnett:
The travis build appears stuck installing a jar, unrelated to this change:

Last log message:
```
[INFO] Installing /home/travis/build/geotools/geotools/modules/unsupported/arcsde/common/target/gt-arcsde-common-23-SNAPSHOT-sources.jar to /home/travis/.m2/repository/org/geotools/gt-arcsde-common/23-SNAPSHOT/gt-arcsde-common-23-SNAPSHOT-sources.jar
```

bradh:
Restarted travis build.

bradh:
The restarted build failed on QA checks, which appear directly related to this change - you have format problems:

```
[ERROR] Found 2 non-complying files, failing build

[ERROR] Non complying file: /home/travis/build/geotools/geotools/modules/library/referencing/src/main/java/org/geotools/referencing/operation/projection/Testing.java

[ERROR] Non complying file: /home/travis/build/geotools/geotools/modules/library/referencing/src/main/java/org/geotools/referencing/operation/projection/Homolosine.java
```

aaime:
You can read more about GeoTools automated checks here: https://docs.geotools.org/latest/developer/conventions/code/qa.html

The formatting is done automatically, you just have to run a maven build on the module, e.g.

- ``mvn clean install`` will do it for you while running the normal build with tests and the like
- ``mvn fmt:format`` explicitly formats the file without doing anything else

It's a good idea to do a full QA run on the referencing module (just the one you modified). For example (did not try it out so you might have to correct it a bit):

```
cd modules/library/referencing
mvn clean install -Dqa
```

ldesousa:
The Testing class was just for scratching, should have not been in the commit. The correct formatting is now applied. Thanks.

aaime:
@ldesousa I was looking at the projection math, which I have no previous experience with, mind.
The referred article is behind a paywall, the wikipedia page does not have details.
That said, we normally just port over the Proj code to Java, and looking at it, it seems quite a bit simpler (assuming I got the right class):
https://github.com/OSGeo/PROJ/blob/master/src/projections/goode.cpp

In particular:

- Handling of thresholds seems simpler, based on simple constants instead of arrays
- I don't see lam_shift/shift handling, seems it's just delegating the projection to mollweide/sinusoidal without further transformations?

About the ``Homolosite.txt`` contents, we normally use the values of a reputable third party software, like ``cs2cs``, as a reference, and make sure GeoTools results are within a reasonable tolerance from them. I've tried with the following, the results are not quite a match (tens of thousands of meters away):

```
cs2cs +init=epsg:4326 +to +datum=WGS84 +proj=goode
-170 50
-12930764.17	5536706.32 0.00
-173 40
-14752692.18	4452779.63 0.00
```

Maybe I have setup the cs2cs command incorrectly or looked up the wrong projection name?

ldesousa:
@aaime The implementation of the Sinusoidal in the Homolosine class of PROJ differs from the standalone implementation. This should not be case and means that at least one of them is not correct (or both). This is an on-going discussion: https://lists.osgeo.org/pipermail/proj/2020-January/009285.html

The values in the `Homolosine.txt` file were computed with GeoTools itself, using the existing implementations of the Sinusoidal and the Homolosine. They are "congruent" with GeoTools.

aaime:
Do you have access to any other software to test the values? I'll otherwise wait for the discussion on proj to be settled.

aaime:
Looking around I've found possibly a way to test with ESRI software.
I have these two codes apparently supporting Homolosine in ESRI software:

```
54052=PROJCS["World_Goode_Homolosine_Land",GEOGCS["GCS_WGS_1984",DATUM["D_WGS_1984",SPHEROID["WGS_1984",6378137.0,298.257223563]],PRIMEM["Greenwich",0.0],UNIT["Degree",0.0174532925199433]],PROJECTION["Goode_Homolosine"],PARAMETER["False_Easting",0.0],PARAMETER["False_Northing",0.0],PARAMETER["Central_Meridian",0.0],PARAMETER["Option",1.0],UNIT["Meter",1.0]]
54053=PROJCS["World_Goode_Homolosine_Ocean",GEOGCS["GCS_WGS_1984",DATUM["D_WGS_1984",SPHEROID["WGS_1984",6378137.0,298.257223563]],PRIMEM["Greenwich",0.0],UNIT["Degree",0.0174532925199433]],PROJECTION["Goode_Homolosine"],PARAMETER["False_Easting",0.0],PARAMETER["False_Northing",0.0],PARAMETER["Central_Meridian",-160.0],PARAMETER["Option",2.0],UNIT["Meter",1.0]]
```

See also: https://desktop.arcgis.com/en/arcmap/10.3/guide-books/map-projections/goodes-homolosine.htm

Also found an arcgis online demo server that seems to support them:

http://sampleserver1.arcgisonline.com/ArcGIS/rest/services/Geometry/GeometryServer/project?inSR=4326&outSR=54052&geometries=%7B%22geometryType%22%3A+%22esriGeometryPoint%22%2C+%22geometries%22%3A+%5B%7B%22x%22%3A+-170%2C+%22y%22%3A+50%2C+%22z%22%3A+-1%7D%5D%7D&f=HTML

However, the results are again pretty different... so not sure, I cannot guarantee the codes are actually matching a Homolosine....

aaime:
Hmm... finding detailed online information about this projection seems hard, even the Snyder handbook just cites it without giving much details. May explain why everyone is doing it differently? :-D

I found also this, which is a bit more substantive (but don't know if correct!): https://proceedings.esri.com/library/userconf/proc98/proceed/TO850/PAP844/P844.HTM

aaime:
Also found this:
https://github.com/d3/d3-geo-projection/blob/master/src/homolosine.js

ldesousa:
@aaime Quick reactions:
- The code from d3-geo-projection is not implementing the Homolosine.
- The description in the ESRI proceedings article is quite detailed. The lobe division is very similar to what I did (apart from the Greenland extensions).
- Since there are no extra transformations in this new class, any errors must be rooted in the existing implementations of the Sinusoidal and/or Mollweide.

ldesousa:
Some interesting follow ups at the PROJ maillist:
https://lists.osgeo.org/pipermail/proj/2020-March/009472.html

Since all transformations between PROJ and GeoTools agree, I believe it all applies to GeoTools too. Main conclusions:
1. Homolosine must be implemented in full spherical mode (at least for now).
2. Questions remain over the accuracy of the ellipsoidal implementation of the Sinusoidal.

I need to force the Sinusoidal in spherical mode in the Homolosine class. We should then get the same values as in PROJ.

ldesousa:
Now we have results in tune with PROJ with the Sinusoidal implemented internally in spherical mode. Two of the checks are failing, but that seems related to third-party libraries. Please let me know if there is something I must correct.

aaime:
Tried again, found that I was using the wrong projection name in cs2cs, "goode" is the uninterrupted holomosine, while you are implementing the interrupted version. Using "igh", "interrupted goode holomosine", provides matching results.

```
 cs2cs +init=epsg:4326 +to +datum=WGS84 +proj=igh
10 0
1113194.91	0.00 0.00
-170 50
-16456381.38	5536683.82 0.00
-173.5 58.7
-16016676.35	6414902.69 0.00
```

modules/library/referencing/src/main/java/org/geotools/referencing/operation/projection/Homolosine.java
```diff
@@ -0,0 +1,253 @@
...
+     */
+    public static class Provider extends AbstractProvider {
+
+        /** For cross-version compatibility. */
+        private static final long serialVersionUID = -7345885830045627291L;
+
+        /** The parameters group. */
+        static final ParameterDescriptorGroup PARAMETERS =
+                createDescriptorGroup(
+                        new NamedIdentifier[] {
```
aaime: I searched the EPSG database, to my surprise the projection is not recognized/documented. 
However, the name as is seems too generic, as we have 3 variants in the wild
- Non interrupted Homolosine (implemented in proj as "goode", see https://proj.org/operations/projections/goode.html. ESRI does not seem to have it.
- Interrupted Homolosine, land centric, implemented in proj as "igh", see "https://proj.org/operations/projections/igh.html", and called "Goode_Homolosine" by ESRI, choosing "option=1"
- Interrupted Homolosine, water centric, not implemented in proj as far as I can see, and called "Goode_Homolosine" by ESRI, choosing "option=2"

For ESRI references see:
* https://desktop.arcgis.com/en/arcmap/10.3/guide-books/map-projections/goodes-homolosine.htm
* https://github.com/Esri/projection-engine-db-doc/blob/master/text/pe_list_projection.txt#L600
* https://raw.githubusercontent.com/Esri/projection-engine-db-doc/master/text/pe_list_projcs.txt (warning, big file, 6+MB, search for homolosine in it, you'll fine two examples)

Suggestions:

- For ESRI named identifier, use "Goode_Homolosine"
- For GeoTools, maybe "Interrupted_Homolosine" to tell it apart from the non interrupted version?

While now they won't present a big problem, I want to avoid issues in a future where the other two variants are implemented, and make sure we don't end up having conflicts.
The different name for GeoTools tries to be a match with the PROJ convention (calling interrupted what interrupted is), while the land/water difference can be handled by a parameter just like ESRI did, the day someone wants to implement the ocean centric math.

ldesousa:
Thank you @aaime for the verifications. In fact the projection that PROJ identifies as `goode` should not be attributed to J.P. Goode, he never proposed such a thing. In an ideal world we would two projections, one named "Homolosine_LandMasses" and another "Homolosine_Oceans". But considering the legacy names, I chose to stick to your suggestions.

aaime:
I've just tried to render a homolosine map in GeoServer, probably entered a too large bbox, but got the following exception:

```
Caused by: org.geotools.data.DataSourceException: Had troubles handling filter reprojection...
	at org.vfny.geoserver.global.GeoServerFeatureSource.reprojectFilter(GeoServerFeatureSource.java:485)
	at org.vfny.geoserver.global.GeoServerFeatureSource.getFeatures(GeoServerFeatureSource.java:394)
	at org.vfny.geoserver.global.GeoServerFeatureSource.getFeatures(GeoServerFeatureSource.java:71)
	at org.geotools.renderer.lite.StreamingRenderer.getFeatures(StreamingRenderer.java:2336)
	at org.geotools.renderer.lite.StreamingRenderer.processStylers(StreamingRenderer.java:2213)
	at org.geotools.renderer.lite.StreamingRenderer.paint(StreamingRenderer.java:900)
	at org.geoserver.wms.map.RenderedImageMapOutputFormat.produceMap(RenderedImageMapOutputFormat.java:596)
	... 124 more
Caused by: java.lang.ArrayIndexOutOfBoundsException: -1
	at org.geotools.referencing.operation.projection.Homolosine.inverseTransformNormalized(Homolosine.java:185)
	at org.geotools.referencing.operation.projection.MapProjection$Inverse.transform(MapProjection.java:1047)
	at org.geotools.referencing.operation.projection.MapProjection$Inverse.transform(MapProjection.java:1106)
	at org.geotools.referencing.operation.transform.AbstractMathTransform.transform(AbstractMathTransform.java:215)
	at org.geotools.referencing.CRS.transform(CRS.java:1397)
	at org.geotools.referencing.CRS.transform(CRS.java:1477)
	at org.geotools.geometry.jts.ReferencedEnvelope.transform(ReferencedEnvelope.java:624)
	at org.geotools.geometry.jts.ReferencedEnvelope.transform(ReferencedEnvelope.java:567)
	at org.geotools.filter.spatial.ReprojectingFilterVisitor.visit(ReprojectingFilterVisitor.java:130)
	at org.geotools.renderer.lite.FastBBOX.accept(FastBBOX.java:70)
	at org.vfny.geoserver.global.GeoServerFeatureSource.reprojectFilter(GeoServerFeatureSource.java:478)
	... 130 more
```

The bbox was bbox=-20000000,-20000000,20000000,20000000 (in homolosine projection meters).
Generally speaking the code should be tolerant to excessive bounding boxes, as WMS client code is very prone to let the user zoom out as much as it wants, and send over the bbox in the view.
A WMS request will always have the bbox in the target CRS and try to reproject it back into the source CRS (wgs84 in my case), the largest I managed to display was a:

"http://localhost:8080/geoserver/wms/reflect?layers=nurc%3Aworld&srs=EPSG:900914&bbox=-15000000,-1500000,8000000,4000000"

image

Where 900914 is defined as:

```
900914=PROJCS["Homolosine",GEOGCS["WGS 84",DATUM["WGS_1984",SPHEROID["WGS 84",6378137,298.257223563 ] ], PRIMEM["Greenwich",0.0], UNIT["degree",0.01745329251994328 ]],PROJECTION["Homolosine"],UNIT["m",1.0] ]
```
I have spent too much time on this one today, don't have a synthetic test that you can try, will try to 
figure it out better later.

ldesousa:
@aaime I passed those and other points outside the co-domain to the projection class and they are handled "well", i.e. no exceptions are issued. However, since they are outside the co-domain, the result makes no sense, and it never will. Note that PROJ (cs2cs) returns NaN for these kind of points.

aaime:
I wanted to check out how other programs handle the homolosine, checked QGIS, but it seems they are using the non interrupted version of it?

image

The ocean version is particularly "fun" to look at:

image

aaime:
Did some more debugging to figure out the error, found two points that fail when added to ScriptTest, added the results from cs2cs:

```
source pt = (-180, 8)
target pt = (-19950840.05, 890555.93)

source pt = (180, 8)
target pt = (20037508.34, 0)
```

The first one causes the "index array out of bounds exception", the second one generates a result that's different enough to cause a failure.

When working on new projections we normally just translate the PROJ C/C++ code to Java with good results. Also helps in maintenance, if a fix is done in PROJ can be easily ported over. Wondering if just doing the port would not be easier in the long term?

aaime:
Hi @ldesousa sorry for the delays in feedback, I have time to look at PRs [more or less once a week](https://www.youtube.com/watch?v=Q187JGeXueA). Any feedback on my last two comments?

ldesousa:
Hi Andrea,

truth is, I have not had the time myself to check why the new tests fail. I will provide thorough feedback as soon as possible. I hope this is not too inconvenient. And thank you for your diligence.

Cheers.

--
Luís

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

ldesousa:
Hi Andrea,

basically I was not expecting lambda angles outside the ]-180º, 180º] interval. The class must be able to handle any angle, the domain of this projection is infinite. With the long weekend ahead I should be able to fix this in the next few days.

Regards.

--
Luís

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

ldesousa:
Hi again. In this latest commit the Homolosine class now handles any kind of latitude or longitude angles, producing the same results as PROJ. Two new small methods take care of the wrapping. A few additional tests demonstrate this. There are no tests for latitude wrapping, the automated testing is not designed for it. But I verified those results with PROJ too.

It might be useful to raise the wrapping methods to the parent (MapProjection class). But better leave that to someone more familiar with GeoTools than me (and possibly to a different PR).

aaime:
Gave it another try in GeoServer, still no joy, though I can visualize a larger area.
The largest request I can do now is:

http://localhost:8080/geoserver/wms/reflect?layers=nurc%3Aworld&srs=EPSG:900914&bbox=-19000000,-10000000,8800000,10000000

image

Of course, there is no projection handler class yet (it was my intention to try and write one, to handle the interruptions and generate a proper output), but before getting there I need to be able to cover the entire planet.

If I push the max x to 8900000 I get the following:

```
Caused by: java.lang.ArrayIndexOutOfBoundsException: -1
	at org.geotools.referencing.operation.projection.Homolosine.inverseTransformNormalized(Homolosine.java:226)
	at org.geotools.referencing.operation.projection.MapProjection$Inverse.transform(MapProjection.java:1047)
	at org.geotools.referencing.operation.projection.MapProjection$Inverse.transform(MapProjection.java:1106)
	at org.geotools.referencing.operation.transform.AbstractMathTransform.transform(AbstractMathTransform.java:215)
	at org.geotools.referencing.CRS.transform(CRS.java:1397)
	at org.geotools.referencing.CRS.transform(CRS.java:1477)
	at org.geotools.geometry.jts.ReferencedEnvelope.transform(ReferencedEnvelope.java:624)
	at org.geotools.geometry.jts.ReferencedEnvelope.transform(ReferencedEnvelope.java:567)
	at org.geotools.filter.spatial.ReprojectingFilterVisitor.visit(ReprojectingFilterVisitor.java:130)
	at org.geotools.renderer.lite.FastBBOX.accept(FastBBOX.java:70)
	at org.vfny.geoserver.global.GeoServerFeatureSource.reprojectFilter(GeoServerFeatureSource.java:478)
```

My guess is that I'm hitting the interruption here, and the code cannot handle the bbox point in "no man's land": 

image

However, for WMS to work, it needs to have a reasonable answer, e.g., one that would make the reprojected bbox fetch the data needed to fill the maps.

aaime:
Oh hey, this actually works:

http://localhost:8080/geoserver/wms/reflect?layers=nurc%3Aworld&srs=EPSG:900914&bbox=-19000000,-10000000,20000000,10000000

image

So maybe the backwards projection may fail any time a bbox corner is inside an interruption?
A WMS client just treats the map as a cartesian space, it will eventually explore it all, having corners land wherever.

ldesousa:
Thank you @aaime. The class should not be spewing out exceptions, but if you pass coordinates outside the co-domain to the inverse the output will make no sense. It is ok for a WMS client to request a bounding box that is outside the co-domain, as long as it keeps to the Cartesian space. But if it tries to compute the inverse things will go inevitably wrong. 

This is the case with any projection. In a perfect world any projection class should return a specific exception when the inverse is invoked outside the co-domain. That would be the mathematically correct thing to do, imho. Note here that one of the difficulties I currently face is that the Unit tests are not able to handle this kind of transformations, for good reason.  

I think I can address this new exception during this weekend.

aaime:
@ldesousa, help me understand here... if the class is not able to handle points outside of the bbox of the valid values, we can make WMS work by adding a projection handler. If it throws exceptions inside the gaps instead, I'm afraid it will be completely useless for WMS usage, because the BBOXes of WMS requests (think tiles) will inevitably include corners in the gaps.

aaime:
Thinking out loud, if I prepared the infrastructure for a homolosine projection handler, would you be interested in filling it? The bits needed are the logic to find the source bbox for queries in maps being rendered in homolosine, and pre-processing logic to slice geometries that happen to be across interruptions on the way forward (that is, from whatever native CRS they are in, towards homolosine).

ldesousa:
@aaime The points you posted are treated "orderly", I get the following results in radians:

(8900000, 10000000) -> (-3.4764012244017013, 1.5707963267948966)

(8900000, -10000000) -> (0.3490658503988659, -1.5707963267948966)

However, I was able to reproduce the exception with points outside the co-domain with very low easting values. The class now handles these properly too. 

This is another kind of points that can not be included in the unit tests. If projections need to survive this sort of hassle, then the unit tests need to be made more flexible, imho.

ldesousa:
@aaime A bit of context first. About 2 years ago I found out PROJ supports this projection. For data restricted to land masses there isn't anything better in FOSS4G at the moment, equal-area with little penalty to angles and distances. With the C-based software my team shifted all the global computation pipelines to the Homolosine. The [2019 pre-release of SoilGrids](https://www.isric.org/news/sneak-preview-new-edition-soil-property-estimates-world-and-request-feedback-soilgrids250m) is a first major result of that shift. 

However, there have been complaints since Google does not support this projection, precisely because they rely on the GeoTools library. I tried to source funds to hire an implementation, but that never materialised. Last summer Jody was here in the Netherlands, we had the time to discuss this and from what he showed it did not looked complex. So I gave it a try. In the meantime I bumped into the questions around PROJ discussed above and everything got delayed. But finally we have an implementation passing all the tests and more. So my initial goal is achieved and the Google people will have to find a different excuse.

Now coming to your particular issue (take into account that I do not really know this WMS client in any sort of detail). After some more consideration, I can only regard the concept of reprojecting bounding boxes as fundamentally flawed. A bounding box is a rectangular area contained within two different values of easting and two different values of northing. In most circumstances, reprojecting an area thus defined to a different CRS will have unexpected results. Take for instance a bounding box defined on the Sinusoidal and reprojected to the Stereographic, the result will be some sort of cartographic Frankenstein. A possible way of achieving a quasi-accurate bounding box transformation would require three steps:
1. Intersect the bounding box with the projection co-domain.
2. Discretise the bounding box into infinitesimally small segments.
3. Transform each segment (or segment vertex) individually.

In conclusion: this is not an issue with GeoTools, but with the downstream programme that is trying such endeavours. Therefore, in my view it would be wiser to not create an handler class in the library and instead invest in the correction of the downstream programme.

My team has been publishing Homolosine maps with MapServer and it all functions pretty much out of the box. MapServer ships with a small WMS client integrated where you can browse Homolosine maps without issue. I have used OWSLib as a client to WCS services, no problems there either. MapServer conducts a certain level of tile caching internally, which again is not troubled by interruptions. Bounding boxes behave in an expected way even outside the co-domain. As long as it keeps to the Cartesian plane and no reprojections are attempted it is all good.

I hope this is clear. And thank you again for nursing this PR.

aaime:
Tested, right now it works without throwing exceptions, excellent. Trying to reproject something like this (water and land):

image

results in:

image

Which is to be expected, without a ProjectionHandler implementation for the Homolosine the rendering subsystem does not know where to cut the geometries before reprojection. I'll try to work on one, in the meantime, this PR goes in.

aaime:
Created Jira ticket, squash merged on master at https://github.com/geotools/geotools/commit/f9a05fe2ee8d0b6bfefeb89195c8904ad1f09b01

aaime:
Started working on a projection handler, it's still rough and cutting the geometries along the equator, which I'd like to avoid, but... looking good so far:

image

aaime:
There... ain't that a beauty?

image

Need to write tests and the like, hopefully this weekend, otherwise later.

jodygarnett:
It is indeed beautiful

aaime:
APH support here: https://github.com/geotools/geotools/pull/2903

PR #2837, opened 2020-03-10
[GEOT-6520] better handling of timezones in geopkg date columns

davidblasby:
The date columns were being handles a bit incorrectly - it has to do with how Java and geopkg deal with dates and local timezones.

The fix was to NOT convert the dates (no changes for timestamps);

old old method: SELECT ... datetime(date,'localtime')
old method:  SELECT ... datetime(date,'utc')
new method: SELECT ... datetime(date) -- no conversion info

I also put in a timezone test that messes with the java timezone to help make sure this problem doesn't come back.

## Checklist

For all pull requests:

      covered by GEOCAT CLA

- [x] There is an issue in GEOT-6520
   NA

### DIFF
--- a/modules/plugin/geopkg/src/main/java/org/geotools/geopkg/GeoPkgFilterToSQL.java
+++ b/modules/plugin/geopkg/src/main/java/org/geotools/geopkg/GeoPkgFilterToSQL.java
@@ -71,10 +71,7 @@ public class GeoPkgFilterToSQL extends PreparedFilterToSQL {
                         + ",'utc')"; // utc -- everything must be consistent -- see
                 // literal visitor
             } else if (java.sql.Date.class.isAssignableFrom(binding)) {
-                return "date("
-                        + super_result
-                        + ",'utc')"; // utc -- everything must be consistent -- see
-                // literal visitor
+                return "date(" + super_result + ")";
             }
         }
         return super_result;
@@ -146,7 +143,7 @@ public class GeoPkgFilterToSQL extends PreparedFilterToSQL {
                 } else if (Timestamp.class.isAssignableFrom(literalValue.getClass())) {
                     sb.append("datetime(?,'utc')");
                 } else if (java.sql.Date.class.isAssignableFrom(literalValue.getClass())) {
-                    sb.append("date(?,'utc')");
+                    sb.append("date(?)");
                 } else if (encodingFunction) {
                     dialect.prepareFunctionArgument(clazz, sb);
                 } else {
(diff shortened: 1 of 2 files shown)


davidblasby:
Big call-out to @aaime for big help in getting this working and all the advise and coding help!

davidblasby:
Request review from @aaime

jodygarnett:
I am not sure I understand why *not* converting is the answer?

Let me try and think through it:

* Q: Is this only converting from the internal geopackage representation (which I understand is utc) to a date the client code can use?
* Q: Do we also consider converting dates into utc when writing content into the geopackage?

Discussion:

* ⚠️ old old method: SELECT ... datetime(date,'localtime')
  
  I could see how this disagrees with the spec as it is storing the wrong thing.

* ⚠️ old method: SELECT ... datetime(date,'utc')
  
  So what is the trouble here? Do we assume the data is already stored in utc inside the geopackage?

* new method: SELECT ... datetime(date) -- no conversion info
  
  This should blindly return the data from the geopackage, which if it is correctly stored in utc would be correct and include the utc timezone information?

modules/plugin/geopkg/src/main/java/org/geotools/geopkg/GeoPkgFilterToSQL.java
```diff
@@ -71,10 +71,7 @@ public String escapeName(String name) {
                         + ",'utc')"; // utc -- everything must be consistent -- see
                 // literal visitor
             } else if (java.sql.Date.class.isAssignableFrom(binding)) {
-                return "date("
-                        + super_result
-                        + ",'utc')"; // utc -- everything must be consistent -- see
-                // literal visitor
+                return "date(" + super_result + ")";
```
jodygarnett: Javadoc is out of sync with this functionality, still describes use of `utc`.

jodygarnett: Q: Do you need to do the same thing for Timestamp.class above?

davidblasby: no, timezone handling of Timestamp is correct.  Java Date "sort of" has a timezone.

jodygarnett: Thanks for the discussion, I have updated the conversation above.

jodygarnett:
Thanks David / Andrea, I will watch the builds go and then backport to unblock the RC release.

nmco:
Thx for summarizing the discussion @jodygarnett!

PR #2840, opened 2020-03-18
[GEOT-6533] Selection of source matrixset when cascading WMTS layers does not take into account Longitudefirst CRSs

singhdan:
- Changed the way in which the source matrixset is selected so that it compares the request SRS and the matrixset SRS names rather than comparing the whole CRS. This will reduce the amount of transformation that will be needed when the tiles are returned
- Added unit tests to verify that the changes work as expected
- Some spelling corrections

## Checklist

For all pull requests:

- [X] Commit message must be in the form "[GEOT-XYZW] Title of the Jira ticket"

### DIFF
--- a/modules/extension/wmts/src/main/java/org/geotools/ows/wmts/request/AbstractGetTileRequest.java
+++ b/modules/extension/wmts/src/main/java/org/geotools/ows/wmts/request/AbstractGetTileRequest.java
@@ -354,10 +354,10 @@ public abstract class AbstractGetTileRequest extends AbstractWMTSRequest impleme
         // See if the layer supports the requested SRS. Matching against the SRS rather than the
         // full CRS will avoid missing matches due to Longitude first being set on the request
         // CRS and therefore minimise transformations that need to be performed on the received tile
+        String requestSRS = CRS.toSRS(requestCRS);
         for (TileMatrixSet matrixSet : capabilities.getMatrixSets()) {
 
             CoordinateReferenceSystem matrixCRS = matrixSet.getCoordinateReferenceSystem();
-            String requestSRS = CRS.toSRS(requestCRS);
             String matrixSRS = CRS.toSRS(matrixCRS);
             if (requestSRS.equals(matrixSRS)) { // matching SRS
                 if (links.containsKey((matrixSet.getIdentifier()))) { // and available for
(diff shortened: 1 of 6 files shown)


modules/extension/wmts/src/test/java/org/geotools/ows/wmts/WebMapTileServerTest.java
```diff
@@ -16,21 +16,27 @@
...
+import org.geotools.ows.wmts.request.GetTileRequest;
 import org.geotools.referencing.CRS;
+import org.geotools.tile.Tile;
 import org.junit.Test;
 import org.opengis.referencing.crs.CoordinateReferenceSystem;
 
 /** @author Emanuele Tajariol (etj at geo-solutions dot it) */
 public class WebMapTileServerTest {
 
+    private static PrintStream out = System.out;
```
aaime: Tests should not be writing to system.out, use the logging framework instead, or if you need it for development, keep the line that does the printout commented out, using System.out directly.

modules/extension/wmts/src/main/java/org/geotools/ows/wmts/request/AbstractGetTileRequest.java
```diff
@@ -351,12 +351,15 @@ private TileMatrixSet selectMatrixSet() throws ServiceException, RuntimeExceptio
...
-        // If the layer provides the requested CRS, use it
+        // See if the layer supports the requested SRS. Matching against the SRS rather than the
+        // full CRS will avoid missing matches due to Longitude first being set on the request
+        // CRS and therefore minimise transformations that need to be performed on the received tile
         for (TileMatrixSet matrixSet : capabilities.getMatrixSets()) {
 
             CoordinateReferenceSystem matrixCRS = matrixSet.getCoordinateReferenceSystem();
-
-            if (CRS.equalsIgnoreMetadata(requestCRS, matrixCRS)) { // matching SRS
+            String requestSRS = CRS.toSRS(requestCRS);
```
aaime: No need to convert it at every round in the loop, could be done outside.

taba90:
> Had a look but I'm not familiar with this module. I've asked @taba90 to double check, he's been playing with the module lately.

Tested this interactively and wmts works as expected.

modules/extension/wmts/src/test/java/org/geotools/ows/wmts/WebMapTileServerTest.java
singhdan: Removed all print statements. I only needed them for development and no need for them to be left in and commented out

modules/extension/wmts/src/main/java/org/geotools/ows/wmts/request/AbstractGetTileRequest.java
singhdan: Agreed, moved to outside the loop

PR #2864, opened 2020-04-08
 [GEOT-2425] update documentation repository references

jodygarnett:
Update the documentation references (pom.xml and RST files).

* [x] update pom.xml examples
* [x] update documentation references
* [x] update update instructions

## Checklist

For all pull requests:

### DIFF
--- a/docs/user/welcome/faq.rst
+++ b/docs/user/welcome/faq.rst
@@ -139,22 +139,25 @@ one associated with the most recent formal release (e.g GeoTools 14-SNAPSHOT) an
 being maintained (e.g. GeoTools 13-SNAPSHOT).
 
 New snapshot jars are built nightly and deployed to a repository separate from the one used for formal releases. If you
-are using Maven as your build tool you can work with a snapshot release by adding the following to your ``pom.xml`` ::
+are using Maven as your build tool you can work with a snapshot release by adding the following to your ``pom.xml`` :
 
-    <repository>
-        <id>boundless</id>
-        <name>Boundless Maven Repository</name>
-        <url>http://repo.boundlessgeo.com/main</url>
-        <snapshots>
-            <enabled>true</enabled>
-        </snapshots>
-    </repository>
+.. code-block:: xml
 
-You can now build your project against a snapshot release by setting it as the your version property as shown here::
+   <repository>
+       <id>osgeo-snapshot</id>
+       <name>OSGeo Snapshot Repository</name>
+       <url>https://repo.osgeo.org/repository/snapshot//url>
+       <snapshots><enabled>true</enabled></snapshots>
+       <releases><enabled>false</enabled></releases>
+   </repository>
 
-    <properties>
-        <geotools.version>8-SNAPSHOT</geotools.version>
-    </properties>
+You can now build your project against a snapshot release by setting it as the your version property as shown here:
+
+.. code-block:: xml
+
+   <properties>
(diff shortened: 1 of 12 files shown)


docs/user/welcome/upgrade.rst
```diff
@@ -26,6 +26,57 @@ But first to upgrade - change your dependency to |release| (or an appropriate st
...
+     <mirror>
+       <id>osgeo-release</id>
+       <name>OSGeo Repository</name>
+       <url>https://repo.osgeo.org/repository/geotools-releases/</url>
+       <mirrorOf>osgeo</mirrorOf>
+     </mirror>
+     <mirror>
+       <id>geoserver-releases</id>
+       <name>Boundless Repository</name>
+       <url>https://repo.osgeo.org/repository/Geoserver-releases/</url>
```
tbarsballe, line 75: This should be `geotools-snapshots`, or just `snapshot` instead of `Geoserver-releases`

docs/user/welcome/faq.rst
```diff
@@ -139,22 +139,25 @@ one associated with the most recent formal release (e.g GeoTools 14-SNAPSHOT) an
...
-        <geotools.version>8-SNAPSHOT</geotools.version>
-    </properties>
+are using Maven as your build tool you can work with a snapshot release by adding the following to your ``pom.xml`` :
+
+.. code-block:: xml
+
+   <repository>
+       <id>osgeo-snapshot</id>
+       <name>OSGeo Snapshot Repository</name>
+       <url>https://repo.osgeo.org/repository/snapshot//url>
```
tbarsballe, line 149: Missing `<` on the closing url tag

docs/user/welcome/upgrade.rst
jodygarnett: Not in this case because the content of repo.boundlessgeo.com were transferred to this specific repository `Geoserver-releases` (and yes it is an accident we cannot fix the capital G).

No `snapshots` were transferred so the above is correct.

tbarsballe: Okay, but the section right above describes the boundless repository as a snapshot repository, so if someone is paying attention this is likely to be a little confusing, Plus, `snapshots` will contain all the geotools snapshots, which might be used if people are only developing on a single module (which seems like a common use-case for the audience most likely to reference these docs)

Basically, this mirroring setup won't quite replicate what was there before. It might be best to just use `releases` and `snapshots` in the config to catch everything for simplicity/reliability.

jodygarnett: Okay I will fix...

docs/user/welcome/upgrade.rst
```diff
@@ -29,15 +31,26 @@ But first to upgrade - change your dependency to |release| (or an appropriate st
 GeoTools 23.x
 -------------
 
-Change to repo.osgeo.org Repository
```
tbarsballe: Small nitpick - this is under GeoTools 23.x, but will also affect 22.x

jodygarnett: Yeah not sure how to address that, ... I could start from the earlier version affected.

tbarsballe: Don't think it matters to much either way, given that 22.x is EOL pretty soon

jodygarnett: That worked, considering this resolved.

PR #2880, opened 2020-04-16
SQL Server Travis-CI online testing

mprins:
This brings SQL Server online testing as a Travis-CI job.

While here:
- upgrade microsoft jdbc driver from pre-release to current release version and cleanup defunct readme (closes https://osgeo-org.atlassian.net/browse/GEOT-6556)
- disable failing `testBBOX3DOutsideLine()` (see/opens https://osgeo-org.atlassian.net/browse/GEOT-6555)

## Checklist

For all pull requests:


aaime:
This is quite interesting work. Maybe something similar could be done for Oracle? Are Travis VMs big enough?
Looking at this ticket https://github.com/wnameless/docker-oracle-xe-11g/issues/118 I ended up on this project, which the author claims has been approved by Oracle, https://github.com/Vincit/travis-oracledb-xe

mprins:
I'll give Oracle a shot next :smiley_cat: 

just need to understand why in this PR is't not running online tests, while om my desktop it does when I run `mvn -B -fae -Dall clean install -T2 -pl :gt-jdbc-sqlserver -Ponline -Dfmt.skip=true` with a `sqlserver.properties` in `~/.geotools`

aaime:
Now runs and fails with:

```
[ERROR]   SQLServer3DOnlineTest>OnlineTestCase.run:112->JDBCGeneric3DOnlineTest.testBBOX3DOutsideLine:181 expected:<0> but was:<2>
```

Believe it's the same test failing for Oracle? It has been added to solve a bug in PostGIS in 2017, there it works fine. I see an implementation of the same test in Hana, which is a recent implementation, so I'm guessing it's passing there too. Looking into the actual query.

aaime:
Hmm... failed to get the logs from the queries, last commit can/should be reverted... but looking at the SQL server docs, I really can't find a way to encode a 3D bbox, maybe it's just something that's not supported by the database.

mprins:
Yes I disabled the testBBOX3DOutsideLine() for Oracle as well, could not get it to work
Opened https://osgeo-org.atlassian.net/browse/GEOT-6534 for that

I have some time tomorrow to look at t h at

mprins:
The actual query sent is:
 `SELECT count(*) FROM "line3d" WHERE "geom".Filter(geometry::STGeomFromText('POLYGON ((2 1, 2 2, 3 2, 3 1, 2 1))', 4326)) = 1`

Which doesn't contain the Z values anymore. These are lost in 2 places...

`FilterToSQL#visit(Literal expression, Object context)` which calls `BBOXImpl.boundingPolygon((Envelope) literal`

https://github.com/geotools/geotools/blob/36f727e5522b52ec997d81ee4ccd99217774f80c/modules/library/jdbc/src/main/java/org/geotools/data/jdbc/FilterToSQL.java#L1437-L1439

 which then looses the 3rd dimension on the envelope

https://github.com/geotools/geotools/blob/36f727e5522b52ec997d81ee4ccd99217774f80c/modules/library/main/src/main/java/org/geotools/filter/spatial/BBOXImpl.java#L206-L214

 even then, because in SQLServerFilterToSQL#visitLiteralGeometry(Literal) there is a call to `Geometry#toText()`

https://github.com/geotools/geotools/blob/36f727e5522b52ec997d81ee4ccd99217774f80c/modules/plugin/jdbc/jdbc-sqlserver/src/main/java/org/geotools/data/sqlserver/SQLServerFilterToSQL.java#L70-L72

which would loose any higher than 2 dimension als well because it uses a WKTWriter without specifying the dimension of the output so that defaults to 2

https://github.com/locationtech/jts/blob/3363990d2204c8a3946e56c4502efff2d9307b73/modules/core/src/main/java/org/locationtech/jts/geom/Geometry.java#L1138-L1140

aaime:
Looks good to me, merging!

aaime:
Feel free to backport changes if you want.

PR #2909, opened 2020-04-30
[GEOT-6578] equalTo support for matchAction and postgis optimization

mbarto:
An optional parameter for MatchAction (ANY, ONE or ALL) has been added to FilterFunction_equalTo.

This is particularly useful if arguments are arrays.

An optimization for the postgis jdbc driver is also introduced, so that the function is translated to the postgresql overlap (&&) operator when both arguments are arrays and the match action is ANY.

## Checklist

For all pull requests:

### DIFF
--- a/modules/library/main/src/main/java/org/geotools/filter/function/FilterFunction_equalTo.java
+++ b/modules/library/main/src/main/java/org/geotools/filter/function/FilterFunction_equalTo.java
@@ -22,6 +22,7 @@ import static org.geotools.filter.capability.FunctionNameImpl.parameter;
 
 import org.geotools.filter.FunctionExpressionImpl;
 import org.geotools.filter.capability.FunctionNameImpl;
+import org.opengis.filter.MultiValuedFilter.MatchAction;
 import org.opengis.filter.capability.FunctionName;
 
 public class FilterFunction_equalTo extends FunctionExpressionImpl {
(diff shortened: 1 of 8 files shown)


jodygarnett:
First up I love the work, thank you for the discussion and clear implementation. 

Next time it would be smart to break up the geotools change from the postgis implementation (as per the checklist) :)

mbarto:
@jodygarnett I pushed a new commit, following your suggestions. I am discovering the full power of converters, by the way :-)
I also tested that it works with array of primitive values, now.

modules/library/main/src/main/java/org/geotools/filter/function/FilterFunction_equalTo.java
```diff
@@ -68,7 +82,21 @@ public Object evaluate(Object feature) {
...
+            }
+        }
+        Filter equalTo =
+                matchAction == null
+                        ? getFilterFactory2().equal(ff.literal(arg0), ff.literal(arg1), false)
+                        : getFilterFactory2()
+                                .equal(ff.literal(arg0), ff.literal(arg1), false, matchAction);
 
-        return Boolean.valueOf(StaticGeometry.equalTo(arg0, arg1));
+        return equalTo.evaluate(null);
```
jodygarnett, line 100: This is *so* much cleaner!

If you would like to be slightly more lazy there is no need to evaulate `arg0` and `arg1`, only to wrap them up as a literal again, if you replace this line with:
```
return equalTo.evaualte(feature)
```

PR #2912, opened 2020-05-02
[GEOT-3377] Charset not used for DBF header

dvfed:
Changes proposed are to make use of charset when read / write DBF header field's names, not only the data.

This also applies to the following tickets:

- [GEOT-3357] DBF header parsing does not use the provided charset
- [GEOT-3093] can't support Chinese fields in shpfiles
- [GEOT-2135] Could shapefile supports coding with DBF header

Some of the above are closed already but in fact the issue is still there. In particular this PR fixes reading in a "memory mapped" case and adds an ability to write field's names according to provided charset.

IMPORTANT NOTE: PR removes **public** method  `DbaseFileHeader.readHeader(ReadableByteChannel, Charset)` in favor of using the constructor to pass a charset. Though the method is public it should not be used widely. Keeping it will complicate the code unnecessary as well as keeping it with unused `charset` parameter will confuse too. Please let me know if there is a better solution.

Test data in chinese as well as overall idea was taken from GEOT-3093 with some small changes. The data are absolutely trivial - just several cities with their lat / lon coordinates, and thus should not be a copyright problem. Other data (en, ru) made similarly to chinese one.

## Checklist

For all pull requests:

### DIFF
--- a/modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/dbf/DbaseFileHeader.java
+++ b/modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/dbf/DbaseFileHeader.java
@@ -100,6 +100,16 @@ public class DbaseFileHeader {
     // lets start out with a zero-length array, just in case
     private DbaseField[] fields = new DbaseField[0];
 
+    private Charset charset;
+
+    public DbaseFileHeader() {
+        this.charset = Charset.defaultCharset();
+    }
+
+    public DbaseFileHeader(Charset charset) {
+        this.charset = charset;
+    }
+
     private void read(ByteBuffer buffer, ReadableByteChannel channel) throws IOException {
         while (buffer.remaining() > 0) {
             if (channel.read(buffer) == -1) {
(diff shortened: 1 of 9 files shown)


dvfed:
@ianturton i've sent CLA as needed.

modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/dbf/DbaseFileHeader.java
```diff
@@ -504,17 +524,6 @@ public int getHeaderLength() {
...
-    }
-
-    /**
-     * Read the header data from the DBF file.
-     *
-     * @param channel A readable byte channel. If you have an InputStream you need to use, you can
-     *     call java.nio.Channels.getChannel(InputStream in).
-     * @throws IOException If errors occur while reading.
-     */
-    public void readHeader(ReadableByteChannel channel, Charset charset) throws IOException {
```
aaime, line 517: A public method in a public class cannot be removed just like this I'm afraid, we try to maintain backwards compatibility as much as possible. Two options:
- Add the method back
- Warn users upgrading of the API break in the [upgrade guide](https://github.com/geotools/geotools/blob/master/docs/user/welcome/upgrade.rst)

Both are acceptable to me.

dvfed: The `charset` is now a class field. I believe that keeping the method with same parameter will be confusing. Going to stick with second option.

aaime:
Checked CLA is there, merging.

PR #2914, opened 2020-05-04
[GEOT-6579]: added array filter function, that can be used to create …

mbarto:
…array literals in ECQL filters

## Checklist

For all pull requests:

### DIFF
--- a/modules/library/main/src/main/java/org/geotools/data/util/ArrayConverterFactory.java
+++ b/modules/library/main/src/main/java/org/geotools/data/util/ArrayConverterFactory.java
@@ -67,7 +67,13 @@ public class ArrayConverterFactory implements ConverterFactory {
             Object result = Array.newInstance(elementType, arrLength);
 
             for (int count = 0; count < arrLength; count++) {
-                Array.set(result, count, Converters.convert(Array.get(source, count), elementType));
+                Object sourceElement = Array.get(source, count);
+                Object converted = Converters.convert(sourceElement, elementType);
+                if (sourceElement != null && converted == null) {
+                    // error in conversion
+                    return null;
+                }
+                Array.set(result, count, converted);
             }
             return (T) result;
         }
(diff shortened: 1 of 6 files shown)


modules/library/main/src/main/java/org/geotools/filter/function/ArrayFunction.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+package org.geotools.filter.function;
+
+import static org.geotools.filter.capability.FunctionNameImpl.parameter;
+
+import org.geotools.filter.FunctionExpressionImpl;
+import org.geotools.filter.capability.FunctionNameImpl;
+import org.opengis.filter.capability.FunctionName;
+
+public class ArrayFunction extends FunctionExpressionImpl {
```
jodygarnett, line 25: public class javadoc

modules/library/main/src/main/java/org/geotools/data/util/ArrayConverterFactory.java
```diff
@@ -55,4 +58,18 @@ public Converter createConverter(Class<?> source, Class<?> target, Hints hints)
...
+
+    private static class ArrayToArrayConverter implements Converter {
+        @Override
+        public <T> T convert(Object source, Class<T> target) throws Exception {
+            int arrLength = Array.getLength(source);
+            Class<?> elementType = target.getComponentType();
+            Object result = Array.newInstance(elementType, arrLength);
+
+            for (int count = 0; count < arrLength; count++) {
+                Array.set(result, count, Converters.convert(Array.get(source, count), elementType));
```
jodygarnett, line 70: This looks great, is it worth checking for `null` here (returned when converters cannot make the conversion) or do you want to just let it ride?

mbarto: @jodygarnett good catch. I think we could check the source element, if it's null then set the target array element to null, if the source is not null and the converted is (error in conversion) we can return null for the whole target array (to notify a conversion error). What do you think?

jodygarnett: Yes that is the correct approach.

Note: We did experiment with returning exceptions for this kind of failure, but the resulting exceptions would end up user visible (resulting in confused unhappy complaints). We ended up with this *return null* approach which IMHO ends up being too silent on details ... a tradeoff.

PR #2916, opened 2020-05-05
[GEOT-6612] Update to jsr 385 units 2.0

ddcruver:
The goal of this pull request is to migrate from the Units of Measure API 1.0 (JSR 363) to the Units of Measure API 2.0 (JSR 385).

## Checklist

For all pull requests:

### DIFF
--- a/modules/extension/xsd/xsd-gml3/src/test/java/org/geotools/gml3/bindings/MeasureTypeBindingTest.java
+++ b/modules/extension/xsd/xsd-gml3/src/test/java/org/geotools/gml3/bindings/MeasureTypeBindingTest.java
@@ -22,7 +22,6 @@ import org.geotools.gml3.GML;
 import org.geotools.measure.Measure;
 import org.opengis.feature.ComplexAttribute;
 import org.w3c.dom.Document;
-import tec.uom.se.unit.BaseUnit;
 
 public class MeasureTypeBindingTest extends ComplexAttributeTestSupport {
 
@@ -34,7 +33,7 @@ public class MeasureTypeBindingTest extends ComplexAttributeTestSupport {
         assertNotNull(measure);
         assertEquals(1234, measure.doubleValue(), 0.1);
 
-        assertEquals("http://someuri", ((BaseUnit) measure.getUnit()).getSymbol());
+        assertEquals("http://someuri", measure.getUnit().getSymbol());
     }
 
     public void testEncode() throws Exception {
--- a/modules/library/metadata/src/test/java/org/geotools/util/GeoToolsUnitFormatRoundTripTest.java
+++ b/modules/library/metadata/src/test/java/org/geotools/util/GeoToolsUnitFormatRoundTripTest.java
@@ -17,15 +17,7 @@
 package org.geotools.util;
 
 import static junit.framework.TestCase.assertEquals;
-import static tec.uom.se.unit.Units.GRAM;
-import static tec.uom.se.unit.Units.JOULE;
-import static tec.uom.se.unit.Units.KELVIN;
-import static tec.uom.se.unit.Units.KILOGRAM;
-import static tec.uom.se.unit.Units.METRE;
-import static tec.uom.se.unit.Units.MOLE;
-import static tec.uom.se.unit.Units.PASCAL;
-import static tec.uom.se.unit.Units.SECOND;
-import static tec.uom.se.unit.Units.WATT;
+import static tech.units.indriya.unit.Units.*;
 
 import java.util.Arrays;
 import java.util.Collection;
@@ -35,8 +27,8 @@ import org.junit.runner.RunWith;
 import org.junit.runners.Parameterized;
 import si.uom.NonSI;
 import si.uom.SI;
-import tec.uom.se.AbstractUnit;
-import tec.uom.se.format.SimpleUnitFormat;
+import tech.units.indriya.AbstractUnit;
+import tech.units.indriya.format.SimpleUnitFormat;
 
 @RunWith(Parameterized.class)
 public class GeoToolsUnitFormatRoundTripTest {
@@ -45,21 +37,22 @@ public class GeoToolsUnitFormatRoundTripTest {
     public static Collection<Object[]> data() {
         return Arrays.asList(
                 new Object[][] {
+                    {GRAM},
                     {GRAM.divide(1_000_000)},
                     {GRAM.divide(1_000_000).divide(METRE.pow(3))},
                     {GRAM.divide(1_000_000_000).divide(METRE.pow(3))},
                     {GRAM.divide(1_000_000).divide(METRE.pow(3))},
                     {METRE.pow(2)},
                     {METRE.pow(3)},
-                    {SI.MOLE.divide(1_000_000)},
+                    {MOLE.divide(1_000_000)},
                     {GRAM.multiply(METRE.pow(-3))},
                     {GRAM.divide(1_000)},
-                    {SI.MOLE.multiply(METRE.pow(-2))},
-                    {SI.PASCAL},
+                    {MOLE.multiply(METRE.pow(-2))},
+                    {PASCAL},
                     {AbstractUnit.ONE},
-                    {METRE.divide(SI.SECOND)},
+                    {METRE.divide(SECOND)},
                     {WATT.multiply(METRE.pow(-2))},
-                    {SI.KILOGRAM.multiply(METRE.pow(-2).multiply(SI.SECOND.pow(-1)))},
+                    {KILOGRAM.multiply(METRE.pow(-2).multiply(SECOND.pow(-1)))},
                     {
                         WATT.divide(1000)
                                 .multiply(METRE.pow(-2))
@@ -72,14 +65,15 @@ public class GeoToolsUnitFormatRoundTripTest {
                                 .multiply(METRE.divide(1_000_000_000).pow(-1))
                     },
                     {MOLE.multiply(METRE.divide(100).pow(-3))},
-                    {SI.PASCAL},
+                    {PASCAL},
                     {AbstractUnit.ONE.divide(100)},
                     {METRE},
+                    {NonSI.DEGREE_ANGLE},
                     {NonSI.DEGREE_ANGLE},
                     {METRE.pow(2).divide(METRE.pow(2))},
                     {METRE.pow(-1).multiply(SECOND)},
                     {KILOGRAM.multiply(METRE.pow(-3))},
-                    {SI.KELVIN.multiply(METRE.pow(-1))},
+                    {KELVIN.multiply(METRE.pow(-1))},
                     {MOLE.multiply(KILOGRAM.pow(-1))},
                     {JOULE.multiply(KILOGRAM.pow(-1))},
                     {KILOGRAM.multiply(METRE.pow(-2).multiply(SECOND.pow(-1)))},
@@ -87,7 +81,8 @@ public class GeoToolsUnitFormatRoundTripTest {
                     {METRE.pow(2).multiply(SECOND.pow(-1))},
                     {KILOGRAM.multiply(METRE.pow(-2))},
                     {MOLE.multiply(METRE.pow(-2))},
-                    {SECOND.pow(-1).multiply(METRE.pow(-3))},
+                    {SECOND.multiply(METRE.pow(-3))},
+                    {METRE.multiply(SECOND.pow(-3))},
                     {GRAM.multiply(KILOGRAM.pow(-1))},
                     {PASCAL.multiply(METRE)},
                     {WATT.multiply(METRE.pow(-2))},
@@ -102,13 +97,13 @@ public class GeoToolsUnitFormatRoundTripTest {
                                 .multiply(KILOGRAM.pow(-1))
                                 .multiply(SECOND.pow(-1))
                     },
-                    {SI.MOLE.divide(1_000_000).multiply(SI.METRE.pow(-2)).multiply(466.2)}
+                    {MOLE.divide(1_000_000).multiply(METRE.pow(-2)).multiply(466.2)}
                 });
     }
 
-    private Unit unit;
+    private Unit<?> unit;
 
-    public GeoToolsUnitFormatRoundTripTest(Unit unit) {
+    public GeoToolsUnitFormatRoundTripTest(Unit<?> unit) {
         this.unit = unit;
     }
 
--- a/modules/library/opengis/pom.xml
+++ b/modules/library/opengis/pom.xml
@@ -177,7 +177,7 @@
     
     <dependency>
       <groupId>systems.uom</groupId>
-      <artifactId>systems-common-java8</artifactId>
+      <artifactId>systems-common</artifactId>
     </dependency>
     
     <!-- Test dependencies -->
--- a/modules/library/referencing/src/main/java/org/geotools/measure/SexagesimalConverter.java
+++ b/modules/library/referencing/src/main/java/org/geotools/measure/SexagesimalConverter.java
@@ -17,10 +17,8 @@
 package org.geotools.measure;
 
 import java.io.ObjectStreamException;
-import java.math.BigDecimal;
-import java.math.MathContext;
 import javax.measure.UnitConverter;
-import tec.uom.se.AbstractConverter;
+import tech.units.indriya.function.AbstractConverter;
 
 /**
  * A converter from fractional degrees to sexagesimal degrees. Sexagesimal degrees are pseudo-unit
@@ -37,7 +35,7 @@ import tec.uom.se.AbstractConverter;
  */
 class SexagesimalConverter extends AbstractConverter {
     /** Serial number for compatibility with different versions. */
-    private static final long serialVersionUID = 3873494343412121773L;
+    private static final long serialVersionUID = -2663951106460584999L;
 
     /** Small tolerance factor for rounding errors. */
     private static final double EPS = 1E-8;
@@ -77,13 +75,28 @@ class SexagesimalConverter extends AbstractConverter {
         this.inverse = inverse;
     }
 
+    @Override
+    public int compareTo(UnitConverter o) {
+        // TODO: ?
+        return 0;
+    }
+
     /** Returns the inverse of this converter. */
-    public final SexagesimalConverter inverse() {
+    public final SexagesimalConverter inverseWhenNotIdentity() {
         return inverse;
     }
 
+    @Override
+    protected boolean canReduceWith(AbstractConverter that) {
+        return that instanceof Inverse;
+    }
+
+    public Number convertWhenNotIdentity(Number value) {
+        return degreeToSexagesimalDegrees(value.doubleValue());
+    }
+
     /** Performs a conversion from fractional degrees to sexagesimal degrees. */
-    public double convert(double value) {
+    public double degreeToSexagesimalDegrees(double value) {
         final int deg, min, sec;
         deg = (int) value; // Round toward 0
         value = (value - deg) * 60;
@@ -94,9 +107,44 @@ class SexagesimalConverter extends AbstractConverter {
         return (((deg * 100 + min) * 100 + sec) + value) / divider;
     }
 
+    /** Performs a conversion from sexagesimal degrees to fractional degrees. */
+    public double sexagesimalDegreesToDegrees(double value) {
+        value *= this.divider;
+        int deg, min;
+        deg = (int) (value / 10000);
+        value -= 10000 * deg;
+        min = (int) (value / 100);
+        value -= 100 * min;
+        if (min <= -60 || min >= 60) { // Accepts NaN
+            if (Math.abs(Math.abs(min) - 100) <= EPS) {
+                if (min >= 0) deg++;
+                else deg--;
+                min = 0;
+            } else {
+                throw new ArithmeticException("Invalid minutes: " + min);
+            }
+        }
+        if (value <= -60 || value >= 60) { // Accepts NaN
+            if (Math.abs(Math.abs(value) - 100) <= EPS) {
+                if (value >= 0) min++;
+                else min--;
+                value = 0;
+            } else {
+                throw new ArithmeticException("Invalid secondes: " + value);
+            }
+        }
+        value = ((value / 60) + min) / 60 + deg;
+        return value;
+    }
+
     /** Returns this converter derivative for the specified {@code x} value. */
     public final double derivative(double x) {
         return 1;
+    }
+
+    @Override
+    public boolean isIdentity() {
+        return false;
     }
 
     /** Returns {@code false} since this converter is non-linear. */
@@ -116,6 +164,11 @@ class SexagesimalConverter extends AbstractConverter {
     @Override
     public int hashCode() {
         return (int) serialVersionUID + divider;
+    }
+
+    @Override
+    protected String transformationLiteral() {
+        return null;
     }
 
     /** On deserialization, returns an existing instance. */
@@ -142,11 +195,20 @@ class SexagesimalConverter extends AbstractConverter {
     /** The inverse of {@link SexagesimalConverter}. */
     private static final class Inverse extends SexagesimalConverter {
         /** Serial number for compatibility with different versions. */
-        private static final long serialVersionUID = -7171869900634417819L;
+        private static final long serialVersionUID = -7145237719599612406L;
 
         /** Constructs a converter. */
         public Inverse(final SexagesimalConverter inverse) {
             super(inverse.divider, inverse);
+        }
+
+        @Override
+        protected boolean canReduceWith(AbstractConverter that) {
+            if (that instanceof SexagesimalConverter.Inverse) {
+                return false;
+            } else {
+                return that instanceof SexagesimalConverter;
+            }
         }
 
         /**
@@ -155,33 +217,8 @@ class SexagesimalConverter extends AbstractConverter {
          * @throws ArithmeticException if the value is out of the valid range for minutes or seconds
          */
         @Override
-        public double convert(double value) {
-            value *= this.divider;
-            int deg, min;
-            deg = (int) (value / 10000);
-            value -= 10000 * deg;
-            min = (int) (value / 100);
-            value -= 100 * min;
-            if (min <= -60 || min >= 60) { // Accepts NaN
-                if (Math.abs(Math.abs(min) - 100) <= EPS) {
-                    if (min >= 0) deg++;
-                    else deg--;
-                    min = 0;
-                } else {
-                    throw new ArithmeticException("Invalid minutes: " + min);
-                }
-            }
-            if (value <= -60 || value >= 60) { // Accepts NaN
-                if (Math.abs(Math.abs(value) - 100) <= EPS) {
-                    if (value >= 0) min++;
-                    else min--;
-                    value = 0;
-                } else {
-                    throw new ArithmeticException("Invalid secondes: " + value);
-                }
-            }
-            value = ((value / 60) + min) / 60 + deg;
-            return value;
+        public Number convertWhenNotIdentity(Number value) {
+            return sexagesimalDegreesToDegrees(value.doubleValue());
         }
 
         /** Returns a hash value for this converter. */
@@ -189,21 +226,5 @@ class SexagesimalConverter extends AbstractConverter {
         public int hashCode() {
             return (int) serialVersionUID + divider;
         }
-    }
-
-    @Override
-    public Number convert(Number value) {
-        return convert(value.doubleValue());
-    }
-
-    @Override
-    public BigDecimal convert(BigDecimal value, MathContext ctx) throws ArithmeticException {
-        double valueAsDouble = value.doubleValue();
-        if (valueAsDouble == Double.NEGATIVE_INFINITY
-                || valueAsDouble == Double.POSITIVE_INFINITY) {
-            throw new ArithmeticException(
-                    "Bigdecimal value magnitude too great to fit in double: " + value);
-        }
-        return new BigDecimal(convert(valueAsDouble));
     }
 }
--- a/modules/library/referencing/src/main/java/org/geotools/referencing/cs/DefaultCoordinateSystemAxis.java
+++ b/modules/library/referencing/src/main/java/org/geotools/referencing/cs/DefaultCoordinateSystemAxis.java
@@ -43,7 +43,7 @@ import org.opengis.referencing.cs.RangeMeaning;
 import org.opengis.util.InternationalString;
 import si.uom.NonSI;
 import si.uom.SI;
-import tec.uom.se.AbstractUnit;
+import tech.units.indriya.AbstractUnit;
 
 /**
  * Definition of a coordinate system axis. This is used to label axes, and indicate the orientation.
@@ -523,7 +523,7 @@ public class DefaultCoordinateSystemAxis extends AbstractIdentifiedObject
      * A default axis for time values in a {@linkplain org.opengis.referencing.cs.TimeCS time CS}.
      *
      * <p>Increasing time go toward {@linkplain AxisDirection#FUTURE future} and units are
-     * {@linkplain NonSI#DAY days}.
+     * {@linkplain SI#DAY days}.
      *
      * <p>The abbreviation is lower case "<var>t</var>".
      */
@@ -748,6 +748,7 @@ public class DefaultCoordinateSystemAxis extends AbstractIdentifiedObject
         ensureNonNull("abbreviation", abbreviation);
         ensureNonNull("direction", direction);
         ensureNonNull("unit", unit);
+
         if (unit.isCompatible(NonSI.DEGREE_ANGLE)) {
             UnitConverter fromDegrees;
             try {
--- a/modules/library/referencing/src/main/java/org/geotools/referencing/factory/OrderedAxisAuthorityFactory.java
+++ b/modules/library/referencing/src/main/java/org/geotools/referencing/factory/OrderedAxisAuthorityFactory.java
@@ -213,7 +213,7 @@ public class OrderedAxisAuthorityFactory extends TransformedAuthorityFactory
         if (userHints != null) {
             final Boolean value = (Boolean) userHints.get(key);
             if (value != null) {
-                return value.booleanValue();
+                return value;
             }
         }
         return false;
@@ -224,8 +224,8 @@ public class OrderedAxisAuthorityFactory extends TransformedAuthorityFactory
      * invoked by constructors only.
      */
     private void completeHints() {
-        hints.put(Hints.FORCE_STANDARD_AXIS_UNITS, Boolean.valueOf(forceStandardUnits));
-        hints.put(Hints.FORCE_STANDARD_AXIS_DIRECTIONS, Boolean.valueOf(forceStandardDirections));
+        hints.put(Hints.FORCE_STANDARD_AXIS_UNITS, forceStandardUnits);
+        hints.put(Hints.FORCE_STANDARD_AXIS_DIRECTIONS, forceStandardDirections);
         // The following hint has no effect on this class behaviour,
         // but tells to the user what this factory do about axis order.
         if (compare(DefaultCoordinateSystemAxis.EASTING, DefaultCoordinateSystemAxis.NORTHING)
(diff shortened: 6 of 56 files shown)


ddcruver:
The code builds but there are a few unit tests that are failing that I am hoping to get some help from the community with.

jodygarnett:
I have been looking forward to this!

modules/extension/xsd/xsd-gml3/src/test/java/org/geotools/gml3/bindings/MeasureTypeBindingTest.java
```diff
@@ -34,7 +33,7 @@ public void testParser() throws Exception {
         assertNotNull(measure);
         assertEquals(1234, measure.doubleValue(), 0.1);
 
-        assertEquals("http://someuri", ((BaseUnit) measure.getUnit()).getSymbol());
+        assertEquals("http://someuri", measure.getUnit().getSymbol());
```
jodygarnett, line 36: Why did the prior code require a cast to BaseUnit? Has the api changed to make getSymbol() visible?

ddcruver: It was also visible in Units of Measure 1.0 as well. Not sure maybe in JSR-275 it wasn't.

modules/library/metadata/src/test/java/org/geotools/util/GeoToolsUnitFormatRoundTripTest.java
```diff
@@ -17,15 +17,15 @@
...
-import static tec.uom.se.unit.Units.WATT;
+import static tech.units.indriya.unit.Units.GRAM;
+import static tech.units.indriya.unit.Units.JOULE;
+import static tech.units.indriya.unit.Units.KELVIN;
+import static tech.units.indriya.unit.Units.KILOGRAM;
+import static tech.units.indriya.unit.Units.METRE;
+import static tech.units.indriya.unit.Units.MOLE;
+import static tech.units.indriya.unit.Units.PASCAL;
+import static tech.units.indriya.unit.Units.SECOND;
+import static tech.units.indriya.unit.Units.WATT;
```
jodygarnett, line 28: I am considering something I should of probably done last time:

1. Add all the unit constants we use to geotools Units class

```java
static Unit<Distance> METRE = tech.units.indriya.unit.Units.METRE;
...
```

2. Update all of these imports with:

```java
import static org.geotools.measure.Units.*;
```

jodygarnett: I considered it last time but I had hoped that the implementation imports would be stable for downstream applications.

This has the advantage of any future implementation change being transparent to applications using GeoTools.

modules/library/metadata/src/test/java/org/geotools/util/GeoToolsUnitFormatRoundTripTest.java
```diff
@@ -45,21 +45,22 @@
...
                 new Object[][] {
+                    {GRAM},
                     {GRAM.divide(1_000_000)},
                     {GRAM.divide(1_000_000).divide(METRE.pow(3))},
                     {GRAM.divide(1_000_000_000).divide(METRE.pow(3))},
                     {GRAM.divide(1_000_000).divide(METRE.pow(3))},
                     {METRE.pow(2)},
                     {METRE.pow(3)},
-                    {SI.MOLE.divide(1_000_000)},
+                    {MOLE.divide(1_000_000)},
```
jodygarnett, line 55: No division between SI and NonSI anymore eh?

modules/library/opengis/pom.xml
```diff
@@ -177,7 +177,7 @@
     
     <dependency>
       <groupId>systems.uom</groupId>
-      <artifactId>systems-common-java8</artifactId>
+      <artifactId>systems-common</artifactId>
```
jodygarnett, line 180: Is there no longer a java 8 specific systems-common?

modules/library/referencing/src/main/java/org/geotools/measure/SexagesimalConverter.java
```diff
@@ -37,7 +35,7 @@
  */
 class SexagesimalConverter extends AbstractConverter {
     /** Serial number for compatibility with different versions. */
-    private static final long serialVersionUID = 3873494343412121773L;
+    private static final long serialVersionUID = -2663951106460584999L;
```
jodygarnett, line 38: Shy is the `serialVersionUID` changing?

modules/library/referencing/src/main/java/org/geotools/measure/SexagesimalConverter.java
```diff
@@ -118,6 +166,12 @@ public int hashCode() {
         return (int) serialVersionUID + divider;
     }
 
+    @Override
+    protected String transformationLiteral() {
+        // TODO
+        return null;
```
jodygarnett, line 172: While `null` is valid for the api, is there anything useful to return here? If not we can remove the todo.

modules/library/referencing/src/main/java/org/geotools/referencing/factory/OrderedAxisAuthorityFactory.java
```diff
@@ -224,8 +224,8 @@ private static boolean booleanValue(final Hints userHints, final Hints.Key key)
      * invoked by constructors only.
      */
     private void completeHints() {
-        hints.put(Hints.FORCE_STANDARD_AXIS_UNITS, Boolean.valueOf(forceStandardUnits));
-        hints.put(Hints.FORCE_STANDARD_AXIS_DIRECTIONS, Boolean.valueOf(forceStandardDirections));
+        hints.put(Hints.FORCE_STANDARD_AXIS_UNITS, forceStandardUnits);
+        hints.put(Hints.FORCE_STANDARD_AXIS_DIRECTIONS, forceStandardDirections);
```
jodygarnett, line 228: Can I confirm you are building in Java 8?

We have had on occasion added code like the above to get around differences between java 8 and java 11 compiler. Differences in the byte code produced has caused us to add extra casts into our code on occasion.

modules/ogc/net.opengis.csw/src/net/opengis/cat/csw20/impl/QueryTypeImpl.java
```diff
@@ -374,7 +375,7 @@ public boolean eIsSet(int featureID) {
             case Csw20Package.QUERY_TYPE__CONSTRAINT:
                 return constraint != null;
             case Csw20Package.QUERY_TYPE__SORT_BY:
-                return SORT_BY_EDEFAULT == null ? sortBy != null : !SORT_BY_EDEFAULT.equals(sortBy);
+                return SORT_BY_EDEFAULT == null ? sortBy != null : !Arrays.equals(SORT_BY_EDEFAULT, sortBy);
```
jodygarnett, line 378: This appears to be cleaning up a warning unrelated to the current activity. Much of this code is generated and thus hard suppress warnings.

If you wish to keep this change you need to set the javadoc `@generated` to `@generated false`, but it may be smarter to back out the change.

modules/plugin/coverage-multidim/coverage-api/pom.xml
```diff
@@ -121,7 +121,7 @@
       <dependency>
           <groupId>com.github.spotbugs</groupId>
           <artifactId>spotbugs-annotations</artifactId>
-          <version>RELEASE</version>
+          <version>4.0.1</version>
```
jodygarnett, line 124: We should really manage this in root dependency management section rather than here.

modules/library/metadata/src/test/java/org/geotools/util/GeoToolsUnitFormatRoundTripTest.java
ddcruver: No they both still exist in si.uom:si-units:2.0.1 but it seemed like some things that were considered non-SI before were now considered SI.  Possibly the document they base the types on was updated since the original Units of Measure 1.0 implementation.

modules/library/opengis/pom.xml
ddcruver: TL;DR They support Java 8 out of the box.

Indriya 2.0 is compiled at source/target level Java 8.  Support for Java8/9+ was one of the reasons they made JSR 385. https://github.com/unitsofmeasurement/indriya/blob/master/pom.xml

It looks like they only support Java 8+:

> Modular environment - While the API should be backward-compatible with at least Java SE 8, both API and RI are expected to expose Java 9 modules
https://www.jcp.org/en/jsr/detail?id=385

Is Java 7 a requirement for GeoTools?

modules/library/referencing/src/main/java/org/geotools/referencing/cs/DefaultCoordinateSystemAxis.java
```diff
@@ -523,7 +523,7 @@
      * A default axis for time values in a {@linkplain org.opengis.referencing.cs.TimeCS time CS}.
      *
      * <p>Increasing time go toward {@linkplain AxisDirection#FUTURE future} and units are
-     * {@linkplain NonSI#DAY days}.
```
ddcruver, line 526: Here is an example of a NonSI type that became an SI unit.

modules/library/referencing/src/main/java/org/geotools/referencing/factory/OrderedAxisAuthorityFactory.java
ddcruver: Java 8 is my development environment for work.  And all my testing of the changes where done in Java 8.

modules/library/referencing/src/main/java/org/geotools/measure/SexagesimalConverter.java
ddcruver: I tried to change as little as possible but the parent this class extends is different.  It appears they move the AbstractConverter from one of their Unit System libraries to their reference implementation.

Wouldn't the class changing its hierarchy mean the serialVersionUID should change?  Your judgement is most likely better then mine but just want to point out the parent class has changed.

modules/library/referencing/src/main/java/org/geotools/measure/SexagesimalConverter.java
ddcruver: `	 * Returns a String describing the transformation that is represented by this converter. 
	 * Contributes to converter's {@code toString} method. If null or empty
	 * {@code toString} output becomes simplified.`

When you toString the converter it will call this method and append the value to the end of the to string if it is not null.  I couldn't think if we should add anything there or not.

modules/library/referencing/src/main/java/org/geotools/referencing/factory/OrderedAxisAuthorityFactory.java
ddcruver: I can't resist following IntelliJ inspection recommendations.  It said this was not required.

modules/ogc/net.opengis.csw/src/net/opengis/cat/csw20/impl/QueryTypeImpl.java
ddcruver: Okay this change is a not as straight forward as the boolean boxing.   I can revert it.

modules/plugin/coverage-multidim/coverage-api/pom.xml
ddcruver: I thought RELEASE was removed from Maven 3.   I can revert it or keep it there its up to you.
https://cwiki.apache.org/confluence/display/MAVEN/Maven+3.x+Compatibility+Notes#Maven3.xCompatibilityNotes-PluginMetaversionResolution. Once again my IDE pointed out to me.   If it works do we want the project to possibly use a different version if build later in time?  Maybe for this one we do. Again let me know if you like me to revert this or not.

modules/ogc/net.opengis.csw/src/net/opengis/cat/csw20/impl/QueryTypeImpl.java
ddcruver: Has been reverted.

modules/library/referencing/src/main/java/org/geotools/measure/SexagesimalConverter.java
jodygarnett: Understood, thanks for the clarification.

modules/library/referencing/src/main/java/org/geotools/measure/SexagesimalConverter.java
jodygarnett: Yes, I checked the base class before commenting, ... so we can remove the `// TODO` ?

modules/library/referencing/src/main/java/org/geotools/referencing/factory/OrderedAxisAuthorityFactory.java
jodygarnett: In this case it probably is not required, I just wanted to double check :)

modules/plugin/coverage-multidim/coverage-api/pom.xml
jodygarnett: I am happy to explicitly depend on 4.0.1, it is just we should put that in a dependency management section in the root pom.xml

modules/library/referencing/src/main/java/org/geotools/referencing/cs/DefaultCoordinateSystemAxis.java
jodygarnett: Gotcha!

modules/library/opengis/pom.xml
jodygarnett: We are Java 8 and Java 11 explicitly; so if it is set to run on Java 8 that is fine.

jodygarnett:
So as a first time contributor (hey thanks!) we do need to ask you to look at CONTRIBUTING.md and send a PDF to the Open Source Geospatial Foundation. We have found that having a CLA makes things easier to work with other open source projects across license differences, and makes it really clear that an individual and/or their employer agree to "open source" :)

The other thing I will need to do (once we past test cases) is prepare a patch for some of the downstream projects (GeoWebCache, GeoServer) so we can keep all the builds happy.

aaime:
@jodygarnett I guess the upgrade guide will have to be updated accordingly, or are the changes completely hidden from the user? The GeoWebCache and GeoServer downstream builds are broken, so I'm guessing not transparent.

modules/library/referencing/src/main/java/org/geotools/measure/SexagesimalConverter.java
ddcruver: Removed TODO statement.

modules/plugin/coverage-multidim/coverage-api/pom.xml
ddcruver: I added a dependency management for version 4.0.2 in the root pom.xml.  I also updated the version of the spotbugs-maven-plugin to match the spotbugs-annotations' version.  I build with the profile -Pspotbugs enabled and it completed with the new plugin version.

ddcruver:
> @jodygarnett I guess the upgrade guide will have to be updated accordingly, or are the changes completely hidden from the user? The GeoWebCache and GeoServer downstream builds are broken, so I'm guessing not transparent.

There is a specific section of the upgrade guide titled "Migrate to JSR-363 Units" so I believe it is appropriate for a similar section to be added.  I believe there are less package name changes but the transitive dependency has a new artifactId and version.  In my opinion the API changes from JSR-275 to JSR-363 are bigger then the changes from JSR-363  to JSR-385.

modules/ogc/net.opengis.csw/src/net/opengis/cat/csw20/impl/QueryTypeImpl.java
ddcruver: Actually when I enable the error prone compiler in IntelliJ i am getting this error and can't compile the project.

`Error:(377, 92) java: [ArrayEquals] Reference equality used to compare arrays
    (see https://errorprone.info/bugpattern/ArrayEquals)
  Did you mean 'return SORT_BY_EDEFAULT == null ? sortBy != null : !Arrays.equals(SORT_BY_EDEFAULT, sortBy);'?`

It is my understanding that the error prone compiler is required by GeoTools and I can't compile the code the way it is in IntelliJ.

jodygarnett: I am never quite sure how to use error prone in IntelliJ, perhaps @aaime has guidance. Does error prone require some configuration to ignore these autogenerated folders?

jodygarnett:
> There is a specific section of the upgrade guide titled "Migrate to JSR-363 Units" so I believe it is appropriate for a similar section to be added. 

Yes, we can add this, I wish to resolve the conversation on use of "static import org.geotools.measure.Units.METRE" first.

modules/ogc/net.opengis.csw/src/net/opengis/cat/csw20/impl/QueryTypeImpl.java
aaime: I don't use errorprone in IntelliJ, just run it in maven explicitly once I'm done with the basic development, and fix whatever pops up.

ddcruver: I used the [IntelliJ IDEA (Official) plugin](https://plugins.jetbrains.com/plugin/7349-error-prone-compiler) and configured it in IntelliJ settings:

File | Settings | Build, Execution, Deployment | Compiler | Java Compiler

jetbrains://idea/settings?name=Build%2C+Execution%2C+Deployment--Compiler--Java+Compiler

GitHub doesn't seem to support their fancy application link to the settings.

Without it I could not compile to run the project, but with it would compile them so I could run tests in the IDE but until I fixed the "error prone" errors it wouldn't let me run tests and if i set the Run Configuration to "Build, no error check" in the Before launch section it would run old stale code.

modules/library/metadata/src/test/java/org/geotools/util/GeoToolsUnitFormatRoundTripTest.java
jodygarnett: I tried the change locally, it was not worth it since only a few test cases internal to geotools notice. And downstream applications mostly create things by text.

See https://github.com/geoserver/geoserver/search?q=tec.uom.se&unscoped_q=tec.uom.se

ddcruver:
> > There is a specific section of the upgrade guide titled "Migrate to JSR-363 Units" so I believe it is appropriate for a similar section to be added.
> 
> Yes, we can add this, I wish to resolve the conversation on use of "static import org.geotools.measure.Units.METRE" first.
> 
> Update: Okay I checked, introducing static constant `org.geotools.measure.Units.METRE` was not worth it.

I might have just pushed up changes you said was not worth it.  Let me know if you like me to revert them.

jodygarnett:
The changes are fine, and long term it will be easier for downstream applications to adjust to any future change of implementation library.

It seems like a design flaw with units that we cannot seamlessly plugin different implementations...

jodygarnett:
So we are down to some test failures and documentation eh?

ddcruver:
... and a few downstream patches.

ddcruver:
The NetCDFUnitParserTest$UnitConversionTest was failing because the Units library is reducing the units to Metric prefix instead of keeping them in the extended form the test expected.

For example it expected mol/(m/100) but got mol/cm.

ddcruver:
Still a few more UnitConverstionTests to fix up.

ddcruver:
Okay that suite is all fixed up.

modules/plugin/coverage-multidim/netcdf/src/main/resources/org/geotools/imageio/netcdf/netcdf-unit-replacements.properties
```diff
@@ -2,4 +2,7 @@ microgrammes\ per\ cubic\ meter=µg*m^-3
 DU=µmol*m^-2*446.2
 m2=m^2
 m3=m^3
-s2=s^2
\ No newline at end of file
+s2=s^2
```
aaime, line 5: This change is going to break, on upgrade, the existing installations using the units replacements customization (visible not only to GeoTools users, but GeoServer users alike, which are often not developers)

What kind of changes are expected and what can be done to fix them? Is it possible to write a guide, and/or is there a table listing before/after differences?

modules/plugin/coverage-multidim/netcdf/src/test/java/org/geotools/coverage/io/netcdf/NetCDFUnitParserTest.java
```diff
@@ -159,10 +160,10 @@ public void reset() {
         public void testDU() throws Exception {
             // could not find a way to make this work with the above test, so doing it another way
             Unit<?> unit = NetCDFUnitFormat.parse("DU");
-            assertEquals("µmol·(1/m²)*446.2", unit.toString());
+            assertEquals("(1/m²)*446.2·μmol", unit.toString());
```
aaime, line 163: The research center sponsoring the changes behind this test required exact matches on expected unit expressions. Is there no way to make it look as before? Why did the factors move?

pom.xml
```diff
@@ -1347,6 +1319,13 @@
         <version>2.10.1</version>
       </dependency>
 
+      <!-- Quality control dependencies -->
+      <dependency>
```
aaime, line 1323: Is this change actually needed for Units upgrade? I don't see the spotbugs annotations being used in the diff.

modules/library/referencing/src/main/java/org/geotools/measure/SexagesimalConverter.java
```diff
@@ -77,13 +75,28 @@ private SexagesimalConverter(final int divider, final SexagesimalConverter inver
...
+        return that instanceof Inverse;
+    }
+
+    public Number convertWhenNotIdentity(Number value) {
+        return degreeToSexagesimalDegrees(value.doubleValue());
+    }
+
     /** Performs a conversion from fractional degrees to sexagesimal degrees. */
-    public double convert(double value) {
+    public double degreeToSexagesimalDegrees(double value) {
```
aaime, line 99: public API break, should be documented on the upgrade guide at a minimum. For best results, deprecate the existin implementation and make it delegate to the new method.

modules/library/referencing/src/main/java/org/geotools/measure/SexagesimalConverter.java
```diff
@@ -191,19 +228,14 @@ public int hashCode() {
...
-    public BigDecimal convert(BigDecimal value, MathContext ctx) throws ArithmeticException {
-        double valueAsDouble = value.doubleValue();
-        if (valueAsDouble == Double.NEGATIVE_INFINITY
-                || valueAsDouble == Double.POSITIVE_INFINITY) {
-            throw new ArithmeticException(
-                    "Bigdecimal value magnitude too great to fit in double: " + value);
-        }
-        return new BigDecimal(convert(valueAsDouble));
-    }
+    //    @Override
```
aaime, line 231: Please don't leave commented out code sections.

pom.xml
aaime: Ah, I see there are lots of other unrelated changes in the diff... the checklist contains this for supported modules:

"The pull request contains changes related to a single objective. If multiple focuses cannot be avoided, each one is in its own commit and has a separate ticket describing it."

For this one we can close an eye, but please remember about it next time.

aaime:
Had a quick look, worried about some changes in NetCDF that might break existing installations on upgrade.

Besides that, the three major points missing here seem to be:

* Fix remaining tests
* An upgrade guide
* Making the downstream projects, GWC and GS, build too. GeoServer is breaking on compile in the gs-main module right now. The three projects act like one, depending on each other snapshots, with GeoServer really "keeping the lights on" for the whole group of three, the other two receiving only a small fraction of independent contributions.

I can try to have a look at GeoServer in the following weeks, right now already using my spare time on another PR with a similar situation (the commons collections one).

aaime:
And oh, forgot: the ticket in Jira. Generally speaking, by the time this PR is ready for merge, all the applicable checklist must be satisifed (in the case of this PR, all of it, besides maybe the "single focus" item, it would be nice to have it, but the unrelated changes have been intermingled with others ,it might be difficult to separate them now).

modules/library/referencing/src/main/java/org/geotools/measure/SexagesimalConverter.java
ddcruver: Actually this class extends AbstractConverter which implements the convert method and uses a final modifier.  This method now calls convertWhenNotIdentity() of this class.  This being said I see no need to document this on the upgrade guide.

modules/plugin/coverage-multidim/netcdf/src/test/java/org/geotools/coverage/io/netcdf/NetCDFUnitParserTest.java
ddcruver: This might need a little investigating, The unit library just seems to sort them in a particular way, even though the _netcdf-unit-replacements.properties_ had a replacement that matched the original output the unit library moved it to the back.

pom.xml
ddcruver: Will do, sorry about that.

modules/plugin/coverage-multidim/netcdf/src/main/resources/org/geotools/imageio/netcdf/netcdf-unit-replacements.properties
ddcruver: Not sure how the units library handled these units in the netcdf format before but adding these ensured that it could parse these as the appropriate temperature units.  I am not familiar with the netcdf format so someone else would have to address updating the documentation.

modules/plugin/coverage-multidim/netcdf/src/test/java/org/geotools/coverage/io/netcdf/NetCDFUnitParserTest.java
ddcruver: Here is where I think the documentation would need to be updated.
https://docs.geoserver.org/latest/en/user/extensions/netcdf/netcdf.html#supporting-custom-netcdf-units

modules/library/referencing/src/main/java/org/geotools/measure/SexagesimalConverter.java
ddcruver: Fixed.

ddcruver:
I can open a separate pull request with the Maven fixes related to findbugs.  Also I don't believe I have access to the JIRA system so I can not make a ticket for either.  Thanks.

pom.xml
ddcruver: I removed it, opened separate pull request.

aaime:
> I can open a separate pull request with the Maven fixes related to findbugs. Also I don't believe I have access to the JIRA system so I can not make a ticket for either. Thanks.

Anyone can open tickets, a registration is needed: https://osgeo-org.atlassian.net/projects/GEOT/summary

aaime:
Merged the small fix on the other PR. That one did not need a CLA as it was minor modifications. This one however is large, a CLA will be needed. I've just checked the CLA repository just in case, I cannot see one on your name, but it could be that your company signed one maybe? Or it was simply not sent yet.

ddcruver:
> Merged the small fix on the other PR. That one did not need a CLA as it was minor modifications. This one however is large, a CLA will be needed. I've just checked the CLA repository just in case, I cannot see one on your name, but it could be that your company signed one maybe? Or it was simply not sent yet.

I will be signing an Individual Contribution agreement shortly, just in the process of double checking with my company.

ddcruver:
> > I can open a separate pull request with the Maven fixes related to findbugs. Also I don't believe I have access to the JIRA system so I can not make a ticket for either. Thanks.
> 
> Anyone can open tickets, a registration is needed: https://osgeo-org.atlassian.net/projects/GEOT/summary

When I click on that link it asks me to login to Atlassian but after I successfully log in I still get an "You need permission" and "Request access from your administrator" message.

jodygarnett:
> When I click on that link it asks me to login to Atlassian but after I successfully log in I still get an "You need permission" and "Request access from your administrator" message.

First up what is your user id, and what permissions do you need? You should be able to report an issue (but not assign it to yourself) without any additional permissions.

ddcruver:
Unfortunetly I don't know how to get my user ID. When I click the link to submit an issue it gives me a permission denied page.

jodygarnett:
> Unfortunetly I don't know how to get my user ID. When I click the link to submit an issue it gives me a permission denied page.

Interesting, are you able to sign up for an account? We have had a couple reports of difficult with those using a corporate google account.

Please try using your browser in "incognito" mode or similar to prevent some kind of OpenAuth cookie from making your life difficult.

aaime:
I've created the ticket: https://osgeo-org.atlassian.net/browse/GEOT-6612

aaime:
Noting this PR will require downstream changes, it's breaking the build in GeoServer:

```
2020-05-10T16:43:57.8909757Z [ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.0:compile (default-compile) on project gs-main: Compilation failure: Compilation failure: 
2020-05-10T16:43:57.8918567Z [ERROR] /home/runner/geoserver/src/main/src/main/java/org/geoserver/catalog/CoverageDimensionCustomizerReader.java:[58,25] package tec.uom.se.format does not exist
2020-05-10T16:43:57.8926288Z [ERROR] /home/runner/geoserver/src/main/src/main/java/org/geoserver/catalog/CoverageDimensionCustomizerReader.java:[597,28] cannot find symbol
```

Before merging this one a corresponding PR in GeoServer with the JSR upgrade must be present.

aaime:
Looking at the changes, they are a major pain to review because one of the checklist requirements was not followed:

"The pull request contains changes related to a single objective. If multiple focuses cannot be avoided, each one is in its own commit and has a separate ticket describing it."

In particular, the PR is filled with typo fixes and other changes probably reported by a static code checker (like IntelliJ's one). Those changes are welcomed in their own separate commit, but are obscuring the substantive changes related to the units migration. As if the PR was not big enough already... Going through it anyways, but please be mindful about it in the future.

aaime:
I hope I have figured out the current build failure. The build fails because it's trying to create a cartesian CRS whose axis unit is in degrees.

This happens in tests created by @bencaradocdavies to support the rotated polar case. The current code ends up building a ProjectedCRS for that case, which is semantically wrong, the result is still expressed in degrees. Before it did not break because the old unit library refused to parse "degrees" as a unit and a fall-back code ended up using meters, which caused the creation of a ProjectedCRS.

Right now the unit library parses the unit correctly and then the referencing subsystem complains the result cannot be a projected CRS. Modified the code so that a DerivedCRS is created instead, and changed the tests to work with it. On my machine, the NetCDF and Grib modules work fine.

Will something be rendered though? Or will WCS work off these? I have no clue, never seen usage of a DerivedCRS before, but it's the only CRS type making sense for a CRS that's still ellipoidal...

jnh5y:
@aaime I'd agree that a note in the upgrade guide is good!  For GeoMesa, I chatted with @elahrvivaz; he notes that we don't use the UOM libraries in GeoMesa.  

Given that, I doubt we'll hit any issues with this upgrade.  It sounds like the care going into GeoServer will cover most bases!  Thanks @aaime for doing that.

ddcruver:
@jodygarnett
TL;DR I will have a signed individual contributor agreement sent to you soon.

Sorry about the delay on the contributor agreement.  I double checked with a higher up at my work if my contributions would have any conflict of interest.  While I am NOT doing this under my company or customers dollar we as employees have signed various legal documents that may potentially limit my ability to work outside the company.  Good news are that he has insured me that as long as may work at CUBRC takes priority and I am contributing to an open source program that is fine.  Sorry again the person I was waiting on just had a personal injury and the timing was bad to contact him.

ddcruver:
> I hope I have figured out the current build failure. The build fails because it's trying to create a cartesian CRS whose axis unit is in degrees.

Thank you for looking into this, I know your already a big contributor in this project and glad you had the time to look into the test failure.  I am not a GeoSpatial export and it may of taken me much longer to determine the root of the problem.  I have approved your pull request to my local fork so it can be included in this pull request.

jodygarnett:
Is the coast now clear for this upgrade to go ahead?

modules/library/referencing/src/main/java/org/geotools/measure/SexagesimalConverter.java
jodygarnett: The logic is that convert(value) is still available, supplied by the superclass; so no net api change.

aaime:
@jodygarnett it won't be mergeable until someone makes corresponding downstream changes.

ddcruver:
How do you submit patches for downstream projects and link them to this pull request? Or is it just a manual process of cliving merge on both pull requests? I think the GeoServer fixes are really simple.

ddcruver:
I submitted another pull request as a patch for geoserver repository, it fails of course because it requires this change to work. https://github.com/geoserver/geoserver/pull/4293

ddcruver:
> @jodygarnett I guess the upgrade guide will have to be updated accordingly, or are the changes completely hidden from the user? The GeoWebCache and GeoServer downstream builds are broken, so I'm guessing not transparent.

Well I believe the patch will fix GeoServer, but GeoWebCache appears to build anyways without any modification.

ddcruver:
I can't verify I have fixed GeoServer completely because I can't run the tests on my machine.  Even a checkout of master fails.

ddcruver:
@jodygarnett or @aaime Did I submit the patch for GeoServer correctly?

aaime:
The PR seems ok, did not have time to look at it today, maybe next Monday (I have a few hours each Monday to look at PRs originating from the community).

ddcruver:
> The PR seems ok, did not have time to look at it today, maybe next Monday (I have a few hours each Monday to look at PRs originating from the community).

Thanks @aaime, yeah that would be great if you could.

ddcruver:
Do not know what was wrong the other day maybe I had an bad snapshot of GeoTools but the tests for GeoServer passed today.  I have a checkout of this branch locally and a checkout of geoserver and updated it to use my local GeoTools snapshots and apply my patches it seems to be all good.

aaime:
The tests in GeoServer are failing. They are all related to the rotated pole projection case. I've spent a few hours trying to fix them, but did not manage to make the tests pass, while some get fixed, other break. I cannot justify spending more working hours on this, may try later in my spare time, but mind, it's very, very limited, could take a long while (e.g. weeks or more).

The gist of it is that the way derived CSs are built in the NetCDF module is not a match for how they are built by the WKT parser, for GeoServer tests to pass the two have to match. And I got there, but then other tests started to break, someone needs to investigate as to why they do. I have partial patches for both GeoTools and GeoServer if anyone is interested in catching up before I do.

ddcruver:
That is weird @aaime, I ran the build again today and nothing failed.  But thanks for looking into this.  Is there someone active in this project that we can ask about NetCDF and rotated pole projection cases?

The coordinate systems and geospatial in general is not yet part of my expertise.

My concern is that we can get this in by the next major release and maybe get a Milestone release including these changes.  The company that I work for is now using a private snapshot with these changes in it but would like to get us to a public release of some type as soon as possible.

aaime:
@ddcruver I'm afraid I'm the only one in the intersection between NetCDF and referencing, but would be pretty happy to be proven wrong about it :-D

ddcruver:
@aaime Lets hopes someone that is interested in Unit 2.0 support in GeoTools and is a NetCDF / referencing expert takes notice of this ticket.  I don't mind learning something new but right now I am trying to find the time, in my limited free time, for my continuous education of a different kind and will be busy until the end of the next month.

aaime:
Oh, I forgot to comment on why the build is not failing for you... because the extension and community modules are not included in the default build, but they are part of all the actual builds on servers and the like.

So your base run should be:

``mvn clean install -Prelease``

which builds core and extension modules. 
And then, separately, you also need to run:

```
cd community
mvn clean install -DskipTests -PcommunityRelease
```

as the community modules are only tested for compiling, so that we can build jars for them, but they don't get the QA effort (for both build server and core devs) that maintained modules do get.

aaime:
Figured out the remaining issues... the changes are not big, but the debugging was epic!
Created a new PR with those extra fixes (new tickets and separate commits, as they were unrelated to JSR 385 proper): https://github.com/geotools/geotools/pull/3017

jodygarnett:
Now that 24.0 is released this work is seeing wider review, including https://osgeo-org.atlassian.net/browse/GEOT-6709 noting problems using reflection access against `indriya`.

PR #2926, opened 2020-05-12
Fix getFeature-request in SAP HANA datastore when working with views …

PaulBiskup:
…and no SRID is found

PR for SAP HANA datastore-plugin when working with views:
issue in JIRA: 
https://osgeo-org.atlassian.net/browse/GEOT-6587

### DIFF
--- a/modules/plugin/jdbc/jdbc-hana/src/main/java/org/geotools/data/hana/HanaDialect.java
+++ b/modules/plugin/jdbc/jdbc-hana/src/main/java/org/geotools/data/hana/HanaDialect.java
@@ -20,6 +20,7 @@ import java.io.IOException;
 import java.math.BigDecimal;
 import java.sql.*;
 import java.sql.Date;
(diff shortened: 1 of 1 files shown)


stefanuhrig:
Thanks for the fix. This will work for HANA 1.

It won't work for HANA 2, but won't do any harm either. I'll try to fix that in a separate PR.

aaime:
Gentle reminder that the [PR checklist](https://github.com/geotools/geotools/blob/master/.github/PULL_REQUEST_TEMPLATE.md) is not optional, should not be removed when opening a PR, and all applicable items should be satisfied before merging. This PR does not contain a test, should not have been merged in a supported module.

aaime:
For this once let's close an eye, but please be mindful about it next time.

PaulBiskup:
> For this once let's close an eye, but please be mindful about it next time.

Alright, sorry. I will I will take this into account in the future.

PR #2930, opened 2020-05-13
Support mbstyle filter expressions [GEOT-6589] 

jodygarnett:
The filter support used by GeoTools has been moved to "Legacy - Other filter":
```
filter: [">", "scalerank", 2]
```

This has been replaced with the use of expressions:

```
filter: [">", ["get","scalerank"], 2]
```

This means we can no longer use style examples available online.

See https://osgeo-org.atlassian.net/browse/GEOT-6589 for details.

## Checklist

For all pull requests:

- [x] There is an issue https://osgeo-org.atlassian.net/browse/GEOT-6589
- [ ] PR for bug fixes and small new features are presented as a single commit: use squash changes when merging
- [ ] Commit message must be in the form "[GEOT-XYZW] Title of the Jira ticket": use squash changes when merging

### DIFF
--- a/modules/extension/mbstyle/src/main/java/org/geotools/mbstyle/expression/MBExpression.java
+++ b/modules/extension/mbstyle/src/main/java/org/geotools/mbstyle/expression/MBExpression.java
@@ -50,7 +50,12 @@ public abstract class MBExpression extends FunctionImpl {
     protected final MBObjectParser parse;
     protected final MBStyleTransformer transformer;
 
-    public MBExpression(JSONArray json) {
+    /**
+     * Please use factory method {@link #create(JSONArray)}
+     *
+     * @param json
+     */
+    protected MBExpression(JSONArray json) {
         this.json = json;
         this.name = (String) json.get(0);
         this.ff = CommonFactoryFinder.getFilterFactory2();
@@ -123,6 +128,12 @@ public abstract class MBExpression extends FunctionImpl {
     /* A list of zoom expression names */
     public static List zoom = Arrays.asList("zoom");
 
+    /**
+     * Factory method used to produce the correct MBExpression subclass for the provided JSONArray.
+     *
+     * @param json definition
+     * @return MBExpression
+     */
     public static MBExpression create(JSONArray json) {
         String name;
         if (json.get(0) instanceof String) {
--- a/modules/extension/mbstyle/src/main/java/org/geotools/mbstyle/expression/MBExpression.java
+++ b/modules/extension/mbstyle/src/main/java/org/geotools/mbstyle/expression/MBExpression.java
@@ -162,24 +162,25 @@ public abstract class MBExpression extends FunctionImpl {
             } else if (zoom.contains(name)) {
                 return new MBZoom(json);
             } else {
-                throw new MBFormatException("Expression \"" + name + "\" invalid.");
+                throw new MBFormatException("Expression \"" + name + "\" not supported.");
             }
         }
         throw new MBFormatException("Requires a string name of the expression at position 0");
     }
 
     public static boolean canCreate(final String name) {
-        return colors.contains(name)
-                || decisions.contains(name)
-                || featureData.contains(name)
-                || heatMap.contains(name)
-                || lookUp.contains(name)
-                || math.contains(name)
-                || ramps.contains(name)
-                || string.contains(name)
-                || types.contains(name)
-                || variableBindings.contains(name)
-                || zoom.contains(name);
+        return name != null
+                && (colors.contains(name)
+                        || decisions.contains(name)
+                        || featureData.contains(name)
+                        || heatMap.contains(name)
+                        || lookUp.contains(name)
+                        || math.contains(name)
+                        || ramps.contains(name)
+                        || string.contains(name)
+                        || types.contains(name)
+                        || variableBindings.contains(name)
+                        || zoom.contains(name));
     }
 
     /** Determines which expression to use. */
--- a/modules/extension/mbstyle/src/main/java/org/geotools/mbstyle/expression/MBExpression.java
+++ b/modules/extension/mbstyle/src/main/java/org/geotools/mbstyle/expression/MBExpression.java
@@ -162,10 +162,14 @@ public abstract class MBExpression extends FunctionImpl {
             } else if (zoom.contains(name)) {
                 return new MBZoom(json);
             } else {
-                throw new MBFormatException("Expression \"" + name + "\" not supported.");
+                throw new MBFormatException(
+                        "Data expression \""
+                                + name
+                                + "\" invalid. It may be misspelled or not supported by this implementation:"
+                                + json);
             }
         }
-        throw new MBFormatException("Requires a string name of the expression at position 0");
+        throw new MBFormatException("Requires a string name of the data expression at position 0");
     }
 
     public static boolean canCreate(final String name) {
--- a/modules/extension/mbstyle/src/main/java/org/geotools/mbstyle/expression/MBLookup.java
+++ b/modules/extension/mbstyle/src/main/java/org/geotools/mbstyle/expression/MBLookup.java
@@ -56,22 +56,26 @@ public class MBLookup extends MBExpression {
     public Expression lookupGet() {
         if (json.size() == 2 || json.size() == 3) {
             if (json.size() == 2) {
+                // Example: ["get", "propertyName"]
                 if (parse.isString(json, 1)) {
                     String propertyName = parse.get(json, 1);
                     return ff.property(propertyName);
-                } else {
+                }
+                // Example: ["get", key]
+                else {
                     // it is unclear from specification if this is even allowed
                     Expression property = parse.string(json, 1);
                     return ff.function("property", property);
                 }
             }
+            // Example: ["get", key, object]
             if (json.size() == 3) {
                 Expression value = parse.string(json, 1);
                 Expression object = parse.string(json, 2);
                 return ff.function("get", value, object);
             }
         }
-        throw new MBFormatException("Expression \"get\" requires a maximum of 3 arguments.");
+        throw new MBFormatException("Data expression \"get\" requires a maximum of 3 arguments.");
     }
 
     /**
--- a/modules/extension/mbstyle/src/main/java/org/geotools/mbstyle/expression/MBLookup.java
+++ b/modules/extension/mbstyle/src/main/java/org/geotools/mbstyle/expression/MBLookup.java
@@ -56,8 +56,14 @@ public class MBLookup extends MBExpression {
     public Expression lookupGet() {
         if (json.size() == 2 || json.size() == 3) {
             if (json.size() == 2) {
-                Expression property = parse.string(json, 1);
-                return ff.function("property", property);
+                if (parse.isString(json, 1)) {
+                    String propertyName = parse.get(json, 1);
+                    return ff.property(propertyName);
+                } else {
+                    // it is unclear from specification if this is even allowed
+                    Expression property = parse.string(json, 1);
+                    return ff.function("property", property);
+                }
             }
             if (json.size() == 3) {
                 Expression value = parse.string(json, 1);
@@ -65,7 +71,7 @@ public class MBLookup extends MBExpression {
                 return ff.function("get", value, object);
             }
         }
-        throw new MBFormatException("Expression \"get\" requires a maximum of 2 arguments.");
+        throw new MBFormatException("Expression \"get\" requires a maximum of 3 arguments.");
     }
 
     /**
--- a/modules/extension/mbstyle/src/main/java/org/geotools/mbstyle/expression/MBLookup.java
+++ b/modules/extension/mbstyle/src/main/java/org/geotools/mbstyle/expression/MBLookup.java
@@ -56,20 +56,18 @@ public class MBLookup extends MBExpression {
     public Expression lookupGet() {
         if (json.size() == 2 || json.size() == 3) {
             if (json.size() == 2) {
-                // Example: ["get", "propertyName"]
                 if (parse.isString(json, 1)) {
+                    // example: ["get", "propertyName"]
                     String propertyName = parse.get(json, 1);
                     return ff.property(propertyName);
-                }
-                // Example: ["get", key]
-                else {
-                    // it is unclear from specification if this is even allowed
+                } else {
+                    // example: ["get", key]
                     Expression property = parse.string(json, 1);
                     return ff.function("property", property);
                 }
             }
-            // Example: ["get", key, object]
             if (json.size() == 3) {
+                // example: ["get", key, object]
                 Expression value = parse.string(json, 1);
                 Expression object = parse.string(json, 2);
                 return ff.function("get", value, object);
--- a/modules/extension/mbstyle/src/main/java/org/geotools/mbstyle/parse/MBFilter.java
+++ b/modules/extension/mbstyle/src/main/java/org/geotools/mbstyle/parse/MBFilter.java
@@ -38,7 +38,19 @@ import org.opengis.style.SemanticType;
  *
  * <p>This wrapper and {@link MBFunction} are a matched set handling dynamic data.
  *
+ * <h2>About MapBox Expression</h2>
+ *
+ * The value for any filter may be specified as an expression. The result type of an expression in
+ * the filter property must be boolean.
+ *
+ * <p>See {@link org.geotools.mbstyle.expression.MBExpression} for details.
+ *
+ * <p>
+ *
  * <h2>About MapBox Filter</h2>
+ *
+ * <p>In previous versions of the style specification, filters were defined using the deprecated
+ * syntax documented here.
  *
  * <p>A filter selects specific features from a layer. A filter is an array of one of the following
  * forms:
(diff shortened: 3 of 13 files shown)


modules/extension/mbstyle/src/main/java/org/geotools/mbstyle/parse/MBFilter.java
```diff
@@ -460,4 +468,59 @@ public Filter filter() {
...
+        final FilterFactory2 ff = parse.getFilterFactory();
+        if (parse.isString(array, 1)) { // legacy filter syntax
+            String key = parse.get(json, 1);
+            Object value = parse.value(json, 2);
+            return ff.notEqual(ff.property(key), ff.literal(value), false);
+        } else {
+            // get the comparables
+            Expression expression1 = parse.string(json, 1);
+            Expression expression2 = parse.string(json, 2);
+            return ff.notEqual(expression1, expression2);
```
jodygarnett, line 515: @vickdw I was going to ask for your input here, this is my first looking into how expressions have been implemented and I wanted to take a consistent approach before changing anything substantially.

The above is based on the MBDecision.decisionNotEqual() which calls `parse.string(json,1)` as shown above. If I understand correctly the string method is willing to convert any literals (strings, numbers, booleans) to a string so we end up with a string equality check only. Rather than a numeric != numeric equality check.

Does this match with your understanding?

modules/extension/mbstyle/src/main/java/org/geotools/mbstyle/parse/MBObjectParser.java
```diff
@@ -332,9 +332,10 @@ public String get(JSONArray json, int index) {
         }
         if (index < json.size() && json.get(index) instanceof String) {
             return (String) json.get(index);
-        }
-        if (index < json.size() && json.get(index) instanceof JSONArray) {
-            return MBExpression.transformExpression((JSONArray) json.get(index)).toString();
+            //        }
+            //        if (index < json.size() && json.get(index) instanceof JSONArray) {
+            //            return MBExpression.transformExpression((JSONArray)
+            // json.get(index)).toString();
```
jodygarnett, line 338: This blind conversion to a string incorrectly resulted in correctly generated Expressions being represented as strings. I suspect it should be removed!

jodygarnett:
@aaime your `Werror` safety checks are earning their keep!

modules/extension/mbstyle/src/main/java/org/geotools/mbstyle/expression/MBLookup.java
```diff
@@ -56,16 +56,22 @@ public Expression lookupAt() {
     public Expression lookupGet() {
         if (json.size() == 2 || json.size() == 3) {
```
davidblasby, line 57: could you put some comments in the code here (maybe an example)?  it's hard to determine what all the json.size() == X mean.

Also, I think get requires one argument - ["get", "propName"]

modules/extension/mbstyle/src/main/java/org/geotools/mbstyle/parse/MBFilter.java
```diff
@@ -460,4 +464,116 @@ public Filter filter() {
...
+            return ff.notEqual(ff.property(key), ff.literal(value), false);
+        } else {
+            Expression expression1 = parse.string(json, 1);
+            Expression expression2 = parse.string(json, 2);
+            return ff.notEqual(expression1, expression2);
+        }
+    }
+
+    private Filter filterLessOrEqual(JSONArray array) {
+        if (json.size() != 3) {
```
davidblasby, line 513: looks like there is a lot of copy-and-paste code here and all the other functions.  They're all the same except the last line.

modules/extension/mbstyle/src/main/java/org/geotools/mbstyle/expression/MBLookup.java
jodygarnett: There is actually an option, documented in the javadoc above for three arguments.
```
[ "get", key, object]
```

This can be used to pull out a specific value not from the feature, but from the results of another expression.  I was really smart I may be able to turn this into an propertyname with xpath - but I am not going to do premature optimization.

For the library we chose the geoapi approach of format shifting the specifrication into javadocs. So you do not have to look up everything each time to see what the code is trying to implement.

modules/extension/mbstyle/src/main/java/org/geotools/mbstyle/parse/MBFilter.java
jodygarnett: Yes, i could probably be smarter on this one, but for now I was trying to exactly match the MBExpression logic. The use of `string` to evaluate expressions is a problem and I would like to have everything consistent before fixing expressions in another pull request.

Still I may be able to factor out a `expression1(json)` and `expression2(json)` method and make this more readable.

modules/extension/mbstyle/src/main/java/org/geotools/mbstyle/parse/MBObjectParser.java
vickdw: I agree, this looks like a good change.

modules/extension/mbstyle/src/main/java/org/geotools/mbstyle/parse/MBFilter.java
vickdw: Yes, that is correct.

modules/extension/mbstyle/src/main/java/org/geotools/mbstyle/parse/MBFilter.java
jodygarnett: This has now been resolved, I think the code is easier to understand.

jodygarnett:
Oracle-XE failure is unrelated, @tbarsballe would you care to review? I have addressed feedback raised thus far.

modules/extension/mbstyle/src/main/java/org/geotools/mbstyle/parse/MBFilter.java
```diff
@@ -378,86 +377,269 @@ public Filter filter() {
...
+    private Expression comparisonExpression1(JSONArray array) {
+        if (parse.isString(array, 1)) { // legacy filter syntax
+            String key = parse.get(array, 1);
+            return ff.property(key);
+        } else {
+            return parse.string(array, 1);
+        }
+    }
+    /**
+     * Comparison value1 defined as an expression (or legacy literal reference).
```
tbarsballe, line 632: "value2" instead of "value1"?

modules/extension/mbstyle/src/main/java/org/geotools/mbstyle/parse/MBFunction.java
```diff
@@ -35,7 +35,10 @@
 import org.opengis.filter.expression.Function;
 
 /**
- * MBFunction json wrapper, allowing conversion to a GeoTools Expression.
+ * MBFunction json wrapper, allowing conversion of function to a GeoTools Expression.
+ *
+ * <p>As of MapBox Style Specification v0.41.0, property expressions are the preferred approach for
```
tbarsballe, line 40: "property expressions" is not a term referenced anywhere in the MBStyle spec (I think the term you are looking for is `data expressions` - https://docs.mapbox.com/mapbox-gl-js/style-spec/expressions/#data-expressions ), and you've referred to them as just "expressions" elsewhere.

Please stick with a consistent name. The distinction between GeoTools Expressions and MapBox expressions is already confusing enough without making up terms.

modules/extension/mbstyle/src/main/java/org/geotools/mbstyle/parse/MBObjectParser.java
```diff
@@ -1145,11 +1150,20 @@ public Expression string(JSONArray json, int index) {
...
+                } else {
+                    throw new MBFormatException(
+                            context
+                                    + " string unavailable: expression' "
+                                    + expressionName
+                                    + "' not supported.");
+                }
             } else {
-                throw new MBFormatException(context + " string from JSONArray not supported");
+                throw new MBFormatException(context + " string from JSONArray not a supported");
```
tbarsballe, line 1166: Not "a supported" what. This error message seems worse then before.

modules/extension/mbstyle/src/main/java/org/geotools/mbstyle/parse/MBObjectParser.java
```diff
@@ -1426,9 +1440,74 @@ public Displacement displacement(JSONObject json, String tag, Displacement fallb
...
 
-    /** @return True if the layer has the provided property explicitly provided, False otherwise. */
+    //
+    // structure checks
+    //
+    /** @return True if json has the provided property explicitly provided, False otherwise. */
     public boolean isPropertyDefined(JSONObject json, String propertyName)
             throws MBFormatException {
+        return isDefined(json, propertyName);
+    }
```
tbarsballe, line 1450: This doesn't add anything - why do both this method and `isDefined(JSONObject json, String propertyName)` exist (I see you've overloaded isDefined below, but why didn't you just overload this method instead)?

modules/extension/mbstyle/src/main/java/org/geotools/mbstyle/parse/MBObjectParser.java
```diff
@@ -1426,9 +1440,74 @@ public Displacement displacement(JSONObject json, String tag, Displacement fallb
...
+    /** @return True if json has the property explicitly provided, False otherwise. */
+    public boolean isDefined(JSONObject json, String propertyName) throws MBFormatException {
         return json.containsKey(propertyName) && json.get(propertyName) != null;
     }
+    /** @return True if array has the property explicitly provided, False otherwise. */
+    public boolean isDefined(JSONArray json, int index) throws MBFormatException {
```
tbarsballe, line 1456: All the `int index` versions of the `is___` methods would benefit from a modified javadoc - arrays don't have properties, they have indices. Something like
```
/** @return True if array has a value at the provided index, False otherwise. */
```
Would make more sense

modules/extension/mbstyle/src/main/java/org/geotools/mbstyle/expression/MBExpression.java
```diff
@@ -151,24 +162,25 @@ public static MBExpression create(JSONArray json) {
             } else if (zoom.contains(name)) {
                 return new MBZoom(json);
             } else {
-                throw new MBFormatException("Expression \"" + name + "\" invalid.");
+                throw new MBFormatException("Expression \"" + name + "\" not supported.");
```
tbarsballe, line 165: This change isn't much of an improvement, as either could be true.
Given that GT doesn't necessarily support the full set of expressions, this distinction is relevant, because the invalid and not supported have different fixes - and invalid expression is user error (most likely a misspelling), and should be simple to fix, whereas an expression that is not supported means they need to upgrade (or wait for) a version of geotools that supports the expression.

Unfortunately I don't think there's much we can do about this - even if we maintained a list of expressions that GT didn't yet support, it would inevitably become outdated. Perhaps a more descriptive error message? Something like:
```
Expression \"" + name + "\" invalid. It may be misspelled or not supported by this implementation"
```

modules/extension/mbstyle/src/main/java/org/geotools/mbstyle/parse/MBFilter.java
```diff
@@ -38,7 +39,34 @@
...
+ *     <li><code>case</code></li>
+ *     <li><code>coalesce</code></li>
+ *     <li><code>match</code></li>
+ *     <li><code>within</code></li>
+ * </ul>
+ *
+ * <h2>Filter Other</h2>
+ *
+ * <p>Implementation Note: In previous versions of the style specification, filters were defined using the deprecated
+ * syntax documented here.
```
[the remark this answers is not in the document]
aaime: Keeping support for the old syntax is very important, the freely available OpenMapTiles styles are based on it: https://github.com/openmaptiles/osm-bright-gl-style/blob/master/style.json#L44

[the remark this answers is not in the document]
jodygarnett: It supports the old syntax, and the old test cases are preserved to ensure it stays that way.

modules/extension/mbstyle/src/main/java/org/geotools/mbstyle/parse/MBFunction.java
jodygarnett: Okay so `data expression` for mapbox style, and `GeoTools Expression` elsewhere.

modules/extension/mbstyle/src/main/java/org/geotools/mbstyle/expression/MBExpression.java
jodygarnett: Thanks, making the change you suggested.

The only positive improvement we could make here is to report the list of `data expressions` we do support; but that is likely to get long over time (and less useful as a result).

modules/extension/mbstyle/src/main/java/org/geotools/mbstyle/parse/MBFilter.java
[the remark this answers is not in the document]
jodygarnett: Changed note to:

> Implementation Note: GeoTools also supports the depreciated syntax documented here (provided by a previous versions of the Mapbox style specification).

To be clear the "Other filter" format being described is indeed supported by GeoTools.

modules/extension/mbstyle/src/main/java/org/geotools/mbstyle/parse/MBFilter.java
```diff
@@ -378,86 +377,269 @@ public Filter filter() {
...
+            Expression coalesce = MBExpression.transformExpression(json);
+            return ff.equals(coalesce, ff.literal(true));
+        } else if ("match".equals(operator)) {
+            Expression match = MBExpression.transformExpression(json);
+            return ff.equals(match, ff.literal(true));
+        } else if ("within".equals(operator)) {
+            Expression within = MBExpression.transformExpression(json);
+            return ff.equals(within, ff.literal(true));
+        } else {
+            throw new MBFormatException("Unsupported filter " + json);
```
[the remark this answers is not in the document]
jodygarnett: The code handles both cases, changing to "Unsupported data expression or filter"

modules/extension/mbstyle/src/main/java/org/geotools/mbstyle/parse/MBFunction.java
jodygarnett: The https://docs.mapbox.com/mapbox-gl-js/style-spec/other/#function page uses:

> As of v0.41.0, property expressions is the preferred method for styling features based on zoom level or the feature's properties. Zoom and property functions are still supported, but will be phased out in a future release.

modules/extension/mbstyle/src/main/java/org/geotools/mbstyle/parse/MBObjectParser.java
jodygarnett: I liked the way the `isDefined( json, propertyName)` method reads better, I should clean up references to this method.

modules/extension/mbstyle/src/main/java/org/geotools/mbstyle/parse/MBObjectParser.java
jodygarnett: Thanks, I have rewritten these.

modules/extension/mbstyle/src/main/java/org/geotools/mbstyle/parse/MBFunction.java
jodygarnett: I had a run at consistently using:
- "data expression"
- "GeoTools Expression"
- "name" (rather than tag or propertyName)
- "value" for object vs "element" for array

jodygarnett:
Thanks for the detailed review @tbarsballe, I believe I have addressed all your feedback.

modules/extension/mbstyle/src/main/java/org/geotools/mbstyle/parse/MBFunction.java
```diff
@@ -37,8 +37,8 @@
 /**
  * MBFunction json wrapper, allowing conversion of function to a GeoTools Expression.
  *
- * <p>As of MapBox Style Specification v0.41.0, property expressions are the preferred approach for
- * dynamically styling features.
+ * <p>As of v0.41.0, data expressions is the preferred method for styling features based on zoom
```
tbarsballe, line 40: "data expessions _are_"

modules/extension/mbstyle/src/main/java/org/geotools/mbstyle/expression/MBLookup.java
```diff
@@ -56,22 +56,26 @@ public Expression lookupAt() {
...
             if (json.size() == 2) {
+                // Example: ["get", "propertyName"]
                 if (parse.isString(json, 1)) {
                     String propertyName = parse.get(json, 1);
                     return ff.property(propertyName);
-                } else {
+                }
+                // Example: ["get", key]
+                else {
```
tbarsballe, line 65: Pretty sure splitting the closing brace from the `else` violates the style guide (although Travis CI seems happy with it, so I guess not?). Not really important either way.

aaime:
@jodygarnett I did not see a reply regarding the backwards compatibility towards old style filters (could be I missed it). Can you ensure those will keep on working? I can give you a data directory to try out with the OpenMapTiles styles if you're interested.

jodygarnett:
> @jodygarnett I did not see a reply regarding the backwards compatibility towards old style filters (could be I missed it). Can you ensure those will keep on working? I can give you a data directory to try out with the OpenMapTiles styles if you're interested.

I replied, but perhaps it is in a resolved conversation. The code preservers the existing support for "other filter", and is only additive in nature (recognizing data expressions).

I will be have to revisit any regressions, but I do not have any capacity to work on this further at present.

modules/extension/mbstyle/src/main/java/org/geotools/mbstyle/parse/MBFilter.java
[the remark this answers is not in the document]
jodygarnett: @aaime I am marking this conversation as unresolved, I updated the docs to make it clear the ongoing support for "other filter"

PR #2931, opened 2020-05-13
[23.x backport] [GEOT-3377] Charset not used for DBF header

dvfed:
Backport of #2912.

## NOTE about what can go wrong with this PR

If this PR will be applied, in some rare cases of when `DbaseFileHeader` is used directly there can be **wrong results**. Consider such a piece of code:
```
DbaseFileHeader header = new DbaseFileHeader(); // default charset
header.readHeader(inChannel, ANOTHER_NON_ENG_CHARSET);
header.addColumn("NON_ENG_COL_NAME", 'C', 100, 0);
writer = new DbaseFileWriter(header, outChannel, ANOTHER_NON_ENG_CHARSET);
writer.write(new Object[] { ... some data ... });
writer.close();
```
Here the `header` will use default charset on write stage, not `ANOTHER_NON_ENG_CHARSET` as could expect an inattentive end-user.

If header is used indirectly, e.g. via `DbaseFileReader` or `ShapefileDataStore` - everything must be fine.

Please let me know how this can be enhanced or perhaps documented additionally. If this is bearable i'll cherry-pick it for 22.x branch too.

### DIFF
--- a/modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/ShapefileDataStore.java
+++ b/modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/ShapefileDataStore.java
@@ -277,7 +277,7 @@ public class ShapefileDataStore extends ContentDataStore implements FileDataStor
             writer.writeHeaders(new Envelope(), shapeType, 0, 100);
         }
 
-        DbaseFileHeader dbfheader = createDbaseHeader(featureType);
+        DbaseFileHeader dbfheader = createDbaseHeader(featureType, charset);
 
         dbfheader.setNumRecords(0);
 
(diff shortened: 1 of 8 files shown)


jodygarnett:
The build tests are having some trouble, appears unrelated to your change...

modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/ShapefileDataStore.java
```diff
@@ -320,10 +320,10 @@ String toSingleLineWKT(CoordinateReferenceSystem crs) {
      * Attempt to create a DbaseFileHeader for the FeatureType. Note, we cannot set the number of
      * records until the write has completed.
      */
-    protected static DbaseFileHeader createDbaseHeader(SimpleFeatureType featureType)
-            throws IOException, DbaseFileException {
+    protected static DbaseFileHeader createDbaseHeader(
+            SimpleFeatureType featureType, Charset charset) throws IOException, DbaseFileException {
```
jodygarnett, line 324: We tend not to backport API changes, I understand that this a protected constructor so the change is "internal".

modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/dbf/DbaseFileHeader.java
```diff
@@ -504,17 +524,26 @@ public int getHeaderLength() {
...
      *
      * @param channel A readable byte channel. If you have an InputStream you need to use, you can
      *     call java.nio.Channels.getChannel(InputStream in).
+     * @param charset A charset that will be used to read header field's names.
      * @throws IOException If errors occur while reading.
+     * @deprecated Use constructor {@link #constructor(Charset)} and {@link
+     *     #readHeader(ReadableByteChannel)} method instead.
      */
+    @Deprecated
     public void readHeader(ReadableByteChannel channel, Charset charset) throws IOException {
```
jodygarnett, line 541: Thanks for preserving backwards compatibility

dvfed:
I've edited the PR description above to make it clear that in some circumstances the changes proposed by this pull request can lead to wrong results.

aaime:
It's looking good to me as well. There is an odd failure in the JDK 11 build but it does not seem related to this change, restarted it anyways, let's see.

PR #2940, opened 2020-05-20
GEOT-6584 - DB2 fix EPSG:4326 axis order to EAST-NORTH

dwadler:
A failure to display anything was occurring in the GeoTools DB2 tutorial program that was developed many years ago. It turns out that this was due to the use of EPSG:4326 which specifies axis order as NORTH-EAST while the data axis order was EAST-NORTH. At some point in time, GeoTools changed to follow the EPSG:4326 specification which hadn't been the case previously.

The behavior was restored by changing CRS.decode(code) to CRS.decode(code,true). to force longitude first.

Minor changes also made to online testcases.

It is not currently possible to run the online tests in the GeoTools test environment due to the lack of an available DB2 environment. I am looking into the possibility of providing a Docker image if that would work.

I am taking over the GeoTools DB2 plugin from Christian Mueller with his agreement.

## Checklist

For all pull requests:


modules/plugin/jdbc/jdbc-db2/src/main/java/org/geotools/data/db2/DB2NGDataStoreFactory.java
```diff
@@ -75,6 +80,24 @@
                     "Use selectivity clause for spatial queries",
                     false,
                     Boolean.TRUE);
+    /* the following values are helpful to pre-set for testing - comment out for release */
```
bradh, line 83: This seems like a bad idea. You should provide a properties option, or do this in the test code.

aaime: Agreed, the values should be in a property file driving the test, like we do for other databases, see here:
https://github.com/geotools/geotools/blob/master/.github/workflows/postgis_online.yml#L37

No manual action takes place during releases.

dwadler:
I agree with the comments above and have reverted DB2NGDataStoreFactory.java to the prior version. The changes were convenient for local testing.

In the DB2 pom.xml file, the dependencies on the DB2 driver need to be commented out or the build will fail as the DB2 driver can not currently be resolved from a repository.

modules/plugin/jdbc/jdbc-db2/src/main/java/org/geotools/data/db2/DB2NGDataStoreFactory.java
dwadler: I'm not clear on where the property file driving the test is located or how it is setup. Currently, automatic testing of online functionality is not possible.

Is it possible to use a Docker image for the online testing?

How are the Oracle JDBC driver files handled?

dwadler: It looks like my pull request is due to the test Oracle XE failing. How do I get around this?

jodygarnett:
> In the DB2 pom.xml file, the dependencies on the DB2 driver need to be commented out or the build will fail as the DB2 driver can not currently be resolved from a repository.

One nice way around that is to use a profile, so you can still test DB2 if the driver is installed locally.

modules/plugin/jdbc/jdbc-db2/src/main/java/org/geotools/data/db2/DB2NGDataStoreFactory.java
jodygarnett: When reviewing we go in and check why the build failed, in this case it is not due travis (and not any change you made in this PR we are reviewing).

jodygarnett:
@dwadler I believe you have addressed all outstanding feedback?

modules/plugin/jdbc/jdbc-db2/src/test/java/org/geotools/data/db2/DB23DOnlineTest.java
```diff
@@ -16,11 +16,21 @@
...
 import org.geotools.jdbc.JDBC3DOnlineTest;
 import org.geotools.jdbc.JDBC3DTestSetup;
+import org.geotools.util.factory.Hints;
+import org.opengis.filter.spatial.BBOX3D;
 
 public class DB23DOnlineTest extends JDBC3DOnlineTest {
 
+    @Override
+    protected void connect() throws Exception {
+        Hints.putSystemDefault(Hints.FORCE_LONGITUDE_FIRST_AXIS_ORDER, Boolean.TRUE);
```
jodygarnett, line 30: These kind of tests can make the build very brittle (as other tests running concurrently with this one may get confused as their environment changes). When running the test as part of a maven build the referencing system may already be setup using the default axis order policy.

If you absolutely need this kind of thing, [WMSCoverageReaderTest](https://github.com/geotools/geotools/blob/master/modules/extension/wms/src/test/java/org/geotools/ows/wms/map/WMSCoverageReaderTest.java) uses:

```
    @Before
    public void setup() {
        System.setProperty("org.geotools.referencing.forceXY", "true");
        Hints.putSystemDefault(Hints.FORCE_AXIS_ORDER_HONORING, "http");
        CRS.reset("all");
    }

    @AfterClass
    public static void teardown() {
        System.clearProperty("org.geotools.referencing.forceXY");
        Hints.putSystemDefault(Hints.FORCE_AXIS_ORDER_HONORING, "");
        CRS.reset("all");
    }
```

jodygarnett: I added a fix based on the above to your branch, hopefully it worked :)

dwadler: Thank you for applying the fix but I don't understand why this was a problem in the first place. If the DB2 online tests aren't being run, how is this becoming a problem in DB23DOnlineTest?

I thought setting System.setProperty("org.geotools.referencing.forceXY", "true") and Hints.putSystemDefault(Hints.FORCE_AXIS_ORDER_HONORING, "http") were somewhat equivalent.

dwadler: Where can I see the fix you put in?

aaime:
Checking the build, seeing a number of compile failure:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.0:testCompile (default-testCompile) on project gt-jdbc-db2: Compilation failure: Compilation failure: 
[ERROR] /home/travis/build/geotools/geotools/modules/plugin/jdbc/jdbc-db2/src/test/java/org/geotools/data/db2/DB23DOnlineTest.java:[28,6] cannot find symbol
[ERROR]   symbol:   class Before
[ERROR]   location: class org.geotools.data.db2.DB23DOnlineTest
[ERROR] /home/travis/build/geotools/geotools/modules/plugin/jdbc/jdbc-db2/src/test/java/org/geotools/data/db2/DB23DOnlineTest.java:[34,6] cannot find symbol
[ERROR]   symbol:   class AfterClass
[ERROR]   location: class org.geotools.data.db2.DB23DOnlineTest
[ERROR] /home/travis/build/geotools/geotools/modules/plugin/jdbc/jdbc-db2/src/test/java/org/geotools/data/db2/DB23DOnlineTest.java:[31,9] cannot find symbol
[ERROR]   symbol:   variable CRS
[ERROR]   location: class org.geotools.data.db2.DB23DOnlineTest
[ERROR] /home/travis/build/geotools/geotools/modules/plugin/jdbc/jdbc-db2/src/test/java/org/geotools/data/db2/DB23DOnlineTest.java:[37,9] cannot find symbol
[ERROR]   symbol:   variable CRS
[ERROR]   location: class org.geotools.data.db2.DB23DOnlineTest
[ERROR] -> [Help 1]
[ERROR] 
```

May have been caused by Jody's last push: https://github.com/geotools/geotools/pull/2940/commits/3e4574ed0aeb0a01718d4fa1096c9255ab9f2672

Jody the JDBC tests are still fully based on Junit 3, I don't think @before will behave properly, if at all.

modules/plugin/jdbc/jdbc-db2/src/test/java/org/geotools/data/db2/DB23DOnlineTest.java
aaime: @jodygarnett we don't use concurrent tests in the same VM, it's an option of Maven we don't use, too many static settings around in GeoTools to go that way.

aaime:
@dwadler I think I've seen somewhere a request of yours about docs explaining how tests work. Not sure there is any, but in short, first time you run the tests locally, on your machine, they will create a new hidden folder with a sample file, something like ``$HOME/.geotools/db2.properties.example``.  

Rename that file into ``$HOME/.geotools/db2.properties`` and change the property values to match your environment, and tests should start running.

The builds we already have create that file just before running the java build, see for example this Github action for PostGIS:
https://github.com/geotools/geotools/blob/master/.github/workflows/postgis_online.yml#L37
or the Travis build machinery for Oracle and SQL Server in these two files:
https://github.com/geotools/geotools/blob/master/.travis.yml
https://github.com/geotools/geotools/tree/master/.travis

dwadler:
The DB2 plugin does generate the file .geotools/db2.properties.example. 
I renamed it to db2.properties and have been able to run all the online 
tests successfully in my environment.

I had been asking about some scripts that seem to be run by Travis that 
control the build/test process, corresponding to the links below. It had 
been recommended that I setup something like this for DB2.

We need to come up with some way of running DB2 in the Travis environment

-------- Forwarded Message --------

Subject: 	Re: [geotools/geotools] GEOT-6584 - Fix EPSG:4326 axis order 
to EAST-NORTH (#2940)
Date: 	Mon, 25 May 2020 01:21:14 -0700

dwadler:
I don't understand what it is about my code that is causing these failures.

aaime:
Neither @before nor @beforeClass will work here, JDBC tests are still based on JUnit 3, Junit 4 annotations won't do anything.

jodygarnett:
gak, I am sorry for breaking things :(

jodygarnett:
I have reverted my change (I appologize for the mistake and not following up).

With the change removed we have one oracle failure:
```
[ERROR] Tests run: 3, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 12.543 s <<< FAILURE! - in org.geotools.data.oracle.OraclePrimaryKeyFinderOnlineTest
4520[ERROR] testSequencedPrimaryKey(org.geotools.data.oracle.OraclePrimaryKeyFinderOnlineTest)  Time elapsed: 4.745 s  <<< FAILURE!
4521junit.framework.AssertionFailedError: actual fid: null does not match expected SEQTABLE.4(\..*)?
4522	at org.geotools.data.oracle.OraclePrimaryKeyFinderOnlineTest.addFeature(OraclePrimaryKeyFinderOnlineTest.java:47)
```

jodygarnett:
Manually merged.

PR #2953, opened 2020-05-29
[GEOT-6577] Allow plugin to handle multiple valid date formats from elastic

travisb-fe:
Updates the `dateFormat` field of `ElasticAttribute`s to be a List of Strings, rather than a single String. Because Elasticsearch can serve back multiple valid date formats, separated by `||`, we need to have a way to handle multiple different possible date formats.

An important note: this update breaks previously existing data stores due to changing the underlying attributes. I'm not sure the proper way to do a data store migration for a change like this,

Coincides with GeoServer PR: https://github.com/geoserver/geoserver/pull/4270

## Checklist

For all pull requests: 

### DIFF
--- a/modules/unsupported/elasticsearch/src/main/java/org/geotools/data/elasticsearch/ElasticFeatureTypeBuilder.java
+++ b/modules/unsupported/elasticsearch/src/main/java/org/geotools/data/elasticsearch/ElasticFeatureTypeBuilder.java
@@ -92,8 +92,8 @@ class ElasticFeatureTypeBuilder extends SimpleFeatureTypeBuilder {
                                 attributeBuilder.buildDescriptor(
                                         attributeName, attributeBuilder.buildType());
                     }
-                    if (att != null && attribute.getDateFormat() != null) {
-                        att.getUserData().put(DATE_FORMAT, attribute.getDateFormat());
+                    if (att != null && attribute.getValidDateFormats() != null) {
+                        att.getUserData().put(DATE_FORMAT, attribute.getValidDateFormats());
                     }
                     if (att != null) {
                         att.getUserData().put(FULL_NAME, attribute.getName());
--- a/modules/unsupported/elasticsearch/src/main/java/org/geotools/data/elasticsearch/ElasticFeatureTypeBuilder.java
+++ b/modules/unsupported/elasticsearch/src/main/java/org/geotools/data/elasticsearch/ElasticFeatureTypeBuilder.java
@@ -22,6 +22,7 @@ import static org.geotools.data.elasticsearch.ElasticConstants.FULL_NAME;
 import static org.geotools.data.elasticsearch.ElasticConstants.GEOMETRY_TYPE;
 import static org.geotools.data.elasticsearch.ElasticConstants.NESTED;
 
+import java.util.ArrayList;
 import java.util.List;
 import java.util.logging.Level;
 import java.util.logging.Logger;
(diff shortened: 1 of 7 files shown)


jodygarnett:
Looks like there is one formatting or code quality error found:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-pmd-plugin:3.12.0:check (default) on project gt-elasticsearch: You have 1 PMD violation. For more details see: /home/travis/build/geotools/geotools/modules/unsupported/elasticsearch/target/pmd.xml -> [Help 1]
12738
```

travisb-fe:
Did not know about the styling check, now I will check that locally from now on first. Updated and fixed, the null check was unneeded, plus I think there was a potential logic error I've corrected.

aaime:
Checking here and in https://github.com/geoserver/geoserver/pull/4270, I have the impression this one is still in progress right?

Did @sjudeng have a look at either PRs?

travisb-fe:
Hi Andrea, yes I've gotten back to working on this one. It's functional, but I am working on making it backwards compatible. Currently it'll break any existing Elasticsearch datastores.

modules/unsupported/elasticsearch/src/main/java/org/geotools/data/elasticsearch/ElasticFeatureTypeBuilder.java
```diff
@@ -92,8 +93,15 @@ public SimpleFeatureType buildFeatureType() {
...
-                        att.getUserData().put(DATE_FORMAT, attribute.getDateFormat());
+                    if (att != null
+                            && (attribute.getValidDateFormats() != null
+                                    || attribute.getDateFormat() != null)) {
+                        if (attribute.getValidDateFormats() == null) {
+                            List<String> validFormats = new ArrayList<String>();
+                            validFormats.add(attribute.getDateFormat());
+                            attribute.setValidDateFormats(validFormats);
+                        }
+                        att.getUserData().put(DATE_FORMAT, attribute.getValidDateFormats());
```
travisb-fe, line 104: On the mailing list I discussed about using a Strategy pattern to interpret the date behaviour. However, due to how small the code sections are that use the date format, I found this way of handling it to be pretty straightforward, but I'm not sure it's the correct approach.

It does work that will be unnecessary for all future DataStores. Essentially this just "corrects" existing DataStores that were using the old date format. It could even be removed in a later release. Otherwise, I am not sure how old DataStores would ever get updated to use the new date format behaviour.

The Strategy option discussed would also work and be a permanent solution, but it wouldn't update old DataStores as far as I can tell. The old DataStores with the old date behaviour is technically incorrect and can fail to interpret Dates correctly. We could leave them alone and instruct users to recreate the DataStores with the new way of handling date formats (which would be handled via the Strategy, allowing both to exist simultaneously).

I am not sure which approach is better here, but hopefully I made the intent clear. If this way is not recommended, and using the Strategy is more proper, I'll update to that.

travisb-fe:
Finally pushed some changes to make it backwards compatible. I made a comment on the code in question since I am unsure what I've done makes sense.

travisb-fe:
Gonna tag @sjudeng again for visibility. It's updated to be backwards compatible now - if what I've done doesn't seem sensible, I can go with the Strategy idea instead.

sjudeng:
Looks good to me. Did you validate the online tests (`-Ponline`) all work with the updates as well?

It's great you've already done the work to maintain backwards compatibility though if you prefer I think it would also be fine to remove dateFormat and just have dateFormats and make it a breaking change. But it's up to you.

I know it's been discussed on the list but generally I agree it's reasonable that there might eventually/already be some breaking changes in this version of the plugin and wouldn't want to see you or other potential contributors get slowed down too much trying to maintain backwards compatibility where it doesn't make sense. Ideally we'd have a migration section in the docs noting any breaking changes.

travisb-fe:
I just ran the online tests on this branch and everything looked good - no errors.

Yeah originally it was going to be useful on my end too, though it ended up not being a huge deal. Either way, if everything looks good for now, then I am satisfied. I'll keep that in mind for the future. :)

aaime:
@sjudeng you could setup a PR build check like we did for [postgis](https://github.com/geotools/geotools/blob/master/.github/workflows/postgis_online.yml), [sqlserver and oracle ](https://github.com/geotools/geotools/blob/master/.travis.yml)

sjudeng:
@travisb-fe Thanks for confirming. The updates (here and on GeoServer side) look good to me and based on the last comment it sounds like it's good you did make the update backwards compatible.

@aaime That would be great and thanks for the reference links. I can try to get that added in the coming weeks unless anyone else (Travis?) wants to take a shot at it.

travisb-fe:
This does not depend on adding the build checks right? I agree that would be a nice thing to have, though I have not been looking at this for a bit.

Everything here is good to merge then? For some reason the GeoServer PR has some build failures, though I'm not sure if it is because this needs to get merged first - the build errors I can't even tell how or if it's related to these changes.

aaime:
Yeah, this dragged too long, time to merge, the build checks are useful but not the focus of the PR itself, should be another one.

PR #2961, opened 2020-06-04
Improvements to the SDMX and ArcGIS ReST API datastores

lmorandini:
- Added support for SDMX ReST  2.1
- Improved support for CRS in ArcGIS ReST API


aaime:
Hi @lmorandini, GeoTools 20.x is dead and buried, the snapshot jars for it are also missing. You should be making a PR against master.

lmorandini:
> Hi @lmorandini, GeoTools 20.x is dead and buried, the snapshot jars for it are also missing. You should be making a PR against master.

No worries, thanks @aaime

PR #2971, opened 2020-06-08
[GEOT-6616]: mapGet FilterFunction to get a value from a map given th…

dromagnoli:
…e mapName.key identifier

## Checklist

For all pull requests:

### DIFF
--- a/modules/library/main/src/main/java/org/geotools/filter/function/FilterFunction_mapGet.java
+++ b/modules/library/main/src/main/java/org/geotools/filter/function/FilterFunction_mapGet.java
@@ -1,0 +1,75 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ *
+ *    (C) 2020, Open Source Geospatial Foundation (OSGeo)
+ *
+ *    This library is free software; you can redistribute it and/or
+ *    modify it under the terms of the GNU Lesser General Public
+ *    License as published by the Free Software Foundation;
+ *    version 2.1 of the License.
+ *
+ *    This library is distributed in the hope that it will be useful,
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ *    Lesser General Public License for more details.
+ */
+package org.geotools.filter.function;
+
+import static org.geotools.filter.capability.FunctionNameImpl.parameter;
+
+import java.util.Map;
+import org.geotools.factory.CommonFactoryFinder;
+import org.geotools.filter.FunctionExpressionImpl;
+import org.geotools.filter.capability.FunctionNameImpl;
+import org.geotools.util.Converters;
+import org.opengis.filter.FilterFactory2;
+import org.opengis.filter.capability.FunctionName;
+import org.opengis.filter.expression.PropertyName;
+import org.opengis.filter.expression.VolatileFunction;
+
+/** Extracts a value from a map given the mapname.key identifier as a parameter */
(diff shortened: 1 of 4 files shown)


modules/library/main/src/main/java/org/geotools/filter/function/FilterFunction_mapGet.java
```diff
@@ -0,0 +1,75 @@
...
+import org.geotools.filter.FunctionExpressionImpl;
+import org.geotools.filter.capability.FunctionNameImpl;
+import org.geotools.util.Converters;
+import org.opengis.filter.FilterFactory2;
+import org.opengis.filter.capability.FunctionName;
+import org.opengis.filter.expression.PropertyName;
+import org.opengis.filter.expression.VolatileFunction;
+
+/** Extracts a value from a map given the mapname.key identifier as a parameter */
+public class FilterFunction_mapGet extends FunctionExpressionImpl implements VolatileFunction {
```
aaime, line 32: @jodygarnett is not happy with these names :-D

```suggestion
public class MapGetFunction extends FunctionExpressionImpl implements VolatileFunction {
```

modules/library/main/src/main/java/org/geotools/filter/function/FilterFunction_mapGet.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    public Object evaluate(Object feature) {
+        String name = getExpression(0).evaluate(feature, String.class);
+
+        if (name == null) {
+            return null;
+        }
+
+        String[] split = name.split("\\.");
```
aaime, line 63: Normally access to property values is done via dedicated property support, e.g., it could be:

``mapGet("myProperty", 'mapKey')``

Where "myProperty" becomes a PropertyName... if it's dynamic, it could also be:

``mapGet(property('myProperty'), mapKey);

At the very least, I would check if the key has a dot in the middle, and if not, evaluate directly assuming the "feature" is a map.

modules/plugin/jdbc/jdbc-postgis/src/test/java/org/geotools/data/postgis/PostGISHStoreOnlineTest.java
```diff
@@ -124,6 +126,18 @@ public void testNullEntry() throws Exception {
         assertNull(object);
     }
 
+    @Test
```
aaime, line 129: Hum ok... but the evaluation is still done in memory right?

modules/library/main/src/main/java/org/geotools/filter/function/FilterFunction_mapGet.java
dromagnoli: ok. Thanks for the hint.
I will rename it. (I have seen a ton of FilterFunction_something so I used that :) Probably old naming convention.)

modules/plugin/jdbc/jdbc-postgis/src/test/java/org/geotools/data/postgis/PostGISHStoreOnlineTest.java
dromagnoli: Yep. it's testing the MapGet filter function functionalities, not the HSTORE native support for extract an item with its key (which is still not implemented). If I have understood your comment, correctly.

dromagnoli:
being replaced by:
https://github.com/geotools/geotools/pull/2977

modules/library/main/src/main/java/org/geotools/filter/function/FilterFunction_mapGet.java
dromagnoli: fixed. thanks

PR #2982, opened 2020-06-11
[GEOT-6623] Add support for jsonPointer to SQL translation in a Postg…

taba90:
…reSQL database

Jira ticket https://osgeo-org.atlassian.net/browse/GEOT-6623

This pr add support to translate jsonPointer function to PostgreSQL dialect.
eg. jsonPointer( property, '/path/to/json/attribute') to property::json -> path -> to -> json ->> attribute.

## Checklist

For all pull requests:

### DIFF
--- a/modules/library/jdbc/src/main/java/org/geotools/data/jdbc/FilterToSQL.java
+++ b/modules/library/jdbc/src/main/java/org/geotools/data/jdbc/FilterToSQL.java
@@ -837,6 +837,16 @@ public class FilterToSQL implements FilterVisitor, ExpressionVisitor {
         encodeBinaryComparisonOperator(filter, extraData, left, right, leftContext, rightContext);
     }
 
+    /**
+     * Encode a BinaryComparisonOperator to SQL
+     *
+     * @param filter the comparison operator to be turned to SQL
+     * @param extraData extraData
+     * @param left left parameter of the binary operator
+     * @param right right parameter of the binary operator
+     * @param leftContext expression type of the right parameter used as context for the left parameter
+     * @param rightContext expression type of the left parameter used as context for the right parameter
+     */
     protected void encodeBinaryComparisonOperator(
             BinaryComparisonOperator filter,
             Object extraData,
--- a/modules/library/jdbc/src/main/java/org/geotools/data/jdbc/FilterToSQL.java
+++ b/modules/library/jdbc/src/main/java/org/geotools/data/jdbc/FilterToSQL.java
@@ -834,6 +834,16 @@ public class FilterToSQL implements FilterVisitor, ExpressionVisitor {
         Class rightContext = getExpressionType(left);
         Class leftContext = getExpressionType(right);
 
+        encodeBinaryComparisonOperator(filter, extraData, left, right, leftContext, rightContext);
+    }
+
+    protected void encodeBinaryComparisonOperator(
+            BinaryComparisonOperator filter,
+            Object extraData,
+            Expression left,
+            Expression right,
+            Class leftContext,
+            Class rightContext) {
         // case sensitivity
         boolean matchCase = true;
         if (!filter.isMatchingCase()) {
@@ -882,7 +892,6 @@ public class FilterToSQL implements FilterVisitor, ExpressionVisitor {
                 f.setParameters(Arrays.asList(right));
                 f.accept(this, Arrays.asList(rightContext));
             }
-
         } catch (java.io.IOException ioe) {
             throw new RuntimeException(IO_ERROR, ioe);
         }
--- a/modules/library/jdbc/src/main/java/org/geotools/data/jdbc/FilterToSQL.java
+++ b/modules/library/jdbc/src/main/java/org/geotools/data/jdbc/FilterToSQL.java
@@ -844,8 +844,10 @@ public class FilterToSQL implements FilterVisitor, ExpressionVisitor {
      * @param extraData extraData
      * @param left left parameter of the binary operator
      * @param right right parameter of the binary operator
-     * @param leftContext expression type of the right parameter used as context for the left parameter
-     * @param rightContext expression type of the left parameter used as context for the right parameter
+     * @param leftContext expression type of the right parameter used as context for the left
+     *     parameter
+     * @param rightContext expression type of the left parameter used as context for the right
+     *     parameter
      */
     protected void encodeBinaryComparisonOperator(
             BinaryComparisonOperator filter,
--- a/modules/library/main/src/main/java/org/geotools/filter/function/JsonPointerFunction.java
+++ b/modules/library/main/src/main/java/org/geotools/filter/function/JsonPointerFunction.java
@@ -30,6 +30,7 @@ import java.io.StringWriter;
 import org.geotools.filter.FunctionExpressionImpl;
 import org.geotools.filter.capability.FunctionNameImpl;
 import org.opengis.filter.capability.FunctionName;
+import org.opengis.filter.expression.ExpressionVisitor;
 
 /** Applies a JSON pointer on a given JSON string, extracting a value out of it */
 public class JsonPointerFunction extends FunctionExpressionImpl {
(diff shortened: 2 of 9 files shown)


modules/library/main/src/main/java/org/geotools/filter/function/JsonPointerFunction.java
```diff
@@ -152,4 +154,9 @@ private void serializeObject(JsonParser parser, JsonGenerator generator) throws
         }
         generator.writeEndObject();
     }
+
+    @Override
+    public Object accept(ExpressionVisitor visitor, Object extraData) {
```
aaime, line 159: Does the same as the overridden method, should be removed.

modules/library/jdbc/src/main/java/org/geotools/data/jdbc/FilterToSQL.java
```diff
@@ -824,6 +824,16 @@ protected void visitBinaryComparisonOperator(BinaryComparisonOperator filter, Ob
         Class rightContext = getExpressionType(left);
         Class leftContext = getExpressionType(right);
 
+        encodeBinaryComparisonOperator(filter, extraData, left, right, leftContext, rightContext);
+    }
+
+    protected void encodeBinaryComparisonOperator(
```
aaime, line 830: This class is extended for all databases, please add javadoc for this method so that subclassers know what it is for.

modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/FilterToSqlHelper.java
```diff
@@ -625,6 +633,42 @@ public boolean visitFunction(Function function, Object extraData) throws IOExcep
...
+            // if not a string need to cast the json attribute
+            boolean needCast =
+                    extraData != null
+                            && extraData instanceof Class
+                            && !extraData.equals(String.class);
+
+            if (needCast) out.write('(');
+            json.accept(delegate, null);
+            out.write(" ::json ");
+            String strPointer = ((Literal) pointer).getValue().toString();
```
aaime, line 649: In general you cannot assume it's going to be a Literal, functions must work with whatever expression they are given as input. The existing function, without the postgresql encoding, would work even if the pointer was specified with a complex expression, possibly attribute based.

If you check [SQLDialect](https://github.com/geotools/geotools/blob/7529923af1143cfe25cd475bb4f05c11fca83957/modules/library/jdbc/src/main/java/org/geotools/jdbc/SQLDialect.java#L1259), it gets called for splitting filters. For this case, I suggest to override it, and check if there is a pointer function, and if the function is called in a "static" way. Static does not mean "literal" directly, it just means that the expression tree referenced by  "pointer" :

- does not use any attribute
- does not use any VolatileFunction

If those conditions are met, you can include the function in the capabilities, otherwise, you exclude it from them and have the splitter move it to the post-filter.
I think (please verify) that the JDBCDataStore code makes the filter go through a SimplfyingFilterVisitor, which is already doing this job for you:
https://github.com/geotools/geotools/blob/029dc5c8ccc503de3cfd7f4d1c939d8056c260c4/modules/library/main/src/main/java/org/geotools/filter/visitor/SimplifyingFilterVisitor.java#L453

If so, all you need to do is to write a DefaultFilterVisitor, if it finds a JSONPointerFunction, it means it was not eliminated by the simplifier, and thus it's dynamic, cannot be encoded in SQL.

modules/plugin/jdbc/jdbc-postgis/src/test/java/org/geotools/data/postgis/PostgisFilterToSQLTest.java
```diff
@@ -248,4 +250,27 @@ public void testFunctionLike() throws Exception {
...
+        Function pointer =
+                ff.function("jsonPointer", ff.property("testJSON"), ff.literal("/arr/0"));
+
+        filterToSql.encode(pointer);
+        String sql = writer.toString().toLowerCase().trim();
+        assertEquals("testjson ::json  -> 'arr' ->> 0", sql);
+    }
+
+    @Test
+    public void testBinaryComparisonWithJsonPointer() throws Exception {
```
aaime, line 266: Please add a test for nested properties, like a/b (instead of array case) and also one for a non-encodable case, like, the pointer value is contained in a second attribute of the feature.

modules/library/main/src/main/java/org/geotools/filter/function/JsonPointerFunction.java
taba90: fixed

modules/library/jdbc/src/main/java/org/geotools/data/jdbc/FilterToSQL.java
taba90: fixed

modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/FilterToSqlHelper.java
taba90: thanks for the point and the explanation, fixed

modules/plugin/jdbc/jdbc-postgis/src/test/java/org/geotools/data/postgis/PostgisFilterToSQLTest.java
taba90: fixed

taba90:
thanks @aaime your feedback have been applied

modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/PostGISDialect.java
```diff
@@ -1490,4 +1499,42 @@ protected String getEstimatedExtentFunction() {
...
+
+                    @Override
+                    public Object visit(Function expression, Object notUsed) {
+                        if (expression instanceof JsonPointerFunction) {
+                            // takes the json pointer param to check if
+                            // can be encoded
+                            Expression param = expression.getParameters().get(1);
+                            if ((expression.getParameters().get(1) instanceof Literal))
+                                // can encode
+                                fcs.addType(JsonPointerFunction.class);
```
aaime, line 1517: Thread safety issue here. The capabilities object is big and normally shared, you should not be modifying it. Since it's large, cloning it is also somehow expensive.
Suggestion: change the PostPreProcessFilterSplittingVisitor so that the "!fcs.supports(expression.getClass())" check is delegated to a separate method that you can override here.

modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/PostGISDialect.java
```diff
@@ -1490,4 +1499,42 @@ protected String getEstimatedExtentFunction() {
...
+                            else {
+                                FilterAttributeExtractor extractor = new FilterAttributeExtractor();
```
aaime, line 1519: The filter to sql will fail if the expression is not a Literal though (class cast exception).

aaime: Never mind, I see you are replacing it with a Literal... however, you should not be modifying the filter here, unless you are sure it has been through a copy first, otherwise you risk changing some cached state (e.g., the structure of a in memory SLD representation, or some such).

modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/PostGISDialect.java
taba90: fixed

modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/PostGISDialect.java
taba90: Thanks for your suggestion @aaime fixed

nmco:
Failure looks unrelated:
image

Log of the failed build:
```
Waiting for SQL Server to start...
.travis/install-mssql2019.sh: 96: .travis/install-mssql2019.sh: 1++: not found
Waiting for SQL Server to start...
.travis/install-mssql2019.sh: 96: .travis/install-mssql2019.sh: 1++: not found
Waiting for SQL Server to start...
.travis/install-mssql2019.sh: 96: .travis/install-mssql2019.sh: 1++: not found
Waiting for SQL Server to start...
The job exceeded the maximum time limit for jobs, and has been terminated.
```

aaime:
Looks good, merging.

PR #2995, opened 2020-06-16
Used release version of SDMX library

lmorandini:


aaime:
There is a QA failure but it's unrelated:

```
[INFO] Processed 179 files (1 non-complying).
[ERROR] Found 1 non-complying files, failing build
[ERROR] Non complying file: /home/travis/build/geotools/geotools/modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/RasterManager.java
[INFO] 
```

I believe this one can be merged, the above issue has already been fixed as far as I know.

PR #3016, opened 2020-06-24
[GEOT-6129] WFS GetPropertyValue requests fail with App-Schema define…

taba90:
…d feature types

This pr fixes a bug issued by a GetPropertyValue operation with a propertyName pointing to a SimpleAttribute in a ComplexFeature. JIRA ticket https://osgeo-org.atlassian.net/browse/GEOT-6129

## Checklist

For all pull requests:

### DIFF
--- a/modules/extension/xsd/xsd-wfs/src/main/java/org/geotools/wfs/PropertyValueCollection.java
+++ b/modules/extension/xsd/xsd-wfs/src/main/java/org/geotools/wfs/PropertyValueCollection.java
@@ -33,10 +33,7 @@ import org.geotools.feature.NameImpl;
 import org.geotools.feature.type.FeatureTypeFactoryImpl;
 import org.geotools.gml3.v3_2.GML;
 import org.geotools.xs.XS;
-import org.opengis.feature.Attribute;
-import org.opengis.feature.Feature;
-import org.opengis.feature.FeatureFactory;
-import org.opengis.feature.Property;
+import org.opengis.feature.*;
 import org.opengis.feature.simple.SimpleFeature;
 import org.opengis.feature.type.AttributeDescriptor;
 import org.opengis.feature.type.AttributeType;
@@ -180,7 +177,8 @@ public class PropertyValueCollection extends AbstractCollection<Attribute> {
                             descriptor.isNillable(),
                             descriptor.getDefaultValue());
 
-            if (next instanceof SimpleFeature) {
+            if (next instanceof SimpleFeature || !(value instanceof ComplexAttribute)) {
+                if (value instanceof Attribute) value = ((Attribute) value).getValue();
                 return factory.createAttribute(value, newDescriptor, null);
             } else {
                 return factory.createComplexAttribute(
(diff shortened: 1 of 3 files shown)


modules/extension/xsd/xsd-wfs/src/main/java/org/geotools/wfs/PropertyValueCollection.java
```diff
@@ -180,7 +177,8 @@ public Object next() {
                             descriptor.isNillable(),
                             descriptor.getDefaultValue());
 
-            if (next instanceof SimpleFeature) {
+            if (next instanceof SimpleFeature || !(value instanceof ComplexAttribute)) {
```
nmco, line 180: Humm, something looks strange here. [``Attribute``](https://github.com/geotools/geotools/blob/41ac131a56b020bd6bf1436e8fd0fa303f7927d1/modules/library/opengis/src/main/java/org/opengis/feature/Attribute.java#L51) is the parent class of [``ComplexAttribute``](https://github.com/geotools/geotools/blob/41ac131a56b020bd6bf1436e8fd0fa303f7927d1/modules/library/opengis/src/main/java/org/opengis/feature/ComplexAttribute.java#L90), hence its more generic.

What we are expressing here is ignore this specific case, but for all the others including this one, perform a  ``((Attribute) value).getValue()``. I think we should test the value class rather than the feature class.

modules/extension/xsd/xsd-wfs/src/main/java/org/geotools/wfs/PropertyValueCollection.java
```diff
@@ -180,7 +177,8 @@ public Object next() {
...
+                if (value instanceof Attribute) value = ((Attribute) value).getValue();
```
nmco, line 181: Caution, ``value instanceof Attribute`` will yield true for almost all attributes, since it's one of the root classes of the hierarchy. If I understood correctly, for simple feature the value should already be a concrete value (string, date, number, ...) but you step up in an use case where the value is an attribute so you want to extract the actual value?

taba90:
@nmco thanks for your review. I gave a better visibility for Attribute creation: the bug is about creating a ComplexAttribute from a Simple Attribute: the getType method will thrown a class cast exception. That is why I'm getting the value to create a simple one in case it is not a ComplexAttribute.

nmco:
The failure looks unrelated:
```
ERROR] Tests run: 3, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 12.46 s <<< FAILURE! - in org.geotools.data.oracle.OraclePrimaryKeyFinderOnlineTest
[ERROR] testSequencedPrimaryKey(org.geotools.data.oracle.OraclePrimaryKeyFinderOnlineTest)  Time elapsed: 4.8 s  <<< FAILURE!
junit.framework.AssertionFailedError: actual fid: null does not match expected SEQTABLE.4(\..*)?
	at org.geotools.data.oracle.OraclePrimaryKeyFinderOnlineTest.addFeature(OraclePrimaryKeyFinderOnlineTest.java:47)
```

I'm going to merge, thx @taba90 for the feedback and update, this was manually tested and seem to work as expected:
```
<wfs:GetPropertyValue valueReference='st:name' service="WFS" version="2.0"
	xmlns:wfs="http://www.opengis.net/wfs/2.0"
	xmlns:fes="http://www.opengis.net/fes/2.0"
	xmlns:gml="http://www.opengis.net/gml/3.2"
	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.opengis.net/wfs/2.0 http://schemas.opengis.net/wfs/2.0/wfs.xsd http://www.opengis.net/gml/3.2 http://schemas.opengis.net/gml/3.2.1/gml.xsd">
	<wfs:Query typeNames="st:Station" srsName="EPSG:4326"></wfs:Query>
</wfs:GetPropertyValue>
```

Since this a bug fix without backwards compatibility issues, I would sugest to backprot right away to 2.17.x and 2.16.x.

PR #3029, opened 2020-07-01
[GEOT-6644] Imagemosaic-JDBC: shift and glitches.

esuarezsantana:
Fixes [GEOT-6644](https://osgeo-org.atlassian.net/browse/GEOT-6644). This bug was originally [posted in the mailing list](https://sourceforge.net/p/geotools/mailman/geotools-devel/thread/20200617133502.GF1738217@itccanarias.org/),
where a partial solution was proposed. This PR uses a deeper approach that seems to solve all problems exposed.

What changes
------------

1. The most important fact about this commit is the change it introduces for the computation of the rendered image. See below for details.

2. Two new tests have been designed: one to test for the right size of the rendered image, and another to test for bad rendered pixels. Both tests fails with current implementation.

3. Class `AbstractThread` has been removed. It was the parent class for `ImageComposerThread` and `ImageDecoderThread`, but those classes has no much in common.

4. More state is stored in `ImageMosaicJDBCReaderState`, reducing the number of parameters in function calls.

5. Many variables have been renamed to ease the reading and understanding of the code. Some unneeded variables have been removed.

6. Some `checkstyle` issues have also been attended.

How rendering was working before
--------------------------------

This plugin is designed to render in two stages. The "decoder stage" consists of the following steps:

1. it queries the DB to get the tiles that intersect the requested region to render (the requested envelope),

2. for every tile, it intersects the requested envelope and the tile envelope,

3. for every intersected envelope, it gets the (approximate) subimage from the tile that correspond to this envelope,

4. then all this info is stored on a queue.

Next, in the "composer stage":

5. for every region in the queue, it retrieves the intersected envelope and its raster data (the subimage from the tile),

6. then, the subimage is copied to a region of a "virtual image" of the same scale than the tile, but aligned with the requested envelope. This process uses some rounding to create the size of the virtual image, and some rounding to guess where to copy the subimage, because the requested envelope "virtual image" grid and the subimage grid do not necessarily match.

7. Finally, the "virtual image" is interpolated to the requested image scale via the interpolation method configured for this dataset (in the xml file). The scaling used is the one precomputed before rounding the virtual image size.

How rendering is working now
----------------------------

Before you get into point (1) above, a new envelope is computed from the requested envelope. This envelope:

- fits the virtual cell grid that the current coverage overview would propagate from itself to infinity if more cells were appended to it,

- and contains the requested envelope, with and extra padding of 1 virtual coverage cell (this is for nicer interpolation). So that, we'll call it expanded envelope.

Then,

1. it queries the DB to get the tiles that intersect the expanded envelope,

2. for every tile, it intersects the expanded envelope and the tile envelope,

3. for every intersected envelope, it gets the (now exact by design) subimage from the tile that correspond to this envelope,

4. then all this info is stored on a queue.

Next, in the "composer stage":

5. for every region in the queue, it retrieves the intersected envelope and its raster data (the subimage from the tile),

6. then, the subimage is copied to a region of a "virtual image" of the same scale than the tile, but aligned with the expanded envelope. This process is precise by definition and needs rounding just to take care of small numeric errors.

7. Finally, the "virtual image" is interpolated to the requested image scale via the interpolation method configured for this dataset (in the xml file). To perform the interpolation, not only the scale but also the correct translation formula is applied.

Caveats
-------

Because of the new rendering method, when rendering a scene where the (upper) coverage overview level has very small resolution compared to the scene (e.g. if you zoom out a lot), a really large image needs to be allocated in memory, so you may run out of memory.

In this case the plugin will return no rendered image and will log a warning message like this:

> Out of memory when trying to render coverage 'mycoverage'.
> Tips: increase memory, add coarser dataset overviews, or do not zoom that out.

## Checklist

For all pull requests:

- [X] The changes are not breaking the build in downstream projects using SNAPSHOT dependencies, GeoWebCache and GeoServer.

- [X] Commit message must be in the form "[GEOT-XYZW] Title of the Jira ticket"
- [X] This PR passes the [QA checks](https://docs.geotools.org/latest/developer/conventions/code/qa.html) (QA checks results will be reported by travis-ci after opening this PR)
- [X] Documentation has been updated accordingly.


esuarezsantana:
I have no idea about the tests that are failing, and whether they are related to this plugin.

aaime:
No, the failure are unrelated and have been/are being taken care of.

I did not review the changes, as per discussion on the devel list, this module has been un-maintained for too long, so after merging this PR it's gonna get moved down to unsupported, where it will stay as part of the build as long as it does not break it (the day it does, it will be kicked out, unless a module maintainer pops up).

PR #3050, opened 2020-07-06
[GEOT-6653] Add InFunction optimization to MemoryFilterOptimizer

fernandor777:
This PR introduces a performance improvement on MemoryFilterOptimizer for in-memory features filter evaluation.

JIRA issue:
https://osgeo-org.atlassian.net/browse/GEOT-6653

## Checklist

For all pull requests:

### DIFF
--- a/modules/library/render/src/main/java/org/geotools/renderer/lite/MemoryFilterOptimizer.java
+++ b/modules/library/render/src/main/java/org/geotools/renderer/lite/MemoryFilterOptimizer.java
@@ -16,11 +16,15 @@
  */
 package org.geotools.renderer.lite;
 
+import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 import java.util.function.BiFunction;
+import org.apache.commons.lang3.tuple.Pair;
+import org.geotools.filter.function.InFunction;
 import org.geotools.filter.visitor.DuplicatingFilterVisitor;
 import org.geotools.util.Converters;
 import org.opengis.annotation.Extension;
@@ -45,6 +49,7 @@ import org.opengis.filter.PropertyIsNotEqualTo;
 import org.opengis.filter.PropertyIsNull;
 import org.opengis.filter.expression.Expression;
 import org.opengis.filter.expression.ExpressionVisitor;
+import org.opengis.filter.expression.Literal;
 import org.opengis.filter.expression.PropertyName;
 import org.opengis.filter.spatial.BBOX;
 import org.opengis.filter.spatial.Beyond;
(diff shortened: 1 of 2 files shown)


modules/library/render/src/main/java/org/geotools/renderer/lite/MemoryFilterOptimizer.java
```diff
@@ -111,7 +117,58 @@ public Object visit(Not filter, Object extraData) {
...
+    /** Checks if an Or filter can be replaced by a single IN condition. */
+    private Object inFilterOptimizer(Or filter, Object extraData) {
+        InFunction inFilter = replaceWithInFilter(filter);
+        if (inFilter == null) return super.visit(filter, extraData);
+        return ff.equals(inFilter, ff.literal(true));
+    }
+
+    private InFunction replaceWithInFilter(Or filter) {
+        List<Filter> children = filter.getChildren();
+        if (children.size() == 0
```
aaime, line 132: This code does repeated passes over data structures as long as the or children, while a traditional loop could have done everything in a single pass, just keeping track of the first property name found, and the list of literals found in the process. 

Just saying, streams are not always an improvement, had this been in a performance critical section of code, the penalty would have been severe. But it's in a one-off, so it does not matter much, you can leave it as is.

modules/library/render/src/test/java/org/geotools/renderer/lite/MemoryFilterOptimizerTest.java
```diff
@@ -144,4 +148,35 @@ public void testEqualFeatureTypes() throws Exception {
         Mockito.verify(spy, Mockito.times(0)).getAttribute(name);
         Mockito.verify(spy, Mockito.times(1)).getAttribute(0);
     }
+
+    public void testInFunctionOptimizer() throws Exception {
```
aaime, line 152: This test covers only the happy case, could you write tests for the exception paths, like or with different properties, and with comparisons other than equality?

modules/library/render/src/main/java/org/geotools/renderer/lite/MemoryFilterOptimizer.java
```diff
@@ -111,7 +117,58 @@ public Object visit(Not filter, Object extraData) {
...
+        List<Expression> inParameters = new ArrayList<>(parameters.size() + 1);
+        inParameters.add(parameters.get(0).getKey());
+        inParameters.addAll(
+                parameters.stream().map(p -> p.getValue()).collect(Collectors.toList()));
+        InFunction inFunction = new InFunction();
+        inFunction.setParameters(inParameters);
+        return inFunction;
+    }
+
+    private Pair<PropertyName, Literal> getEqualsParameters(PropertyIsEqualTo equals) {
```
aaime, line 157: Nitpicking here, an "in" optimization can be performed over pairs of "Expression, Literal", e.g., ``strLow(myAtt) = 'abc' or strLow(myAtt) = 'def'`` or ``(a + b) = 10 or (a + 10) = 12`` can be both optimized with the in function. The only requirement is that all non literal bits in the comparison are equal. Again, probably not the most common case, up to you if you want to generalize or not.

modules/library/render/src/main/java/org/geotools/renderer/lite/MemoryFilterOptimizer.java
fernandor777: Indeed a single iteration over the collection is possible and faster, so I did the fix, thanks!

modules/library/render/src/test/java/org/geotools/renderer/lite/MemoryFilterOptimizerTest.java
fernandor777: Agreed, added test cases when the In function optimization is not accomplished.

modules/library/render/src/main/java/org/geotools/renderer/lite/MemoryFilterOptimizer.java
fernandor777: Nice hint, added expression support.

aaime:
Looking good, merging, thanks!

PR #3063, opened 2020-07-09
[GEOT-6657] Fix data URL protocol validation

fernandor777:
This PR adds 'data:' schema validation for data protocol handling code on GeoTools URL converter.

JIRA issue:
https://osgeo-org.atlassian.net/browse/GEOT-6657

## Checklist

For all pull requests:

### DIFF
--- a/modules/library/main/src/main/java/org/geotools/data/util/CommonsConverterFactory.java
+++ b/modules/library/main/src/main/java/org/geotools/data/util/CommonsConverterFactory.java
@@ -257,12 +257,22 @@ public class CommonsConverterFactory implements ConverterFactory {
                 URL parsed = new URL(string);
                 return target.cast(parsed);
             } catch (Exception e) {
+                // try to fetch a data protocol URL, or return null
+                return tryGetDataURL(string, target);
+            }
+        }
+
+        private <T> T tryGetDataURL(String string, Class<T> target) {
+            // check if it's a data protocol URL
+            if (string.startsWith(DataUrlConnection.DATA_PREFIX)) {
                 try {
                     URL parsed = new URL(null, string, new DataUrlHandler());
                     return target.cast(parsed);
                 } catch (Exception e2) {
                     return null;
                 }
+            } else {
+                return null;
             }
         }
     }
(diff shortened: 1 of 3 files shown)


modules/library/main/src/main/java/org/geotools/data/util/CommonsConverterFactory.java
```diff
@@ -257,12 +257,22 @@
...
+                // try to fetch a data protocol URL, or return null
+                return tryGetDataURL(string, target);
+            }
+        }
+
+        private <T> T tryGetDataURL(String string, Class<T> target) {
+            // check if it's a data protocol URL
+            if (string.startsWith(DataUrlConnection.DATA_PREFIX)) {
                 try {
                     URL parsed = new URL(null, string, new DataUrlHandler());
```
nmco, line 269: I think this needs a bit more of context, after chatting with @Fernando over this issue I understood that the original problem that we are fixing here is a performance issue with WMS, in some environments we had maps where the rendering times performance went from ~600 ms to ~5 second, sometimes restarting the server fixed the issue.

Long story short, when the support for the ``:data`` protocol [this line was added](https://github.com/geotools/geotools/blob/master/modules/library/main/src/main/java/org/geotools/data/util/CommonsConverterFactory.java#L261), it basically uses a constructor that skips the URL validation, this allows URL using ``data:`` protocol to pass thought, but unfortunately it will also allow any invalid URL to pass.

The RenderableSVGCache is using the converter, where the line above was added, to validate that it got a valid URL and should try to fetch an SVG, since the converter will not validate anything, all URL's are consider valid, and the SVG factory will try to fetch all of them, causing the performance issue. Depending on which factory is loaded first on a certain environment, this issue may or not happen, for example if we are using something like this ``<WellKnownName>wkt://COMPOUNDCURVE((0 0, 0.25 0), CIRCULARSTRING(0.25 0, 0.5 0.5, 0.75 0), (0.75 0, 1 0))</WellKnownName>`` and the factory capable of handling WKT marks is loaded first then no performance problem.

One way of fixing this would be to register the ``data:`` protocol using the Java official way, but unfortunately the URL URLStreamHandlerFactory can be registered only once, and seems like we can have another registration for it overlapping its usage. This makes the entire first line converters registering fails and making the URL converter not being found. Reading a bit seems like since Java 9 we have an extension point for the URL handlers, but we are not there yet.

Hence the proposed solution on the PR, which basically explicit only allow the ``:data`` protocol to be validated blindly.

fernandor777: This is right.  This performance bug happens when the SVG extension points are loaded in first place before, for example, the WKT one for Graphic Mark factories, making SVG tries to load data from the wrong URL.  So this doesn't happen always but the root issue could lead to other problems in the codebase.  I agree Java 9 URL handlers new extension point will be a better solution instead handling this manually on converters.

PR #3078, opened 2020-07-20
[GEOT-6665] Add percentages computation for StandardDeviationClassificationFunction

taba90:
This pull request adds percentages computation to the StandardDeviationFunction as for the others ClassificationFunctions.
Jista ticket https://osgeo-org.atlassian.net/browse/GEOT-6665

## Checklist

For all pull requests:

### DIFF
--- a/modules/library/main/src/main/java/org/geotools/filter/function/StandardDeviationFunction.java
+++ b/modules/library/main/src/main/java/org/geotools/filter/function/StandardDeviationFunction.java
@@ -120,24 +120,34 @@ public class StandardDeviationFunction extends ClassificationFunction {
             throws IOException {
         int classSize = classifier.getSize();
         Object firstMax = classifier.getMax(0);
-        Filter less = FF.less(attr, FF.literal(firstMax));
-        int sizeFirstClass = features.subCollection(less).size();
         int totalSize = features.size();
+        Filter greaterThanOrEqualTo = FF.greaterOrEqual(attr, FF.literal(firstMax));
+        FeatureCollection subCollection = features.subCollection(greaterThanOrEqualTo);
+        int sizeFirstClass = totalSize - subCollection.size();
         double[] percentages = new double[classSize];
         // we don't know the min value in the collection because the
         // the first interval is open to infinity to the left.
         // needs a query to get the classMembers
-        percentages[0] = (sizeFirstClass / totalSize) * 100;
+        percentages[0] = ((double) sizeFirstClass / (double) totalSize) * 100;
 
         double min = ((Number) classifier.getMin(1)).doubleValue();
-        return computeGroupByPercentages(features, percentages, totalSize, min, standardDeviation);
+        percentages =
+                computeGroupByPercentages(
+                        subCollection, percentages, totalSize, min, standardDeviation);
+        computeLastPercentage(percentages, totalSize);
+        return percentages;
+    }
+
+    private void computeLastPercentage(double[] percentages, double totalSize) {
+        double sum = Arrays.stream(percentages).sum();
+        percentages[percentages.length - 1] = 100.0 - sum;
     }
 
     @Override
     protected void computePercentage(
             double[] percentages, double classMembers, double totalSize, int index) {
-        // since we are grouping by using the min of the second interval adds +1 to the index
-        super.computePercentage(percentages, classMembers, totalSize, index + 1);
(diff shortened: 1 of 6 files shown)


modules/library/main/src/main/java/org/geotools/filter/function/StandardDeviationFunction.java
```diff
@@ -96,4 +110,65 @@ private Double getMax(int index, int numClasses, double average, double standard
         if (index < 0 || index >= numClasses - 1) return null;
         return Double.valueOf(average - (((numClasses / 2.0) - 1 - index) * standardDeviation));
     }
+
+    public double[] getPercentages(
```
aaime, line 114: Should not be public. Besides that, this large block of code (this method and the ones below) seem like it could be shared with EqualIntervalFunction, or not? Guess you can do the same as with ``AbstractQuantityClassificationFunction`` and roll a common superclass.

taba90: Thanks for your review @aaime. I've put it as private. Anyway this is not the same block of code. The EqualIntervalFunction use a GroupByVisitor to query the data and compute percentages. In this case it is creating filters mapping the breaks to query the data. Also both are different from the computePercentages method in the `AbstractQuantityClassificationFunction`

aaime: A group by visitor seems indeed more efficient, the current one seems to be loading the entire data set from the database to run the counts? If so, that's quite inefficient.

taba90: thanks for your point @aaime . Code has been modified so that GroupBy is used also for StandardDeviation percentages.

taba90:
fails on downstream integration seems unrelated. Got once [ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.22.2:test (default-test) on project gs-importer: There are test failures
and now after having restart the build [ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.22.2:test (default-test) on project gs-web-sec-core: There are test failures

modules/library/main/src/main/java/org/geotools/filter/function/StandardDeviationFunction.java
```diff
@@ -96,4 +111,41 @@ private Double getMax(int index, int numClasses, double average, double standard
...
+            double standardDeviation)
+            throws IOException {
+        int classSize = classifier.getSize();
+        Object firstMax = classifier.getMax(0);
+        Filter less = FF.less(attr, FF.literal(firstMax));
+        int sizeFirstClass = features.subCollection(less).size();
+        int totalSize = features.size();
+        double[] percentages = new double[classSize];
+        // we don't know the min value in the collection because the
+        // the first interval is open to infinity to the left.
```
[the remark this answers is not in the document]
taba90: Thanks @aaime I've added a check on percentages one by one. The openess to infinity of the last class is handled here https://github.com/geotools/geotools/pull/3078/commits/9678b0934a080c3a4e343c6a456ff193a4efef87#diff-2319f1367ac925907808cd3d8f0fa864R283 where the index returned from the query is checked on the percentages length and in case  the resulting percentage is merged on the one of the last class. I've added also a test to check this behaviour.

taba90:
failures are on Oracle build.

taba90:
failures on gt-render and on gt-graph looks unrelated to this pr.

[ERROR] testMark(org.geotools.renderer.lite.LabelObstacleTest)  Time elapsed: 0.166 s  <<< FAILURE!
java.lang.AssertionError: expected:<1.0> but was:<0.0>
	at org.geotools.renderer.lite.LabelObstacleTest.checkNoIntersection(LabelObstacleTest.java:239)
	at org.geotools.renderer.lite.LabelObstacleTest.testMark(LabelObstacleTest.java:208)


[ERROR] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:3.1.1:check (default) on project gt-charts: Failed during checkstyle execution: Failed during checkstyle configuration: unable to parse configuration stream: Connection timed out (Connection timed out) -> [Help 1]

aaime:
Indeed they do. Squash-merging.

PR #3089, opened 2020-07-29
[GEOT-6672] gt-jdbc-hana - Spatial filters against views don't always…

stefanuhrig:
… work

If the HANA dialect is asked for the SRID of a geometry column in a view
(via `HanaDialect#getGeometrySRID()`), it returns null.

This leads to a faulty query generation in
`HanaDialect#prepareGeometryValue()` because `-1` is used as SRID instead of
the SRID of the geometries in the view's column. As a result the
database returns an error stating that a spatial reference system with
SRID `-1` does not exist.

We extend the `getGeometrySRID()` function to return the SRID of the first
non-null geometry in the view's column to fix the issue.


stefanuhrig:
The failing test in one build on Travis CI is not related to this change.

PR #3098, opened 2020-07-30
[GEOT-6672] gt-jdbc-hana - Spatial filters against views don't always…

stefanuhrig:
… work

If the HANA dialect is asked for the SRID of a geometry column in a view
(via HanaDialect#getGeometrySRID()), it returns null.

This leads to a faulty query generation in
HanaDialect#prepareGeometryValue() because -1 is used as SRID instead of
the SRID of the geometries in the view's column. As a result the
database returns an error stating that a spatial reference system with
SRID -1 does not exist.

We extend the getGeometrySRID() function to return the SRID of the first
non-null geometry in the view's column to fix the issue.

### DIFF
--- a/modules/plugin/jdbc/jdbc-hana/src/main/java/org/geotools/data/hana/HanaDialect.java
+++ b/modules/plugin/jdbc/jdbc-hana/src/main/java/org/geotools/data/hana/HanaDialect.java
@@ -180,8 +180,7 @@ public class HanaDialect extends PreparedStatementSQLDialect {
         return GEOMETRY_TYPE_NAME;
     }
 
-    @Override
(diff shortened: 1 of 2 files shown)


stefanuhrig:
The Travis CI build failed with error

```
Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:3.1.1:check (default) on project gt-jts-wrapper:
Failed during checkstyle execution:
Failed during checkstyle configuration: unable to parse configuration stream: Connection timed out (Connection timed out)
```

Besides that, everything builds fine. The timeout seems not related to this change. Hence, I'm going to merge it.

PR #3113, opened 2020-08-07
What can we do about situations like this?

jnh5y:
Signed-off-by: Jim Hughes <jnh5y@ccri.com>

## Checklist

For all pull requests:

### DIFF
--- a/modules/library/main/src/main/java/org/geotools/filter/FidFilterImpl.java
+++ b/modules/library/main/src/main/java/org/geotools/filter/FidFilterImpl.java
@@ -56,7 +56,7 @@ public class FidFilterImpl extends AbstractFilter implements Id {
     private Set<String> ids = new LinkedHashSet<>();
 
     /** Constructor which takes {@link org.opengis.filter.identity.Identifier}, not String. */
-    protected FidFilterImpl(Set /* <Identifier> */ fids) {
+    protected FidFilterImpl(Set<Identifier> fids) {
         // check these are really identifiers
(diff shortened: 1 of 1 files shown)


jnh5y:
@aaime thoughts about what to do on the unchecked branch for situations like this where a type has been commented out? 

In the case of the protected method, it seems fairly ok-ish to go ahead with the change.  I'm wondering if we'll break other projects in the other case.

aaime:
Often the case is the opposite: the code is old enough that the dev wanted to use generics, but we were still using a version of Java that did not support them, so they did add them as comments. I would uncomment them, and if it's publicly visible, add a note in the migration guide.

jnh5y:
@aaime awesome.  If the pattern of removing comments like that is one that is worth following, I may just roll through the warnings looking for those.  (I'm thinking of doing that since I'd have some confidence of being able to make the change and know that it is ok, etc.)

Separately, with that kind of thing in mind, I wonder if it'd be worth implementing a JAI wrapper so that the Suppressions for JAI would be located in one place?  (Also, if the RPE/JAI replacement ever gets sorted, one would just replace/update that little bit of wrapping code?)

aaime:
I'm pretty skeptical about wrapping JAI, but maybe do an example that shows better what you'd do?

jnh5y:
Sounds good.  I'll save that idea for later, and if I can get to it, I'll produce an example.  Seems like knocking out the commented types is an easier and profitable change.

aaime:
Closing, this one seemed mostly like a conversation, rather than something actually meant to be merged.

PR #3118, opened 2020-08-20
[GEOT-6682] WFS 2.0 GetFeature reponse parsing fails on "boundedBy"

awaterme:
Pull request for [GEOT-6682](https://osgeo-org.atlassian.net/browse/GEOT-6682). 

## Checklist

For all pull requests:

- [x] This PR passes the [QA checks](https://docs.geotools.org/latest/developer/conventions/code/qa.html) (QA checks results will be reported by travis-ci after opening this PR) _I am not sure, but hope so_
- [ ] Documentation has been updated accordingly. _Not required in this case_

### DIFF
--- a/modules/extension/xsd/xsd-wfs/src/main/java/org/geotools/wfs/v2_0/bindings/EnvelopePropertyTypeBinding.java
+++ b/modules/extension/xsd/xsd-wfs/src/main/java/org/geotools/wfs/v2_0/bindings/EnvelopePropertyTypeBinding.java
@@ -19,20 +19,41 @@ package org.geotools.wfs.v2_0.bindings;
 import java.util.ArrayList;
 import java.util.List;
 import javax.xml.namespace.QName;
+import net.opengis.wfs20.EnvelopePropertyType;
+import net.opengis.wfs20.Wfs20Factory;
+import net.opengis.wfs20.Wfs20Package;
+import org.eclipse.emf.ecore.EAttribute;
+import org.eclipse.emf.ecore.EClass;
+import org.eclipse.emf.ecore.EcoreFactory;
+import org.eclipse.emf.ecore.util.FeatureMap;
+import org.eclipse.emf.ecore.util.FeatureMap.Entry;
+import org.eclipse.emf.ecore.util.FeatureMapUtil;
 import org.eclipse.xsd.XSDElementDeclaration;
 import org.geotools.geometry.jts.ReferencedEnvelope;
 import org.geotools.gml3.v3_2.GML;
 import org.geotools.wfs.v2_0.WFS;
 import org.geotools.xsd.AbstractComplexBinding;
+import org.geotools.xsd.ElementInstance;
+import org.geotools.xsd.Node;
 
 public class EnvelopePropertyTypeBinding extends AbstractComplexBinding {
+
+    private Class<ReferencedEnvelope> referencedEnvelopeClass = ReferencedEnvelope.class;
+    private EClass newClass = EcoreFactory.eINSTANCE.createEClass();
+    private EAttribute newAttribute = EcoreFactory.eINSTANCE.createEAttribute();
+    private EAttribute anyAttribute = Wfs20Package.eINSTANCE.getEnvelopePropertyType_Any();
+
+    public EnvelopePropertyTypeBinding() {
+        newClass.setInstanceClass(referencedEnvelopeClass);
+        newAttribute.setEType(newClass);
+    }
 
     public QName getTarget() {
         return WFS.EnvelopePropertyType;
     }
 
-    public Class getType() {
-        return ReferencedEnvelope.class;
+    public Class<?> getType() {
+        return referencedEnvelopeClass;
     }
 
     @Override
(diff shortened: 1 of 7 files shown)


modules/ogc/net.opengis.wfs/src/net/opengis/wfs20/impl/SimpleFeatureCollectionTypeImpl.java
```diff
@@ -178,7 +178,7 @@ public Object eGet(int featureID, boolean resolve, boolean coreType) {
     public void eSet(int featureID, Object newValue) {
     switch (featureID) {
       case Wfs20Package.SIMPLE_FEATURE_COLLECTION_TYPE__BOUNDED_BY:
-        setBoundedBy((EnvelopePropertyType)newValue);
+        setBoundedBy(newValue instanceof EnvelopePropertyType ? (EnvelopePropertyType) newValue : null);
```
jodygarnett, line 181: Q: What is it getting back that is not an EnvelopePropertyType?

jodygarnett: I see, it is explicitly allowed to be null:

```
<complexType name="BoundingShapeType">
  <sequence>
    <choice>
      <element ref="gml:Envelope"/>
      <element ref="gml:Null"/>
    </choice>
  </sequence>
  <attribute name="nilReason" type="gml:NilReasonType"/>
</complexType>
```

awaterme: > 
> 
> Q: What is it getting back that is not an EnvelopePropertyType?

Hi Jody, it get an empty String, because the parsing failed.

awaterme: > 
> 
> I see, it is explicitly allowed to be null:
> 
> ```
> <complexType name="BoundingShapeType">
>   <sequence>
>     <choice>
>       <element ref="gml:Envelope"/>
>       <element ref="gml:Null"/>
>     </choice>
>   </sequence>
>   <attribute name="nilReason" type="gml:NilReasonType"/>
> </complexType>
> ```

When I look into the actual schema (http://schemas.opengis.net/wfs/2.0/wfs.xsd) I see the following:

grafik

Are GeoTools using this schema?

I think the problem is the "any" type...

aaime: GeoTools is using that schema. Being a schema driven parser, it might well be that its having problems when dealing with a "any" target, it probably does not know what bindings to use. 
@awaterme if you're interested, try adding a breakpoint in XSAnyTypeBinding and see if that's the one it is using to parse stuff.
@nmco and @fernandor777 deal with this code more than I do, maybe they have a more educated guess as to what is happening.

awaterme:
I have adjusted the solution. I hope it is acceptable now: 

* I have rolled back the modifications of the generated code
* I have adjusted the bindings instead. The ClassCastException is resolved. The prior GT implementation always had a "null" bounding box. Now the bounding box is nested into the FeatureMap. The double nesting is the only solution I found which is accepted by the generated EMF model. 

I would be glad if the fix could make it into the GeoTools 24 release, although it is a little late already.

 BTW: The `XSAnyTypeBinding` is not executed here, because the nested element is an GML Envelope. The binding for the envelope is used instead.

ianturton (approved):

modules/extension/xsd/xsd-wfs/src/main/java/org/geotools/wfs/v2_0/bindings/EnvelopePropertyTypeBinding.java
```diff
@@ -19,20 +19,41 @@
...
 import org.geotools.geometry.jts.ReferencedEnvelope;
 import org.geotools.gml3.v3_2.GML;
 import org.geotools.wfs.v2_0.WFS;
 import org.geotools.xsd.AbstractComplexBinding;
+import org.geotools.xsd.ElementInstance;
+import org.geotools.xsd.Node;
 
 public class EnvelopePropertyTypeBinding extends AbstractComplexBinding {
 
+    private Class<ReferencedEnvelope> referencedEnvelopeClass = ReferencedEnvelope.class;
```
aaime, line 41: All of these... they look like they should be constants, no?

awaterme: Hi Andrea,

could be constants, but in that case on class loading of EnvelopePropertyTypeBinding the EMF Factories would also be bootstrapped. In my experience that sometimes causes trouble, maybe in testing scenarios, etc.
On the other hand I was not sure about the lifespan of the EnvelopePropertyTypeBinding. If it is created often, statics might make sense...
As you like. Let me know what you prefer please.

Best regards,
Andreas

aaime: If memory serves me right, bindings are instantiated anew every time we parse or encode something. Each parse/encode gets its own dedicated Pico Container context, bindings are registered in there, every time. I'm not sure if they are just registered as classes, and then instantiated every time they are needed, or not. If it was the former, that would mean they get instantiated multiple times per parse/encode operation, too.

awaterme: Do you prefer static fields then?

aaime: Personally I do, yes.

awaterme:
Hi Andrea,

next attempt with static fields...

Best regards,
Andreas

aaime:
Think it looks good. Constants are not named after Java coding conventions, should look like "THIS_IS_A_CONSTANT", but they are private, so whatever... guess I'll have to add a Checkstyle check for it, later down the road, so that it cannot happen any longer.
In the meantime, squash merging.

PR #3121, opened 2020-08-26
[GEOT-6270] SimplifyingFilterVisitor does not account for null values

huseyin-bozkurt:
when simplifying negations

"not(a between 10 and 20)" results in "a < 10 or a > 20", but it should be "a < 10 or a > 20 or a is null" in SimplifyingFilterVisitor

## Checklist

For all pull requests:

### DIFF
--- a/modules/library/main/src/main/java/org/geotools/filter/visitor/SimplifyingFilterVisitor.java
+++ b/modules/library/main/src/main/java/org/geotools/filter/visitor/SimplifyingFilterVisitor.java
@@ -20,11 +20,8 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
-import java.util.Objects;
 import java.util.Set;
 import java.util.regex.Pattern;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
 import org.geotools.filter.FilterAttributeExtractor;
 import org.opengis.feature.simple.SimpleFeatureType;
 import org.opengis.feature.type.FeatureType;
@@ -419,19 +416,19 @@ public class SimplifyingFilterVisitor extends DuplicatingFilterVisitor {
             } else if (simplified == Filter.EXCLUDE) {
                 return Filter.INCLUDE;
             } else if (simplified instanceof PropertyIsBetween) {
+                List<Filter> orFilters = new ArrayList<>();
                 PropertyIsBetween pb = (PropertyIsBetween) simplified;
                 Filter lt = ff.less(pb.getExpression(), pb.getLowerBoundary());
                 Filter gt = ff.greater(pb.getExpression(), pb.getUpperBoundary());
+                orFilters.add(lt);
+                orFilters.add(gt);
                 PropertyName pn = (PropertyName) pb.getExpression();
                 String pName = pn.getPropertyName();
-                Filter nullFilter = null;
                 if (isNillable(pName)) {
-                    nullFilter = ff.isNull(pb.getExpression());
+                    // the property may have a NULL value, we need to vouch for it
+                    orFilters.add(ff.isNull(pb.getExpression()));
                 }
-                return ff.or(
-                        Stream.of(lt, gt, nullFilter)
-                                .filter(Objects::nonNull)
-                                .collect(Collectors.toList()));
+                return ff.or(orFilters);
             } else if (simplified instanceof PropertyIsEqualTo) {
                 PropertyIsEqualTo pe = (PropertyIsEqualTo) simplified;
                 return ff.notEqual(pe.getExpression1(), pe.getExpression2(), pe.isMatchingCase());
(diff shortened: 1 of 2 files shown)


huseyin-bozkurt (commented):
I added only NULL filter in case of not(between ( Literal, Literal)) and updated the unit test accordingly.

modules/library/main/src/main/java/org/geotools/filter/visitor/SimplifyingFilterVisitor.java
```diff
@@ -418,7 +421,8 @@ public Object visit(Not filter, Object extraData) {
                 PropertyIsBetween pb = (PropertyIsBetween) simplified;
                 Filter lt = ff.less(pb.getExpression(), pb.getLowerBoundary());
                 Filter gt = ff.greater(pb.getExpression(), pb.getUpperBoundary());
-                return ff.or(lt, gt);
+                Filter nullFilter = ff.isNull(pb.getExpression());
```
aaime, line 424: If the ``featureType`` field has been set, check if the property is nullable before adding this extra. Add a test accordingly.

huseyin-bozkurt: I have just made related changes and added the test as well.

nmco:
Looks like a valid format valdiation failure:
```
[ERROR] Failed to execute goal com.coveo:fmt-maven-plugin:2.4.0:check (default) on project gt-main: Found 2 non-complying files, failing build -> [Help 1]
```
Doing a full Maven build before commuting should fix all the formatting issues.

modules/library/main/src/main/java/org/geotools/filter/visitor/SimplifyingFilterVisitor.java
```diff
@@ -418,7 +423,13 @@ public Object visit(Not filter, Object extraData) {
...
                 Filter lt = ff.less(pb.getExpression(), pb.getLowerBoundary());
                 Filter gt = ff.greater(pb.getExpression(), pb.getUpperBoundary());
-                return ff.or(lt, gt);
+                PropertyName pn = (PropertyName) pb.getExpression();
+                String pName = pn.getPropertyName();
+                Filter nullFilter = null;
+                if(isNillable(pName)) {
+                	nullFilter = ff.isNull(pb.getExpression());
+                }
+                return ff.or(Stream.of(lt,gt,nullFilter).filter(Objects::nonNull).collect(Collectors.toList()));
```
nmco, line 432: If I'm not mistaken the propose of the expression ``Stream.of(lt,gt,nullFilter).filter(Objects::nonNull).collect(Collectors.toList()) `` is to get rid of ``nullFilter`` if it's ``NULL``, I think ti would be more efficient and clear to do something like this instead:
```java
if (isNillable(pName)) {
    // the property may have a NULL value, we need to vouch for it
    nullFilter = ff.isNull(pb.getExpression());
    return ff.or(lt, gt, nullFilter);
}
// the property value is not nullable
return ff.or(lt, gt);
```

huseyin-bozkurt: Not possible to use ff.or(lt, gt, nullFilter); because it accepts 2 parameters or an ArrayList (for more than 2 filters).

huseyin-bozkurt:
> Looks like a valid format validation failure:
> 
> ```
> [ERROR] Failed to execute goal com.coveo:fmt-maven-plugin:2.4.0:check (default) on project gt-main: Found 2 non-complying files, failing build -> [Help 1]
> ```
> 
> Doing a full Maven build before commuting should fix all the formatting issues.

Re-format is done and commited.

modules/library/main/src/main/java/org/geotools/filter/visitor/SimplifyingFilterVisitor.java
nmco: You're correct, my mistake. In that case I would suggest to separate the concerns; first you build the list of filters then you build the filter. Something like this:
```java
List<Filter> orFilters = new ArrayList<>();
orFilters.add(lt);
orFilters.add(gt);
if (isNillable(pName)) {
    // the property may have a NULL value, we need to vouch for it
    orFilters.add(pb.getExpression());
}
return ff.or(orFilters);
```

huseyin-bozkurt: Dear @nmco, I updated the file according to your suggestion.

nmco:
Thank you for the update @huseyin-bozkurt , I have required the Travis build, it was not clear for me if the build failure was related or not. Apart from that the PR looks good to me, what do you think @aaime ?

aaime:
Looks good. Work is done on behalf of GeoSolutions, our CLA covers it. Merging.

PR #3122, opened 2020-08-28
[GEOT-6687] AzimulthalEquidistantProjectionHandlerFactory is not handling false origins

aaime:
See: https://osgeo-org.atlassian.net/browse/GEOT-6687

## Checklist

For all pull requests:

### DIFF
--- a/modules/library/main/src/main/java/org/geotools/renderer/crs/AzimulthalEquidistantProjectionHandlerFactory.java
+++ b/modules/library/main/src/main/java/org/geotools/renderer/crs/AzimulthalEquidistantProjectionHandlerFactory.java
@@ -161,12 +161,37 @@ public class AzimulthalEquidistantProjectionHandlerFactory implements Projection
                 Geometry difference = renderingGeometry.difference(bufferedDateline);
                 List<Polygon> polygons = PolygonExtracter.getPolygons(difference);
                 for (Polygon p : polygons) {
-                    Geometry transformed =
-                            JTS.transform(
-                                    p,
-                                    CRS.findMathTransform(
-                                            renderingEnvelope.getCoordinateReferenceSystem(),
-                                            sourceCRS));
+                    Geometry transformed = null;
+                    try {
+                        transformed =
+                                JTS.transform(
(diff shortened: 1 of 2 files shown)


modules/library/main/src/main/java/org/geotools/renderer/crs/AzimulthalEquidistantProjectionHandlerFactory.java
```diff
@@ -161,12 +161,37 @@ private void initializeDatelineCutter(CoordinateReferenceSystem crs, Point2D.Dou
...
+                                        DefaultGeographicCRS.WGS84,
+                                        false);
+                        if (handler == null) throw e;
+                        Geometry preProcessed = handler.preProcess(polygonWGS84);
+                        transformed =
+                                JTS.transform(
+                                        preProcessed,
+                                        CRS.findMathTransform(
+                                                renderingEnvelope.getCoordinateReferenceSystem(),
+                                                sourceCRS));
```
jodygarnett, line 193: I thought through putting a try catch around these, in order to throw the original TransformException exception. It is more useful to have this failure so you can tell exactly how far it got.

jodygarnett (approved):

jodygarnett:
I saw a `maven-metadata.xml` error here- oh no!

modules/library/main/src/main/java/org/geotools/renderer/crs/AzimulthalEquidistantProjectionHandlerFactory.java
aaime: The method declares to throw TransformException, so in case JTS.transform throws, it's going to go out of the method and handled at higher level, logged or re-wrapped. Don't think anything is lost. Have I misunderstood your comment?

jodygarnett: Nope it is fine, was trying to decide if there was any value in throwing the origional exception (there was not)

PR #3134, opened 2020-09-04
DRAFT: counting with longs

aaime:
Draft PR to verify how much things change by counting using longs instead of ints.

## Checklist

For all pull requests:

### DIFF
--- a/modules/library/main/src/main/java/org/geotools/data/FeatureSource.java
+++ b/modules/library/main/src/main/java/org/geotools/data/FeatureSource.java
@@ -206,8 +206,36 @@ public interface FeatureSource<T extends FeatureType, F extends Feature> {
      *         or {@code -1} if this cannot be calculated.
      *
      * @throws IOException if there are errors getting the count
+     * @deprecated Please use {@link #count(Query)} instead
      */
-    int getCount(Query query) throws IOException;
+    @Deprecated
+    default int getCount(Query query) throws IOException {
+        return Math.toIntExact(count(query));
+    }
+
+    /**
+     * Gets the number of the features that would be returned by the given
+     * {@code Query}, taking into account any settings for max features and
+     * start index set on the {@code Query}.
+     * <p>
+     * It is possible that this method will return {@code -1} if the calculation
+     * of number of features is judged to be too costly by the implementing class.
+     * In this case, you might call <code>getFeatures(query).size()</code>
+     * instead.
+     * <p>
+     * Example use:<pre><code> int count = featureSource.getCount();
+     * if( count == -1 ){
+     *    count = featureSource.getFeatures( "typeName", count ).size();
+     * }
+     *
+     * @param query the query to select features
+     *
+     * @return the numer of features that would be returned by the {@code Query};
+     *         or {@code -1} if this cannot be calculated.
+     *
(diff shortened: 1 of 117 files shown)


modules/library/main/src/main/java/org/geotools/data/FeatureSource.java
```diff
@@ -206,8 +206,36 @@
      *         or {@code -1} if this cannot be calculated.
      *
      * @throws IOException if there are errors getting the count
+     * @deprecated Please use {@link #count(Query)} instead
      */
-    int getCount(Query query) throws IOException;
+    @Deprecated
+    default int getCount(Query query) throws IOException {
```
jodygarnett, line 212: Document use of Integer.MAX_VALUE:
```
If this source contains more than Integer.MAX_VALUE elements, returns Integer.MAX_VALUE.
```

modules/library/main/src/main/java/org/geotools/data/FeatureSource.java
```diff
@@ -206,8 +206,36 @@
...
+     * start index set on the {@code Query}.
+     * <p>
+     * It is possible that this method will return {@code -1} if the calculation
+     * of number of features is judged to be too costly by the implementing class.
+     * In this case, you might call <code>getFeatures(query).size()</code>
+     * instead.
+     * <p>
+     * Example use:<pre><code> int count = featureSource.getCount();
+     * if( count == -1 ){
+     *    count = featureSource.getFeatures( "typeName", count ).size();
```
jodygarnett, line 228: Is a cast or `Math.toIntExact` needed (assuming `count` is a long here).

```
      *  if( count == -1 ){
      *    count = Math.toIntExact( featureSource.getFeatures( "typeName", count ).size() );
```

jodygarnett (approved):

modules/library/main/src/main/java/org/geotools/data/FeatureSource.java
aaime: I think it should throw an exception instead?

modules/library/main/src/main/java/org/geotools/data/FeatureSource.java
aaime: Yes, this is the result of a "inline method" refactor, I did not check call point by call point, the PR is most to try and get some grip on the discussion on gt-devel. The way it's going, it's not very much compatible with my timeframes, the DGGS activity is just a prototype at the moment and is going to end soon, I was hoping to get some useful API improvement out of it, on the side.

modules/library/main/src/main/java/org/geotools/data/FeatureSource.java
mprins: If throwing it should really be [`java.lang.ArithmeticException`](https://docs.oracle.com/javase/8/docs/api/java/lang/ArithmeticException.html) to stay in line with what happens in eg. `java.lang.Math` when the result overflows an int, but method only declares `IOException` which seems counter-intuitive to throw. `ArithmeticException` being unchecked doesn't really need to be declared of course, but the unwitting programmer could be in for a surprise in that case.
That being said, I would prefer if it throws anything to having it return a magic value (even if it is `IOException` wrapping `ArithmeticException`)

PR #3161, opened 2020-09-24
cql-json-module-start

turingtestfail:

## Checklist

For all pull requests:


modules/unsupported/cql-json/src/main/java/org/geotools/filter/text/cqljson/CQLJson.java
```diff
@@ -0,0 +1,76 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ *
+ *    (C) 2016, Open Source Geospatial Foundation (OSGeo)
```
jodygarnett, line 5: 2020 ?

modules/unsupported/cql-json/src/main/java/org/geotools/filter/text/cqljson/CQLJson.java
```diff
@@ -0,0 +1,76 @@
...
+package org.geotools.filter.text.cqljson;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.geotools.filter.text.commons.CompilerUtil;
+import org.geotools.filter.text.cql2.CQLException;
+import org.geotools.filter.text.cqljson.model.Predicates;
+import org.opengis.filter.Filter;
+import org.opengis.filter.FilterFactory;
+
```
jodygarnett, line 27: We try for public javadocs, this is an unsupported module so you can wait until this is ready to be consider for the library.

modules/unsupported/cql-json/src/main/java/org/geotools/filter/text/cqljson/CQLJsonFilterBuilder.java
```diff
@@ -0,0 +1,610 @@
...
+import org.geotools.filter.text.generated.parsers.ParseException;
+import org.opengis.filter.Filter;
+import org.opengis.filter.FilterFactory;
+import org.opengis.filter.expression.Expression;
+import org.opengis.filter.expression.Function;
+import org.opengis.filter.expression.Literal;
+import org.opengis.filter.expression.PropertyName;
+import org.opengis.geometry.Geometry;
+
+final class CQLJsonFilterBuilder {
```
jodygarnett, line 64: This is a bit odd to have as a builder, are you sure you are not looking for a FilterVisitor? Traverse the data structure and covert ...

modules/unsupported/cql-json/src/main/java/org/geotools/filter/text/cqljson/GeomUtil.java
```diff
@@ -0,0 +1,206 @@
...
+import org.opengis.geometry.aggregate.MultiCurve;
+import org.opengis.geometry.aggregate.MultiPoint;
+import org.opengis.geometry.aggregate.MultiSurface;
+import org.opengis.geometry.coordinate.PointArray;
+import org.opengis.geometry.primitive.Curve;
+import org.opengis.geometry.primitive.Point;
+import org.opengis.geometry.primitive.Surface;
+import org.opengis.geometry.primitive.SurfaceBoundary;
+
+public class GeomUtil {
```
jodygarnett, line 38: I expect this class will be used a lot, may wish to go for a more explicit class name :)

modules/unsupported/cql-json/src/main/java/org/geotools/filter/text/cqljson/model/Contains.java
```diff
@@ -0,0 +1,20 @@
...
+ *
+ *    This library is distributed in the hope that it will be useful,
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ *    Lesser General Public License for more details.
+ */
+
+package org.geotools.filter.text.cqljson.model;
+
+public class Contains extends SpatialExpressions {}
```
jodygarnett, line 20: This used to be how enumerations were represented in Java (so they could be strongly typed).

Is making SpatialExpressions an enum worthwhile?

jodygarnett (commented):
This is an unsupported module, so does not require a formal review (until you are ready for it to be added to the library).

modules/unsupported/cql-json/src/test/java/org/geotools/filter/text/cqljson/CQLJsonParsingTest.java
```diff
@@ -0,0 +1,598 @@
...
+                        + "             ]"
+                        + "          }";
+
+        CQLJsonCompiler cqlJsonCompiler =
+                new CQLJsonCompiler(swimmingpool, new FilterFactoryImpl());
+        cqlJsonCompiler.compileFilter();
+        Filter filter = cqlJsonCompiler.getFilter();
+        Assert.assertEquals(
+                "[[ swimming_pool = true ] AND [[ floor > 5 ] OR [ material is like brick% ] OR [ material is like %brick ]]]",
+                filter.toString());
```
jodygarnett, line 53: I strongly recommend using ` ECQL.toCQL(filter)` rather than `filter.toString()` to have a stable string representation to compare against. As a bonus it should also show the difference between `5` and '"5"`.

modules/unsupported/cql-json/src/test/java/org/geotools/filter/text/cqljson/CQLJsonParsingTest.java
```diff
@@ -0,0 +1,598 @@
...
+    }
+
+    @Test
+    public void convertAndIntersects() throws JsonProcessingException, CQLException {
+        String beam =
```
jodygarnett, line 58: These methods can be hard to maintain with having to quote content.

* Please see MapBoxTestUtils test for one idea to make this easier, a utility method that changes `\`` to `\"` and then parses.

```
    public static JSONObject object(String json) throws ParseException {
        JSONParser parser = new JSONParser();
        String text = json.replace('\'', '\"');
        Object object = parser.parse(text);
        return (JSONObject) object;
    }
```

Used like:
```
        String jsonStr = "{'name1': 'fred', 'name2' : 1, 'name3' : null, 'name4': { 'age': 1 }}";
        JSONObject object = MapboxTestUtils.object(jsonStr);
```

* Another good approach is to have a property file of reference ECQL expressions to round trip (convert ECQL -parse-> Filter -encode-> CQL-JSON -parse-> Filter and then double check the Filters are the same. Which is what a lot of the duplication in this class is about.

```
convertAndIntersects=(beamMode = "ScanSAR Narrow" ] AND [ swathDirection = "ascending") ...
```

This feedback assume the module will be long lived and plan for adding more and more tests over time as problems are reported. See example [here](https://github.com/geotools/geotools/blob/master/modules/extension/ysld/src/test/java/org/geotools/ysld/parse/YsldParseCookbookTest.java).

turingtestfail: As suggested I moved the JSON into its own file.

modules/unsupported/cql-json/src/test/java/org/geotools/filter/text/cqljson/CQLJsonParsingTest.java
turingtestfail: Good point.  I changed the test to use ECQL.toCQL rather than toString for stability.

modules/unsupported/cql-json/src/main/java/org/geotools/filter/text/cqljson/model/Contains.java
turingtestfail: I did this as a bit of future proofing and to match the CQLJson standard.  If you think this clutter things I can remove it.

modules/unsupported/cql-json/src/main/java/org/geotools/filter/text/cqljson/GeomUtil.java
turingtestfail: Yeah, that makes a lot of sense.  Changed to "MapToOpenGISGeomUtil"

modules/unsupported/cql-json/src/main/java/org/geotools/filter/text/cqljson/CQLJsonFilterBuilder.java
turingtestfail: This is the pattern that the original CQL to filter developers followed.  My understanding is that the FilterVisitor is for going from Filter to something else (like SQL) vs. what I am doing here, which is going from JSON to Filter.

modules/unsupported/cql-json/src/main/java/org/geotools/filter/text/cqljson/CQLJson.java
turingtestfail: Fixed.

jodygarnett:
Travis indicates a Java 11 QA test wants to clean up your code:
```
[ERROR] Failed to execute goal com.coveo:fmt-maven-plugin:2.4.0:check (default) on project gt-cql-json: Found 2 non-complying files, failing build -> [Help 1]
[ERROR] 
```

aaime:
More info about QA checks here: https://docs.geotools.org/latest/developer/conventions/code/qa.html
In order to fix formatting just run a maven build locally, it does the right formatting for you.

turingtestfail:
Done.  All tests passing now.

turingtestfail:
I have not been following the debate about hierarchical vs array but I can
reach out to the Gitter channel to check in on the latest.

jodygarnett:
@turingtestfail as this is an unsupported module can I confirm you have commit access to merge this PR?

turingtestfail:
Apologies, I don't really understand the question.  I have commit access to
the source repo of this PR (
https://github.com/turingtestfail/geotools/tree/cql-json-module-start ) but
I do not have any GeoTools commit access.

modules/unsupported/cql-json/pom.xml
```diff
@@ -0,0 +1,37 @@
...
+        <dependency>
+            <groupId>org.geotools</groupId>
+            <artifactId>gt-cql</artifactId>
+            <version>24-SNAPSHOT</version>
+            <scope>compile</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.geotools</groupId>
+            <artifactId>gt-jts-wrapper</artifactId>
+            <version>24-SNAPSHOT</version>
```
mprins, line 27: There should be no version here (and it's wrong anyway), also as  `compile` scope is the default the scope should be omitted as well. Same for a few lines above

modules/unsupported/cql-json/src/main/java/org/geotools/filter/text/cqljson/MapToOpenGISGeomUtil.java
```diff
@@ -0,0 +1,217 @@
...
+     * @param geoJsonAsMap GeoJSON Map
+     * @return Geometry for filter creation
+     * @throws ParseException
+     */
+    public static org.opengis.geometry.Geometry parseMapToGeometry(Map<String, Object> geoJsonAsMap)
+            throws ParseException {
+        String type = "Bbox";
+        ArrayList coordinates = null;
+        if (geoJsonAsMap.get("bbox") != null) {
+            type = "Bbox";
```
mprins, line 58: This statement can be omitted

turingtestfail: Fixed

modules/unsupported/cql-json/pom.xml
turingtestfail: Fixed

turingtestfail:
@aaime FYI I just heard this from Clemens Portele in the WFS-FES Gitter:
The current draft specifies the object flavor (and has a few implementations), but also points to the array flavor. The SWG is open to which option will eventually been specified, but needs more input why one of the options is preferable to make a good decision. The current plan is to wait at least for the comments from the public review. Preparing part 3 for the public review is the top priority for the group for now.

turingtestfail:
@jodygarnett I understand now.  Yes, I would like commit access for this unsupported module and agree with the approach laid out in https://docs.geotools.org/latest/developer/procedures/create.html

aaime:
@turingtestfail you should ask on the geotools-devel list, in particular ask for "commit access of the cql-json community module", of which you'll be the maintainer, and hence need commit access.

jodygarnett:
@turingtestfail I would like to confirm you have commit access to take care of this community module.

aaime:
@jodygarnett he merged his own PR, that can be done only with commit rights, no? :-D

PR #3162, opened 2020-09-25
[GEOT-6708] Rendering issue while using scientific notation of number…

alemarte:
…s inside SLD filters

## Checklist

For all pull requests:

- [x] You have sent a Contribution Licence Agreement (CLA) as necessary (not required for small changes, e.g., fixing typos in documentation) (Working in GeoSolutions)

### DIFF
--- a/modules/library/main/src/main/java/org/geotools/data/util/NumericConverterFactory.java
+++ b/modules/library/main/src/main/java/org/geotools/data/util/NumericConverterFactory.java
@@ -19,6 +19,8 @@ package org.geotools.data.util;
 import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.util.HashMap;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 import org.geotools.util.Converter;
 import org.geotools.util.ConverterFactory;
 import org.geotools.util.factory.Hints;
@@ -43,6 +45,9 @@ import org.geotools.util.factory.Hints;
  * @since 2.4
  */
 public class NumericConverterFactory implements ConverterFactory {
+
+    private static final Logger LOGGER =
+            org.geotools.util.logging.Logging.getLogger(NumericConverterFactory.class);
 
     public Converter createConverter(Class source, Class target, Hints hints) {
         // convert to non-primitive class
@@ -181,6 +186,8 @@ public class NumericConverterFactory implements ConverterFactory {
 
         public Object convert(Object source, Class target) throws Exception {
             target = primitiveToWrapperClass(target);
+            source = cleanSource(source, target);
+
             if (source instanceof Number) {
                 Number s = (Number) source;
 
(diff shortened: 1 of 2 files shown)


modules/library/main/src/main/java/org/geotools/data/util/NumericConverterFactory.java
```diff
@@ -325,6 +327,34 @@ static String toIntegral(String s) {
...
+     * @return a BigDecimal in case the source is a String in scientific notation, null
+     *     otherwise
+     */
+    static BigDecimal getBigDecimalFromScientificNotation(Object source) {
+        if (source instanceof String) {
+            try {
+                return ((String) source).toUpperCase().contains("E")
+                        ? new BigDecimal((String) source)
+                        : null;
+            } catch (NumberFormatException ex) {
```
ianturton, line 350: Can we at least log to finest that this failed - otherwise you get no clue as to what went wrong with out opening the debugger up.

modules/library/main/src/test/java/org/geotools/data/util/NumericConverterFactoryTest.java
```diff
@@ -311,4 +311,9 @@ public void testPrimitiveTypes() throws Exception {
         assertEquals(1, convert("1", Integer.class));
         assertEquals(Integer.valueOf(1), convert("1", Integer.class));
     }
+
+    public void testStringInScientificNotationToNumber() throws Exception {
+        assertEquals(Long.valueOf("1103442244"), convert("1.1034422448E9", Long.class));
+        assertEquals(Double.valueOf("1103442244.8"), convert("1.1034422448E9", Double.class));
```
ianturton, line 317: Please add a lower case 'e' test too for peace of mind

ianturton (changes requested):

ianturton (approved):

modules/library/main/src/main/java/org/geotools/data/util/NumericConverterFactory.java
```diff
@@ -325,6 +333,37 @@ static String toIntegral(String s) {
         }
     }
 
+    /**
+     * Clean the source handling possible edge cases.
```
aaime, line 337: Should probably clarify what the edge cases are? My guess is that the issue happens when you have a target of type Long/Integer (integral in general) and the value is using scientific notation?

aaime (changes requested):

mprins:
This PR introduces a formatting error
```
[ERROR] Non complying file: /home/travis/build/geotools/geotools/modules/library/main/src/main/java/org/geotools/data/util/NumericConverterFactory.java
```
https://travis-ci.org/github/geotools/geotools/jobs/730238968#L3101

modules/library/main/src/main/java/org/geotools/data/util/NumericConverterFactory.java
alemarte: At the moment only the case you mention justifies the presence of this extra code, and I tried to keep the change as clean as possible. Given that the computational cost is high related to the targeted cases. An alternative approach I was thinking could be make the conversion fails and then retry with a cleansed source.

alemarte:
The cleanSource method has been documented with an ordered list of the handled edge cases.
The formatted error should have been fixed.
There remains the perplexity about the efficiency of performing the cleaning at every conversion operation.

aaime:
About the efficiency, it is my understanding that the cleaning is needed only if the target is a integral number (byte, short, integer, long)... maybe you could check that is the case, before doing the cleanup?

alemarte:
> About the efficiency, it is my understanding that the cleaning is needed only if the target is a integral number (byte, short, integer, long)... maybe you could check that is the case, before doing the cleanup?

And the source is of type String.

aaime (approved):

aaime:
Merged. A note for the other core devs, Alessandro works for GeoSolutions, so he's covered by our corporate CLA. 
@alemarte do we need backport PRs too?

PR #3165, opened 2020-10-01
[GEOT-6713] StreamingRenderer doesn't extract property transforms for xpath expressions

elahrvivaz:
* Add xpath descriptors to style attribute extraction

This pull request fixes an issue in the StreamingRenderer where property transforms are not extracted correctly for complex xpath expressions, resulting in a failure rendering.

## Checklist

For all pull requests:

### DIFF
--- a/modules/library/render/src/test/java/org/geotools/renderer/lite/StreamingRendererTest.java
+++ b/modules/library/render/src/test/java/org/geotools/renderer/lite/StreamingRendererTest.java
@@ -21,6 +21,7 @@ import static org.easymock.EasyMock.createNiceMock;
 import static org.easymock.EasyMock.expect;
 import static org.easymock.EasyMock.replay;
 import static org.geotools.referencing.crs.DefaultGeographicCRS.WGS84;
+import static org.junit.Assert.assertArrayEquals;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertSame;
@@ -59,6 +60,8 @@ import org.geotools.data.store.ContentFeatureSource;
 import org.geotools.feature.DefaultFeatureCollection;
 import org.geotools.feature.simple.SimpleFeatureBuilder;
 import org.geotools.feature.simple.SimpleFeatureTypeBuilder;
+import org.geotools.filter.expression.PropertyAccessor;
+import org.geotools.filter.expression.PropertyAccessorFactory;
 import org.geotools.geometry.GeneralEnvelope;
 import org.geotools.geometry.jts.LiteCoordinateSequence;
 import org.geotools.geometry.jts.LiteShape2;
@@ -80,6 +83,7 @@ import org.geotools.styling.StyleFactoryImpl;
 import org.geotools.styling.StyleImpl;
 import org.geotools.styling.Symbolizer;
 import org.geotools.test.TestData;
+import org.geotools.util.factory.Hints;
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
@@ -92,7 +96,9 @@ import org.mockito.ArgumentCaptor;
 import org.mockito.Mockito;
 import org.opengis.feature.simple.SimpleFeature;
 import org.opengis.feature.simple.SimpleFeatureType;
+import org.opengis.feature.type.AttributeDescriptor;
 import org.opengis.filter.Filter;
+import org.opengis.filter.expression.Expression;
 import org.opengis.filter.spatial.BBOX;
 import org.opengis.geometry.BoundingBox;
 import org.opengis.parameter.GeneralParameterValue;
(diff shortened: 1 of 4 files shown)


elahrvivaz:
There's an error in the downstream build but as far as I can tell it's not related to this change.

modules/library/render/src/test/java/org/geotools/renderer/lite/StreamingRendererTest.java
```diff
@@ -927,4 +933,136 @@ public void errorOccurred(Exception e) {
...
+                    @Override
+                    public SimpleFeatureCollection getFeatures(Query query) {
+                        properties.add(query.getPropertyNames());
+                        return super.getFeatures(query);
+                    }
+                };
+
+        StyleBuilder sb = new StyleBuilder();
+        // add a complex expression to the style
+        Expression color = StreamingRenderer.filterFactory.property("/complex/props/color");
```
aaime, line 971: So the actual attribute structure of the feature type has a "complex" property that is a string, which is marked to be "xml" using a user attribute in the descriptor, and then there  is a property extractor that can drill into the string and extract bits out of it. That is a "creative" use of property accessors, no wonder StreamingRenderer was not prepared for it. We can access JSON structures stored as strings in a similar way using a filter function, that required no changes in the rest of the code.

modules/library/render/src/test/java/org/geotools/renderer/lite/StreamingRendererTest.java
```diff
@@ -927,4 +933,136 @@ public void errorOccurred(Exception e) {
...
+                        throws IllegalArgumentException {
+                    String root = getRoot(xpath);
+                    if (root == null) {
+                        throw new IllegalArgumentException("Invalid xpath: " + xpath);
+                    }
+                    if (object instanceof SimpleFeature) {
+                        String xml = (String) ((SimpleFeature) object).getAttribute(root);
+                        // we don't care about actually extracting the xpath value in our test
+                        return (T) xml;
+                    } else if (object instanceof SimpleFeatureType) {
```
aaime, line 1041: @jodygarnett I'd like to hear some opinion about this bit. Is it correct for a property accessor to return the descriptor of a property that does not match what is actually being accessed, but a xpath parent of it?

modules/library/render/src/test/java/org/geotools/renderer/lite/StreamingRendererTest.java
elahrvivaz: Yes, you captured it perfectly. I believe we came up with this approach based on some of the existing geotools code, but it's been a while and I no longer recall exactly what.
Are there any examples of the existing json filter function you could link to? We also have a jsonPath filter function, but at least in our implementation, we use the same property name expression so I think we hit the same issue (i.e. `jsonPath('$.foo.bar')`. Maybe we could rewrite the filter function to take `jsonPath('foo', 'bar')` instead.

aaime: Yes, the filter function should take attributes that are visible in the feature type, something that can be described as is, that can be given a descriptor of its own (see my question about the property accessor).

The function I was referring to is a JSONPointer one, a quick GitHub search provides references and examples:
https://github.com/geotools/geotools/search?q=jsonpointer

elahrvivaz: thanks, based on that function, it seems like separating out the path from the property name would probably work for us if this PR is ultimately deemed undesirable.

modules/library/render/src/test/java/org/geotools/renderer/lite/StreamingRendererTest.java
jodygarnett: If I am reading this code correctly `root` is the the initial attribute name in the xpath,
so this is going to return one of the property descriptors from the SimpleFeatureType. Should the the property descriptor matching the value returned by the "instance SimpleFeature" case above...

Where do you see " property that does not match what is actually being accessed, but a xpath parent of it?"

aaime: The path in the test is "/complex/props/color", the attribute descriptor returned is for the "/complex" property, because the "props/color" part is in the XML contained in the "complex" property, which is a string. So the path is pointing to one attribute, but the descriptor returned is not for it, it's for the "parent's parent" of it.

jodygarnett: Oh I see, what a pain.

To be honest:
- each property accessor was supposed to access one thing (complex feature, simple feature type, java.util.map, java bean, json object, whatever)
- we had an xpath library (JXPath?) would use property accessors to drill down through this stuff
- Just added the ability to use XPath to drill into FeatureType, SimpleFeatureType, ComplexFeatureType ... and I kind of went with it.

So if I have the design correct, the XPath should drill into the type system should drill down as far as it is able.

In this case it should return a PropertyDescriptor for color binding to Color.class etc...

aaime: Right... so it would need to have a schema of the XML in question and look it up, or just give up and assume everything is a string (the little XML snipped in the test is "<props><color>blue</color></props>", which does not carry any info).
@elahrvivaz I guess this means going the "function way" is probably better.

elahrvivaz: Thanks, I will close the PR. I think our approach was designed to get around having a well-defined schema, to allow schema-less json blobs, since e.g. there would not be a property descriptor for color in the feature type. seems to have worked surprisingly well until hitting this issue :)

PR #3167, opened 2020-10-05
[GEOT-6704] BaseRequestTypeImpl.extendedProperties must not be shared…

awaterme:
PR for [GEOT-6704](https://osgeo-org.atlassian.net/browse/GEOT-6704). 

I also wrote to the developers list regarding this issue, but I received no response.

Problem:
**WFS 2.0** `net.opengis.wfs20.impl.BaseRequestTypeImpl.extendedProperties` is initialized with a static default map from the same type. As result all instance of `BaseRequestTypeImpl` share the same map.
This was introduced with EMF upgrade, AFAICS in GT 21.

Analysis:
* When I regenerate `net.opengis.wfs20.impl.BaseRequestTypeImpl` the static default field is null. This is different from the current state in GT. This is the first part of the solution. Now `extendedProperties` is always null, not shared any more.
* Looking at the **WFS 1** variant of the class I saw the JavaDoc comment `@generated NOT` on the field. A new map is assigned here:

From **WFS 1** `net.opengis.wfs.impl.BaseRequestTypeImpl`:
```
	/**
	 * The cached value of the '{@link #getExtendedProperties() <em>Extended Properties</em>}' attribute.
	 * <!-- begin-user-doc -->
	 * <!-- end-user-doc -->
	 * @see #getExtendedProperties()
	 * @generated NOT
	 * @ordered
	 */
	protected Map extendedProperties = new HashMap();
```

* So I suppose that prior to GeoTools 21 the `extendedProperties` were adjusted by hand, to have an own, not-null map instance assigned. 
* This PR restores the state before GeoTools 21 for WFS 2, which is as one would expect.

I think the solution to modify generated code is not optimal. I  added two unit test cases to prevent from future problems.

An alternative approach would be to keep the `extendedProperties` in WFS 1 _and_ WFS 2 BaseRequestTypeImpl null as intended by the generator. However this is quite ugly for users.

I am open to other suggestions.

PS:
I used Eclipse 2020-09 for generating the class, which contains EMF 2.21.0.v20200708-0547.

## Checklist

For all pull requests:

### DIFF
--- a/modules/ogc/net.opengis.wfs/pom.xml
+++ b/modules/ogc/net.opengis.wfs/pom.xml
@@ -34,6 +34,15 @@
       <artifactId>gt-main</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
+       <groupId>junit</groupId>
+       <artifactId>junit</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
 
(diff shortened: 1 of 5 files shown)


aaime:
Looks good to me. Rerunning the downstream integration build, which had failed due to a maven central related issue.

aaime:
Merged, backport as you see fit.

awaterme:
Thank you for merging, Andrea!

PR #3169, opened 2020-10-05
[GEOT-6715]: COG Support in GeoTIFF

dromagnoli:
https://osgeo-org.atlassian.net/browse/GEOT-6715
Supporting ImageIO-EXT COG Streams into GeoTIFF reader.

Note that:
The PR contains a temporary pom.xml update to use imageio-ext snapshot.
The final PR will be updated to use a newly released imageio-ext 1.3.X version.

## Checklist

For all pull requests:

### DIFF
--- a/modules/library/coverage/pom.xml
+++ b/modules/library/coverage/pom.xml
@@ -122,7 +122,7 @@
   	  <artifactId>imageio-ext-tiff</artifactId>
       <!-- The version number is specified in the parent POM. -->      
     </dependency>
-    
+
     <!--  JAI Tools -->  
     <dependency>
        <groupId>org.jaitools</groupId>
--- a/modules/library/coverage/src/main/java/org/geotools/coverage/grid/io/imageio/MaskOverviewProvider.java
+++ b/modules/library/coverage/src/main/java/org/geotools/coverage/grid/io/imageio/MaskOverviewProvider.java
@@ -16,6 +16,7 @@
  */
 package org.geotools.coverage.grid.io.imageio;
 
+import it.geosolutions.imageio.core.SourceSPIProvider;
 import it.geosolutions.imageio.maskband.DatasetLayout;
 import it.geosolutions.imageio.utilities.ImageIOUtilities;
 import java.awt.Rectangle;
@@ -46,18 +47,19 @@ import org.geotools.util.URLs;
  */
 public class MaskOverviewProvider {
 
+    private static final boolean DEFAULT_SKIP_EXTERNAL_FILES_LOOKUP =
+            ImageIOUtilities.isSkipExternalFilesLookup();
+
     private static final Logger LOGGER =
             org.geotools.util.logging.Logging.getLogger(MaskOverviewProvider.class);
 
     public static final String OVR_EXTENSION = ".ovr";
 
-    private ImageReaderSpi readerSpi;
-
     private ImageReaderSpi overviewReaderSpi;
 
-    private ImageInputStreamSpi streamSpi;
-
     private ImageInputStreamSpi overviewStreamSpi;
+
+    private SourceSPIProvider sourceSpiProvider;
 
     private DatasetLayout layout;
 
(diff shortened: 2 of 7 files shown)


pom.xml
```diff
@@ -89,10 +89,11 @@
     <test.forkMode>once</test.forkMode>
     <test.args></test.args>
     <src.output>${basedir}/target</src.output>
-    <imageio.ext.version>1.3.2</imageio.ext.version>
+    <imageio.ext.version>1.3-SNAPSHOT</imageio.ext.version>
```
dromagnoli, line 92: This is temporary for a first review of the PR.

modules/library/coverage/pom.xml
```diff
@@ -122,7 +122,6 @@
   	  <artifactId>imageio-ext-tiff</artifactId>
       <!-- The version number is specified in the parent POM. -->      
     </dependency>
-
```
aaime, line 125: What happened here? :-D

modules/library/coverage/src/main/java/org/geotools/coverage/grid/io/imageio/MaskOverviewProvider.java
```diff
@@ -787,16 +803,15 @@ public SpiHelper(
             }
         }
 
-        public boolean isMultidim() {
-            return isMultidim;
+        public SpiHelper(ReadingAccessObject readObject) throws IOException {
+            this.readingAccessObject = readObject;
```
aaime, line 807: Why duplicate fields, if the readingAccessObject (ugly name btw) already has them and it's a field, best to remove the other 3 no?

modules/plugin/geotiff/pom.xml
```diff
@@ -119,7 +119,12 @@
       <groupId>it.geosolutions.imageio-ext</groupId>
   	  <artifactId>imageio-ext-tiff</artifactId>
       <!-- The version number is specified in the parent POM. -->      
-    </dependency>   
+    </dependency>
+    <dependency>
```
aaime, line 123: Is this dependency big? Would it make sense to make it optional here in GeoTools, maybe not everyone needs to read COGs. Could be part of the standard deps in GeoServer, or maybe even better a plugin (if memory serves me right, S3 deps are really big, though maybe this cog-reader artifact does not depend on them directly).

modules/plugin/geotiff/src/test/java/org/geotools/gce/geotiff/GeoTiffReaderTest.java
```diff
@@ -1310,6 +1313,32 @@ public void testRescaleWithNodataFloat() throws IOException {
         }
     }
 
+    @Test
+    public void testCOGRead() throws URISyntaxException, IOException {
+        String url =
```
aaime, line 1318: This is an online test, may not work all the time. Does image mosaic already have a "online" test category/profile?

aaime (changes requested):

modules/library/coverage/pom.xml
dromagnoli: I was getting a transitive dependency POM error so I played a bit with the poms.
looks like the reset to its original version didn't worked fine. Will fix it

modules/plugin/geotiff/pom.xml
dromagnoli: Actually, the S3-range-reader (and related aws deps) is not a dependency of the reader

modules/library/coverage/src/main/java/org/geotools/coverage/grid/io/imageio/MaskOverviewProvider.java
dromagnoli: done. also renamed the base classes 
(see https://github.com/geosolutions-it/imageio-ext/pull/228/)

modules/plugin/geotiff/src/test/java/org/geotools/gce/geotiff/GeoTiffReaderTest.java
dromagnoli: I have moved that test in a self contained XOnlineTest class so that it will be excluded when not using the online profile

aaime:
Looks good to me. I've merged the changes in imageio-ext, should be able to build once there are new snapshots available.

pom.xml
```diff
@@ -2310,6 +2321,12 @@
       <url>https://repo.osgeo.org/repository/snapshot/</url>
       <snapshots><enabled>true</enabled></snapshots>
       <releases><enabled>false</enabled></releases>
+    </repository>
+	<repository>
+      <id>geosolutions</id>
+      <name>GeoSolutions libraries repository</name>
+      <url>https://maven.geo-solutions.it/</url>
+	  <snapshots><enabled>true</enabled></snapshots>
```
dromagnoli, line 2329: temporary too for the draft PR. It will be removed on the final PR

aaime:
Building GeoServer I've found that gs-main was trying to download the cog support, checking the dependencies I've found indeed:

```
[INFO] |  \- it.geosolutions.imageio-ext:imageio-ext-cog-reader:jar:1.3-SNAPSHOT:compile
[INFO] |     +- it.geosolutions.imageio-ext:imageio-ext-cog-commons:jar:1.3-SNAPSHOT:compile
[INFO] |     \- it.geosolutions.imageio-ext:imageio-ext-cog-streams:jar:1.3-SNAPSHOT:compile
[INFO] |        \- org.ehcache:ehcache:jar:3.4.0:compile
```

The dependencies here in GeoTools should be optional, maybe mark them as "test" if possible, otherwise as "provided"? Otherwise GeoServer core war will contain COG support out of the box (which might be a medium term objective).

aaime:
Configuring a COG I get the following in GeoServer:

image

This is less than ideal, the URL was provided, so the coverage name would better be the last element of the path provided in the URL instead. We also had issues in the past with readers returning the dreaded "geotiff_coverage" name.

dromagnoli:
> Building GeoServer I've found that gs-main was trying to download the cog support, checking the dependencies I've found indeed:
> 
> ```
> [INFO] |  \- it.geosolutions.imageio-ext:imageio-ext-cog-reader:jar:1.3-SNAPSHOT:compile
> [INFO] |     +- it.geosolutions.imageio-ext:imageio-ext-cog-commons:jar:1.3-SNAPSHOT:compile
> [INFO] |     \- it.geosolutions.imageio-ext:imageio-ext-cog-streams:jar:1.3-SNAPSHOT:compile
> [INFO] |        \- org.ehcache:ehcache:jar:3.4.0:compile
> ```
> 
> The dependencies here in GeoTools should be optional, maybe mark them as "test" if possible, otherwise as "provided"? Otherwise GeoServer core war will contain COG support out of the box (which might be a medium term objective).

Those 3 jars are around 38Kb (in total) with ehcache already part of standard GeoServer.
All the others cog-related jars (the heavy ones, such as S3 aws sdk related) only come with the cog community module.
So that should be safe.

dromagnoli:
> Configuring a COG I get the following in GeoServer:
> 
> image
> 
> This is less than ideal, the URL was provided, so the coverage name would better be the last element of the path provided in the URL instead. We also had issues in the past with readers returning the dreaded "geotiff_coverage" name.

Locally fixed. will be part of the final PR.

PR #3172, opened 2020-10-07
[GEOT-6716]: Support COG on ImageMosaic

dromagnoli:
Note that:
The PR contains a temporary pom.xml update to use imageio-ext snapshot.
The final PR will be updated to use a newly released imageio-ext 1.3.X version.

The PR contains GEOT-6715 for reviewing purposes before merging GEOT-6715.

## Checklist

For all pull requests:


modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/ImageMosaicConfigHandler.java
```diff
@@ -161,6 +158,8 @@
 
     private boolean useExistingSchema;
 
+    private boolean isCog;
```
aaime, line 161: Recommend calling this one just "cog", "isCog" is the name of a getter method, should not be be used for a field.
Also in the property file, there is nothing else starting with "is", having a ``cog=true`` makes things probably easier.

Config wise, wondering if having a cog=true could come with some defaults, like for example:

```
CogRangeReader=it.geosolutions.imageioimpl.plugins.cog.HTTPRangeReader
Heterogeneous=false
cog=true
SuggestedSPI=it.geosolutions.imageioimpl.plugins.cog.CogImageReaderSpi
```
If those two are the common range reader and SPI, I would just assume them if not specified, and force configuration only if extra behavior is needed, e.g., trying to access a COG is a secured S3 bucket, for example.

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/ImageMosaicConfigHandler.java
```diff
@@ -1346,6 +1348,27 @@ private void createPropertiesFiles(MosaicConfigurationBean mosaicConfiguration)
             properties.setProperty(Prop.SUGGESTED_FORMAT, cachedFormat.getClass().getName());
         }
 
+        if (catalogConfigurationBean.isCog()) {
+            properties.setProperty(Prop.IS_COG, Boolean.toString(catalogConfigurationBean.isCog()));
+            CogConfigurationBean cogBean = catalogConfigurationBean.getCogBean();
+            if (cogBean != null) {
+                String rangeReader = cogBean.getRangeReader();
+                if (rangeReader != null) {
+                    properties.setProperty(Prop.COG_RANGE_READER, rangeReader);
```
aaime, line 1357: All these new properties should be documented somewhere (I get this is just a draft, just saying that doc updates should be included in the final version of the PR).

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/catalog/CogConfigurationBean.java
```diff
@@ -0,0 +1,145 @@
...
+import java.net.URI;
+import java.net.URISyntaxException;
+import org.apache.commons.beanutils.BeanUtils;
+import org.geotools.gce.imagemosaic.Utils;
+import org.geotools.gce.imagemosaic.catalog.index.Indexer;
+import org.geotools.gce.imagemosaic.catalog.index.IndexerUtils;
+import org.geotools.util.Utilities;
+
+/** Bean containing all COG related configuration properties */
+public class CogConfigurationBean {
```
aaime, line 30: Can we please just call this a "CogConfiguration"? Nowhere else in GeoTools we use this "Bean" suffix convention, nowhere in Java own platform either.

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/catalog/CogConfigurationBean.java
```diff
@@ -0,0 +1,145 @@
...
+    public CogUri createUri(String url) {
+        // Create basic uri
+        URI uri = URI.create(url);
+        String userInfo = uri.getUserInfo();
+        if (userInfo == null) {
+            // Add credentials to URI when provided
+            String user = getUser();
+            String password = getPassword();
+            if (user != null && password != null) {
+                userInfo = user + ":" + password;
```
aaime, line 122: Wondering what would happen with S3 or Azure though? I guess this is for standard HTTP. However, putting the username and password in the URL does not seem awfully secure, what if they end up in some logs for example?

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/GranuleAccessProvider.java
```diff
@@ -0,0 +1,68 @@
...
+import javax.imageio.stream.ImageInputStream;
+import org.geotools.coverage.grid.io.AbstractGridCoverage2DReader;
+import org.geotools.coverage.grid.io.AbstractGridFormat;
+import org.geotools.coverage.grid.io.imageio.MaskOverviewProvider;
+import org.geotools.util.factory.Hints;
+
+/**
+ * An interface providing multiple types of entities involved when getting access to a Granule.
+ *
+ * <p>An ImageReaderSpi to create an ImageReader to decode that granule An ImageInputStreamSpi to
```
aaime, line 32: I guess this was meant to be a list, use <ul><li>item</li>...</ul> to avoid the formatter from squashing it in one line.

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/CogGranuleAccessProvider.java
```diff
@@ -0,0 +1,127 @@
...
+    public void setGranuleInput(Object input) throws IOException {
+        this.inputUrl = (URL) input;
+        CogUri cogUri = cogConfig.createUri(inputUrl.toString());
+        this.input =
+                new CogSourceSPIProvider(
+                        cogUri, imageReaderSpi, imageInputStreamSpi, cogConfig.getRangeReader());
+    }
+
+    @Override
+    public MaskOverviewProvider getMaskOverviewsProvider() throws IOException {
```
aaime, line 95: This seems to duplicate the entire parent method for a boolean flag in the MaskOverviewProvider. Maybe have a protected method that takes the boolean in the base class, and call it from the base class and from this class with different boolean values instead?

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/GranuleDescriptor.java
```diff
@@ -1735,8 +1711,8 @@ GranuleOverviewLevelDescriptor getLevel(final int index) {
                 throw new IllegalArgumentException(
                         "Unable to get an ImageReader for the provided file "
                                 + granuleUrl.toString());
-            final boolean ignoreMetadata = customizeReaderInitialization(reader, null);
-            reader.setInput(inStream, false, ignoreMetadata);
+
+            reader.setInput(inStream, false, false);
```
aaime, line 1715: Hum... why is it fixed to false here, but a similar call use ignoreMetadata before in the same file?

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/GranuleDescriptor.java
```diff
@@ -902,19 +872,41 @@ public GranuleDescriptor(
         final String granuleLocation = (String) feature.getAttribute(locationAttribute);
         final ReferencedEnvelope granuleBBox = getFeatureBounds(feature);
 
-        // If the granuleDescriptor is not there, dump a message and continue
-        final URL rasterFile = pathType.resolvePath(parentLocation, granuleLocation);
-        if (rasterFile == null) {
+        boolean hasCustomGranuleProvider =
```
aaime, line 875: This block of code (not just the boolean, but also the choice based on the boolean) seems to have commonalities with the other constructor, but not the same. Is there any chance to reduce duplication?

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/GranuleDescriptor.java
```diff
@@ -593,53 +597,6 @@ private void checkPamDataset() throws IOException {
         pamDataset = pamParser.parsePAM(auxFile);
     }
 
-    private boolean customizeReaderInitialization(ImageReader reader, Hints hints) {
```
aaime, line 596: So this one moved in the InitializingReader extension point implementation? Nice. But the COG readers have no use for it?

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/ImageMosaicDatastoreWalker.java
```diff
@@ -233,4 +257,151 @@ public void run() {
             }
         }
     }
+
+    protected void handleCogGranule(final String granuleBeingProcessed) throws IOException {
```
aaime, line 261: I'm guessing this more or less mirrors the handleFile parent method? Any chance this class can be more or less unaware of COGs and just work based on the usage of existing (or new) interfaces encapsulating the different behaviors?

aaime (commented):
Generally speaking, I'm worried that there are too many "if (cog)" in the code. As long as it's configuration, that's unavoidable, but the granule harvesting and mosaic management should be completely unaware of it, and just use interfaces.

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/CogGranuleAccessProvider.java
dromagnoli: Actually the real difference is in the SpiHelper constructor using a SourceSPIProvider in the COG implementation whilst the default one uses a URL.

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/GranuleAccessProvider.java
dromagnoli: Yep. Not sure why it added a <p>. 
Your comments come parsed by github :) I will use a UL and LI Items :)

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/GranuleDescriptor.java
dromagnoli: I'll check. there are some differences in exception throwing vs null return and url validation in the 2 constructors.

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/GranuleDescriptor.java
dromagnoli: an oversight while refactoring the initialization. will fix it. thx

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/ImageMosaicConfigHandler.java
dromagnoli: Right.
At the bottom of the code it is actually working like that more or less. When config is missing, defaults are used.

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/ImageMosaicConfigHandler.java
dromagnoli: @aaime I was actually planning to add them in the geoserver doc as part of the incoming GeoServer PR since most part of these fields and values is defined there.
https://docs.geoserver.org/stable/en/user/data/raster/imagemosaic/configuration.html#configuration-files

Will it make sense?

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/ImageMosaicDatastoreWalker.java
dromagnoli: I'll take a look but there are at least 2 or 3 different points as well as different types/classes as input being used.

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/catalog/CogConfigurationBean.java
dromagnoli: will do.

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/catalog/CogConfigurationBean.java
dromagnoli: @aaime 
This is how both the HTTP and the S3 COG implementation works on imageio-ext. 
They are getting the user and pass from the URI.
What about having all the Imageio-ext classes currently dealing with URI be updated to directly use CogUri instead so that they can get user and pass directly from there?

aaime: Could be an option, but then it would not be any more a "COG" URI, but hum... something else, more general. BasicAuthURI maybe?

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/CogGranuleAccessProvider.java
aaime: One thing that came to mind after the review. Checking the external files is expensive in other occasions, could it be a configuration parameter? And maybe if cog is enabled, you just raise it by default unless explicitly specified? (convenient, but magic, should be well documented). This would make the COG path less special.

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/ImageMosaicConfigHandler.java
aaime: Hum... I thought there was more documentation about mosaics in GeoTools, but what I can find points to the GeoServer docs: https://docs.geotools.org/stable/userguide/library/coverage/mosaic.html
Looks good to me, makes little sense to keep the same documentation duplicated in two places.

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/ImageMosaicConfigHandler.java
aaime: Would be interesting to verify the defaults are working as intended then.

aaime:
From the build results, it seems a "GranuleDescriptorModifier" class is missing in the PR.

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/CogGranuleAccessProvider.java
dromagnoli: I can add it as a config param so it will "override" the System.property right?

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/ImageMosaicConfigHandler.java
dromagnoli: Will add a test taking the indexer and stripping out those entries to double check

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/catalog/CogConfigurationBean.java
dromagnoli: @aaime The CogUri also has caching flag info though
https://github.com/geosolutions-it/imageio-ext/blob/master/plugin/cog/cog-commons/src/main/java/it/geosolutions/imageioimpl/plugins/cog/CogUri.java

What about CogAuthUri?
BTW, user and pass might be null

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/GranuleDescriptor.java
dromagnoli: Done

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/GranuleDescriptor.java
dromagnoli: Done

dromagnoli:
> From the build results, it seems a "GranuleDescriptorModifier" class is missing in the PR.

Damn... fixing it :)

> Generally speaking, I'm worried that there are too many "if (cog)" in the code. As long as it's configuration, that's unavoidable, but the granule harvesting and mosaic management should be completely unaware of it, and just use interfaces.

Actually, the only "relevant" isCog check is the one providing a CogGranuleAccessorProvider which need to take into account all the COG related classes. Most part of the other if(cog) around are simply in configuration propagation between the different places (properties loader/writer), configHandler, indexer init and so on in order to eventually look for/populate/load/write the cogConfiguration.

I'll see if I can refactor a bit the handleCogGranule but that does very specific things with respect to default handleFile which is heavily file-centric

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/catalog/CogConfigurationBean.java
dromagnoli: Updated as BasicAuthURI.

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/catalog/CogConfigurationBean.java
dromagnoli: Done

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/ImageMosaicConfigHandler.java
dromagnoli: Done

dromagnoli:
Superseeded by https://github.com/geotools/geotools/pull/3173

PR #3181, opened 2020-10-18
[GEOT-6722] Fix lost precision bug when saving a Double to GeoPackage

piomar123:
The intention of this PR is to fix the problem with a lost precision while preserving backward-compatibility where possible. Mapping between Java classes and SQL types has been modified from:
```
// Saving:
Float.class -> REAL
Double.class -> REAL
// Reading:
FLOAT -> Double.class (implicit by JDBC read value)
DOUBLE -> Double.class
REAL -> Double.class (implicit by JDBC read value)
```
into:
```
Saving:
Float.class -> FLOAT
Double.class -> DOUBLE
Reading:
FLOAT -> Double.class (explicit)
DOUBLE -> Double.class
REAL -> Double.class (explicit)
```
According to the GeoPackage specification http://www.geopackage.org/spec/ there is no difference between `DOUBLE` and `REAL` types - both can be used interchangeably (Table 1. GeoPackage Data Types).  Moreover, SQLite types are only hints and even `FLOAT` column can contain double-precision floating point numbers (https://sqlite.org/floatingpoint.html). Taking this into consideration and the fact that currently `geoPackage.reader().next().getAttribute()` will return a `Double` value, I believe it makes sense to preserve this behavior and read the attribute as a `Double` even for `FLOAT` type in GeoPackage.

## Checklist

For all pull requests:

### DIFF
--- a/modules/plugin/geopkg/src/main/java/org/geotools/geopkg/GeoPkgDialect.java
+++ b/modules/plugin/geopkg/src/main/java/org/geotools/geopkg/GeoPkgDialect.java
@@ -234,7 +234,10 @@ public class GeoPkgDialect extends PreparedStatementSQLDialect {
     @Override
     public void registerSqlTypeNameToClassMappings(Map<String, Class<?>> mappings) {
         super.registerSqlTypeNameToClassMappings(mappings);
+        // preserve SQLite full precision when dealing with a GeoPackage:
+        mappings.put("FLOAT", Double.class);
         mappings.put("DOUBLE", Double.class);
+        mappings.put("REAL", Double.class);
         mappings.put("BOOLEAN", Boolean.class);
         mappings.put("DATE", java.sql.Date.class);
         mappings.put("TIMESTAMP", java.sql.Timestamp.class);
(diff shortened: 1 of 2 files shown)


modules/plugin/geopkg/src/main/java/org/geotools/geopkg/GeoPkgDialect.java
```diff
@@ -234,7 +234,10 @@ public String getGeometryTypeName(Integer type) {
     @Override
     public void registerSqlTypeNameToClassMappings(Map<String, Class<?>> mappings) {
         super.registerSqlTypeNameToClassMappings(mappings);
+        // preserve SQLite full precision when dealing with a GeoPackage:
+        mappings.put("FLOAT", Double.class);
```
jodygarnett, line 238: Should this be `Float.class` ?

aaime (approved):
Looks good to me

aaime:
Merged without CLA due to minimal changes to a single "non test" file. @piomar123 this will make the bugfix appear on GeoTools 25.0, to be released in March 2021. If you want it to be included sooner, feel free to do backport PRs against the 24.x and 23.x branches.

PR #3187, opened 2020-10-22
[GEOT-6726] Add new FilteringVectorProcess

taba90:
This pr adds a new vector rendering transformation allowing to define features' groups and to return for each one, only one feature having the MIN/MAX value of the attribute defined by the client.

Jira ticket https://osgeo-org.atlassian.net/browse/GEOT-6726

## Checklist

For all pull requests:

### DIFF
--- a/docs/user/unsupported/process/process-feature.rst
+++ b/docs/user/unsupported/process/process-feature.rst
@@ -62,10 +62,10 @@ This is a very flexible process which can be used to:
 
 
 
-Filtering Features
-^^^^^^^^^^^^^^^^^^
+Group candidate selection
+^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Filters a FeatureCollection according to the aggregate operation chosen (MIN or MAX) and the groups defined through attribute names: given a collection of features for each group defined only the feature having the MIN or MAX value for the chosen attribute will be included in the final output.
+Filters a FeatureCollection according to the aggregate operation chosen (MIN or MAX) and the groups defined through attribute names. Given a feature collection, groups according to the defined grouping attributes, and returns the feature having the MIN or MAX value for the chosen attribute. One feature will be chosen for each group.
 
 This vector process accepts four parameters:
 
--- a/docs/user/unsupported/process/process-feature.rst
+++ b/docs/user/unsupported/process/process-feature.rst
@@ -59,4 +59,22 @@ This is a very flexible process which can be used to:
      description: String   description                    description: String
                            area=area( the_geom)           area: Double
                            text=concatenate(name,'-',id)  text: String
- 
+
+
+
+Filtering Features
+^^^^^^^^^^^^^^^^^^
+
+Filters a FeatureCollection according to the aggregate operation chosen (MIN or MAX) and the groups defined through attribute names: given a collection of features for each group defined only the feature having the MIN or MAX value for the chosen attribute will be included in the final output.
+
+This vector process accepts four parameters:
+
+* ``data``: the data on which perform the computation.
+
+* ``aggregation``: the type of operation required to filter data (MIN or MAX).
+
+* ``operationAttribute``: the xpath to the attribute whose value will be used to perform the MIN or MAX operation.
+
+* ``groupingAttributes``: a lists of xpath pointing to the attirbutes defining the features' groups for which perform the filtering process.
+
+The process ensures a consistent results only if the FeatureCollection has already been sorted by the feature's attributes passed as the ``groupingAttributes``  parameter.
--- a/modules/unsupported/process-feature/src/main/java/org/geotools/process/vector/FilteringVectorProcess.java
+++ b/modules/unsupported/process-feature/src/main/java/org/geotools/process/vector/FilteringVectorProcess.java
@@ -1,0 +1,423 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ *
+ *    (C) 2020, Open Source Geospatial Foundation (OSGeo)
+ *
+ *    This library is free software; you can redistribute it and/or
+ *    modify it under the terms of the GNU Lesser General Public
+ *    License as published by the Free Software Foundation;
+ *    version 2.1 of the License.
+ *
+ *    This library is distributed in the hope that it will be useful,
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ *    Lesser General Public License for more details.
+ */
+package org.geotools.process.vector;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.NoSuchElementException;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import org.apache.commons.lang3.ArrayUtils;
+import org.geotools.data.Query;
+import org.geotools.factory.CommonFactoryFinder;
+import org.geotools.feature.FeatureCollection;
+import org.geotools.feature.FeatureIterator;
+import org.geotools.feature.collection.DecoratingFeatureCollection;
+import org.geotools.feature.collection.DecoratingFeatureIterator;
+import org.geotools.feature.collection.PushBackFeatureIterator;
+import org.geotools.feature.type.Types;
+import org.geotools.filter.AttributeExpressionImpl;
+import org.geotools.filter.SortByImpl;
+import org.geotools.metadata.i18n.ErrorKeys;
+import org.geotools.metadata.i18n.Errors;
+import org.geotools.process.ProcessException;
+import org.geotools.process.factory.DescribeParameter;
+import org.geotools.process.factory.DescribeProcess;
+import org.geotools.util.factory.GeoTools;
+import org.opengis.coverage.grid.GridGeometry;
+import org.opengis.feature.Feature;
+import org.opengis.feature.type.FeatureType;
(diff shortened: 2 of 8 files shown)


docs/user/unsupported/process/process-feature.rst
```diff
@@ -59,4 +59,22 @@ This is a very flexible process which can be used to:
      description: String   description                    description: String
                            area=area( the_geom)           area: Double
                            text=concatenate(name,'-',id)  text: String
- 
+
+
+
+Filtering Features
```
aaime, line 65: Tad too generic, it's not just filtering, it's more like "choosing a particular feature in a group". "Group candidate selection" maybe?

docs/user/unsupported/process/process-feature.rst
```diff
@@ -59,4 +59,22 @@ This is a very flexible process which can be used to:
...
+^^^^^^^^^^^^^^^^^^
+
+Filters a FeatureCollection according to the aggregate operation chosen (MIN or MAX) and the groups defined through attribute names: given a collection of features for each group defined only the feature having the MIN or MAX value for the chosen attribute will be included in the final output.
```
aaime, line 68:
```suggestion
Filters a FeatureCollection according to the aggregate operation chosen (MIN or MAX) and the groups defined through attribute names. Given a feature collection, groups according to the defined grouping attributes, and returns the feature having the MIN or MAX value for the chosen attribute. One feature will be chosen for each group.
```

modules/unsupported/process-feature/src/main/java/org/geotools/process/vector/FilteringVectorProcess.java
```diff
@@ -0,0 +1,423 @@
...
+        // produces new PropertyName to add to the query
+        List<PropertyName> propertiesToAdd =
+                Stream.of(sortBy).map(s -> s.getPropertyName()).collect(Collectors.toList());
+        propertiesToAdd.add(ff.property(operationAttribute));
+
+        // eventually merge with existing ones
+        List<PropertyName> pns = getNewProperties(propertiesToAdd, properties);
+        q.setProperties(pns);
+
+        // features with null values for groupingAttributes and operationAttribute
```
aaime, line 165: A database in general can group over nulls, it's just a different value. IMHO this should not be hard-coded, but part of the filtering done in the SLD. If this is a hard requirement, it needs to be clearly documented, as it goes against normal practice (TBH, I would consider this behavior a bug).

For reference: https://learnsql.com/blog/null-values-group-clause/

modules/unsupported/process-feature/src/main/java/org/geotools/process/vector/FilteringVectorProcess.java
```diff
@@ -0,0 +1,423 @@
...
+                toCompareValues.add(p.evaluate(f));
+            }
+            if (groupingValues.size() > 0 && groupingValues.equals(toCompareValues)) return true;
+            return false;
+        }
+
+        private boolean addGroupingValues(List<Object> groupingValues, Feature f) {
+            for (PropertyName p : groupByAttributes) {
+                Object result = p.evaluate(f);
+                if (result != null) groupingValues.add(p.evaluate(f));
```
aaime, line 368: Same as above, null is a valid grouping value, in general.

modules/unsupported/process-feature/src/main/java/org/geotools/process/vector/FilteringVectorProcess.java
```diff
@@ -0,0 +1,423 @@
...
+        private boolean featureComparison(List<Object> groupingValues, Feature f) {
+            List<Object> toCompareValues = new ArrayList<>(groupingValues.size());
+            for (PropertyName p : groupByAttributes) {
+                toCompareValues.add(p.evaluate(f));
+            }
+            if (groupingValues.size() > 0 && groupingValues.equals(toCompareValues)) return true;
+            return false;
+        }
+
+        private boolean addGroupingValues(List<Object> groupingValues, Feature f) {
```
aaime, line 365: This is not really adding the values to the list, it's more like gathering them. Seems also dangerous, I see no cleaning happening, if a feature is skipped for any reason, the groupingValues would keep on accumulating?
IMHO better to have this method return a feature list of values, or null if they are not found.

modules/unsupported/process-feature/src/main/java/org/geotools/process/vector/FilteringVectorProcess.java
```diff
@@ -0,0 +1,423 @@
...
+            this.groupByAttributes = groupByAttributes;
+            this.operationAttribute = operationValue;
+            this.aggregation = aggregation;
+        }
+
+        @Override
+        public boolean hasNext() {
+            List<Object> groupingValues = new ArrayList<>(groupByAttributes.size());
+            Map<Object, Feature> beingFiltered = new HashMap<>();
+            List<Double> valuesToCompare = new ArrayList<>();
```
aaime, line 310: Why do they need to be doubles? Any object supporting equals would work as a grouping mechanism here.

modules/unsupported/process-feature/src/main/java/org/geotools/process/vector/FilteringVectorProcess.java
```diff
@@ -0,0 +1,423 @@
...
+            Object result = this.operationAttribute.evaluate(f, Number.class);
+            if (result instanceof Number) {
+                Double value = ((Number) result).doubleValue();
+                valuesToCompare.add(value);
+                beingFiltered.put(value, f);
+
+            } else {
+                if (result != null)
+                    // not a numeric value. Throwing exception
+                    throw new ProcessException(
```
aaime, line 386: Why in the world is that? Any Comparable will do no?

modules/unsupported/process-feature/src/main/java/org/geotools/process/vector/FilteringVectorProcess.java
```diff
@@ -0,0 +1,423 @@
...
+                return false;
+            return true;
+        }
+
+        private void addDoubleValueFromFeature(
+                Feature f, List<Double> valuesToCompare, Map<Object, Feature> beingFiltered) {
+            Object result = this.operationAttribute.evaluate(f, Number.class);
+            if (result instanceof Number) {
+                Double value = ((Number) result).doubleValue();
+                valuesToCompare.add(value);
```
aaime, line 380: Accumulating features in memory like this is dangerous, could lead to OOM if the groups are large, if all the operations needed are min and max, just keep track of the current value and current feature, and replace if the new feature is "better" than the old one. No need to keep the entire list. Then the method could be renamed to something like "updateSelection" or something like that.

aaime (changes requested):

modules/unsupported/process-feature/src/main/java/org/geotools/process/vector/FilteringVectorProcess.java
aaime: Thinking about it, the MinVisitor and MaxVisitor are already doing the right thing, make them visit the features one by one, get the result, reset them before visiting the next group, done.

docs/user/unsupported/process/process-feature.rst
taba90: fixed

docs/user/unsupported/process/process-feature.rst
taba90: fixed

modules/unsupported/process-feature/src/main/java/org/geotools/process/vector/FilteringVectorProcess.java
taba90: fixed

modules/unsupported/process-feature/src/main/java/org/geotools/process/vector/FilteringVectorProcess.java
taba90: fixed

modules/unsupported/process-feature/src/main/java/org/geotools/process/vector/FilteringVectorProcess.java
taba90: fixed

modules/unsupported/process-feature/src/main/java/org/geotools/process/vector/FilteringVectorProcess.java
taba90: fixed

modules/unsupported/process-feature/src/main/java/org/geotools/process/vector/FilteringVectorProcess.java
taba90: Using Comparable now.  In case of ComplexFeatures the evaluation without providing a Context will return a Property. Added some code to check the type and eventually return the Property value as a comparable

docs/user/unsupported/process/process-feature.rst
aaime: The title still looks the same to me?

image

modules/unsupported/process-feature/src/main/java/org/geotools/process/vector/FilteringVectorProcess.java
aaime: I can still see a null check in the current code, getGroupingValues method.

taba90:
thanks for your review @aaime your feedback have been applied

docs/user/unsupported/process/process-feature.rst
taba90: fixed

modules/unsupported/process-feature/src/main/java/org/geotools/process/vector/FilteringVectorProcess.java
taba90: fixed

aaime (approved):

PR #3196, opened 2020-11-03
RFC: Transform data store

jnh5y:
The goal of this work is to enable the easy addition of new columns defined by Expressions.  Additionally, this DataStore allows for projecting to a subset of the columns in the FeatureStore. 

## Checklist

For all pull requests:

### DIFF
--- a/modules/extension/transform/pom.xml
+++ b/modules/extension/transform/pom.xml
@@ -66,17 +66,23 @@
 
     <dependency>
       <groupId>org.geotools</groupId>
-      <artifactId>gt-property</artifactId>
-      <version>${project.version}</version>
-      <scope>test</scope>
-    </dependency>
-
-    <dependency>
-      <groupId>org.geotools</groupId>
       <artifactId>gt-cql</artifactId>
       <version>${project.version}</version>
     </dependency>
 
+    <dependency>
+      <groupId>org.geotools</groupId>
+      <artifactId>gt-process-feature</artifactId>
+      <version>${project.version}</version>
+      <scope>compile</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>org.geotools</groupId>
+      <artifactId>gt-property</artifactId>
+      <version>${project.version}</version>
+      <scope>test</scope>
+    </dependency>
 
     <dependency>
       <groupId>org.geotools</groupId>
@@ -103,12 +109,6 @@
       <version>${project.version}</version>
       <classifier>tests</classifier>
       <scope>test</scope>
-    </dependency>
-    <dependency>
-      <groupId>org.geotools</groupId>
-      <artifactId>gt-process-feature</artifactId>
-      <version>25-SNAPSHOT</version>
-      <scope>compile</scope>
     </dependency>
 
   </dependencies>
--- a/modules/extension/transform/pom.xml
+++ b/modules/extension/transform/pom.xml
@@ -75,7 +75,6 @@
       <groupId>org.geotools</groupId>
       <artifactId>gt-cql</artifactId>
       <version>${project.version}</version>
-      <scope>test</scope>
     </dependency>
 
 
@@ -104,6 +103,12 @@
       <version>${project.version}</version>
       <classifier>tests</classifier>
       <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.geotools</groupId>
+      <artifactId>gt-process-feature</artifactId>
+      <version>25-SNAPSHOT</version>
+      <scope>compile</scope>
     </dependency>
 
   </dependencies>
--- a/modules/extension/transform/src/main/java/org/geotools/data/transform/TransformDataStoreFactory.java
+++ b/modules/extension/transform/src/main/java/org/geotools/data/transform/TransformDataStoreFactory.java
@@ -1,3 +1,19 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ *
+ *    (C) 2020, Open Source Geospatial Foundation (OSGeo)
+ *
+ *    This library is free software; you can redistribute it and/or
+ *    modify it under the terms of the GNU Lesser General Public
+ *    License as published by the Free Software Foundation;
+ *    version 2.1 of the License.
+ *
+ *    This library is distributed in the hope that it will be useful,
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ *    Lesser General Public License for more details.
+ */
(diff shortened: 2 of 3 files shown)


aaime:
The QA build reports a PMD error:

``[INFO] PMD Failure: org.geotools.data.transform.TransformDataStoreFactory:76 Rule:UnusedLocalVariable Priority:3 Avoid unused local variables such as 'namespace'..``

aaime:
@jnh5y this has been stuck for a while it seems, any progress in the plans?

jnh5y:
@aaime thanks for asking!  I got redirected a little toward the end of 2020.  

What needs to happen to finish this contribution well?  

I know that I definitely need to rebase / fix the PMD issues.  What else should be done to make this as-useful-as-possible?  (Acknowledging that a handy Wicket wizard in GeoServer would be awesome, but beyond my ken.:))

modules/extension/transform/src/main/java/org/geotools/data/transform/TransformDataStoreFactory.java
```diff
@@ -0,0 +1,181 @@
...
+                            .stream()
+                            .map(ad -> new Definition(ad.getLocalName()))
+                            .collect(Collectors.toList());
+        }
+
+        String transformDefinition = (String) TRANSFORM_DEFINITION.lookUp(params);
+
+        if (transformDefinition != null) {
+            List<Definition> transforms =
+                    Arrays.stream(transformDefinition.split(";"))
```
aaime, line 109: How about using a property file format instead? The GUI in GeoServer allows for textarea input, if the Param is marked as "large text", example here: https://github.com/geotools/geotools/blob/14d27a2db448454c64e8f38e9977565cc400d52b/modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStoreFactory.java#L218

modules/extension/transform/pom.xml
```diff
@@ -66,18 +66,23 @@
...
-      <scope>test</scope>
     </dependency>
 
     <dependency>
       <groupId>org.geotools</groupId>
-      <artifactId>gt-cql</artifactId>
+      <artifactId>gt-process-feature</artifactId>
       <version>${project.version}</version>
-      <scope>test</scope>
+      <scope>compile</scope>
```
aaime, line 77: Why the extra scope here? And does the module really need to depend on process-feature, I don't understand why?

modules/extension/transform/src/main/java/org/geotools/data/transform/TransformDataStoreFactory.java
```diff
@@ -0,0 +1,181 @@
...
+    public DataStore createDataStore(Map<String, Serializable> params) throws IOException {
+
+        URI namespace = (URI) NAMESPACEP.lookUp(params);
+        Repository repository = (Repository) REPOSITORY.lookUp(params);
+
+        String datastoreName = (String) DATASTORE_NAME.lookUp(params);
+        String featureTypeName = (String) FEATURE_TYPE_NAME.lookUp(params);
+        String newFeatureTypeName = (String) NEW_FEATURE_TYPE_NAME.lookUp(params);
+
+        DataStore ds = repository.dataStore(new NameImpl(datastoreName));
```
aaime, line 83: This is somewhat dangerous... the repository controls the lifecycle of the objects it returns, meaning, it can decide to close the store leaving the transformed version dangling. This actually happens in GeoServer, where there is a _cache_ of stores with limited size, if there is enough turn-around the base store might be closed while this one is still open, bricking it.  (or maybe the admin just opens the base store, saves its configuration, and GeoServer kills the old store and creates a new one, also bricking the transformed one).

On the other side, this is not a new issue, "features-pregeneralized" suffers from the same issue. In general, the wrappers should pick the store from the repository every time they need it, instead of holding onto it.... of course that might have other potential issues, like performance degradation. We don't have a good answer to this issue yet (e.g., cheap check if a store is still alive).

Solving it is not trivial, maybe document about the issue somewhere.

aaime:
I'd also recommend rebasing the PR, the builds are running fully on GitHub actions now.

aaime:
@jnh5y should we keep the PR open?

jnh5y:
> @jnh5y should we keep the PR open?

Sorry to have been slow to reply.  Let me ask a few more questions on this PR and if we agree on an approach, I'll rebase and knock out the changes.

modules/extension/transform/src/main/java/org/geotools/data/transform/TransformDataStoreFactory.java
jnh5y: This is a great point.  I think I see a few options...

1.  Write up a "ReconnectingFeatureStore" to get a new delegate FeatureSource for each request.  Seems like this class would need to verify that the SimpleFeatureType has not shifted underneath...  It seems like this would address the "pregeneralized" situation as well?  As you mention, it seems reasonable to worry a little about performance; that said, I cannot put my finger on exactly what I'd expect to be slow.  (I hope the Repository lookup is fast.)

2.  Riffing on the idea to have Transforms as properties, another option would be to enter the datastore connection parameters as properties block?  That way the Transform DataStore could just manage its own direct connection?

3.  Use as-is and document the issues.

4.  Could a GeoServer CatalogListener (or similar) be created to help bridge these issues?  E.g. when a store is saved/changed/etc, the CatalogListener could poke all the TransformDataStores to refresh their connection?

Are there any other callbacks or options I'm missing?  Of the above, are any of them sounding better than others?

jnh5y: @aaime any feedback on these questions?  (Admittedly, they are tricky.:))

jnh5y:
@aaime any more suggestions on this PR (before the stale bot closes it)?

modules/extension/transform/src/main/java/org/geotools/data/transform/TransformDataStoreFactory.java
aaime: Whew, found some time to look at the questions.

1. Yes, such  FeatureSource/FeatureStore could be helpful. The potential performance issue is just related to how the Repository works (which is not known). If the repository was to create a new instance for each call, that would be bad performance-wise. The repository provided by GeoServer is backed by the ResourcePool cache, so no worries. ImageMosaic has an option to pull the index store from a repository that, so far, seems to have worked.

2. I either don't understand this suggestion or don't like it :-D  Repository is also about sharing and benefiting from it, like, avoiding creating N connection pools when one would do.

3. An option that I'd avoid too :-D

4. That would work, although making it specific to TransformDataStores is a bit narrow. Maybe the store could be implementing an interface to make it recognizable as a Repository user, and provide a method to clear the caches.

Generally speaking I like option 1. better, mostly because I haven't seen cases of Repository that actually do re-create stores at every call.

PR #3205, opened 2020-11-10
[GEOT-6730]: Supporting external overviews on COG

dromagnoli:
## Checklist

For all pull requests:


modules/plugin/geotiff/src/test/java/org/geotools/gce/geotiff/GeoTiffReaderCogOnlineTest.java
```diff
@@ -61,4 +75,49 @@ public void testCogRead() throws URISyntaxException, IOException {
...
+        gg.setValue(new GridGeometry2D(range, envelope));
+        params[0] = gg;
+
+        GridCoverage2D coverage = reader.read(params);
+        assertNotNull(coverage);
+        RenderedImage image = coverage.getRenderedImage();
+        int numTileX = image.getNumXTiles();
+        int numTileY = image.getNumYTiles();
+
+        CogImageReader imageReader = (CogImageReader) image.getProperty("JAI.ImageReader");
```
aaime, line 106: Interesting bit... made me wonder about a few things that are related to COG, but not the ovr usage per se.

1) The coverage generated by image mosaic normally contains a list of files used to generated it. Is that still the case for COG? 
But I guess it's the main file, not the ovr, that is listed there... still useful I believe.

2) Also wondering how CSW direct download would play with COGs.

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/GranuleDescriptor.java
```diff
@@ -1060,15 +1061,13 @@ public GranuleLoadingResult loadRaster(
                 granuleURLUpdated = ovrProvider.getOvrURL();
                 assert ovrProvider.getExternalOverviewInputStreamSpi() != null
                         : "no cachedStreamSPI available for external overview!";
-                inStream =
+                SourceSPIProvider sourceSpiProvider =
```
aaime, line 1064: I'm not familiar with this bit of code, but the change gives me the impression that before the code allowed to have two different types of format for image and ovr (which is a thing, I believe that GDAL handles jpeg as main image but tiff as ovr for example). While the updated code assumes they are the same?

dromagnoli: Fixed.

modules/plugin/geotiff/src/test/java/org/geotools/gce/geotiff/GeoTiffReaderCogOnlineTest.java
dromagnoli: 1. Updated to deal with that, reporting the COG Url. Note that ImageMosaic reports .ovr instead, so it will be the same for COG

2. All GT APIs used by DirectDownload are File-related (FileGroups, FileGroupProvider, FileInfo and so on). We need to revisit the GT API to have it working for cog too

aaime (approved):
LGTM

PR #3212, opened 2020-11-16
[GEOT-6738] Mongo ComplexFeatures are not reprojected

taba90:
Jira ticket https://osgeo-org.atlassian.net/browse/GEOT-6738

this pr fix a bug with MongoComplexFeatures not being reprojected.

## Checklist

For all pull requests:

### DIFF
--- a/modules/plugin/mongodb/src/main/java/org/geotools/data/mongodb/complex/MongoComplexUtilities.java
+++ b/modules/plugin/mongodb/src/main/java/org/geotools/data/mongodb/complex/MongoComplexUtilities.java
@@ -16,6 +16,8 @@
  */
 package org.geotools.data.mongodb.complex;
 
+import static org.geotools.referencing.CRS.findMathTransform;
+
 import com.mongodb.BasicDBList;
 import com.mongodb.DBCursor;
 import com.mongodb.DBObject;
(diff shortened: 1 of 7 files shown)


modules/plugin/mongodb/src/main/java/org/geotools/data/mongodb/complex/MongoComplexUtilities.java
```diff
@@ -41,6 +52,10 @@
     // key used to store the parent JSON path in a feature user data map
     public static final String MONGO_PARENT_PATH = "MONGO_PARENT_PATH";
 
+    private static ThreadLocal<
```
aaime, line 55: Thread locals can turn into memory leaks, and web containers do complain about TL that are not freed on web application un-deploy. That said, we already use something similar in JDBC stores, so it would not be a first. If there is no way to keep these as collection fields (creating the transformations once per collection instead of once per feature) it's not going to be the end of the world if it stays as is.

taba90: Thanks for your point @aaime . I strove to found another way but looked to me a forced choice: it was possible to not keep it at a feature level when the static methods in the MongoConplexUtilities class are called from the [MongoNestedMapping class](https://github.com/geotools/geotools/blob/d101d8ccae34a1d9afba1bd103a8550c99fd4e73/modules/plugin/mongodb/src/main/java/org/geotools/data/mongodb/complex/MongoNestedMapping.java#L79) but I did not find way to handle the case when the static method is called from the [JsonSelectFunction](https://github.com/geotools/geotools/blob/82af4ca92eaefc092e809a89377da6b5dbae73b9/modules/plugin/mongodb/src/main/java/org/geotools/data/mongodb/complex/JsonSelectFunction.java#L64)

PR #3213, opened 2020-11-16
[GEOT-6739] GroupCandidateSelectionProcess should return feature with null attribute

taba90:
Jira ticket https://osgeo-org.atlassian.net/browse/GEOT-6739

- This pr fix a problem with GroupCandidateSelection process, making it returning a feature aslso if in a defined group the operationValue is always NULL for all the members.

## Checklist

For all pull requests:

### DIFF
--- a/modules/unsupported/process-feature/src/main/java/org/geotools/process/vector/GroupCandidateSelectionProcess.java
+++ b/modules/unsupported/process-feature/src/main/java/org/geotools/process/vector/GroupCandidateSelectionProcess.java
@@ -297,14 +297,13 @@ public class GroupCandidateSelectionProcess implements VectorProcess {
             Feature bestFeature = null;
             while (super.hasNext()) {
                 Feature f = super.next();
-                Comparable operationValue = getComparableFromEvaluation(f);
-                if (bestFeature == null && operationValue != null) {
+                if (bestFeature == null) {
                     // no features in the list this is the first of the group
                     // takes the values to check the following features if belong to the same
                     // group
                     groupingValues = getGroupingValues(groupingValues, f);
                     bestFeature = f;
-                } else if (bestFeature != null && operationValue != null) {
+                } else {
                     // is the feature in the group?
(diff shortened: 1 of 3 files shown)


modules/unsupported/process-feature/src/main/java/org/geotools/process/vector/GroupCandidateSelectionProcess.java
```diff
@@ -343,23 +342,26 @@ else if (groupingValues != null
         }
 
         private Feature updateBestFeature(Feature best, Feature f) {
+            Comparable bestValue = getComparableFromEvaluation(best);
+            Comparable value = getComparableFromEvaluation(f);
+            if (bestValue == null && value == null) return best;
```
aaime, line 347: So, 

```
if (value == null) return best;
else if (bestValue == null) return f;
```

modules/unsupported/process-feature/src/test/resources/org/geotools/process/vector/test-data/featuresToGroupWithGroupAllNull.properties
```diff
@@ -0,0 +1,13 @@
+_=the_geom:MultiPolygon:srid=4326,cat:java.lang.Long,group:String,group2:String,numericVal:java.lang.Integer
+features.1=MULTIPOLYGON(((2 2,2 3,3 3,3 2,2 2)), ((3 9,3 12,5 12,5 9,3 9)))|10|group1|secondGroup1|null
```
aaime, line 2: Is it just me, or there is no group with all values at null? I can only see groups with a mix of nulls and non nulls (which needs to be tested for sure, but the title talks about features with null attribute).

modules/unsupported/process-feature/src/main/java/org/geotools/process/vector/GroupCandidateSelectionProcess.java
taba90: indeed fixed

PR #3265, opened 2020-12-21
[GEOT-6762]: CSS to SLD translation supporting ColorMapEntries with CQL Expression

dromagnoli:
Supporting CQL Expressions in ColorMapEntries attributes on CSS so that they get preserved when translated to SLD

# Checklist

For all pull requests:

### DIFF
--- a/modules/library/cql/src/main/java/org/geotools/filter/text/commons/ExpressionToText.java
+++ b/modules/library/cql/src/main/java/org/geotools/filter/text/commons/ExpressionToText.java
@@ -138,7 +138,6 @@ public class ExpressionToText implements ExpressionVisitor {
         expression.getExpression1().accept(this, output);
         output.append(" + ");
         expression.getExpression2().accept(this, output);
-
         return output;
     }
 
@@ -147,13 +146,20 @@ public class ExpressionToText implements ExpressionVisitor {
      */
     @Override
     public Object visit(Divide expression, Object extraData) {
-
         StringBuilder output = asStringBuilder(extraData);
-        expression.getExpression1().accept(this, output);
+        visitWithBrackets(expression.getExpression1(), output);
         output.append(" / ");
-        expression.getExpression2().accept(this, output);
-
+        visitWithBrackets(expression.getExpression2(), output);
         return output;
+    }
+
+    private void visitWithBrackets(Expression expression, StringBuilder output) {
+        boolean needsBrackets = (expression instanceof Subtract || expression instanceof Add);
+        // Make sure to include Subtract or Add expression between brackets to preserve
+        // operator precedences.
+        output.append(needsBrackets ? "(" : "");
+        expression.accept(this, output);
+        output.append(needsBrackets ? ")" : "");
     }
 
     /* (non-Javadoc)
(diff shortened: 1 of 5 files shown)


modules/library/cql/src/main/java/org/geotools/filter/text/commons/ExpressionToText.java
```diff
@@ -147,15 +146,22 @@ public Object visit(Add expression, Object extraData) {
...
-        expression.getExpression1().accept(this, output);
+        visitWithBrackets(expression.getExpression1(), output);
         output.append(" / ");
-        expression.getExpression2().accept(this, output);
-
+        visitWithBrackets(expression.getExpression2(), output);
         return output;
     }
 
+    private void visitWithBrackets(Expression expression, StringBuilder output) {
```
aaime, line 156: This should have its own test, and maybe its own ticket too.

modules/unsupported/css/src/test/java/org/geotools/styling/css/ParserSyntheticTest.java
```diff
@@ -820,4 +830,51 @@ public void testMultiNestedSelectors() {
...
     }
+
+    @Test
+    public void testDynamicColorMapEntryFromCss() throws Exception {
+        String css =
+                "* { raster-channels: 'auto'; raster-color-map: color-map-entry("
+                        + "[env('color','#FF0000')], "
+                        + "[env('min', 1) + 2 * (env('max',10) - env('min',1)) / 3],"
+                        + "[env('customOpacity', 1)])  ;}";
+        LinearColorMap colorMap = buildColorMapFromCss(css);
```
aaime, line 841: So this builds the final result and checks the default values for env are respected?
I would also verify that the ColorMap/ColorMapEntries have the expected structure, similar to the one we'd get out of a SLD parse.

aaime (changes requested):

modules/library/cql/src/main/java/org/geotools/filter/text/commons/ExpressionToText.java
dromagnoli: The ParseSyntheticTest is actually already containing that test.
Would that be ok?

modules/unsupported/css/src/test/java/org/geotools/styling/css/ParserSyntheticTest.java
dromagnoli: Yep. it does 2 passes.
one checking for the default value and one right after having set some ENV values so that the result get different.
I have updated the test to check that the colormap has the expected structure.

modules/library/cql/src/main/java/org/geotools/filter/text/commons/ExpressionToText.java
dromagnoli: I have added a separate CQL text validating the expressions structure/order.

dromagnoli:
@aaime I have added a dedicate test for the expression-brackets management and refined the colormap test too.

aaime (approved):

PR #3285, opened 2020-12-23
[GEOT-6769] Refact ClippingFeatureCollection

taba90:
Jira ticket https://osgeo-org.atlassian.net/browse/GEOT-6769

- Move ClippingFeatureCollection from inside the ClipProcess to library/main

# Checklist

For all pull requests:


modules/library/main/src/main/java/org/geotools/feature/collection/ClipIntersectsFeatureCollection.java
```diff
@@ -0,0 +1,80 @@
...
+ */
+package org.geotools.feature.collection;
+
+import org.geotools.data.simple.SimpleFeatureCollection;
+import org.geotools.data.simple.SimpleFeatureIterator;
+import org.locationtech.jts.geom.Geometry;
+import org.opengis.feature.simple.SimpleFeatureType;
+
+/**
+ * A SimpleFeatureCollection that can filter features' geometries by a clip (crop) spatialFilter and
```
aaime, line 25: I'm confused as to why we need both.. filtering by intersection is just a filter, FilteringFeatureCollection can be used for the task without creating a dedicated collection for that functionality? Also, the filter is best sent down to the data source.

So it should be merged with other filters coming from the current request, if possible. Thinking about a WFS/WMS case for example, say the WMS bbox and the intersects filter have a small intersection, working at the filter level would allow to load only the features in the intersection between the two, loading significantly less data. And if the intersection is empty, maybe don't load data at all.

taba90: thanks for your point @aaime  please check this comment on the related gs pr https://github.com/geoserver/geoserver/pull/4627#discussion_r554445769

taba90: @aaime  I changed the pr and the jira ticket to simply refact the ClipCollection, and moved the IntersectClipCollection to geoserver

aaime: Thanks!

aaime:
Looks good, merged.

PR #3308, opened 2021-01-04
initial jdk11 update work (draft, wip, not for merge)

bradh:
This updates the arcgis-rest module to build with jdk11. 

The main fix is to update the dependencies to include javax ws api. Possibly some impl is also needed to make it run correctly.

I've also bumped mockito (to match parent pom) and powermock.

I have disabled 4 tests that are still failing. This is clearly the wrong approach, but I don't see how to fix it without rewrite, and I definitely don't understand the code well enough to do that.

### DIFF
--- a/modules/unsupported/arcgis-rest/pom.xml
+++ b/modules/unsupported/arcgis-rest/pom.xml
@@ -39,8 +39,7 @@
     </developers>
 
     <properties>
-        <powermock.version>1.6.6</powermock.version>
-        <mockito.version>1.10.19</mockito.version>
+        <powermock.version>2.0.9</powermock.version>
     </properties>
 
     <repositories>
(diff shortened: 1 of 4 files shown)


modules/unsupported/arcgis-rest/pom.xml
```diff
@@ -121,6 +113,11 @@
             <version>23-SNAPSHOT</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
mprins, line 116: Better to create a profile for this that is activated by the JDK in use, that way you won't run into conflicts when running on JDK8.

see for example https://github.com/geoserver/geofence/pull/154 and https://github.com/geoserver/geofence/commit/67c982239c59f080444830b6000dd0421d2

modules/unsupported/arcgis-rest/pom.xml
```diff
@@ -121,6 +113,11 @@
             <version>23-SNAPSHOT</version>
```
mprins, line 113:
```suggestion
            <version>${project.version}</version>
```

bradh:
Closing in favour of an alternative to be proposed by Luca.

PR #3317, opened 2021-01-15
[GEOT-6817] Extend usage of varargs in core module, make the API easier to use

aaime:
See discussion on gt-devel

# Checklist

For all pull requests:

### DIFF
--- a/modules/library/main/src/main/java/org/geotools/data/DataUtilities.java
+++ b/modules/library/main/src/main/java/org/geotools/data/DataUtilities.java
@@ -793,7 +793,7 @@ public class DataUtilities {
      * @throws ArrayIndexOutOfBoundsException If the number of provided values does not match the
      *     featureType
      */
-    public static SimpleFeature template(SimpleFeatureType featureType, Object[] providedValues) {
+    public static SimpleFeature template(SimpleFeatureType featureType, Object... providedValues) {
         return SimpleFeatureBuilder.build(
                 featureType, defaultValues(featureType, providedValues), null);
     }
@@ -808,7 +808,7 @@ public class DataUtilities {
      *     featureType
      */
     public static SimpleFeature template(
-            SimpleFeatureType featureType, String featureID, Object[] providedValues) {
+            SimpleFeatureType featureType, String featureID, Object... providedValues) {
         return SimpleFeatureBuilder.build(
                 featureType, defaultValues(featureType, providedValues), featureID);
     }
@@ -820,7 +820,7 @@ public class DataUtilities {
      * @throws ArrayIndexOutOfBoundsException If the number of provided values does not match the
      *     featureType
      */
-    public static Object[] defaultValues(SimpleFeatureType featureType, Object[] values) {
+    public static Object[] defaultValues(SimpleFeatureType featureType, Object... values) {
         if (values == null) {
             values = new Object[featureType.getAttributeCount()];
         } else if (values.length != featureType.getAttributeCount()) {
@@ -955,7 +955,7 @@ public class DataUtilities {
      * @throws IOException If provided features Are null or empty
      */
     public static FeatureReader<SimpleFeatureType, SimpleFeature> reader(
-            final SimpleFeature[] features) throws IOException {
+            final SimpleFeature... features) throws IOException {
         if ((features == null) || (features.length == 0)) {
             throw new IOException("Provided features where empty");
         }
@@ -994,7 +994,7 @@ public class DataUtilities {
      * @param featureArray Array of features
      * @return FeatureSource
      */
-    public static SimpleFeatureSource source(final SimpleFeature[] featureArray) {
+    public static SimpleFeatureSource source(final SimpleFeature... featureArray) {
         final SimpleFeatureType featureType;
 
         if ((featureArray == null) || (featureArray.length == 0)) {
@@ -1166,7 +1166,7 @@ public class DataUtilities {
      * @param features Array of features
      * @return FeatureCollection
      */
-    public static SimpleFeatureCollection collection(SimpleFeature[] features) {
+    public static SimpleFeatureCollection collection(SimpleFeature... features) {
         // JG: There may be some performance to be gained by using ListFeatureCollection here
         DefaultFeatureCollection collection = new DefaultFeatureCollection(null, null);
         for (SimpleFeature feature : features) {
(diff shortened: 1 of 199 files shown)


aaime:
Was trying to self-review the morning after... the code generally looks better to me.
Trying to find guidance on the net, got these two interesting references:
* https://stackoverflow.com/questions/5405673/java-varargs-method-param-list-vs-array
* https://www.thefinestartist.com/effective-java/42

What I gather at a first read, is to stay away from varargs when the original array parameter involved primitives, or was collecting any type of object, instead of a specific type (as passing an array would be considered an item, instead of "the array").

There are a few cases of those, going to highlight them.

modules/library/opengis/src/main/java/org/opengis/feature/simple/SimpleFeature.java
```diff
@@ -143,7 +143,7 @@
      *
      * @param values The attribute values to set.
      */
-    void setAttributes(Object[] values);
```
aaime, line 146: Dangerous conversion here, will likely trip existing code that was not refactored to avoid the array.

modules/library/referencing/src/main/java/org/geotools/referencing/factory/epsg/AbstractEpsgFactory.java
```diff
@@ -2747,7 +2745,7 @@ public synchronized Set generateFromCoordinateReferenceSystemCodes(
      *     element.
      */
     // TODO: Use generic type for "Object[] codes" with J2SE 1.5.
-    private void sort(final Object[] codes) throws SQLException, FactoryException {
```
aaime, line 2750: Private method, not an issue

modules/library/main/src/main/java/org/geotools/data/DataUtilities.java
```diff
@@ -793,7 +793,7 @@ public static SimpleFeature template(SimpleFeatureType featureType, String featu
      * @throws ArrayIndexOutOfBoundsException If the number of provided values does not match the
      *     featureType
      */
-    public static SimpleFeature template(SimpleFeatureType featureType, Object[] providedValues) {
```
aaime, line 796: Dangerous conversion, will trip client code that is not converted to not using arrays.

modules/library/main/src/main/java/org/geotools/data/DataUtilities.java
```diff
@@ -808,7 +808,7 @@ public static SimpleFeature template(SimpleFeatureType featureType, Object[] pro
      *     featureType
      */
     public static SimpleFeature template(
-            SimpleFeatureType featureType, String featureID, Object[] providedValues) {
```
aaime, line 811: Same as above

modules/library/main/src/main/java/org/geotools/data/DataUtilities.java
```diff
@@ -820,7 +820,7 @@ public static SimpleFeature template(
      * @throws ArrayIndexOutOfBoundsException If the number of provided values does not match the
      *     featureType
      */
-    public static Object[] defaultValues(SimpleFeatureType featureType, Object[] values) {
```
aaime, line 823: Same as above.

modules/library/main/src/main/java/org/geotools/feature/FeatureCollection.java
```diff
@@ -237,5 +237,5 @@
     Object[] toArray();
 
     /** @see java.util.Collection#toArray(Object[]) */
-    <O> O[] toArray(O[] a);
```
aaime, line 240: Dangerous conversion here.

modules/library/main/src/main/java/org/geotools/feature/simple/SimpleFeatureBuilder.java
```diff
@@ -211,7 +211,7 @@ public void addAll(List<Object> values) {
     }
 
     /** Adds an array of attributes. */
-    public void addAll(Object[] values) {
```
aaime, line 214: Dangerous conversion here

modules/library/main/src/main/java/org/geotools/feature/simple/SimpleFeatureBuilder.java
```diff
@@ -324,7 +324,7 @@ public SimpleFeature buildFeature(String id) {
     }
 
     /** Quickly builds the feature using the specified values and id */
-    public SimpleFeature buildFeature(String id, Object[] values) {
```
aaime, line 327: Dangerous conversion here

modules/library/main/src/main/java/org/geotools/feature/visitor/CalcUtil.java
```diff
@@ -126,7 +126,7 @@ static Number average(Number[] numbers) {
      *
      * @return the most
      */
-    static Class bestClass(Object[] objects) {
```
aaime, line 129: Package private, probably not dangerous

modules/library/metadata/src/main/java/org/geotools/metadata/PropertyAccessor.java
```diff
@@ -506,7 +506,7 @@ private static void set(
             final Method getter,
             final Method setter,
             final Object metadata,
-            final Object[] arguments)
```
aaime, line 509: Also to be removed.

modules/library/main/src/main/java/org/geotools/data/DataUtilities.java
jodygarnett: Will it? It is too bad as this is one case where it may be a usability improvement ...

aaime: That's what I understood reading articles around, but then tried in a debugger, and the array was not provided as an item, but as the full argument.
The case on which I based the above comment is "Array.asList", which did bite me in the past, you think you're passing in an array to get a list, and get a list with a single value, the array itself. Looking again at the issue, it seems to be happening only if the users of the method tried to pass arrays of primitives:
https://javadevnotes.com/java-array-to-list-examples

modules/library/opengis/src/main/java/org/opengis/feature/simple/SimpleFeature.java
aaime: Actually no dangerous, the only issue is with primitive arrays, that were not converted to use varargs

aaime:
@jodygarnett this one should be good to go, I've added a few examples in the upgrade guide.

aaime:
Merging so that I can move on with the other cleanups, happy to handle eventual feedback (possibly on docs, that's the only new part) post-merge.

PR #3319, opened 2021-01-15
[GEOT-6791] Include dependencies in definition of API change when backporting

tbarsballe:
See https://osgeo-org.atlassian.net/browse/GEOT-6791

This PR documents the inclusion of major dependency upgrades in the definition of an API change for the purposes of backporting, to avoid breaking downstream projects when upgrading the minor version of a stable GeoTools branch.

I've also documented that additive only API changes may be documented after one month on master, since as far as I'm aware, this is implicitly permitted, but not documented anywhere in GeoTools (only in GeoServer), and made a few minor improvements to the rest of the "stable branch" documentation.

# Checklist

For all pull requests:

### DIFF
--- a/docs/developer/procedures/stable.rst
+++ b/docs/developer/procedures/stable.rst
@@ -6,25 +6,35 @@ For one we are not adding new features, and we need to back out changes that fai
 
 Let us start with what is restricted:
 
-* Changing **any** API.
-* Breaking any client code that previously worked
+* Changing **any** API. Additive-only API changes may be backported from master after one month, where technically feasible.
+* Breaking any client code that previously worked.
 
 With that in mind here is what we can do on a stable branch:
 
 * applying a fix
 * creating a new plugin
 * create an additional plug-in
+
+
+Applying a fix may also involve upgrading a dependency. When doing so, make sure the API of the dependency has not changed - you should treat an API change in a dependency the same as an API change in GeoTools, as code that depends on your module may also depend upon the funcionality of its dependencies. In some cases, upgrading a dependency to fix a bug may also bring with it API changes in the dependency. Consider whether a backport is really necessary in such cases. Security fixes may need to force an api change.
   
 Since plugins are optional, and the functionality is available through the interfaces of the core library new plugins can be added to the library as there will be required change in client code.
 
 Applying a Fix to the Stable Branch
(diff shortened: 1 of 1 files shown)


docs/developer/procedures/stable.rst
```diff
@@ -39,8 +47,7 @@ Do you have a Jira issue? Chances are any change worth doing on the branch has a
 Applying your Change to master
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-It is nice to do this after every fix, or you may want to save up a couple.
-Please try to apply your fixes to trunk before the next trunk Milestone release (we like to release the best code we can - and that includes your fixes).
+If you've applied a fix to the stable branch, and it can be ported to master, then do so. Please try to apply your fixes to master before the next Release Candidate (we like to release the best code we can - and that includes your fixes).
```
aaime, line 50: Hmm... how one gets to the fix is not the business of our documentation, and agree it's common to reproduce and fix on whatever is the target version first, but mind, PRs are only accepted on master first, and then backported on stable.

tbarsballe: I tried to clean up the existing docs a bit here (they already went into process a bit, but felt outdated). It might make more sense to restructure the order of them so that the "Apply to master" section comes before "Apply to stable" to drive the point home a bit more, but that starts conflicting with the ostensible purpose of the page - "Working on the stable branch".
Do you think changes are needed here?

Also, there are rare cases in which we do accept PRs on a stable branch before master, when a fix cannot be applied to master (either because it was fixed on master by a non-backportable dependency change, or affects a module that no longer exists), but that is nitpicking a bit.

aaime: The cases are there indeed, but very, very rare.

About the contents of the page, I believe we just need to clarify that committig on the master branch is a pre-requisite to merging on the stable branch, unless there is a reason making the commit to master un-needed, or impossible.

tbarsballe: I've striped out some of the process suggestions and made it more clear that you must make a PR against master before stable.
I also tried to make the existing instructions a bit more consistent, and removed several lines that suggested pushing directly to master (replacing those with "open a pull request" instead). Hopefully everything is still pretty clear.

The recommendations around API changes remain the same.

aaime:
Thanks, merged!

PR #3328, opened 2021-02-01
[GEOT-6799] WMTS online tests

roarbra:
This PR tries to fix up in WMTS online tests. Two goals.

- Extend OnelineTestCase and place them in online package
- Split into classes for separate server environments

The first one is to clarify what is online tests, and what isn't. The other one is for running some of the tests, but not necessary all.

# Checklist

For all pull requests:

- [X] The changes are not breaking the build in downstream projects using SNAPSHOT dependencies, GeoWebCache and GeoServer (there is an automatic PR check verifying this, check this when it turns green).


modules/extension/wmts/src/test/java/org/geotools/ows/wmts/online/LantmaterietWMTSServerOnlineTest.java
```diff
@@ -45,22 +40,27 @@
...
+public class LantmaterietWMTSServerOnlineTest extends OnlineTestCase {
 
     private URL server;
 
     private URL serverWithStyle;
 
-    @Before
-    public void setUpInternal() throws MalformedURLException {
+    @Override
+    protected String getFixtureId() {
```
aaime, line 58: So this one needs a property file, but it's not actually going to use it?

modules/extension/wmts/src/test/java/org/geotools/ows/wmts/online/WMTSTransparentTileOnlineTest.java
```diff
@@ -36,30 +38,24 @@
...
-    protected void tearDownInternal() throws Exception {
-        service = null;
+    public void setUpInternal() throws Exception {
+        service = createKVPService();
     }
 
     @Override
     protected String getFixtureId() {
-        return "wmts";
+        return "wmts-unesco";
```
aaime, line 58: Guess this one does not really use the file either?

aaime (commented):
It's probably best if @ianturton and/or @etj have a look at this one, just added a couple of comments.

One thing that I believe is missing, if we actually want the tests to be executed, is a github action that does exactly that ([here is a postgis one as an example](https://github.com/geotools/geotools/blob/master/.github/workflows/postgis_online.yml)). 

Otherwise nobody will actually run them, besides maybe when working on the module.
I guess that these tests will have some tendency to randomly fail, as they depend on external servers that we have no control over.

On Travis we had a way to mark a check as not mandatory... I have not used this before, but it seems there is a "continue-on-fail" setting that might help, described in this issue:
https://github.com/actions/toolkit/issues/399
From what I understand, there are issues if there is a build matrix, but it may work well for an action that has a single build action in it.

roarbra:
My intention was to clean up the tests within gt-wmts. First off was to mark those tests using external servers as so, by extending OnlineTestCase as this seems to be the GeoTools way of doing it as for now.

Next thing I noticed was that if several tests use the same fixtureId, and one of them fails within setUpInternal. Then this function wan't be called for the rest of the tests. Which occasionally made some tests fail at random. The same tests would work fine if started separately. This makes it confusing to work with the test cases.

I see the problem with many property files, but I'm not tempted to change the internals of OnlineTestCase either. The whole matter could maybe be solved differently, and I haven't actually looked into what these tests are testing.

What I do know is that someone, if that is me, @ianturton or @etj , should look into the tests of gt-wmts.

aaime:
@roarbra whether the PR is going to be merged, or not, it's not my call, [I'm not the maintainer nor a developer of the gt-wmts module.](https://github.com/geotools/geotools/blob/master/modules/extension/wmts/pom.xml#L48) 

I'll refrain from adding further comments.

aaime:
Looks like @etj won't be able to review this one for a while, so having a second look and actually trying to run the tests.

The "lantmariet" case does not seem to require a specific url, but just the creation of the property file (an example file is not created either). Is that right?
The actual test run is failing for me (running them with -Ponline from maven):

```
[ERROR]   LantmaterietWMTSServerOnlineTest>OnlineTestCase.run:117->testWMTSCoverageReaderWithStylePlaceholder:193 Images are visibly different, found 520 different pixels, against a threshold of 100
You can add -Dorg.geotools.image.test.interactive=true to show a dialog comparing them (requires GUI support)
```

Comparing with master, it fails the same way there....

The WMTSRestTest example file assumes a "raspberry" URL in the example file (@ianturton is that your doing?), from the looks of the test code it would seem it's looking for a MapProxy deploy? I could be wrong.
I guess the rest case would require a capabilities document that's purely resource oriented... I don't have one, nor the inclination to set up one. Tried giving it the caps document of a embedded GWC, on master (not this branch):

```
[ERROR]   WebMapTileServerOnlineTest>OnlineTestCase.run:117->testIssueGetTileRequestKVP:113->issueGetTileRequest:142 » NullPointer
[ERROR]   WebMapTileServerOnlineTest>OnlineTestCase.run:117->testIssueGetTileRequestREST:119->issueGetTileRequest:142 » NullPointer
[ERROR]   WMTSTransparentTileTest>OnlineTestCase.run:117->testTransparentTile:100 NullPointer
```

Oh well, the online tests are pretty broken to start with, and does not seem this change is making things any worse.

PR #3360, opened 2021-02-18
Update ErrorProne to 2.5.1

aaime:
During the code sprint we have had people having difficulties setting up IntelliJ on Mac, with some issue related to errorprone.
Wondering if updating it to the latest release helps? They seem to have updated the range of Java versions it supports.

This version does not really like code doing object serialization, as that can be leveraged to perform arbitrary code execution.
I've removed a couple of methods that we don't really need, doing it, while suppressed the other cases.

@jodygarnett do you have a way to check if this version of ErrorProne behaves any better on the Mac?

# Checklist

For all pull requests:

- [ ] The changes are not causing two modules to share the same Java packages (to avoid [Java 9+ split package](http://tutorials.jenkov.com/java/modules.html#split-packages-not-allowed) issues)

### DIFF
--- a/modules/library/main/src/main/java/org/geotools/data/Base64.java
+++ b/modules/library/main/src/main/java/org/geotools/data/Base64.java
@@ -404,6 +404,7 @@ public class Base64 {
      * @see Base64#DONT_BREAK_LINES
      * @since 2.0
      */
+    @Deprecated // deserialization is unsafe, so deprecated serialization too
     public static String encodeObject(java.io.Serializable serializableObject, int options) {
         // Streams
         java.io.ByteArrayOutputStream baos = null;
@@ -899,6 +900,8 @@ public class Base64 {
      * @return The decoded and deserialized object
      * @since 1.5
      */
+    @SuppressWarnings("BanSerializableRead")
+    @Deprecated // unsafe
     public static Object decodeToObject(String encodedObject) {
         // Decode and gunzip if necessary
         byte[] objBytes = decode(encodedObject);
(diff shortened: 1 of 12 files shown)


modules/library/main/src/main/java/org/geotools/data/Base64.java
```diff
@@ -368,99 +366,6 @@ private Base64() {}
...
-    /**
-     * Serializes an object and returns the Base64-encoded version of that serialized object. If the
-     * object cannot be serialized or there is another error, the method will return <tt>null</tt>.
-     * The object is not GZip-compressed before being encoded.
-     *
-     * @param serializableObject The object to encode
-     * @return The Base64-encoded object
-     * @since 1.4
-     */
-    public static String encodeObject(java.io.Serializable serializableObject) {
```
jodygarnett, line 380: So these methods are just being removed? No deprecation cycle ...

jodygarnett (approved):
Build checks okay, wanted to ask about the removed Base64 methods.

modules/library/main/src/main/java/org/geotools/data/Base64.java
aaime: Nothing was using them, and they are generally dangerous. This class has been incorportated from some other code base, keeping it as is. I can add them back, deprecate, and annotate them so that PMD ignores them.

aaime: Done as suggested above.

aaime:
Closing, will re-open later with a newer version of error prone and based on up to date code (some classes involved in this one were deprecated and removed)

PR #3370, opened 2021-02-24
[GEOT-6808] -  Http headers set for WMTS isn't used for GetCapabiliti…

roarbra:
The Headers for HTTP must be set during the constructor, because there is a call for GetCapabilities therein.

It will break build downstream, because Geoserver are using a constructor with null for it's third argument. Now we are introducing two constuctors with similar signatures.

# Checklist

For all pull requests:


aaime:
Ok so you are introducing a backwards incompatible change. I don't see an obvious way to avoid it, but since it's a breaking change, it must be reported in the ``upgrade.rst``, just like you did for the HTTPClient.

modules/extension/wmts/src/main/java/org/geotools/ows/wmts/WebMapTileServer.java
```diff
@@ -186,6 +250,11 @@ public WMTSServiceType getType() {
         return type;
     }
 
+    /**
```
aaime, line 253: This should be reported in the upgrade guide too.

aaime:
The rest of the changes looks good to me.

roarbra:
Thanks @aaime ! I've updated the documentation.

docs/user/welcome/upgrade.rst
```diff
@@ -211,6 +211,27 @@ In addition a new plugin ``gt-http-commons`` has been added for MultithreadedHtt
    HTTPClient http = HTTPClientFinder.createClient(hints);
    HTTPResponse response = http.get();
 
+WMTS - WebMapTileServer initialisation
+--------------------------------------
+
+We have introdused a new contructor for the WebMapTileServer.
```
mprins, line 217:
```suggestion
We have introduced a new contructor for the WebMapTileServer.
```

docs/user/welcome/upgrade.rst
```diff
@@ -211,6 +211,27 @@ In addition a new plugin ``gt-http-commons`` has been added for MultithreadedHtt
...
+See list of available constructors:
+
+.. code-block:: java
+
+  public WebMapTileServer(URL serverURL, HTTPClient httpClient)
+  public WebMapTileServer(URL serverURL, HTTPClient httpClient, Map<String, String> headers) // <- NEW CONSTRUCTOR
+  public WebMapTileServer(URL serverURL, HTTPClient httpClient, WMTSCapabilities capabilities)
+  public WebMapTileServer(URL serverURL, HTTPClient httpClient, WMTSCapabilities capabilities, Map<String, Object> hints)
+
+By the same reason we will not allow changes to the headers after initialisation,
```
mprins, line 231:
```suggestion
For the same reason we will not allow changes to the headers after initialisation,
```

aaime:
PR was not compiling due to a collection type mismatch. Stepped in to fix, as @jodygarnett is about to cut 25-RC (any API change past that would go in 26.x by default, to be released in September, unless one can convince the PSC otherwise and backport).

```
diff --git a/modules/extension/wmts/src/test/java/org/geotools/ows/wmts/map/WMTSCoverageReaderTest.java b/modules/extension/wmts/src/test/java/org/geotools/ows/wmts/map/WMTSCoverageReaderTest.java
index 767db6a982..346fa5cddd 100644
--- a/modules/extension/wmts/src/test/java/org/geotools/ows/wmts/map/WMTSCoverageReaderTest.java
+++ b/modules/extension/wmts/src/test/java/org/geotools/ows/wmts/map/WMTSCoverageReaderTest.java
@@ -241,7 +241,7 @@ public class WMTSCoverageReaderTest {
                 templateURL);
     }
 
-    protected List<Tile> testInitMapRequest(WMTSCoverageReader wcr, ReferencedEnvelope bbox)
+    protected Set<Tile> testInitMapRequest(WMTSCoverageReader wcr, ReferencedEnvelope bbox)
             throws Exception {
 
         int width = 400;
```

PR #3403, opened 2021-03-18
[GEOT-6846] MongoDB plugin refact FilterToMongo and FilterSplitter

taba90:
Jira ticket https://osgeo-org.atlassian.net/browse/GEOT-6846
This pull requests 

- introduce an AbstractFilterToMongo to let client that doesn't use a CollectionMapper be able to use the already implemented Filter to MongoQuery mapping.
- made the Mongo filter splitter to be a separate class from being an anonymous class.

AppSchema MongoDB intergration tests GeoServer side have also been runned locally.

# Checklist

For all pull requests:


modules/plugin/mongodb/src/main/java/org/geotools/data/mongodb/MongoFilterSplitter.java
```diff
@@ -0,0 +1,99 @@
...
+                && expression1 instanceof Literal) {
+            preStack.push(filter);
+        }
+    }
+
+    @Override
+    public Object visit(PropertyIsLike filter, Object notUsed) {
+        if (original == null) original = filter;
+
+        if (!fcs.supports(PropertyIsLike.class)) {
```
aaime, line 68: Confused, MongoDB can support PropertyIsLike, or not? How is that working?

taba90: I simply moved the anonymous class to a separe class, I don't have any precise idea of why that if is there. Looking at the comment that is there seems a leftover. Removing it did not caused any test failures, and indeed in the [createFilterCapabilities ](https://github.com/geotools/geotools/blob/1fed12f88ab0c9d4ef92aa9cf73af8f94d5c1291/modules/plugin/mongodb/src/main/java/org/geotools/data/mongodb/MongoDataStore.java#L302) the PropertyIsLike filter is alsways added.

modules/plugin/mongodb/src/main/java/org/geotools/data/mongodb/MongoFilterSplitter.java
```diff
@@ -0,0 +1,99 @@
...
+        Expression expression2 = filter.getExpression2();
+        if ((expression1 instanceof JsonSelectFunction
+                        || expression1 instanceof JsonSelectAllFunction)
+                && expression2 instanceof Literal) {
+            preStack.push(filter);
+        } else if ((expression2 instanceof JsonSelectFunction
+                        || expression2 instanceof JsonSelectAllFunction)
+                && expression1 instanceof Literal) {
+            preStack.push(filter);
+        }
```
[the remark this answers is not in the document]
taba90: fixed

PR #3405, opened 2021-03-18
[GEOT-2972] Character set from CPG files

dvfed:
This PR will add an ability to specify the shapefile's DBF file charset via [CPG](https://en.wikipedia.org/wiki/Shapefile) file.

How it is configured and works:

- `ShapefileDataStore` does not look up CPG file by default. If CPG needed then `tryCPGFile` property must be set.
- `ShapefileDatastoreFactory` checks the parameters given and if no explicit charset provided, it will allow CPG file usage by setting appropriate property on created `ShapefileDataStore`.
- Similarly `DirectoryDataStore` will force using the charset from parameters on all the underlying shapefiles if charset provided, or otherwise each created `ShapefileDataStore` will use its own CPG file that belongs to a particular shapefile.
- If no CPG was found, a charset provided by the store will be used.
- Creating CPG not supported at the moment.

**What can go wrong with existing code that depends on shapefile plugin?**

If user does not provide a charset parameter, assuming the default charset will be used, this is not true anymore if CPG file exists. 
The charset from CPG file will be used instead and this charset must be correct in this case.
I believe there are not many such cases with CPG file containing wrong charset.

Alternatively, we could introduce an explicit `Param` like `USE_CPG_FILE`, but I think using CPG by default is better solution.

# Checklist

For all pull requests:


modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/ShapefileSetManager.java
```diff
@@ -82,13 +91,39 @@ protected DbaseFileReader openDbfReader(boolean indexed) throws IOException {
...
 
+        Charset charset = store.getCharset();
+
+        if (store.isTryCpgFile()
+                && shpFiles.get(CPG) != null
+                && (!shpFiles.isLocal() || (shpFiles.isLocal() && shpFiles.exists(CPG)))) {
+            try (BufferedReader br =
+                    new BufferedReader(
+                            new InputStreamReader(
+                                    shpFiles.getInputStream(CPG, this), "ISO-8859-1"))) {
```
mprins, line 102: better to use  `java.nio.charset.StandardCharsets.ISO_8859_1` instead of a string

mprins (changes requested):
some small remarks

dvfed:
Correct remarks, thanks! Just edited and pushed.
Question: should the documentation be updated somehow if the PR will be accepted?

mprins (approved):
LGTM, I like that it tries to autodetect the charset is none is specified

mprins:
I think a note in the upgrade notes would be nice, but also think this would be a nice enhancement for stable versions (24.x/25.x); 

You could add a line to the API doc for the ShapefileDataStoreFactory that it will try to detect the charset based on the CPG file, and perhaps note on the DBFCHARSET param that setting it will override the autodetect

let' see what other people think.

modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/ShapefileDataStore.java
```diff
@@ -117,6 +117,8 @@
 
     long maxDbfSize = ShapefileFeatureWriter.DEFAULT_MAX_DBF_SIZE;
 
+    private boolean tryCpgFile = false;
```
aaime, line 120: Common Java naming approach is to keep acronyms in all caps. See also: https://stackoverflow.com/questions/2236807/java-naming-convention-with-acronyms#:~:text=From%20sun%20java%20docs%3A,such%20as%20URL%20or%20HTML).

```suggestion
    private boolean tryCPGFile = false;
```

modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/ShapefileDataStore.java
```diff
@@ -219,6 +221,14 @@ void setMaxDbfSize(long maxDbfSize) {
         this.maxDbfSize = maxDbfSize;
     }
 
+    public boolean isTryCpgFile() {
```
aaime, line 224: ShapefileDataStore is really old and many public methods lack Javadoc. That said, recently we are trying to give new public API minimal docs. Could you add javadocs to these new getter and setter?

modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/ShapefileDataStoreFactory.java
```diff
@@ -242,6 +242,9 @@ public DataStore createDataStore(Map<String, ?> params) throws IOException {
             store.setMemoryMapped(useMemoryMappedBuffer);
             store.setBufferCachingEnabled(cacheMemoryMaps);
             store.setCharset(dbfCharset);
+            if (!hasParam(DBFCHARSET, params)) {
```
aaime, line 245: Ah ha, good idea.

modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/ShapefileSetManager.java
```diff
@@ -82,13 +92,39 @@ protected DbaseFileReader openDbfReader(boolean indexed) throws IOException {
...
+                        store.setCharset(charset);
+                    } catch (Exception e) {
+                        if (LOGGER.isLoggable(Level.WARNING)) {
+                            LOGGER.warning(
+                                    "Can't figure out charset from CPG file. Will use provided by the store.");
+                        }
+                    }
+                }
+            } catch (IOException e) {
+                // could happen if cpg file does not exist remotely
```
aaime, line 117: Can you log this failures at a low logging level, like, FINER?

modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/ShapefileSetManager.java
```diff
@@ -82,13 +92,39 @@ protected DbaseFileReader openDbfReader(boolean indexed) throws IOException {
...
+                && shpFiles.get(CPG) != null
+                && (!shpFiles.isLocal() || (shpFiles.isLocal() && shpFiles.exists(CPG)))) {
+            try (BufferedReader br =
+                    new BufferedReader(
+                            new InputStreamReader(
+                                    shpFiles.getInputStream(CPG, this), ISO_8859_1))) {
+                String charsetName;
+                if ((charsetName = br.readLine()) != null) {
+                    try {
+                        charset = Charset.forName(charsetName.trim());
```
aaime, line 107: Ok, so here you're taking a gamble, and hoping the codepage has the same name as a Java charset... I guess they will not always line up, but better than nothing, right? I quickly searched for an existing source of cpg->charset mappings but could not find one. Well, it's called "try CPG" after all :-D

aaime (changes requested):

modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/ShapefileSetManager.java
dvfed: I was thinking about that, did some research, not very deep though, and didn't find any relevant information. `Charset` class itself is smart enough to deals with a number of charset's names and aliases, as javadoc states. If someone will report an exotic mapping we could add it later. ;-)

dvfed:
Fixed, also added javadoc, and a couple of words to upgrade instruction.

aaime (approved):

aaime:
Looking good, merged.

dvfed:
@aaime Is it worth making backport to 24.x/25.x? It seems, there is at least [one vote for](#issuecomment-803180760).

mprins:
@dvfed I posted a message to the dev mailinglist to see if there is any traction for a backport: https://sourceforge.net/p/geotools/mailman/geotools-devel/thread/862f6bcb-3ed0-3168-4be3-d095e90a5cd1%40gmail.com/#msg37260914

PR #3428, opened 2021-03-31
[GEOT-6859] Speed up geotools DbaseFileWriter.FieldFormatter.getFieldString(int size, String s)

gwalbran:
Shapefile data is stored in dbase file format and text columns have a limit of 254 bytes.

getFieldString(int size, String s) is the method responsible for truncating an
input string to one that is small enough to fit into the column.

It does this by putting the string into a string buffer and then removing a character
at a time from the end, until the string fits. But for very long strings this can
take a rather long time. The reason that characters are removed one at a time is that
the limit is in bytes, and many charsets require mutliple bytes per character.

Careful reading of the StringBuffer.replace javadocs says that the sequence (buffer)
will be 'lengthened to accommodate the specified String if necessary' hence the buffer
becoming far bigger than expected.

Solution: use StringBuffer.insert(int dstOffset, CharSequence s, int start, int end) to
set the string into the buffer. This will set no more the maxBytes characters into
the buffer, which will mean that there shouldn't be an excessive amount of characters
to remove.

# Checklist

For all pull requests:

### DIFF
--- a/modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/dbf/DbaseFileWriter.java
+++ b/modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/dbf/DbaseFileWriter.java
@@ -345,27 +345,30 @@ public class DbaseFileWriter implements Closeable {
         }
 
         public String getFieldString(int size, String s) {
-            buffer.replace(0, size, emptyString);
-            buffer.setLength(size);
             // international characters must be accounted for so size != length.
-            if (s != null) {
-                int maxBytes = Math.min(size, s.length());
-                buffer.replace(0, maxBytes, s);
-                int currentBytes = buffer.toString().getBytes(charset).length;
-                if (currentBytes > size) {
-                    while (currentBytes > size) {
-                        int index = buffer.length() - 1;
-                        buffer.deleteCharAt(index);
-                        currentBytes = buffer.toString().getBytes(charset).length;
-                    }
+            if (s == null) {
+                buffer.replace(0, size, emptyString);
+                buffer.setLength(size);
+            } else {
+                buffer.setLength(0);
+
+                // every character is at least one byte. So we start by only adding at most size
+                // characters to the buffer. This is so we don't have too many excess characters
+                // to remove later.
+                int maxChars = Math.min(size, s.length());
+                buffer.insert(0, s, 0, maxChars);
+
+                // we remove characters from the buffer until it contains at most size bytes.
+                while (size < buffer.toString().getBytes(charset).length) {
+                    int index = buffer.length() - 1;
+                    buffer.deleteCharAt(index);
                 }
-                // in sequence to the above step, rather than as an alternative, as the step above
-                // might have removed a last char that was multi-byte, making the buffer short
-                if (currentBytes < size) {
-                    int diff = size - currentBytes;
-                    for (int i = 0; i < diff; i++) {
-                        buffer.append(' ');
-                    }
+
+                // Now we ensure that the buffer is padded with spaces if it is smaller than size.
+                // This could be because the input string isn't as big as size, or that it contains
+                // multi-byte characters that have been removed.
+                while (size > buffer.toString().getBytes(charset).length) {
+                    buffer.append(' ');
                 }
             }
 
--- a/modules/plugin/shapefile/src/test/java/org/geotools/data/shapefile/DbaseFileTest.java
+++ b/modules/plugin/shapefile/src/test/java/org/geotools/data/shapefile/DbaseFileTest.java
@@ -25,6 +25,8 @@ import java.nio.charset.StandardCharsets;
 import java.util.TimeZone;
 import java.util.logging.Level;
 import java.util.logging.Logger;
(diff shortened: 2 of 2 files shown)


modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/dbf/DbaseFileWriter.java
```diff
@@ -350,7 +350,7 @@ public String getFieldString(int size, String s) {
             // international characters must be accounted for so size != length.
             if (s != null) {
                 int maxBytes = Math.min(size, s.length());
-                buffer.replace(0, maxBytes, s);
+                buffer.insert(0, s, 0, maxBytes);
```
aaime, line 353: So the old code was trying to replace the chars between 0 and maxBytes with the target string... and if the string ``s`` was longer, it would have expanded the buffer to fit it fully, e.g.:

```
    public static void main(String[] args) {
        StringBuffer sb = new StringBuffer("abcde");
        sb.replace(1, 3, "A_LONG_STRING");
        System.out.println(sb);
    }
```

prints ``aA_LONG_STRINGde``.  Using insert indeed limits the chars being added, for example the following:

```
    public static void main(String[] args) {
        StringBuffer sb = new StringBuffer("abcde");
        sb.insert(1, "A_LONG_STRING", 0, 2);
        System.out.println(sb);
    }
```

prints ``aA_bcde``.

So it works better for very long strings indeed. But you can see, it's worsening the situation for small strings and for cases there are no multybyte chars in the string, as the above ensures that the resulting string buffer is longer (the string is inserted by moving the chars to the left), while before no reduction would have been needed. 
In the end, the PR is replacing one performance issue with another.

I would suggest to rework this so that the insert is used against an empty buffer, and then:
- if the buffer lenght is smaller, pad the buffer with the emptystring, using another insert (the empty string is guarateed to have single byte chars, so it can be done in a single line)
- if the buffer lenght is greater, apply the same char by char reduction as now

aaime (changes requested):

modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/dbf/DbaseFileWriter.java
gwalbran: Good spotting, I hadn't thought about that.

I've gone back to using the replace, but before that checking the string length and trimming it down if it is too big.

modules/plugin/shapefile/src/test/java/org/geotools/data/shapefile/DbaseFileTest.java
```diff
@@ -186,31 +188,79 @@ public void testEmptyFields() throws Exception {
...
+        assertEquals(sizeInBytes, formattedString.getBytes().length);
+        assertEquals("\u0412\u0412", formattedString);
+
+        // if a multi-byte character is the last to be removed then padding may be required to
+        formattedString = formatter.getFieldString(sizeInBytes, "\u0412A\u0412\u0412");
+        assertEquals(sizeInBytes, formattedString.getBytes().length);
+        assertEquals("\u0412A ", formattedString);
+    }
+
+    @Test(timeout = 10000)
```
aaime, line 245: Please increase the timeout. The GitHub PR checks work in virtualized environments whose performance is highly unpredictable, while we want to avoid a borderline stuck build, we cannot demand a test to "go fast".

modules/plugin/shapefile/src/test/java/org/geotools/data/shapefile/DbaseFileTest.java
```diff
@@ -186,31 +188,79 @@ public void testEmptyFields() throws Exception {
...
+        // build up a very large input string. The test string is also formed so that the 8th char
+        // is also multibyte
+        String test =
+                IntStream.range(0, 100000)
+                        .mapToObj(i -> "\u0412A cat\u0412jumped over the dog")
+                        .collect(Collectors.joining(","));
+
         String formattedString = formatter.getFieldString(8, test);
-        assertEquals("        ".getBytes().length, formattedString.getBytes().length);
+        // for this test we are more concerned with getting the answer in a reasonable time, rather
```
aaime, line 261: But you are testing the actual answer too, so the comment here is confusing, suggest removing it.

aaime (changes requested):
Thanks, the code changes look good.

modules/plugin/shapefile/src/test/java/org/geotools/data/shapefile/DbaseFileTest.java
gwalbran: I'll remove

modules/plugin/shapefile/src/test/java/org/geotools/data/shapefile/DbaseFileTest.java
gwalbran: will remove

aaime (approved):

mprins:
updated jira ticket title to match this PR and set to solve in 26-RC

PR #3429, opened 2021-04-01
[GEOT-6860] SchemaCache doesn't follow redirect when downloading from HttpConnection

taba90:
Jira ticket https://osgeo-org.atlassian.net/browse/GEOT-6860

this pr: 
- makes the SchemaCache able to follow a redirect when downloading;
- fix an https url in the online-test replacing a not working one causing the test to fail

# Checklist

For all pull requests:

### DIFF
--- a/modules/library/xml/src/main/java/org/geotools/xml/resolver/SchemaCache.java
+++ b/modules/library/xml/src/main/java/org/geotools/xml/resolver/SchemaCache.java
@@ -246,7 +246,9 @@ public class SchemaCache {
             connection.setReadTimeout(downloadTimeout);
             connection.setUseCaches(false);
             connection.connect();
-            if (connection.getResponseCode() != HttpURLConnection.HTTP_OK) {
+            int responseCode = connection.getResponseCode();
+            if (responseCode != HttpURLConnection.HTTP_OK) {
+                if (hasRedirect(responseCode)) return followRedirect(connection, blockSize);
                 LOGGER.warning(
                         String.format(
                                 "Unexpected response \"%d %s\" while downloading %s",
@@ -255,8 +257,8 @@ public class SchemaCache {
                                 location.toString()));
                 return null;
             }
-            // read all the blocks into a list
             InputStream input = null;
+            // read all the blocks into a list
             List<byte[]> blocks = new LinkedList<>();
             try {
                 input = connection.getInputStream();
@@ -462,5 +464,30 @@ public class SchemaCache {
         File defaultXmlFile = new File(workspacesDir, "default.xml");
         if (!defaultXmlFile.exists()) return false;
         return true;
+    }
+
+    private static boolean hasRedirect(int responseCode) {
+        return responseCode == HttpURLConnection.HTTP_SEE_OTHER
+                || responseCode == HttpURLConnection.HTTP_MOVED_PERM
+                || responseCode == HttpURLConnection.HTTP_MOVED_TEMP;
+    }
+
+    private static byte[] followRedirect(HttpURLConnection connection, int blockSize) {
+        String redirect = connection.getHeaderField("Location");
+        byte[] result = null;
+        if (redirect == null) {
+            LOGGER.warning("Tried to follow redirect but no url was provided in Location header");
+        } else {
+            try {
+                URI redirectURL = new URI(redirect);
+                LOGGER.info("Following redirect to " + redirect);
+                result = download(redirectURL, blockSize);
+            } catch (URISyntaxException uri) {
+                LOGGER.warning(
+                        "Tried to follow redirect but invalid url was provided in Location header: "
+                                + redirect);
+            }
+        }
+        return result;
     }
 }
--- a/modules/library/xml/src/main/java/org/geotools/xml/resolver/SchemaCache.java
+++ b/modules/library/xml/src/main/java/org/geotools/xml/resolver/SchemaCache.java
@@ -109,6 +109,10 @@ public class SchemaCache {
         }
     }
 
+    /** Max number of attempts to follow a redirect to avoid redirect loop * */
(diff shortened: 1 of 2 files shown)


modules/library/xml/src/main/java/org/geotools/xml/resolver/SchemaCache.java
```diff
@@ -463,4 +465,17 @@ static boolean isSuitableDirectoryToContainCache(File directory) {
...
+    private static boolean hasRedirect(int responseCode) {
+        return responseCode == HttpURLConnection.HTTP_SEE_OTHER
+                || responseCode == HttpURLConnection.HTTP_MOVED_PERM
+                || responseCode == HttpURLConnection.HTTP_MOVED_TEMP;
+    }
+
+    private static byte[] followRedirect(HttpURLConnection connection, int blockSize)
+            throws URISyntaxException {
+        String redirect = connection.getHeaderField("Location");
+        URI redirectURL = new URI(redirect);
```
nmco, line 478: I would add a log message here at INFO level.

modules/library/xml/src/test/java/org/geotools/xml/resolver/SchemaCacheOnlineTest.java
```diff
@@ -119,16 +121,16 @@ public void downloadHttps() throws Exception {
...
     private void check(byte[] bytes) {
-        Assert.assertTrue(bytes.length > 0);
+        assertTrue(bytes.length > 0);
         String text = new String(bytes);
-        Assert.assertTrue(text.contains("GeoSciML"));
-        Assert.assertTrue(text.contains("<schema"));
-        Assert.assertTrue(text.contains("</schema>"));
+        assertTrue(text.contains("GeoSciML"));
+        assertTrue(text.contains("<schema"));
+        assertTrue(text.contains("</schema>"));
```
nmco, line 128: Looks like these are unrelated changes done by the IDE.

nmco:
@taba90 should not the schema cache by delegating HTTP requests on GeoTools HTTP client? Then the HTTP client would be the one taking care of the redirect, fixing this issue for all the code that needs to performs an HTTP request, e.g. WMS client, WFS client, etc ...?

modules/library/xml/src/main/java/org/geotools/xml/resolver/SchemaCache.java
taba90: fixed. Added an additional null check also

modules/library/xml/src/test/java/org/geotools/xml/resolver/SchemaCacheOnlineTest.java
taba90: fixed

modules/library/xml/src/main/java/org/geotools/xml/resolver/SchemaCache.java
```diff
@@ -246,7 +246,9 @@ static void store(File file, byte[] bytes) {
             connection.setReadTimeout(downloadTimeout);
             connection.setUseCaches(false);
             connection.connect();
-            if (connection.getResponseCode() != HttpURLConnection.HTTP_OK) {
+            int responseCode = connection.getResponseCode();
+            if (responseCode != HttpURLConnection.HTTP_OK) {
+                if (hasRedirect(responseCode)) return followRedirect(connection, blockSize);
```
aaime, line 251: While not very common, it's possible that the second response is another redirect. And so on.
Browsers follow them, but have some limits to avoid being caught in a redirect loop, e.g.:
https://www.liquidweb.com/kb/troubleshooting-too-many-redirects/

taba90: fixed. Seemed reasonable to me to put a max number of redirect of 5.

taba90:
> @taba90 should not the schema cache by delegating HTTP requests on GeoTools HTTP client? Then the HTTP client would be the one taking care of the redirect, fixing this issue for all the code that needs to performs an HTTP request, e.g. WMS client, WFS client, etc ...?

thanks for your point @nmco. The SchemaCache seems to me the correct place to follow redirect, as long as it is it taking care of downloading the xsd, that looks to me more a "low level" operation that the HTTPClient of WMS WFS ecc, should no care about as it is meant more to be used OGC service specific request.

taba90:
@nmco @aaime thanks for your reviews. Your feedbacks have been applied.

modules/library/xml/src/main/java/org/geotools/xml/resolver/SchemaCache.java
```diff
@@ -463,4 +475,31 @@ static boolean isSuitableDirectoryToContainCache(File directory) {
...
+            try {
+                URI redirectURL = new URI(redirect);
+                LOGGER.info("Following redirect to " + redirect);
+                result = download(redirectURL, blockSize, redirectionCount);
+            } catch (URISyntaxException uri) {
+                LOGGER.warning(
+                        "Tried to follow redirect but invalid url was provided in Location header: "
+                                + redirect);
+            }
+        }
```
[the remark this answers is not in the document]
taba90: fixed

modules/library/xml/src/main/java/org/geotools/xml/resolver/SchemaCache.java
```diff
@@ -97,6 +97,9 @@
     /** Default download timeout. Change it with -Dschema.cache.download.timeout=<milliseconds> */
     private static int downloadTimeout = 60000;
 
+    /** Max number of attempts to follow a redirect to avoid redirect loop * */
+    private static final int MAX_FOLLOW_REDIRECT = 5;
```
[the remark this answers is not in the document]
taba90: fixed

PR #3444, opened 2021-04-05
[GEOT-6717] IndexOutOfBoundsException at CapabilitiesFilterSplitter.visit(Function,...)

groldan:
Fix for an `IndexOutOfBoundsException` thrown at
`CapabilitiesFilterSplitter.visit(Function,...)` when the post-stack
size is bigger than the number of function arguments.

---

# Checklist

For all pull requests:

### DIFF
--- a/modules/library/main/src/main/java/org/geotools/filter/visitor/CapabilitiesFilterSplitter.java
+++ b/modules/library/main/src/main/java/org/geotools/filter/visitor/CapabilitiesFilterSplitter.java
@@ -939,33 +939,36 @@ public class CapabilitiesFilterSplitter implements FilterVisitor, ExpressionVisi
 
     /** @see org.geotools.filter.FilterVisitor#visit(org.geotools.filter.FunctionExpression) */
     @Override
-    public Object visit(Function expression, Object notUsed) {
-        if (!fcs.fullySupports(expression)) {
-            postStack.push(expression);
+    public Object visit(Function function, Object notUsed) {
+        if (!fcs.fullySupports(function)) {
+            postStack.push(function);
             return null;
         }
 
-        if (expression.getName() == null) {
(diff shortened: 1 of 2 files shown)


aaime:
@groldan I have no experience with this class, from a quick check it's used only by wfs-ng and app-schema.
It's a pity that it was contributed in such a poor way, dropped there, deprecated the other splitter, but not actually used in any store. Nowadays that would be hard, to deprecate a class one needs to remove all calling points and replace with the non deprecated version (or suppress all usages of the old one, something we'd be against, part of preparing a new non deprecated class is to ensure it's actually working and the migration process is manageable).

jodygarnett (approved):
This looks good, minor feedback on using `POST_SIZE` and `PRE_SIZE` as they are final.

groldan:
> @groldan I have no experience with this class, from a quick check it's used only by wfs-ng and app-schema.
> It's a pity that it was contributed in such a poor way, dropped there, deprecated the other splitter, but not actually used in any store. Nowadays that would be hard, to deprecate a class one needs to remove all calling points and replace with the non deprecated version (or suppress all usages of the old one, something we'd be against, part of preparing a new non deprecated class is to ensure it's actually working and the migration process is manageable).

Hey Andrea, it was a silly mistake, easily avoidable if using variable names that make sense instead of i,j,k. Can happen to anyone back in 2011 :). Agreed on a better process to deprecate a class though.

It's actually used by `jdbcconfig`, and the reason why I had to duplicate the `ConfigDatabase` class in geoserver-cloud here: https://github.com/geoserver/geoserver-cloud/blob/480afd80/starters/catalog-backend-starter/src/main/java/org/geoserver/cloud/config/jdbcconfig/CloudJdbcConfigDatabase.java#L62

Thanks for the review @jodygarnett

PR #3475, opened 2021-04-26
[GEOT-6876] Check for limit or offset in isFullBoundsQuery

GeorgeDewar:
This PR resolves GEOT-6876 by ensuring that isFullBoundsQuery returns false if the query contains a limit or an offset.

If isFullBoundsQuery returns true, then GeoTools will use the dialect's getOptimizedBounds to get the bounds of the whole table, and return this. Where there is a limit or offset in the query, the bounds of the whole table is not the correct response.

I have added tests to cover the cases that have been fixed. In the PostgreSQL plugin, where estimated extents are disabled in the tests by default, I had to add additional tests that enable the estimated extents feature and call these tests.

And, I would love to hear if anyone knows why the ability to turn on and off estimated extents in PostGIS is there at all (i.e. why would anybody ever turn it off). Is there a chance it's actually just to support the tests? If that is the case, we could remove it as it is no longer required for the tests to pass.

# Checklist

For all pull requests:

### DIFF
--- a/modules/plugin/jdbc/jdbc-postgis/src/test/java/org/geotools/data/postgis/PostGISFeatureSourceOnlineTest.java
+++ b/modules/plugin/jdbc/jdbc-postgis/src/test/java/org/geotools/data/postgis/PostGISFeatureSourceOnlineTest.java
@@ -108,6 +108,16 @@ public class PostGISFeatureSourceOnlineTest extends JDBCFeatureSourceOnlineTest 
         assertTrue(areCRSEqual(CRS.decode("EPSG:4326"), bounds.getCoordinateReferenceSystem()));
     }
 
+    public void testEstimatedBoundsWithLimit() throws Exception {
+        ((PostGISDialect) dataStore.getSQLDialect()).setEstimatedExtentsEnabled(true);
+        super.testBoundsWithLimit();
+    }
+
+    public void testEstimatedBoundsWithOffset() throws Exception {
+        ((PostGISDialect) dataStore.getSQLDialect()).setEstimatedExtentsEnabled(true);
+        super.testBoundsWithOffset();
+    }
+
     public void testSridFirstGeometry() throws Exception {
         SimpleFeatureType schema = dataStore.getSchema(tname("ft3"));
         GeometryDescriptor gd = schema.getGeometryDescriptor();
--- a/modules/plugin/jdbc/jdbc-postgis/src/test/java/org/geotools/data/postgis/PostGISFeatureSourceOnlineTest.java
+++ b/modules/plugin/jdbc/jdbc-postgis/src/test/java/org/geotools/data/postgis/PostGISFeatureSourceOnlineTest.java
@@ -75,37 +75,12 @@ public class PostGISFeatureSourceOnlineTest extends JDBCFeatureSourceOnlineTest 
         assertTrue(areCRSEqual(CRS.decode("EPSG:4326"), bounds.getCoordinateReferenceSystem()));
     }
 
-    public void testEstimatedBounds() throws Exception {
-        // enable fast bbox
-        ((PostGISDialect) dataStore.getSQLDialect()).setEstimatedExtentsEnabled(true);
+    // Exercise the getBounds code path where getOptimizedBounds is disabled
+    public void testNonEstimatedBounds() throws Exception {
+        // disabled fast bbox
+        ((PostGISDialect) dataStore.getSQLDialect()).setEstimatedExtentsEnabled(false);
 
-        ReferencedEnvelope bounds = dataStore.getFeatureSource(tname("ft1")).getBounds();
-        assertEquals(0l, Math.round(bounds.getMinX()));
-        assertEquals(0l, Math.round(bounds.getMinY()));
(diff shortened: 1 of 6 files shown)


modules/plugin/jdbc/jdbc-postgis/src/test/java/org/geotools/data/postgis/PostGISFeatureSourceOnlineTest.java
```diff
@@ -75,37 +75,12 @@ public void testBBOXOverlapsEncoding() throws Exception {
         assertTrue(areCRSEqual(CRS.decode("EPSG:4326"), bounds.getCoordinateReferenceSystem()));
     }
 
-    public void testEstimatedBounds() throws Exception {
```
mprins, line 78: maybe overlooking something, but why remove testing with estimated bounds?

aaime: Indeed must not be removed. I also don't see new tests for the fix.

GeorgeDewar: As far as I can tell, this test existed because estimated bounds was disabled by default in these tests for PostgreSQL (because "the tests assume non estimated extents", line 57 of PostGISTestSetup.java). But with this fix, disabling the feature for the tests to work is not required.

That means, now the testBounds test in the superclass is testing what was being tested by testEstimatedBounds, and the same for testEstimatedBoundsWithQuery (now achieved by testBoundsWithQuery).

That is also why you do not see new tests for the fix. The tests for this functionality already exist - they were just being bypassed before by setting setEstimatedExtentsEnabled to false.

GeorgeDewar: If you are not comfortable with that or not sure that my understanding is correct (it might not be, after all), it is possible to retain all the existing tests, keep setEstimatedExtentsEnabled false by default, and add a new test similar to testEstimatedBounds which turns it on and calls through to testGetFeaturesWithOffset to test the new functionality. That would be a smaller and simpler change.

GeorgeDewar:
It looks like the Oracle tests are failing for a build-related reason I don't understand. But I can see that the Linux QA tests are failing due to a formatting error so I will look into that.

modules/plugin/jdbc/jdbc-postgis/src/test/java/org/geotools/data/postgis/PostGISFeatureSourceOnlineTest.java
aaime: Yes please :-D

GeorgeDewar: Done!

modules/plugin/jdbc/jdbc-postgis/src/test/java/org/geotools/data/postgis/PostGISTestSetup.java
```diff
@@ -55,6 +55,8 @@ protected void setUpDataStore(JDBCDataStore dataStore) {
         ((PostGISDialect) dataStore.getSQLDialect()).setLooseBBOXEnabled(false);
 
         // the tests assume non estimated extents
+        // Note: this was due to GEOT-6876, and the tests will now also pass with estimated extents
```
aaime, line 58: How did you come to this factually incorrect, misleading conclusion? 

This line of code has been there since the beginning of the github history, so you did not get it from it.
As you probably noticed, the JDBC tests are largely shared across all databases, thus, bound calculation tests are written to target precise, feature by feature aggregations, which can be performed, one way or the other, in all databases.

Extent _estimation_ enables a trade off, a fast but potentially incorrect calculation (which is fine for common usage, hence it being the default).  For example PostGIS computes it using [ST_Estimated_Extent](http://www.postgis.org/docs/ST_Estimated_Extent.html) which uses statistics to instrantly return a value.
This can result in calculated bounds having different values across databases. Hence, two sets of tests are needed, the shared ones, using a database aggregation function, and the databse specific ones, only for stores that support estimated extents, with results that are unique to them.

Feel free to improve the comment if you want, providing a longer explanation of why estimated extents are turned off, but remove the one you added, a relationship with GEOT-6876 is accidental at best.

GeorgeDewar: Thank you @aaime for the great explanation. I apologize for the misunderstanding. I came to that conclusion because all of the PostGIS tests still pass with estimated extents turned on, when the fix is in place.

I think in this instance it's best that I remove the comment, lest I write something else misguided in an attempt to improve it.

aaime:
Thanks, squash-merged. If you need the fix on 25.x and 24.x please create backport PRs.

GeorgeDewar:
Thanks @aaime!

PR #3497, opened 2021-05-05
[25.x backport] [GEOT-2972] Character set from CPG files

dvfed:
[GEOT-2972](https://osgeo-org.atlassian.net/browse/GEOT-2972) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

Backport of #3405.

There was small discusion on backporting in [geotools-devel list.](https://sourceforge.net/p/geotools/mailman/geotools-devel/thread/862f6bcb-3ed0-3168-4be3-d095e90a5cd1%40gmail.com/#msg37260914)
I am late to discusion but still: IMO there is no need for system variable "switch" that gets old behavior back.
If a client does not provide a charset, relying on datastore's default, and experiences some trouble with CPG, she could just remove these files. There is no reason to keep CPG files which are simply unused as in an existing plugin version, and moreover which contain wrong charset.

Not sure if upgrade notes should be moved / copied to "25.x" section of `upgrade.rst` in case of backporting from "main" (actually, from "26.x"). I've copied upgrade notes to the "25.x" section, and even added some more words to help plugin's clients avoid potential issues with wrong charset.
Please let me know what also (instead of) should be done.

# Checklist

For all pull requests:


mprins:
> IMO there is no need for system variable "switch" that gets old behavior back.

I agree, but I'm not the boss.

Any existing code already setting the codepage is unaffected, any code relying on using default codepage where the sidecar file specifies different than default relies on chance (we actually have an application that suffers from this that would improve with this PR), the case of a missing sidecar file is also unaffected as the default is still the same.

having a note somewhere on the release notes is good

aaime:
I disagree on not needing a system variable. I'm also not "the boss", just one of PMC members, others opinion might differ, but I won't merge it in the current state.

Scenario:
- A user has a product based on GeoTools 25.x, say GeoServer. 
- They have shapefiles that work as expected
- In order to get a bug fix, they upgrade GeoServer to the next bugfix release on the stable series (one where we promise to maintain backwards compatibility)
- All of a sudden, shapefiles stop working, he has no idea why.

How can that happen? Maybe the CPG files have faulty entry (there are lots of .prj files with wrong information, would not be surprised if the same happened to cpg files), or maybe our assumption that the cpg file match a Charset is not warranted for all cases, and something unforeseen happens (a cpg spec matches a java charset, but it's not the one we expect).

Yes, the data is broken, but our general take is to be tolerant to broken data whenever possible, and we try very hard to preserve backwards compatibility.
The user could go and set a working charset on each shapefile... that could take a lot of time, depending on how many shapefiles are there (the REST API can be used to automate that, writing the script also takes time).

Adding a simple system variable here takes 5 minutes, and will allow oddball cases of backwards incompatibilty in minutes for users too. Please add it? :-D

dvfed:
I will fix the PR as soon as I get back to work.

Mostly I use the shape plugin in a custom format-to-format transformation
tool, so I used to deal with broken data. This is not the case for
Geoserver user, the scenario above shows it clearly.

Now I even think would it be better turn auto cpg detection off by default
on backport. Reading shape in wrong charset is a problem, but writing could
be even worse.

aaime:
@dvfed uh did not consider the write side but you're right, it might break other tools that are assuming the CPG file contains a valid codepage. This also affects the main branch though, no?

dvfed:
@aaime yes, main branch affected too, changes has been merged already.

That the other reason why I tried to apply same code for a backport. The
changes seemed harmless, but now they don't.

aaime:
@dvfed you're stil in time to make changes to main branch as well, it won't be released to the public until September 2021.

dvfed:
Fixed the PR. What have done here: introduced a system property that enables CPG file usage. A charset autodetection is **disabled** by default.

@aaime For the main branch i would do the same, but rather **enable** autodetection by default. In case of trouble a client could use the property to turn autodetection off, bringing old behavior back. Basically, that is the approach we planned for the backport itself. 

Thus we'll achieve the goal - to simplify DBF charset manipulation: get rid of explicit charset setting in favor of CPG sidecar, less parameters, less system properties. After all, the release of a new major version seems the best time to introduce some breaking change?

aaime:
That all sounds pretty reasonable, yep. Thanks!

PR #3499, opened 2021-05-06
[GEOT-6886] Raster to Vector rendering transformations with reprojection can generate incorrect images

sikeoka:
[GEOT-6886](https://osgeo-org.atlassian.net/browse/GEOT-6886) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

This PR moves the projection handling code for raster to vector rendering transformations from the plain renderer into a separate method that is called by both the plain renderer and the optimized renderer.
https://osgeo-org.atlassian.net/browse/GEOT-6886

# Checklist

For all pull requests:

- [X] The changes are not breaking the build in downstream projects using SNAPSHOT dependencies, GeoWebCache and GeoServer (there is an automatic PR check verifying this, check this when it turns green).

- [X] This PR passes all existing unit tests (test results will be reported by Continuous Integration after opening this PR)
- [X] This PR passes the [QA checks](https://docs.geotools.org/latest/developer/conventions/code/qa.html) (QA checks results will be reported by Continuous Integration after opening this PR)

### DIFF
--- a/modules/library/render/src/main/java/org/geotools/renderer/lite/StreamingRenderer.java
+++ b/modules/library/render/src/main/java/org/geotools/renderer/lite/StreamingRenderer.java
@@ -2654,55 +2654,8 @@ public class StreamingRenderer implements GTRenderer {
                 }
                 boolean cloningRequired = isCloningRequired(lfts);
                 RenderableFeature rf = createRenderableFeature(layerId, cloningRequired);
-                rf.layer = liteFeatureTypeStyle.layer;
-
-                ProjectionHandler handler = liteFeatureTypeStyle.projectionHandler;
-                // Check if a reprojection has been made, in that case, let's update the projection
-                // Handler
-                CoordinateReferenceSystem featureCrs =
-                        features.getSchema().getCoordinateReferenceSystem();
-                ScreenMap screenMap = liteFeatureTypeStyle.screenMap;
-                if (handler != null
-                        && featureCrs != null
-                        && !CRS.equalsIgnoreMetadata(handler.getSourceCRS(), featureCrs)) {
-                    try {
-                        handler =
-                                ProjectionHandlerFinder.getHandler(
-                                        mapExtent, featureCrs, isMapWrappingEnabled());
-                        if (screenMap != null) {
-                            Envelope mapArea = mapExtent;
-                            if (getRenderingBuffer() == 0) {
-                                int metaBuffer = findRenderingBuffer(lfts);
-                                if (metaBuffer > 0) {
-                                    mapArea =
-                                            expandEnvelope(
-                                                    mapArea, worldToScreenTransform, metaBuffer);
-                                }
-                            }
-                            ReferencedEnvelope envelope =
-                                    expandEnvelopeByTransformations(
-                                            lfts, new ReferencedEnvelope(mapArea, destinationCrs));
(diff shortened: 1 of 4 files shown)


modules/library/render/src/main/java/org/geotools/renderer/lite/StreamingRenderer.java
```diff
@@ -2757,13 +2710,25 @@ private void drawOptimized(
             // one is there to make sure a single feature error does not ruin the rendering
             // (best effort) whilst an exception in hasNext() + ignoring catch results in
             // an infinite loop
+            boolean firstFeature = true;
+            ProjectionHandler[] handlers = new ProjectionHandler[lfts.size()];
```
aaime, line 2714: I don't understand why projection handlers and screenmaps are accumulated in arrays: they are not used after the loop, nor across iterations of the loop. Seems a simple local variable in the loop would also work.

modules/unsupported/process-raster/src/test/java/org/geotools/process/raster/ContourProcessTest.java
```diff
@@ -244,24 +244,34 @@ private GridCoverage2D createVerticalGradient(
...
 
-    private int countWhiteSamples(String styleName, int width, int height) throws Exception {
+    @Test
+    public void testContourWithTransformAndReprojection_WithoutOpacityMultiFTS() throws Exception {
+        // The style with multiple FeatureTypeStyles will use optimized rendering.
+        // The style without opacity will use the screenmap.
+        countWhiteSamples("contour_multi_fts_without_opacity.sld");
+    }
+
+    private void countWhiteSamples(String styleName) throws Exception {
```
aaime, line 274: Love the reuse...  but the method is now asserting something, not just counting white pixels. This sounds like a better name (open to suggestions):

```suggestion
    private void assertWhiteSamples(String styleName) throws Exception {
```

aaime (changes requested):

modules/library/render/src/main/java/org/geotools/renderer/lite/StreamingRenderer.java
sikeoka: It is so that the projection handler and screenmaps only have to be calculated for the first feature in the loop and then they are reused for the rest of the features.  It seemed to me like this caching could be useful if there were a lot of features.

aaime: Oh, now I see it. Yes, it makes sense. Maybe add a comment in that respect before the arrays and firstfeature? :-D

modules/unsupported/process-raster/src/test/java/org/geotools/process/raster/ContourProcessTest.java
sikeoka: I made this change.

modules/library/render/src/main/java/org/geotools/renderer/lite/StreamingRenderer.java
sikeoka: I added a comment here.

aaime:
Merged thanks. Backport as you see fit.

PR #3506, opened 2021-05-10
[GEOT-6889] Configuration of JNDI and exception handling

roarbra:
[GEOT-6889](https://osgeo-org.atlassian.net/browse/GEOT-6889) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

The function getInitialContext threw an IllegalArgumentException that caller didn't catch. The functionality behind new InitialContext is rather mysterious, so I've tried to log those variables that I've seen influence it. That might change by time.

Lastly the argument hints wasn't used, and probably that was the whole problem for me.

# Checklist

For all pull requests:

### DIFF
--- a/modules/library/jdbc/src/main/java/org/geotools/data/jdbc/datasource/JNDIDataSourceFactory.java
+++ b/modules/library/jdbc/src/main/java/org/geotools/data/jdbc/datasource/JNDIDataSourceFactory.java
@@ -64,8 +64,7 @@ public class JNDIDataSourceFactory extends AbstractDataSourceFactorySpi {
     public DataSource createNewDataSource(Map<String, ?> params) throws IOException {
         String refName = (String) JNDI_REFNAME.lookUp(params);
         try {
-            return (DataSource)
-                    GeoTools.getInitialContext(GeoTools.getDefaultHints()).lookup(refName);
+            return (DataSource) GeoTools.getInitialContext().lookup(refName);
         } catch (Exception e) {
             throw new DataSourceException("Could not find the specified data source in JNDI", e);
         }
@@ -85,7 +84,7 @@ public class JNDIDataSourceFactory extends AbstractDataSourceFactorySpi {
     @Override
     public boolean isAvailable() {
         try {
-            GeoTools.getInitialContext(GeoTools.getDefaultHints());
+            GeoTools.getInitialContext();
             return true;
         } catch (Exception e) {
             return false;
--- a/modules/library/metadata/src/main/java/org/geotools/util/factory/GeoTools.java
+++ b/modules/library/metadata/src/main/java/org/geotools/util/factory/GeoTools.java
@@ -26,10 +26,14 @@ import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Modifier;
 import java.net.MalformedURLException;
 import java.net.URL;
+import java.security.AccessController;
+import java.security.PrivilegedAction;
 import java.util.ArrayList;
 import java.util.Collections;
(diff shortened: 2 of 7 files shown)


aaime:
This looks... weird. Hints is designed to contain tons of little things that control the behavior of GeoTools, as one can see from the plenty of [statically defined keys](https://github.com/geotools/geotools/blob/main/modules/library/metadata/src/main/java/org/geotools/util/factory/Hints.java#L72) in the class. Also keys can only be subclasses or ``RenderingHints.Key``, which is another odd thing to have in an InitialContext... actually it seems broken, since ``InitialContext.lookup(Name)`` wants a ``java.naming.Name`` as the lookup key, while ``RenderingHints.Key`` does not even have a name.. not sure how the two can be reconciled.

Hints are not used, but if they were, I would just look into the Hints for special keys controlling the initial context creation, like which implementation to use (subclasses are possible), and maybe have a key for its initial contents.

roarbra:
I've always thought of those hints acting as a String within the Map, but I see now that it isn't.
Since that argument wasn't used, I took the liberty to introduce a function without arguments. That should save someone else from frustration, but I don't know what happened to the original bug.

aaime:
@jodygarnett any opinion on this one?

modules/library/metadata/src/main/java/org/geotools/util/factory/GeoTools.java
```diff
@@ -938,15 +941,109 @@ public static EntityResolver getEntityResolver(Hints hints) {
...
+            if (javaHome != null) {
+                sb.append("Or from a file specified by system property java.home:\n")
+                        .append(javaHome)
+                        .append("\n");
+            }
+            LOGGER.log(Level.WARNING, sb.toString());
+        }
+        if (e instanceof NamingException) {
+            return (NamingException) e;
+        } else {
```
groldan, line 1040: Unnecessary else clause.

modules/library/jdbc/src/main/java/org/geotools/data/jdbc/datasource/JNDIDataSourceFactory.java
```diff
@@ -64,8 +64,7 @@ public boolean canProcess(Map<String, ?> params) {
     public DataSource createNewDataSource(Map<String, ?> params) throws IOException {
         String refName = (String) JNDI_REFNAME.lookUp(params);
         try {
-            return (DataSource)
-                    GeoTools.getInitialContext(GeoTools.getDefaultHints()).lookup(refName);
+            return (DataSource) GeoTools.getInitialContext().lookup(refName);
         } catch (Exception e) {
             throw new DataSourceException("Could not find the specified data source in JNDI", e);
         }
```
groldan, line 70: Aside, if `getInitialContext()` declares it throws a `NamingException`, that's what should be catch. Any other thing will be a `RuntimeException` and hence totally unrelated. The message in the `DataSourceException` will be misleading, a runtime exception indicates some other kind of problem and needs a different treatment.

roarbra: I totally agree, but I'm not sure if I should temper this one at the moment. My only change is to remove the hints argument.

modules/library/metadata/src/main/java/org/geotools/util/factory/GeoTools.java
roarbra: That's ok.

PR #3507, opened 2021-05-10
[GEOT-6890] Move StAX GML parser from wfs-ng to a new module

matthijsln:
[GEOT-6890](https://osgeo-org.atlassian.net/browse/GEOT-6890) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

The internal GML StAX parser from the wfs-ng module is faster and simpler than the gt-xsd based ones. It is useful to promote this to a public API.

See the [geotools-devel](https://sourceforge.net/p/geotools/mailman/geotools-devel/thread/6b5af809-e71f-e9db-f595-7def3d5156a5%40b3partners.nl/#msg37276548) mailing list messages.

This module moves the duplicated `parseGeom()` methods from [XmlSimpleFeatureParser](https://github.com/geotools/geotools/blob/main/modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/internal/parsers/XmlSimpleFeatureParser.java#L248) and the [XmlFeatureParser](https://github.com/geotools/geotools/blob/main/modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/internal/parsers/XmlFeatureParser.java#L179) (where `XmlComplexFeatureParser` descends from) to a separate module named `gml-streaming`. These have already diverged because #3393 added MultiCurve support to one but not the other, and only one supports coordinate inversion based on CRS.

Some GML testcases have been added for a test coverage of 77%, this is of course not an exhaustive test suite of the entie GML schema.

**Note: the new module is already included in the top-level POM, because the code was moved from `wfs-ng` and in this PR now depends on it.**

Some questions:
- Commits are currently not squashed, is it better to do so?
- ~~The new module has a dependency on gt-xsd-gml3, but this is only required for the GML element names. Should it be removed and the constants added to this module?~~

# Checklist

For all pull requests:


matthijsln:
I did not squash the commits yet so the changes I made after copying the parseGeom() and related methods are easier to follow.

mprins (approved):
LGTM.

We can squash + merge, not really needed to do this in the PR.

@matthijsln is covered by the @B3Partners CLA

modules/unsupported/gml-streaming/pom.xml
```diff
@@ -0,0 +1,23 @@
...
+  <parent>
+    <groupId>org.geotools</groupId>
+    <artifactId>unsupported</artifactId>
+    <version>26-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>gt-gml-streaming</artifactId>
+  <packaging>jar</packaging>
+  <name>GML XML support using Streaming API for XML (StAX)</name>
+  <description>Efficiently read GML elements using StAX in a streaming manner. No writing.</description>
```
aaime, line 14: At first I though the whole feature reading was moved, but I see only gml geometries parsing is implemneted?
GML is a huge spec, covers even more than features, but it's commonly associated to features.

The module description should be more specific and clarify only geometry parsing is supported. If there is no intention to parse features and feature collections before graduating it, the name of the module should also be changed, as the focus is very specific, unlike the more complex parsers available in GeoTools.

I guess it's also version specific, as geometry encoding is the part that changed over time. 
A clearer name might be ``gt-gml3-geometry-streaming``, if that's the ambition of the module.

modules/unsupported/gml-streaming/src/main/java/org/geotools/gml/stream/GML.java
```diff
@@ -0,0 +1,54 @@
...
+ * This library is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without
+ * even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ */
+
+package org.geotools.gml.stream;
+
+import javax.xml.namespace.QName;
+
+/** Only the constants from the GML schema which are used by XmlStreamGeometryReader. */
```
aaime, line 20:
```suggestion
/** Names of GML elements parseable with XmlStreamGeometryReader. */
```

modules/unsupported/gml-streaming/src/main/java/org/geotools/gml/stream/XmlStreamGeometryReader.java
```diff
@@ -0,0 +1,764 @@
...
+import org.locationtech.jts.geom.LinearRing;
+import org.locationtech.jts.geom.MultiLineString;
+import org.locationtech.jts.geom.Point;
+import org.locationtech.jts.geom.Polygon;
+import org.opengis.referencing.FactoryException;
+import org.opengis.referencing.NoSuchAuthorityCodeException;
+import org.opengis.referencing.crs.CoordinateReferenceSystem;
+
+/**
+ * Parse GML from a StAX XMLStreamReader. Code migrated from the XmlSimpleFeatureParser class from
```
aaime, line 47: Don't think users will care about the history of where this came from.

```suggestion
 * Parse GML 3 geometries from a StAX XMLStreamReader.
```

aaime (commented):
Looks good to me, believe the focus of the module should be clarified in name and description. See comments below.

modules/unsupported/gml-streaming/pom.xml
matthijsln: Good points, I will clarify it's only parsing geometries. I do not have the intention to add feature and featureCollection parsing to it and definitely not general application schemas. This would keep this module small and simple but still useful. I'll see if I can write some user guide documentation to clarify the intention and when this parser can be used.

Not sure if the '3' suffix is needed in `gt-gml3-geometry-streaming`, because for the elements it supports it will parse GML 2.x elements which are deprecated in GML 3.x but retained for backwards compatibility. A clearer distinction would be GML 3.2 which has a different namespace. I'll have to look into the specific geometry element differences, but I think this parser could support both the `http://www.opengis.net/gml` and `http://www.opengis.net/gml/3.2` namespaces.

aaime: Good point indeed, no version number needed if it can succesfully parse geometries from all GML versions.

aaime:
Squash-merged, thanks!

PR #3516, opened 2021-05-17
[GEOT-6892] ExpressionDomParser early NPE

fgdrf:
[GEOT-6892](https://osgeo-org.atlassian.net/browse/GEOT-6892) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

Signed-off-by: Frank Gasdorf <fgdrf@users.sourceforge.net>

<Include a few sentences describing the overall goals for this Pull Request>

# Checklist

Reviewing is a process done by project maintainers, **mostly on a volunteer basis** (thus limited in time). We need to keep the review overhead as small as possible, and appreciate if you help us to do so by completing the following items. Feel free to ask in a comment if you have troubles with any of them.

For all pull requests:

- [x] Confirm you have read the [contribution guidelines](https://github.com/geotools/geotools/blob/main/CONTRIBUTING.md)
- [x] You have sent a Contribution Licence Agreement (CLA) as necessary (not required for small changes, e.g., fixing typos in documentation)
- [ ] Make sure the first PR targets the `main` branch, eventual backports will be managed later. This can be ignored if the PR is fixing an issue that only happens in a specific branch, but not in newer ones.
- [x] The changes are not causing two modules to share the same Java packages (to avoid [Java 9+ split package](http://tutorials.jenkov.com/java/modules.html#split-packages-not-allowed) issues)
- [x] The changes are not breaking the build in downstream projects using SNAPSHOT dependencies, GeoWebCache and GeoServer (there is an automatic PR check verifying this, check this when it turns green).

The following are required only for core and extension modules (they are welcomed, but not required, for unsupported modules):

- [x] There is an issue in [Jira](https://osgeo-org.atlassian.net/projects/GEOT) describing the bug/task/new feature (a notable exemptions is, changes not visible to end users). The ticket is for the GeoTools project, if the issue was found elsewhere it's a good practice to link to the origin ticket/issue. Issue is https://osgeo-org.atlassian.net/browse/GEOT-6892
- [x] The pull request contains changes related to a single objective. If multiple focuses cannot be avoided, each one is in its own commit and has a separate ticket describing it.
- [x] PR for bug fixes and small new features are presented as a single commit
- [x] Commit message(s) must be in the form "[GEOT-XYZW] Title of the Jira ticket"
- [x] New unit tests have been added covering the changes
- [x] This PR passes all existing unit tests (test results will be reported by Continuous Integration after opening this PR)
[...]

PR #3523, opened 2021-05-25
[DRAFT] SLDParser initalization

fgdrf:
of ExpressionDOMParser and usage of it

Signed-off-by: Frank Gasdorf <fgdrf@users.sourceforge.net>

# Checklist

For all pull requests:

- [ ] Confirm you have read the [contribution guidelines](https://github.com/geotools/geotools/blob/main/CONTRIBUTING.md)
- [ ] Make sure the first PR targets the `main` branch, eventual backports will be managed later. This can be ignored if the PR is fixing an issue that only happens in a specific branch, but not in newer ones.
- [ ] The changes are not causing two modules to share the same Java packages (to avoid [Java 9+ split package](http://tutorials.jenkov.com/java/modules.html#split-packages-not-allowed) issues)

### DIFF
--- a/modules/library/xml/src/main/java/org/geotools/xml/styling/SLDParser.java
+++ b/modules/library/xml/src/main/java/org/geotools/xml/styling/SLDParser.java
@@ -29,6 +29,7 @@ import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.logging.Level;
 import java.util.regex.Pattern;
 import javax.imageio.ImageIO;
@@ -100,7 +101,6 @@ import org.geotools.util.SimpleInternationalString;
 import org.geotools.util.factory.GeoTools;
 import org.opengis.filter.Filter;
 import org.opengis.filter.FilterFactory;
-import org.opengis.filter.FilterFactory2;
 import org.opengis.filter.expression.Expression;
 import org.opengis.filter.expression.Function;
 import org.opengis.filter.expression.Literal;
@@ -123,8 +123,6 @@ import org.xml.sax.SAXException;
  * @author jgarnett
  */
 public class SLDParser {
-
-    private static final FilterFactory2 FF = CommonFactoryFinder.getFilterFactory2();
 
     /** HISTOGRAM */
     private static final String HISTOGRAM = "histogram";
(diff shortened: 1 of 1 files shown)


modules/library/xml/src/main/java/org/geotools/xml/styling/SLDParser.java
```diff
@@ -201,80 +199,95 @@
...
-    }
-
-    public SLDParser(StyleFactory factory, FilterFactory filterFactory) {
-        this.factory = factory;
+    public SLDParser(StyleFactory styleFactory, FilterFactory filterFactory) {
+        Objects.requireNonNull(styleFactory);
+        Objects.requireNonNull(filterFactory);
+        this.factory = styleFactory;
         this.ff = filterFactory;
+        this.expressionDOMParser = new ExpressionDOMParser(CommonFactoryFinder.getFilterFactory2());
```
aaime, line 216: Shouldn't it use the filter factory provided as an argument, at least if that one is a FilterFactory2 (which should be most of the time, the only actual implementation in GeoTools implements both FF2 and thus FF).

aaime (changes requested):
Overall looks good, suggesting a small change.

modules/library/xml/src/main/java/org/geotools/xml/styling/SLDParser.java
fgdrf: Actually with merge of #3516 it could be just 
`this.expressionDOMParser = new ExpressionDOMParser()`

However ExpressionDOMParser second constructor uses FilterFactory2 interface.

IMHO using filterFactory from SLDParser constructor arg assumes internal knowlege about the implementiontion of FilterFactoryImpl. But if someone else provides a FilterFactory via META-INF/services which doesn't implement both, FF and FF2 this class could not be used anymore. With other words a cast from FF to FF2 is from interface perspective not a good choise because you know details about implementation. This would not be a good example for the whole codebase because this kind of patterns ist copied and pasted and makes it hard to understand the code and to maintain as well. Looking forward for proposal @jodygarnett set up few years ago

aaime: That's why I said "if". the code can be modified to test _if_ the provided factory is a FF2, and use it, otherwise create a different one. In the last 16 years spent working every day on GT/GS I have yet to see another implementation of FilterFactory, but yes, it may happen, and might even have aleady happened :-D

PR #3527, opened 2021-05-26
[GEOT-6899] Resource (CPU/Heap) starvation building shapefile spatial index

groldan:
[GEOT-6899](https://osgeo-org.atlassian.net/browse/GEOT-6899) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

Shapefile quadtree index building is triggered on each calling thread,
eventually exhausting available CPU/Memory resources.

Acquire a write lock on the QIX file before building the quad-tree, and
check for stale-ness right after acquiring the lock, to avoid building
the same quad-tree concurrently and/or unnecessarily.

Before:
image

After:

image

# Checklist

For all pull requests:

### DIFF
--- a/modules/plugin/shapefile/src/test/java/org/geotools/data/shapefile/IndexManagerTest.java
+++ b/modules/plugin/shapefile/src/test/java/org/geotools/data/shapefile/IndexManagerTest.java
@@ -1,0 +1,220 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ *
+ *    (C) 2021, Open Source Geospatial Foundation (OSGeo)
+ *
+ *    This library is free software; you can redistribute it and/or
+ *    modify it under the terms of the GNU Lesser General Public
+ *    License as published by the Free Software Foundation;
+ *    version 2.1 of the License.
+ *
+ *    This library is distributed in the hope that it will be useful,
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ *    Lesser General Public License for more details.
+ */
+package org.geotools.data.shapefile;
+
+import static org.geotools.data.shapefile.files.ShpFileType.QIX;
+import static org.geotools.data.shapefile.files.ShpFileType.SHP;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+import java.io.File;
+import java.io.IOException;
+import java.net.URL;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+import org.geotools.data.shapefile.files.ShpFiles;
+import org.geotools.feature.NameImpl;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+public class IndexManagerTest extends TestCaseSupport {
+
+    private ShapefileDataStore mockDataStore;
+    private ShpFiles shpFiles;
+
+    @Before
+    public void setUp() throws IOException {
+        File statesShp = copyShapefiles(STATE_POP);
+        shpFiles = new ShpFiles(statesShp);
+
+        mockDataStore = mock(ShapefileDataStore.class);
+        when(mockDataStore.getTypeName()).thenReturn(new NameImpl("statepop"));
+    }
+
+    @After
+    @Override
+    public void tearDown() throws Exception {
+        super.tearDown();
+        shpFiles.dispose();
+    }
+
+    @Test
+    public void testCreateSpatialIndex() throws Exception {
+        deleteQixFile();
+        IndexManager indexManager = new IndexManager(shpFiles, mockDataStore);
+
+        assertFalse(shpFiles.exists(QIX));
+        final boolean force = false;
+        assertTrue(indexManager.createSpatialIndex(force));
+        assertTrue(shpFiles.exists(QIX));
+        assertFalse(indexManager.createSpatialIndex(force));
+    }
+
+    @Test
+    public void testCreateSpatialIndex_force() throws Exception {
+        deleteQixFile();
+        IndexManager indexManager = new IndexManager(shpFiles, mockDataStore);
+
+        final boolean force = true;
+        assertTrue(indexManager.createSpatialIndex(force));
+
+        File file = null;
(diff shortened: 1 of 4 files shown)


modules/plugin/shapefile/src/test/java/org/geotools/data/shapefile/IndexManagerTest.java
```diff
@@ -0,0 +1,228 @@
...
+        try {
+            file = shpFiles.acquireReadFile(QIX, indexManager.writer);
+            assertNotNull(file);
+        } finally {
+            shpFiles.unlockRead(file, indexManager.writer);
+        }
+
+        final long currentLastModified = file.lastModified();
+
+        for (int i = 0;
```
aaime, line 96: A bit worried about this bit, in the context of build servers and PR checks... which can be very unreliable in terms of time taken to do things. 1 second is a lot of time on a dev machine, but might not be enough on an overworked build server or small PR check VM. We have been bitten by this kind of assumption several times, with tests assuming quick execution randomly failing every now and then.

I suggest the following change: allow the loop to run for eventually a long time (e.g., 30 seconds?), but bail out of the loop as soon as the condition you're looking for (new file created). If you exit the loop without the file being re-created, then it's a failiure.
With this approach it's going to be fast (even faster) on a dev machine, but it's lenient towards a slow PR check that's being kicked out of the CPUs by other VMs (or by other tests running in parallel, even the build checks run with -T2).

modules/plugin/shapefile/src/test/java/org/geotools/data/shapefile/IndexManagerTest.java
```diff
@@ -0,0 +1,228 @@
...
+        final boolean force = false;
+        assertTrue("stale index should force rebuild", indexManager.createSpatialIndex(force));
+        assertTrue(shpFiles.exists(QIX));
+
+        assertTrue(indexFile.lastModified() > lastModified);
+
+        assertFalse("index is not stale anymore", indexManager.createSpatialIndex(force));
+    }
+
+    @Test(timeout = 10_000) // timeout at 10s to fail in case of deadlock
```
aaime, line 137: Same as above, allow longer timeouts... it should only time out if there is an issue, normally this test runs pretty quick, no?

modules/plugin/shapefile/src/test/java/org/geotools/data/shapefile/IndexManagerTest.java
```diff
@@ -0,0 +1,228 @@
...
+        long buildCount = builds.stream().filter(Boolean::booleanValue).count();
+
+        String msg =
+                String.format(
+                        "Expected only 1 true result from %d concurrent createSpatialIndex calls",
+                        taskCount);
+        assertEquals(msg, 1, buildCount);
+    }
+
+    @Test(timeout = 10_000) // timeout at 10s to fail in case of deadlock
```
aaime, line 157: Same as above.

modules/plugin/shapefile/src/test/java/org/geotools/data/shapefile/IndexManagerTest.java
```diff
@@ -0,0 +1,228 @@
...
+                    @Override
+                    protected void doCreateSpatialIndex() throws Exception {
+                        int concurrency = concurrentBuilds.incrementAndGet();
+                        super.doCreateSpatialIndex();
+                        concurrentBuilds.decrementAndGet();
+                        maxConcurrentBuilds.set(Math.max(maxConcurrentBuilds.get(), concurrency));
+                    }
+                };
+
+        final int threadCount = 8;
```
aaime, line 173: Try to relate the number of threads to the actual CPUs, on a build server or PR check we're unlikely to have that many. 
Maybe something like this:

```suggestion
        final int threadCount = Math.min(8, Runtime.getRuntime().availableProcessors());
```

modules/plugin/shapefile/src/test/java/org/geotools/data/shapefile/IndexManagerTest.java
```diff
@@ -0,0 +1,228 @@
...
+            final boolean force)
+            throws Exception {
+        final ExecutorService executor = Executors.newFixedThreadPool(threadCount);
+        List<Callable<Boolean>> tasks =
+                IntStream.range(0, taskCount)
+                        .mapToObj(
+                                i ->
+                                        (Callable<Boolean>)
+                                                () -> indexManager.createSpatialIndex(force))
+                        .collect(Collectors.toList());
```
aaime, line 202: An old style for loop ends up taking less space and requires no casts, be mindful of the 80 columns limit.

```suggestion
        List<Callable<Boolean>> tasks  = new ArrayList(); 
        for(int i = 0; i < taskCount; i++) 
           tasks.add(indexManager.createSpatialIndex(force));
```

Just 3 lines ;-) (yes, curly braces are not mandatory in Google AOSP codingin conventions)

aaime (changes requested):
Code changes look good, some aspects of the tests might be problematic.

modules/plugin/shapefile/src/test/java/org/geotools/data/shapefile/IndexManagerTest.java
groldan: Done! https://github.com/geotools/geotools/pull/3527/files#diff-cca0c8a31b2bd3b8cf2a2473aae2301caf3f80def759a5d36a2d82b3cc4e319bR94-R103

modules/plugin/shapefile/src/test/java/org/geotools/data/shapefile/IndexManagerTest.java
groldan: https://github.com/geotools/geotools/pull/3527/files#diff-cca0c8a31b2bd3b8cf2a2473aae2301caf3f80def759a5d36a2d82b3cc4e319bR137-R138

modules/plugin/shapefile/src/test/java/org/geotools/data/shapefile/IndexManagerTest.java
groldan: https://github.com/geotools/geotools/pull/3527/files#diff-cca0c8a31b2bd3b8cf2a2473aae2301caf3f80def759a5d36a2d82b3cc4e319bR157-R158

modules/plugin/shapefile/src/test/java/org/geotools/data/shapefile/IndexManagerTest.java
groldan: https://github.com/geotools/geotools/pull/3527/files#diff-cca0c8a31b2bd3b8cf2a2473aae2301caf3f80def759a5d36a2d82b3cc4e319bR173-R174

modules/plugin/shapefile/src/test/java/org/geotools/data/shapefile/IndexManagerTest.java
groldan: https://github.com/geotools/geotools/pull/3527/files#diff-cca0c8a31b2bd3b8cf2a2473aae2301caf3f80def759a5d36a2d82b3cc4e319bR196-R211

aaime (approved):

PR #3528, opened 2021-05-28
[GEOT-6900] Shapefile quadtree build performance improvements

groldan:
[GEOT-6900](https://osgeo-org.atlassian.net/browse/GEOT-6900) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

Use a strategy object (`BoundsReader`) to assist `ShapeFileIndexer` in speeding up the 
`QuadTree` optimization  phase, providing quick access to each shapefile
 record envelope, potentially avoiding an immense amount of random disk I/O calls through
`ShapefileReader`, as the quad tree internal nodes get split/shrank.

Since the `QuadTree` leaf nodes hold only the shapefile record ids, and not their
 bounds, the tree layout optimization phase may incur into too much random disk reads on the
`.shp` file, which has a bigger impact the bigger the shapefile is, especially related to the size of the
geometries more than the number of records itself.

The `BoundsReader` strategy object is meant to avoid that to the extent possible.

To a given point, record bounds will be stored in heap memory (up to 1MiB, accounting for 32K
 records, or 64K records if it's a points shapefile).

For a bigger number of shapefile records, the strategy is
 to store the bounds in a temporary file (named 
 `GeoTools_shp_qix_bounds_<random number>.tmp`  under `${java.io.tmpdir}`), 
 which is memory mapped and deleted at `BoundsReader.close()`. This leverages the Operating 
 System's native paging, and due to the reduced size of the bounds file compared to the actual 
 `.shp` and avoiding the parsing performed by `ShapefileReader.nextRecord()`, results in 
 dramatically less random I/O and computing.

Note, however, that if there's not enough temporary space in the file system where the 
`java.io.tmpdir` directory resides, a fall back strategy that reads directly from the 
`ShapefileReader` will be used. This should be a very edge case though, since with a bounds record
 size of 32 bytes, the required temporary storage is 30.1MiB per million features.

## Index build time

The polygon test dataset that was used is attached to [GEOT-6900](https://osgeo-org.atlassian.net/browse/GEOT-6900).

Running `ShapefileIndexer` on it from `main` at 0adffaf3278ca8ff7f6e653b33749620f6cbf59c, prints:
```
341769 features indexed in 193826ms.
```
The same execution with this patch prints:
```
341769 features indexed in 920ms.
```
and produces the exact same results.

The new `ShapefileIndexerStressTest.java` test, ran against the above-mentioned baseline and this patch, results in:

|# Features| Baseline | This patch |
|------------------|----------------| ----------------|
|1,000            | 15 ms| 16 ms|
|10,000         | 41 ms| 24 ms|
|100,000       | 204 ms| 106 ms|
|1,000,000   | 1821 ms| 600 ms|
|10,000,000| 21994 s| 5720 ms|

This shows that the impact of the improvement is proportional to the number of records, but more importantly to the complexity/size of the data itself. Since the test data produced by the stress test is very simple, with 10M records we get only a 4x improvement, but with 300K complex polygons the improvement is 2 orders of magnitude.

# Checklist

For all pull requests:

### DIFF
--- a/modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/ShapefileIndexerBoundsHelper.java
+++ b/modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/ShapefileIndexerBoundsHelper.java
@@ -1,0 +1,332 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ *
+ *    (C) 2021, Open Source Geospatial Foundation (OSGeo)
+ *
+ *    This library is free software; you can redistribute it and/or
+ *    modify it under the terms of the GNU Lesser General Public
+ *    License as published by the Free Software Foundation;
+ *    version 2.1 of the License.
+ *
+ *    This library is distributed in the hope that it will be useful,
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ *    Lesser General Public License for more details.
+ */
+package org.geotools.data.shapefile;
+
+import static org.geotools.data.shapefile.shp.ShapeType.POINT;
+import static org.geotools.data.shapefile.shp.ShapeType.POINTM;
+import static org.geotools.data.shapefile.shp.ShapeType.POINTZ;
+
+import java.io.Closeable;
+import java.io.IOException;
+import java.io.RandomAccessFile;
+import java.nio.DoubleBuffer;
+import java.nio.MappedByteBuffer;
+import java.nio.channels.FileChannel.MapMode;
+import java.nio.file.FileStore;
+import java.nio.file.FileSystemException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import org.geotools.data.shapefile.index.quadtree.QuadTree;
+import org.geotools.data.shapefile.shp.IndexFile;
+import org.geotools.data.shapefile.shp.ShapeType;
+import org.geotools.data.shapefile.shp.ShapefileReader;
+import org.geotools.data.shapefile.shp.ShapefileReader.Record;
+import org.geotools.util.NIOUtilities;
+import org.geotools.util.logging.Logging;
+import org.locationtech.jts.geom.Envelope;
+import org.locationtech.jts.index.quadtree.Quadtree;
+
+/**
+ * Utility class to assist {@link ShapeFileIndexer} in speeding up the {@link QuadTree} optimization
+ * phase through a {@link BoundsReader strategy object} that provides quick access to each shapefile
+ * record envelope, potentially avoiding an immense amount of random disk I/O calls through {@link
+ * ShapefileReader}, as the quad tree internal nodes get split/shrank.
+ *
+ * <p>Since the {@link QuadTree} leaf nodes hold only the shapefile record ids, and not their
+ * bounds, the tree layout optimization phase may incur into too much random disk reads on the
+ * {@code .shp} file.
+ *
+ * <p>The strategy object {@link #create provided} is meant to avoid that to the extent possible.
+ *
+ * <p>To a given point, record bounds will be stored in heap memory (up to 1MiB, accounting for 32K
+ * records, or 64K records if it's a points shapefile).
+ *
+ * <p>For a bigger number of {@link IndexFile#getRecordCount() shapefile records}, the strategy is
+ * to store the bounds in a temporary file (named {@code GeoTools_shp_qix_bounds_<random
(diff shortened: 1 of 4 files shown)


modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/ShapefileIndexerBoundsHelper.java
```diff
@@ -0,0 +1,325 @@
...
+    private static final Logger LOGGER = Logging.getLogger(ShapefileIndexerBoundsHelper.class);
+
+    /**
+     * A record count threshold past which bounds storage falls back to a temporary file.
+     *
+     * <p>Set to the equivalent of 1MB of heap memory, accounting for 32768 records (for a 32 byte
+     * serialized envelope size), or 65536 records if the shapefile geometry type is POINT (using 16
+     * bytes per envelope).
+     */
+    private static final int FALLBACK_TO_FILE_RECCOUNT_THRESHOLD = (1024 * 1024) / 32;
```
aaime, line 83:
```suggestion
    private static final int FALLBACK_TO_FILE_REC_COUNT_THRESHOLD = (1024 * 1024) / 32;
```

(at first I told myself... is that a typo? looks like "raccoon"!)

modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/ShapefileIndexerBoundsHelper.java
```diff
@@ -0,0 +1,325 @@
...
+            reader.read(recNumber, env);
+        }
+
+        @Override
+        public void expand(int recNumber, Envelope env) throws IOException {
+            reader.expand(recNumber, env);
+        }
+
+        private void checkAvailableSpace(final Path path, final int fileSize)
+                throws IOException, FileSystemException {
```
aaime, line 314: FileSystemException is a subclass of IOException, so, redundant in the throw.

modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/ShapefileIndexerBoundsHelper.java
```diff
@@ -0,0 +1,325 @@
...
+
+        if (numRecs <= fileRecCountThreshold) {
+            return new HeapBoundsReader(numRecs, pointBounds);
+        }
+        try {
+            return new FileBoundsReader(numRecs, pointBounds);
+        } catch (IOException e) {
+            LOGGER.log(
+                    Level.INFO,
+                    "Unable to create a temporary file backed bounds reader, falling back to slower ShapefileReader randome access",
```
aaime, line 109: Typo:

```suggestion
                    "Unable to create a temporary file backed bounds reader, falling back to slower ShapefileReader random access",
```

modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/ShapefileIndexerBoundsHelper.java
```diff
@@ -0,0 +1,325 @@
...
+    static interface BoundsReader {
+
+        /** Records the bounds of the given shapefile record to be read after. */
+        void insert(int recNumber, Envelope env);
+
+        /** Initializes the provided envelope to the bounds of the requested shapefile record. */
+        void read(int recNumber, Envelope env) throws IOException;
+
+        /** Expands the provided envelope to cover the bounds of the shapefile record number. */
+        void expand(int recNumber, Envelope env) throws IOException;
```
aaime, line 138: Would recommend renaming to "expandEnvelope", this name seems like it would expand the reader somehow instead (at first I was looking for code that was growing the backing buffer/file, which did not make sense...)

aaime (changes requested):
Nicely done, the speedup is indeed massive, the code is well documented and readable too. Added a couple of notes below.

modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/ShapefileIndexerBoundsHelper.java
groldan: lol, done. Thanks for the review.

modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/ShapefileIndexerBoundsHelper.java
groldan: Done.

modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/ShapefileIndexerBoundsHelper.java
groldan: Fixed.

groldan:
Amended to fix merge conflicts and respect the new PMD rules

aaime (approved):

PR #3539, opened 2021-06-04
[GEOT-6902]: NetCDF dataset access performance improvements

dromagnoli:
[GEOT-6902](https://osgeo-org.atlassian.net/browse/GEOT-6902) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

A few caching improvements on NetCDF plugin
https://osgeo-org.atlassian.net/browse/GEOT-6902

# Checklist

For all pull requests:

### DIFF
--- a/modules/plugin/coverage-multidim/netcdf/src/main/java/org/geotools/imageio/netcdf/AncillaryFileManager.java
+++ b/modules/plugin/coverage-multidim/netcdf/src/main/java/org/geotools/imageio/netcdf/AncillaryFileManager.java
@@ -62,6 +62,7 @@ import org.geotools.gce.imagemosaic.properties.PropertiesCollectorFinder;
 import org.geotools.gce.imagemosaic.properties.PropertiesCollectorSPI;
 import org.geotools.imageio.netcdf.Slice2DIndex.Slice2DIndexManager;
 import org.geotools.imageio.netcdf.utilities.NetCDFUtilities;
+import org.geotools.util.SoftValueHashMap;
 import org.geotools.util.URLs;
 import org.geotools.util.Utilities;
 import org.geotools.util.logging.Logging;
@@ -189,6 +190,14 @@ public class AncillaryFileManager implements FileSetManager {
     // dimension name
     private final Map<String, MultipleBandsDimensionInfo> multipleBandsDimensionsInfo =
             new HashMap<>();
+
+    // Indexer and datastore config can be considered static so we can cache them
+    // in order to avoid their repeated unmarshalling when accessing a dataset.
+
+    protected static final Map<String, Indexer> INDEXER_CACHE = new SoftValueHashMap<>();
+
+    protected static final Map<String, DataStoreConfiguration> DATASTORE_CONFIG_CACHE =
+            new SoftValueHashMap<>();
 
     static {
         try {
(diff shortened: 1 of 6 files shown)


modules/plugin/coverage-multidim/netcdf/src/main/java/org/geotools/imageio/netcdf/AncillaryFileManager.java
```diff
@@ -515,29 +524,35 @@ public void resetSliceManager() throws IOException {
...
-                if (schemas != null) {
-                    // Map schema names to schema attributes string
-                    List<SchemaType> schemaElements = schemas.getSchema();
-                    for (SchemaType schemaElement : schemaElements) {
-                        schemaMapping.put(schemaElement.getName(), schemaElement.getAttributes());
-                    }
+                INDEXER_CACHE.put(indexerFile.getAbsolutePath(), indexer);
+            }
+        }
+        if (indexer != null) {
```
aaime, line 537: In the previous code the indexer was assumed to be there, in case the unmarshaller could be created.
Besides that... what happens if an unmarshaller or an index cannot be created? Should'nt an exception be thrown anyways?

modules/plugin/coverage-multidim/netcdf/src/main/java/org/geotools/imageio/netcdf/utilities/NetCDFUtilities.java
```diff
@@ -1489,4 +1518,18 @@ public static ParameterBehaviour getParameterBehaviour(String parameter) {
...
+
+    private static void clearCache() {
+        URI_FORMAT_CACHE.clear();
+    }
+
+    /**
+     * Clear all the internal caches. Call this method if the datastore config has been updated or
+     * the NetCDF XML file has been modified.
+     */
+    public static void clearCaches() {
```
aaime, line 1530: Who's gonna use this? A listener for GeoServer "reset" button?

aaime (commented):
I see no tests, however, they could be written: check that the second time you ask for one object, you get the same exact instance (the one in cache) rather than another one.

Also, wondering about the cache sizes. The default is 20 hard references and everything else is going to be soft. Is that good? I guess it might, just wondering if some thought was put into it.

modules/plugin/coverage-multidim/netcdf/src/main/java/org/geotools/imageio/netcdf/utilities/NetCDFUtilities.java
dromagnoli: Exactly

dromagnoli:
> I see no tests, however, they could be written: check that the second time you ask for one object, you get the same exact instance (the one in cache) rather than another one.
> 
> Also, wondering about the cache sizes. The default is 20 hard references and everything else is going to be soft. Is that good? I guess it might, just wondering if some thought was put into it.


Tests added.
About the soft/hard reference usage, yep, I have used default ones to not have many hard references around.

modules/plugin/coverage-multidim/netcdf/src/main/java/org/geotools/imageio/netcdf/AncillaryFileManager.java
dromagnoli: Added exceptions on those cases

dromagnoli:
GeoServer downstream integration is expected to fail.
it will be fixed with the dependent PR:
https://github.com/geoserver/geoserver/pull/5095

PR #3540, opened 2021-06-07
[GEOT-6906] WFS DescribeFeatureType doesn't account for outputFormats defined globally

roarbra:
[GEOT-6906](https://osgeo-org.atlassian.net/browse/GEOT-6906) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

… defined globally

As explained in JIRA StrictWFS_2_0_Strategy.getServerSupportedOutputFormats doesn't look for Parameter-elements within the element OperationsMetadata.

The rest of the PR is concerned with adopting AbstractWfsDataStoreOnelineTest for the WFS server with this capabilities problem. The main issues are that CRS should be COMPLIANT, and that a GetFeature request returning everything is massive.

# Checklist

For all pull requests:

### DIFF
--- a/modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/internal/parsers/PullParserFeatureReader.java
+++ b/modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/internal/parsers/PullParserFeatureReader.java
@@ -18,6 +18,7 @@ package org.geotools.data.wfs.internal.parsers;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.xml.namespace.QName;
 import javax.xml.stream.XMLStreamException;
@@ -55,6 +56,10 @@ public class PullParserFeatureReader implements GetParser<SimpleFeature> {
 
     GeometryCoordinateSequenceTransformer transformer;
 
+    private long lastLogMessage = 0;
+    private long logCounter = 0;
+    private final long WAIT_LOG = 5000;
+
     public PullParserFeatureReader(
             final Configuration wfsConfiguration,
             final InputStream getFeatureResponseStream,
@@ -82,6 +87,7 @@ public class PullParserFeatureReader implements GetParser<SimpleFeature> {
         if (inputStream != null) {
             try {
                 inputStream.close();
+
             } finally {
                 inputStream = null;
                 parser = null;
@@ -95,6 +101,18 @@ public class PullParserFeatureReader implements GetParser<SimpleFeature> {
         Object parsed;
         try {
             parsed = parser.parse();
+            if (parsed != null && LOGGER.isLoggable(Level.FINE)) {
+                logCounter++;
+                long time = System.currentTimeMillis();
+                if (lastLogMessage == 0) {
+                    LOGGER.fine("First feature parsed.");
+                    lastLogMessage = time;
+                } else if (time - lastLogMessage > WAIT_LOG) {
+                    LOGGER.fine(String.format("Number of features parsed: %d", logCounter));
+                    lastLogMessage = time;
+                }
+            }
+
         } catch (XMLStreamException | SAXException e) {
             throw new IOException(e);
         }
--- a/modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/internal/v2_0/StrictWFS_2_0_Strategy.java
+++ b/modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/internal/v2_0/StrictWFS_2_0_Strategy.java
@@ -665,12 +665,7 @@ public class StrictWFS_2_0_Strategy extends AbstractWFSStrategy {
         return ftypeCrss;
     }
 
-    /**
-     * Returns parameters defined in OperationsMetadata
-     *
-     * @param parameterName
-     * @return
-     */
+    /** Returns parameters defined in OperationsMetadata */
     @SuppressWarnings("unchecked")
     private Set<String> findParameters(final String parameterName) {
         final OperationsMetadataType operationsMetadata = capabilities.getOperationsMetadata();
@@ -678,11 +673,11 @@ public class StrictWFS_2_0_Strategy extends AbstractWFSStrategy {
         List<DomainType> parameters = operationsMetadata.getParameter();
         for (DomainType parameter : parameters) {
             if (parameterName.equals(parameter.getName())) {
-                Set<String> outputFormats = new HashSet<>();
+                Set<String> foundValues = new HashSet<>();
                 for (ValueType value : (List<ValueType>) parameter.getAllowedValues().getValue()) {
-                    outputFormats.add(value.getValue());
+                    foundValues.add(value.getValue());
                 }
-                return outputFormats;
+                return foundValues;
             }
         }
         return Collections.emptySet();
@@ -698,12 +693,12 @@ public class StrictWFS_2_0_Strategy extends AbstractWFSStrategy {
             String paramName = param.getName();
 
             if (parameterName.equals(paramName)) {
-                Set<String> outputFormats = new HashSet<>();
+                Set<String> foundValues = new HashSet<>();
 
                 for (ValueType value : (List<ValueType>) param.getAllowedValues().getValue()) {
-                    outputFormats.add(value.getValue());
+                    foundValues.add(value.getValue());
                 }
-                return outputFormats;
+                return foundValues;
             }
         }
         return Collections.emptySet();
--- a/modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/internal/v2_0/StrictWFS_2_0_Strategy.java
+++ b/modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/internal/v2_0/StrictWFS_2_0_Strategy.java
@@ -561,6 +561,9 @@ public class StrictWFS_2_0_Strategy extends AbstractWFSStrategy {
         final OperationType operationMetadata = getOperationMetadata(operation);
 
         Set<String> serverSupportedFormats = findParameters(operationMetadata, parameterName);
+        if (serverSupportedFormats.isEmpty()) {
+            serverSupportedFormats = findParameters(parameterName);
+        }
         return serverSupportedFormats;
     }
 
@@ -662,10 +665,32 @@ public class StrictWFS_2_0_Strategy extends AbstractWFSStrategy {
         return ftypeCrss;
     }
 
+    /**
+     * Returns parameters defined in OperationsMetadata
+     *
+     * @param parameterName
+     * @return
+     */
+    @SuppressWarnings("unchecked")
+    private Set<String> findParameters(final String parameterName) {
+        final OperationsMetadataType operationsMetadata = capabilities.getOperationsMetadata();
+
+        List<DomainType> parameters = operationsMetadata.getParameter();
+        for (DomainType parameter : parameters) {
+            if (parameterName.equals(parameter.getName())) {
+                Set<String> outputFormats = new HashSet<>();
+                for (ValueType value : (List<ValueType>) parameter.getAllowedValues().getValue()) {
+                    outputFormats.add(value.getValue());
+                }
+                return outputFormats;
+            }
+        }
+        return Collections.emptySet();
+    }
+
     @SuppressWarnings("unchecked")
     protected Set<String> findParameters(
             final OperationType operationMetadata, final String parameterName) {
-        Set<String> outputFormats = new HashSet<>();
 
         List<DomainType> parameters = operationMetadata.getParameter();
         for (DomainType param : parameters) {
@@ -673,13 +698,15 @@ public class StrictWFS_2_0_Strategy extends AbstractWFSStrategy {
             String paramName = param.getName();
 
             if (parameterName.equals(paramName)) {
+                Set<String> outputFormats = new HashSet<>();
 
                 for (ValueType value : (List<ValueType>) param.getAllowedValues().getValue()) {
                     outputFormats.add(value.getValue());
                 }
+                return outputFormats;
             }
         }
-        return outputFormats;
+        return Collections.emptySet();
     }
 
     protected AbstractTransactionActionType createInsert(Wfs20Factory factory, Insert elem)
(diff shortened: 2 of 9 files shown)


modules/unsupported/wfs-ng/src/test/java/org/geotools/data/wfs/online/AbstractWfsDataStoreOnlineTest.java
```diff
@@ -467,13 +453,24 @@ public BoundingBox getBounds() {
...
         }
+        */
     }
 
+    /**
+     * There must be a better way to check for axis flipping than counting all features.
+     *
+     * @throws Exception
+     */
+    @Ignore
```
mprins, line 473: please don't switch off working tests, if you think there is a better way you are free to implement that in a 
 new PR

mprins (changes requested):
why not use the default value of describeFeatureType's outputFormat as described in the spec?

modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/internal/parsers/PullParserFeatureReader.java
```diff
@@ -95,6 +101,18 @@ public SimpleFeature parse() throws IOException {
         Object parsed;
         try {
             parsed = parser.parse();
+            if (parsed != null && LOGGER.isLoggable(Level.FINE)) {
```
matthijsln, line 104: This logging code should probably be removed, left after debugging?

roarbra: It is left after debugging, because the original code was hanging and not showing any progress at all. This is a way to turn on a little feed-back on what's going on without interfering to much.
And the LOGGER.isLoggable(Level.FINE) makes it easy to switch off.

roarbra:
> why not use the default value of describeFeatureType's outputFormat as described in the spec?

@mprins 
The getDefaultOutputFormat does several things. It first checks what outputFormats the server handles, then it checks the configuration if a special outputFormat should be used. And at last it selects the first outputFormat that bot the server and this installation of GeoTools can handle.

The "application/gml+xml"-format should definitely be handled both by the server and certainly by GeoTools. And will for most cases be the result of this function.

We could return null instead of throwing an exception. If I understand the meaning of a default value for an attribute correct, it means the value that will be used if the attribute isn't specified. And if we set outputFormat=null, I don't think the attribute will be added.

At the other side. This exception disclosed a shortcoming of the code. According to the specification of OWS. It's legitimate to define outputFormat's for all operations in the root element of OperationsMetadata. And that shortcoming I mean is solved by this PR.

mprins (approved):

modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/internal/v2_0/StrictWFS_2_0_Strategy.java
```diff
@@ -662,24 +665,48 @@ protected OperationType getOperationMetadata(final WFSOperationType operation) {
         return ftypeCrss;
     }
 
+    /**
+     * Returns parameters defined in OperationsMetadata
+     *
+     * @param parameterName
```
aaime, line 671: This could be stripped

modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/internal/v2_0/StrictWFS_2_0_Strategy.java
```diff
@@ -662,24 +665,48 @@ protected OperationType getOperationMetadata(final WFSOperationType operation) {
...
+     * @return
+     */
+    @SuppressWarnings("unchecked")
+    private Set<String> findParameters(final String parameterName) {
+        final OperationsMetadataType operationsMetadata = capabilities.getOperationsMetadata();
+
+        List<DomainType> parameters = operationsMetadata.getParameter();
+        for (DomainType parameter : parameters) {
+            if (parameterName.equals(parameter.getName())) {
+                Set<String> outputFormats = new HashSet<>();
```
aaime, line 681:
```suggestion
                Set<String> values = new HashSet<>();
```

The method is generic, we don't know they are output formats here.

modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/internal/v2_0/StrictWFS_2_0_Strategy.java
```diff
@@ -662,24 +665,48 @@ protected OperationType getOperationMetadata(final WFSOperationType operation) {
...
             final OperationType operationMetadata, final String parameterName) {
-        Set<String> outputFormats = new HashSet<>();
 
         List<DomainType> parameters = operationMetadata.getParameter();
         for (DomainType param : parameters) {
 
             String paramName = param.getName();
 
             if (parameterName.equals(paramName)) {
+                Set<String> outputFormats = new HashSet<>();
```
aaime, line 701: Same as above (I see it was already like that, but it's a rename-refactor away from being fixed).

modules/unsupported/wfs-ng/src/test/java/org/geotools/data/wfs/online/AbstractWfsDataStoreOnlineTest.java
```diff
@@ -327,29 +332,6 @@ public void testTypes() throws IOException, NoSuchElementException {
         List<String> typeNames = Arrays.asList(types);
         assertTrue(typeNames.contains(testType.FEATURETYPENAME));
 
-        for (String typeName : types) {
```
aaime, line 330: Hmm... I get this one had no assertions, but it was at least checking the code would not throw exceptions.

modules/unsupported/wfs-ng/src/test/java/org/geotools/data/wfs/online/WFSOnlineTestSupport.java
```diff
@@ -144,12 +145,6 @@ public static void doFeatureReaderWithQuery(DataStore wfs, String typeName) thro
             }
             assertNotNull("must have 1 feature ", feature);
             fid = feature.getID();
-            int j = 0;
```
aaime, line 147: This was actually testing there is at least one feature in the reader, skipping the counting machinery... now there is nothing?

modules/unsupported/wfs-ng/src/test/java/org/geotools/data/wfs/online/AbstractWfsDataStoreOnlineTest.java
roarbra: I had to remove it, because it spent all too much time. I do think that it checks something that is checked elsewhere, but I will override in my class instead.

modules/unsupported/wfs-ng/src/test/java/org/geotools/data/wfs/online/WFSOnlineTestSupport.java
roarbra: It was to avoid iterating over the whole result set. My intention was to set maxFeatures and then check that I at least got that many features. At some point the iteration vanished. Fixed it now.

modules/unsupported/wfs-ng/src/test/java/org/geotools/data/wfs/online/AbstractWfsDataStoreOnlineTest.java
aaime: Thanks!

modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/internal/v2_0/StrictWFS_2_0_Strategy.java
roarbra: I've renamed to foundValues.

modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/internal/v2_0/StrictWFS_2_0_Strategy.java
roarbra: Oh, didn't notice there was a suggestion. I used foundValues.

modules/unsupported/wfs-ng/src/test/java/org/geotools/data/wfs/online/WFSOnlineTestSupport.java
aaime: Given the nature of the test, you could break out of the loop once j reaches 5, that's going to save a lot of time for large datasets

roarbra: @aaime Would you like me to change this code?

I don't really see the point in a unit test that just iterates over a result set. It should test for something. If there is a problem to iterate over a result set I do believe it would be revealed in several of the tests.

aaime: The issue with testing something, besides the count, is that it would require some knowledge of the particular feature type being tested, or how ids are implemented, stuff like that... I'd leave it there, if you really want to fix something, make it exit after 6 iterations, just enough to make the count test fail and indicate there is something broken in either the max feature request encoding, or the server response.

roarbra: I placed the assertTrue(j<=5) inside the while.

modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/internal/parsers/PullParserFeatureReader.java
jodygarnett: It should be useful, a little off topic from the initial bug report, but useful.

modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/internal/v2_0/StrictWFS_2_0_Strategy.java
```diff
@@ -662,24 +665,43 @@ protected OperationType getOperationMetadata(final WFSOperationType operation) {
         return ftypeCrss;
     }
 
+    /** Returns parameters defined in OperationsMetadata */
+    @SuppressWarnings("unchecked")
+    private Set<String> findParameters(final String parameterName) {
+        final OperationsMetadataType operationsMetadata = capabilities.getOperationsMetadata();
```
jodygarnett, line 671: I am a bit confused and need to double check the specification; `operations` plural should be the list of operations like `GetCapabilities`, `DescribeFeatureType`, etc... right?

jodygarnett: When you look for output formats defined globally (for feature content) I assumed you were only going to check the `GetFeature` operation...

jodygarnett: Checking...

http://docs.opengeospatial.org/is/09-025r2/09-025r2.html#52
  
Common request parameters has Standard presentation parameters including outputFormat parameter so the different operations handle the concept in the same manner. This just describes how outputFormat parameter can be used by the operations; if they have the idea they should use the parameter in a consistent fashion.

It does not imply that each operation support the same output formats (for example GetFeature and DescribeFeatureType provide different information).

* http://docs.opengeospatial.org/is/09-025r2/09-025r2.html#table_12 Table 12 — Parameter domains for WFS operations

jodygarnett: Okay I see in the example they have:
```
      <ows:Parameter name="version">
         <ows:AllowedValues>
         <ows:Value>2.0.2</ows:Value>
         <ows:Value>2.0.0</ows:Value>
         </ows:AllowedValues>
      </ows:Parameter>
```

So version is a parameter common to all the different operations - you are right!

I have never seen this done for outputFormats; but since one server does it we should support it.

jodygarnett (approved):

modules/unsupported/wfs-ng/src/test/java/org/geotools/data/wfs/online/KartverketStedsnavnDataStoreOnlineTest.java
```diff
@@ -0,0 +1,108 @@
...
+import org.opengis.feature.type.GeometryDescriptor;
+import org.opengis.filter.Filter;
+import org.opengis.filter.Id;
+import org.opengis.referencing.crs.CoordinateReferenceSystem;
+
+/** @author Roar Brænden */
+public class KartverketStedsnavnDataStoreOnlineTest extends AbstractWfsDataStoreOnlineTest {
+
+    static final String SERVER_URL =
+            "https://wfs.geonorge.no/skwms1/wfs.stedsnavn?request=GetCapabilities&service=WFS";
```
jodygarnett, line 43: Thanks for an online test case to verify the functionality.

It may be worth covering this in a static get capabilities test case if you run into any issues in the future (as online tests are not often run).

aaime: Either that or we could have an online test actually being run as a PR check, similar to what is done for postgresql and the like. Only worry is reliability: in the case of databases we set them up and control them, here the services are managed by external entities, tests might start failing either randomly or even permanently due to changes to their configs.

Overall having a mock test that relies on saved documents seems better, it can be always run and it's reliable.

aaime: Well, since there is plenty of online tests already anyways, we can leave "moving to a mock test" to a future PR. Merging.

PR #3558, opened 2021-06-18
[GEOT-6918]: Implementing now function

dromagnoli:
[GEOT-6918](https://osgeo-org.atlassian.net/browse/GEOT-6918) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

See [GEOT-6918]:

# Checklist

For all pull requests:

### DIFF
--- a/modules/library/main/src/main/java/org/geotools/filter/function/NowFunction.java
+++ b/modules/library/main/src/main/java/org/geotools/filter/function/NowFunction.java
@@ -22,7 +22,21 @@ import org.geotools.filter.FunctionExpressionImpl;
 import org.geotools.filter.capability.FunctionNameImpl;
 import org.opengis.filter.capability.FunctionName;
 
-/** Return the current time (milliseconds) as a Date */
+/**
+ * Return the current time as a Date. When evaluated directly, it returns a new value at each
+ * invocation. When evaluated as part of a filter againsta a DataStore, the value might be one, and
+ * fixed, during the data access. For example:
+ *
+ * <ul>
+ *   <li>If the datastore uses {@link org.geotools.filter.visitor.SimplifyingFilterVisitor}, the
+ *       function will be called once, and replaced with the returned value
+ *   <li>If the datastore encodes the function down in a native query language, the value could also
+ *       be the same for all processed features. For example, the PostgreSQL <a
+ *       href="https://www.postgresql.org/docs/13/functions-datetime.html#FUNCTIONS-DATETIME-CURRENT">now</a>
+ *       function always returns the time at the beginning of the current transaction, to give a
+ *       consistent notion of current time across it.
+ * </ul>
+ */
 public class NowFunction extends FunctionExpressionImpl {
     public static FunctionName NAME = new FunctionNameImpl("now", Date.class);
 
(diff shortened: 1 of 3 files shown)


modules/library/main/src/main/java/org/geotools/filter/function/NowFunction.java
```diff
@@ -0,0 +1,44 @@
...
+package org.geotools.filter.function;
+
+import java.util.Calendar;
+import org.geotools.filter.FunctionExpressionImpl;
+import org.geotools.filter.capability.FunctionNameImpl;
+import org.opengis.filter.capability.FunctionName;
+
+/** Return the current time (milliseconds) as an expression */
+public class NowFunction extends FunctionExpressionImpl {
+    public static FunctionName NAME = new FunctionNameImpl("now", Long.class);
```
aaime, line 27: Seems more natural to return a Date, datedifference would still work with it.

dromagnoli: Fixed. thanks for the feedback

modules/library/main/src/main/java/org/geotools/filter/function/NowFunction.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+package org.geotools.filter.function;
+
+import java.util.Date;
+import org.geotools.filter.FunctionExpressionImpl;
+import org.geotools.filter.capability.FunctionNameImpl;
+import org.opengis.filter.capability.FunctionName;
+
+/** Return the current time (milliseconds) as a Date */
+public class NowFunction extends FunctionExpressionImpl {
```
ianturton, line 26: Should this not implement `VolatileFunction` otherwise you run the risk of it getting cached

ianturton (changes requested):

modules/library/main/src/main/java/org/geotools/filter/function/NowFunction.java
aaime: It does not, otherwise different features would be compared to different times. From the javadoc of VolatileFunction:

```
Functions whose value changes over time but not within the same feature collection filtering
are considered to be <em>stable</em> and as such their result can be considered a constant during
the single data access operation
```

There is no cross-request caching, the SimplifyingFilterVisitor transforms filters incoming from the user request on the fly each time.

ianturton: OK, but I would make that clear in the documentation - I would expect to get a different value back each time I called a function called Now :-)

aaime: It does give a different value... except when it's optimized out in SimplifyingFilterVisitor. It does that for every function not taking arguments, or taking static arguments. But yes, does not hurt to remind it in its javadoc.

aaime: I have updated the javadoc, what do you think @ianturton ?

ianturton (approved):

PR #3570, opened 2021-07-07
[GEOT-5859] upgrade httpclient

ianturton:
[GEOT-5859](https://osgeo-org.atlassian.net/browse/GEOT-5859) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

# Checklist

For all pull requests:

### DIFF
--- a/modules/plugin/http-commons/src/main/java/org/geotools/http/commons/MultithreadedHttpClient.java
+++ b/modules/plugin/http-commons/src/main/java/org/geotools/http/commons/MultithreadedHttpClient.java
@@ -171,6 +171,7 @@ public class MultithreadedHttpClient implements HTTPClient, HTTPConnectionPoolin
         }
 
         HttpMethodResponse response = new HttpMethodResponse(resp);
+
         return response;
     }
 
--- a/modules/plugin/http-commons/src/main/java/org/geotools/http/commons/MultithreadedHttpClient.java
+++ b/modules/plugin/http-commons/src/main/java/org/geotools/http/commons/MultithreadedHttpClient.java
@@ -40,6 +40,7 @@ import org.apache.http.client.methods.HttpGet;
 import org.apache.http.client.methods.HttpPost;
 import org.apache.http.client.methods.HttpRequestBase;
 import org.apache.http.client.protocol.HttpClientContext;
+import org.apache.http.entity.InputStreamEntity;
 import org.apache.http.entity.StringEntity;
 import org.apache.http.impl.client.BasicCredentialsProvider;
 import org.apache.http.impl.client.HttpClientBuilder;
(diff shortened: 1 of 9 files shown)


ianturton:
It would be good @roarbra had a look at this too - I think that all the local/nonlocal proxy stuff is handled by the HttpClient code now, but I'm not sure how to test it to make sure.

mprins (approved):
Great, LGTM!
I seem to recall this as a lot of work having logged the initial issue, but perhaps the refactoring alleviated that somewhat.

mprins (approved):

modules/plugin/http-commons/src/main/java/org/geotools/http/commons/MultithreadedHttpClient.java
```diff
@@ -77,123 +78,83 @@
...
-            HttpState state = client.getState();
-            if (state == null) {
-                state = new HttpState();
-                client.setState(state);
-            }
-            AuthScope authscope = AuthScope.ANY;
-            Credentials credentials = new UsernamePasswordCredentials(proxyUser, proxyPassword);
-            state.setProxyCredentials(authscope, credentials);
-        }
+        clientBuilder.useSystemProperties();
```
aaime, line 106: Oh neat... and it accounts for even more system properties than the old code did, see the [class javadoc](https://hc.apache.org/httpcomponents-client-4.5.x/current/httpclient/apidocs/index.html?org/apache/http/impl/client/HttpClientBuilder.html) (it's not in the method javadoc for some reason).

modules/plugin/http-commons/src/main/java/org/geotools/http/commons/MultithreadedHttpClient.java
```diff
@@ -250,59 +222,68 @@ public void setPassword(String password) {
...
-        client.getState().clearCredentials();
         if (user != null && password != null) {
             AuthScope authscope = AuthScope.ANY;
             Credentials credentials = new UsernamePasswordCredentials(user, password);
-
-            client.getParams().setAuthenticationPreemptive(true);
-            client.getState().setCredentials(authscope, credentials);
-        } else {
-            client.getParams().setAuthenticationPreemptive(false);
+            // TODO - check if this works for all types of auth or do we need to look it up?
```
aaime, line 228: I see another bit of code above where credentials are base64 encoded and set in the headers... so is this not working as it should?

aaime (approved):
Overall looks pretty good. I have a question below.

aaime:
@roarbra any chance you could have a look at this one? Upgrades the multithreaded HTTP client to Commons HTTP client 4.5.x

modules/plugin/http-commons/src/main/java/org/geotools/http/commons/MultithreadedHttpClient.java
ianturton: It seems to depend on if you are using GET or POST and I'm not entirely sure what setting the credentials here does or doesn't affect. Also not sure if this code can handle anything but BASIC auth (as that's all that is tested) or if in fact we support anything else

modules/plugin/http-commons/src/main/java/org/geotools/http/commons/MultithreadedHttpClient.java
ianturton: Actually, we should probably add that list to our docs somewhere

roarbra:
Hi,

I’m at vacation for the moment, and haven’t a opportunity to look at this
for the next two weeks.

Best regards,

Roar Brænden

man. 12. jul. 2021 kl. 15:43 skrev Ian Turton ***@***.***>:

modules/plugin/http-commons/src/main/java/org/geotools/http/commons/MultithreadedHttpClient.java
ianturton: I've managed to fix this up - it seems the problem was the wiremock doesn't ask for basic authentication so the client doesn't send it!

jodygarnett:
Writing up release notes, cannot tell what version of http client is actually updated to - is it a transitive dependency now?

mprins:
Commons-httpclient 3.1 upgraded to apache httpclient 4.5.13 as far as I can tell

PR #3583, opened 2021-07-14
Deadlock on concurrent WFS at start

tiago-s-vieira:
https://osgeo-org.atlassian.net/browse/GEOS-10132

Simultaneous requests after geoserver starts, creates some deadlocks on getSchemas().

### DIFF
--- a/modules/extension/xsd/xsd-core/src/main/java/org/geotools/xsd/XSD.java
+++ b/modules/extension/xsd/xsd-core/src/main/java/org/geotools/xsd/XSD.java
@@ -183,8 +183,10 @@ public abstract class XSD {
         if (schema == null) {
             synchronized (this) {
                 if (schema == null) {
-                    LOGGER.fine("building schema for schema: " + getNamespaceURI());
-                    schema = buildSchema();
+                    synchronized (Schemas.class) {
+                        LOGGER.fine("building schema for schema: " + getNamespaceURI());
+                        schema = buildSchema();
+                    }
                 }
             }
         }


aaime:
I have re-instated the checklist. For a one line change the CLA is not required, but a unit test is.
So the test should try to reproduce what's happening in GeoServer, right after the start-up, when hit by concurrent requests... by the thread stack traces I have the impression it's locking up while trying to parse GetFeature requests provided as XML POST requests. If that's the case, then maybe a simple test with an Executor, enough threads, and callables doing parsing of GetFeature requests that could be located at ``src/test/resources`` might do the trick. The test can be placed in ``gt-xsd-wfs``.

Oh, I've also created a GeoTools ticket for you and updated the PR title accordingly.

dev-rit:
@aaime we tried to replicate the deadlock using the Executor approach to parse the GetFeature requests, but there were no deadlocks. Created a simple class extending the WFSTestSupport and then in the testParse() did the parsing of the request provided by a xml file.  In this scenario XSD class schema variable gets initialized during the initialization of the class WFSTestSupport, hence when the threads starts the parsing the schema is never null and no deadlocks. 

Secondly  I also tried two approaches to obtain the schema object by using the Executor with a list of Callable returning schema object as :

1. 

for (Iterator d = new org.geotools.wfs.v1_1.WFSConfiguration().getXSD().getDependencies().iterator(); d.hasNext(); ) {
                                XSD xsd = (XSD) d.next();
                                xsd.getSchema();
                            }

2.
 () -> {
                        XSDSchema schema = null;
                        try {
                            schema = GML.getInstance().getSchema();
                        } catch (IOException e) {
                            throw e;
                        }
                        return schema;
                    }

300 threads were launched for these test scenarios and I was unable to generate the deadlock using unit tests.

aaime:
First of all, the original issue was reported against a GeoServer 2.18.1, and here you're using a newer version of GeoTools. Just to make sure, can you reproduce the issue with GeoServer [main branch nightly builds](https://build.geoserver.org/geoserver/)?

Second observation, if you are using Tomcat for the deploys, you might get in a situation where requests are queuing against the Tomcat that's starting up, and then they all start exactly at the same time, the moment the startup is done.
Reproducing this level of exact concurrency requires a bit more effort... you need to use a CountdownLatch that all executors wait onto, and then, once they are all queued on the executor service, release the latch, so that they really all start in the exact same moment. If you do this, could you please add to the PR the test code you've developed, even if it does not work?

dev-rit:
@aaime  attached [
[SchemasDeadLockTest.zip](https://github.com/geotools/geotools/files/6974609/SchemasDeadLockTest.zip)
]  is an attempt to generate the deadlock. But I am not able to do so with this test. 

Second of all in our environment we are using geotools 24.1 with geoserver 2.18.1. It was in this environment where the deadlock appeared.

PR #3590, opened 2021-07-30
[GEOT-6947] Copy Constructor not adding null Locale

ydeliorman:
[GEOT-6947](https://osgeo-org.atlassian.net/browse/GEOT-6947) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

This pull request adds the possibility to have a null value to copy constructor.

# Checklist

### DIFF
--- a/modules/library/metadata/src/main/java/org/geotools/util/GrowableInternationalString.java
+++ b/modules/library/metadata/src/main/java/org/geotools/util/GrowableInternationalString.java
@@ -111,8 +111,7 @@ public class GrowableInternationalString extends AbstractInternationalString
                         : Stream.of(Locale.getAvailableLocales()).collect(Collectors.toSet());
         for (Locale locale : locales) {
             String value = internationalString.toString(locale);
-            if (value != null && locale != null && !value.equals(defaultValue))
-                localMap.put(locale, value);
+            if (value != null && !value.equals(defaultValue)) localMap.put(locale, value);
         }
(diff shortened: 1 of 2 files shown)


modules/library/metadata/src/main/java/org/geotools/util/GrowableInternationalString.java
```diff
@@ -142,13 +141,11 @@ public synchronized void add(final Locale locale, final String string)
                     }
             }
             String old = localMap.get(locale);
-            if (old != null) {
```
taba90, line 145: So when adding a content for a Locale that already has it, it will be now replaced.
I guess this was done to avoid exception GeoServer side on the UI component. But this is not the desired behaviour, since the client code should be the one making sure to not add duplicated entries to avoid exception to be thrown.
I would put this code back as it was and fix the validation for duplicated entries in the component on the GeoServer related pr.

taba90 (changes requested):

modules/library/metadata/src/main/java/org/geotools/util/GrowableInternationalString.java
ydeliorman: done

ydeliorman:
Code is fixed according to review.

taba90:
@ydeliorman thanks, looks good to me

PR #3602, opened 2021-08-12
[GEOT-6958]: imageMosaic fix on mixed coverages (heterogeneous+homogeneous)

dromagnoli:
[GEOT-6958](https://osgeo-org.atlassian.net/browse/GEOT-6958) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

See https://osgeo-org.atlassian.net/browse/GEOT-6958

# Checklist


modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/ImageMosaicReader.java
```diff
@@ -385,7 +386,18 @@ private void initReaderFromURL(final Object source, final Hints hints) throws Ex
                             ImageMosaicConfigHandler.createGranuleCatalogFromDatastore(
                                     parentDirectory, datastoreProperties, true, getHints());
                 } else {
-                    CatalogConfigurationBean bean = beans.get(0).getCatalogConfigurationBean();
+                    // We need to figure out if there is an heterogeneous mosaic in the list and use
+                    // it as reference. (Otherwise, read on heterogeneous coverages may not work
+                    // properly)
+                    List<MosaicConfigurationBean> referenceBeans =
```
aaime, line 392: You actually just want the first no? Then it's slightly better to use findFirst and then check the returned Optional, and use the "orElse" to pick the first from the original list as a fallback.

modules/plugin/imagemosaic/src/test/java/org/geotools/gce/imagemosaic/ImageMosaicReaderTest.java
```diff
@@ -4632,6 +4634,40 @@ public void testCoverageOnBands() throws Exception {
         reader.dispose();
     }
 
+    @Test
+    public void testHeterogeneousConfigs() throws Exception {
+        // we have 2 mosaics here:
```
aaime, line 4639: Two mosaics... or two coverages in the same mosaic?

aaime (changes requested):
Makes sense, proposed a small improvement.

modules/plugin/imagemosaic/src/test/java/org/geotools/gce/imagemosaic/ImageMosaicReaderTest.java
dromagnoli: Right, they are actually 2 coverages of the same mosaic store.

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/ImageMosaicReader.java
dromagnoli: Oki, Will do.
Thanks for the feedback

dromagnoli:
@aaime all feedbacks have been addressed.

aaime (approved):

PR #3606, opened 2021-08-16
[GEOT-6959] SimpleHttpClient does support proxy

roarbra:
[GEOT-6959](https://osgeo-org.atlassian.net/browse/GEOT-6959) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

Correcting a wrong assumption that SimpleHttpClient didn't support proxy servers.

# Checklist

### DIFF
--- a/modules/library/http/src/test/java/org/geotools/http/HTTPClientFinderTest.java
+++ b/modules/library/http/src/test/java/org/geotools/http/HTTPClientFinderTest.java
@@ -119,8 +119,6 @@ public class HTTPClientFinderTest {
         try {
             HTTPClient client = HTTPClientFinder.createClient(HTTPProxy.class);
             assertTrue(client instanceof SimpleHttpClient);
-        } catch (HTTPFactoryException e) {
-            // Check message?
         } finally {
             if (nullInitially) {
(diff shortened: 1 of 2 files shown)


modules/library/http/src/test/java/org/geotools/http/HTTPClientFinderTest.java
```diff
@@ -118,8 +117,8 @@ public void createClientWithSystemProxyHost() throws Exception {
             System.setProperty("http.proxyHost", "http://proxy.dummy/");
         }
         try {
-            HTTPClientFinder.createClient(HTTPProxy.class);
-            Assert.fail("Should throw HTTPFactoryException exception.");
+            HTTPClient client = HTTPClientFinder.createClient(HTTPProxy.class);
+            assertTrue(client instanceof SimpleHttpClient);
```
aaime, line 121: Thanks. Is the try/catch still necessary? The test would pass if a ``HTTPFactoryException`` is thrown, I don't think it's what we want, right?

roarbra:
You're right. The catch shouldn't be there, but the try/finally is kept for resetting system property.

PR #3607, opened 2021-08-16
[GEOT-6960] Amend condition in which force decimals will be applied

prushforth:
[GEOT-6960](https://osgeo-org.atlassian.net/browse/GEOT-6960) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

Amend range in which force decimals is applied, add some tests to cover the changes, leaving existing tests intact. 

# Checklist

- [x] I have sent a [Contribution Licence Agreement](https://docs.geotools.org/latest/developer/procedures/contribution_license.html) (not required for small changes, e.g., fixing typos in documentation). (For GeoServer.  Do you require a separate agreement for GeoTools?  OTOH this could be considered a small change).

### DIFF
--- a/modules/library/xml/src/main/java/org/geotools/gml/producer/CoordinateFormatter.java
+++ b/modules/library/xml/src/main/java/org/geotools/gml/producer/CoordinateFormatter.java
@@ -99,7 +99,7 @@ public final class CoordinateFormatter {
     /** Formats a number with the configured number of decimals */
     public StringBuffer format(double x, StringBuffer sb) {
         String formatted;
-        if ((Math.abs(x) >= DECIMAL_MIN && x < DECIMAL_MAX) || x == 0) {
+        if ((Math.abs(x) >= DECIMAL_MIN && Math.abs(x) < DECIMAL_MAX) || x == 0) {
             x = truncate(x);
             long lx = (long) x;
             if (lx == x) {
--- a/modules/library/xml/src/test/java/org/geotools/gml/producer/CoordinateFormatterTest.java
+++ b/modules/library/xml/src/test/java/org/geotools/gml/producer/CoordinateFormatterTest.java
@@ -28,6 +28,54 @@ public class CoordinateFormatterTest {
     }
 
     @Test
+    public void testFormatScientificDECIMAL_MIN() {
+        CoordinateFormatter formatter = new CoordinateFormatter(6);
+        formatter.setPadWithZeros(true);
+        formatter.setForcedDecimal(true);
+        Assert.assertEquals("0.000010", formatter.format(0.00001));
+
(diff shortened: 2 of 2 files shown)


modules/library/xml/src/main/java/org/geotools/gml/producer/CoordinateFormatter.java
```diff
@@ -99,7 +99,7 @@ private String repeatZeros(int num) {
     /** Formats a number with the configured number of decimals */
     public StringBuffer format(double x, StringBuffer sb) {
         String formatted;
-        if ((Math.abs(x) >= DECIMAL_MIN && x < DECIMAL_MAX) || x == 0) {
+        if ((Math.abs(x) >= DECIMAL_MIN && Math.abs(x) < DECIMAL_MAX) || x == 0) {
```
jodygarnett, line 102: While this logic is correct, it does show an issue with truncate(x) below. 

If I understand correctly the following is incorrect:
```
scaled += 0.5;
```

And should be:

```
scaled += Math.signnum(x) * 0.5;
```

Still the above feedback goes a bit beyond what your PR offered to fix.

jodygarnett (approved):
While I noticed a small glitch in truncate that this PR makes visible the PR fixes the issue it intended to address, thank you for your contribution.

modules/library/xml/src/main/java/org/geotools/gml/producer/CoordinateFormatter.java
prushforth: I didn't delve into truncate. How did you notice the issue?

jodygarnett: I looked at testFormatScientificDECIMAL_MIN and noticed you did not test negative numbers, so I wondered if it would work and reviewed truncate.

My understanding may be wrong fo course.

jodygarnett:
Q: As this is a bug fix do you need it backported?

modules/library/xml/src/main/java/org/geotools/gml/producer/CoordinateFormatter.java
prushforth: I would / will make that change, but I'm not at my dev machine (for about a week).

prushforth:
> Q: As this is a bug fix do you need it backported?

No, I don't think so, I am waiting to commit code to GeoServer that depends on it, but there's no dependency at the moment.

jodygarnett:
Thanks, you are welcome to test truncate, as indicated I may be reading the code wrong.

prushforth:
Thanks, Jody!

PR #3612, opened 2021-08-26
[GEOT-6964] Java 17 compatibility

bjornharrtell:
[GEOT-6964](https://osgeo-org.atlassian.net/browse/GEOT-6964) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

  

# Checklist

### DIFF
--- a/modules/extension/xsd/xsd-core/pom.xml
+++ b/modules/extension/xsd/xsd-core/pom.xml
@@ -56,6 +56,11 @@
       </roles>
     </developer>
   </developers>
+
+  <properties>
+    <!-- TODO: remove once EsyMock is updated, see https://github.com/easymock/easymock/issues/274 -->
+    <test.otherJVMParams>--add-opens=java.base/java.lang=ALL-UNNAMED</test.otherJVMParams>
+  </properties>
   <!-- =========================================================== -->
   <!--     Dependency Management                                   -->
   <!-- =========================================================== -->
--- a/modules/extension/ysld/pom.xml
+++ b/modules/extension/ysld/pom.xml
@@ -33,6 +33,11 @@
     </developer>
   </developers>
 
+  <properties>
+    <!-- TODO: remove once EsyMock is updated, see https://github.com/easymock/easymock/issues/274 -->
+    <test.otherJVMParams>--add-opens=java.base/java.lang=ALL-UNNAMED</test.otherJVMParams>
+  </properties>
+
   <dependencyManagement>
     <dependencies>
       <dependency>
--- a/modules/unsupported/css/pom.xml
+++ b/modules/unsupported/css/pom.xml
@@ -45,6 +45,11 @@
     </developer>
   </developers>
 
+  <properties>
+    <!-- This is needed by Parboiled generated code, see https://github.com/sirthias/parboiled/issues/175 -->
+    <test.otherJVMParams>--add-opens=java.base/java.lang=ALL-UNNAMED</test.otherJVMParams>
+  </properties>
+
   <dependencies>
     <dependency>
       <groupId>org.geotools</groupId>
--- a/modules/unsupported/s3-geotiff/pom.xml
+++ b/modules/unsupported/s3-geotiff/pom.xml
@@ -42,6 +42,11 @@
     </developer>
   </developers>
 
+  <properties>
+    <!-- Allow old version of EHCache to work, I presume? This module is on the chopping block anyways -->
+    <test.otherJVMParams>--add-opens=java.base/java.lang=ALL-UNNAMED --add-opens=java.base/java.lang.ref=ALL-UNNAMED --add-opens=java.base/java.lang.reflect=ALL-UNNAMED --add-opens=java.base/java.util=ALL-UNNAMED --add-opens=java.base/java.io=ALL-UNNAMED --add-opens=java.base/java.util.concurrent=ALL-UNNAMED --add-opens=java.base/java.util.concurrent.locks=ALL-UNNAMED</test.otherJVMParams>
+  </properties>
+
   <dependencies>
     <dependency>
       <groupId>com.amazonaws</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -475,6 +475,7 @@
     <javadoc.maxHeapSize>1536M</javadoc.maxHeapSize>
     <test.forkMode>once</test.forkMode>
     <test.args></test.args>
+    <test.otherJVMParams></test.otherJVMParams>
     <src.output>${basedir}/target</src.output>
     <imageio.ext.version>1.3.10</imageio.ext.version>
     <jts.version>1.18.2</jts.version>
@@ -1647,7 +1648,7 @@
             <exclude>${test.exclude.pattern}</exclude>
             <exclude>${jdk11test.skip.pattern}</exclude>
           </excludes>
-          <argLine>-Xmx${test.maxHeapSize} ${jvm.opts} -Dfile.encoding=UTF-8 -Dorg.geotools.test.extensive=${extensive.tests} -Dorg.geotools.test.interactive=${interactive.tests} -Dorg.geotools.image.test.skip=${skip.image.tests} -Dorg.geotools.image.test.interactive=${interactive.image} -Djava.awt.headless=${java.awt.headless} -Dsun.java2d.d3d=${sun.java2d.d3d} -Djava.io.tmpdir="${java.io.tmpdir}" -Djava.library.path="${java.library.path}" ${test.args}  -XX:+TieredCompilation -XX:TieredStopAtLevel=1 -XX:+IgnoreUnrecognizedVMOptions -Djava.util.logging.config.file="${geotoolsBaseDir}"/${logging-profile}.properties --add-opens=java.base/java.lang=ALL-UNNAMED --add-opens=java.base/java.lang.ref=ALL-UNNAMED --add-opens=java.base/java.lang.reflect=ALL-UNNAMED --add-opens=java.base/java.util=ALL-UNNAMED --add-opens=java.base/java.io=ALL-UNNAMED --add-opens=java.base/java.util.concurrent=ALL-UNNAMED --add-opens=java.base/java.util.concurrent.locks=ALL-UNNAMED --add-exports=java.desktop/sun.awt.image=ALL-UNNAMED</argLine>
+          <argLine>-Xmx${test.maxHeapSize} ${jvm.opts} -Dfile.encoding=UTF-8 -Dorg.geotools.test.extensive=${extensive.tests} -Dorg.geotools.test.interactive=${interactive.tests} -Dorg.geotools.image.test.skip=${skip.image.tests} -Dorg.geotools.image.test.interactive=${interactive.image} -Djava.awt.headless=${java.awt.headless} -Dsun.java2d.d3d=${sun.java2d.d3d} -Djava.io.tmpdir="${java.io.tmpdir}" -Djava.library.path="${java.library.path}" ${test.args}  -XX:+TieredCompilation -XX:TieredStopAtLevel=1 -XX:+IgnoreUnrecognizedVMOptions -Djava.util.logging.config.file="${geotoolsBaseDir}"/${logging-profile}.properties --add-exports=java.desktop/sun.awt.image=ALL-UNNAMED ${test.otherJVMParams}</argLine>
           <!-- Ignores test failure only if we are generating a       -->
           <!-- report for publication on the web site. See the        -->
           <!-- profiles section at the beginning of this pom.xml file. -->


bjornharrtell:
The Downstream integration build fails due to that GeoServer also have a test using reflection to override a *private final static field*. Pretty nasty in my opinion and should use the new public API I introduce here. But not sure how to make that happen unless merging this and adapt GeoServer ASAP.

aaime:
@bjornharrtell just prepare twin PRs on the two projects, that build fine on your machine. Someone will confirm they do and then we can merge them together.

Btw, [this is one example of things I was expecting to break solid in Java 17](https://osgeo-org.atlassian.net/browse/GEOT-6967)... and they don't?

bjornharrtell:
> @bjornharrtell just prepare twin PRs on the two projects, that build fine on your machine. Someone will confirm they do and then we can merge them together.
> 
> Btw, [this is one example of things I was expecting to break solid in Java 17](https://osgeo-org.atlassian.net/browse/GEOT-6967)... and they don't?

Agree, that is suspect.. could it be that we somehow do not cover it in tests?

aaime:
@bjornharrtell maybe it's because [we're telling the compiler to target Java 8](https://github.com/geotools/geotools/blob/main/pom.xml#L1589)?
Maybe try locally to up the version number there to 17 and see... that said, even if we decide to drop support for Java 8, the target would become java 11, not 17.... I've tried that, with the Java 17 RC release, and... nothing, it builds fine. Duh?? :rofl:

bjornharrtell:
@aaime some more thought on this... it's reflection so it make sense that it doesn't cause compilation issues, however I would expect runtime issues covered by tests when run with the Java 17 JRE. Perhaps this will become more clear when I can invest som more time into the corresponding GeoServer issue with some more real life runtime experience.

fgdrf:
Maybe I can add some details here...

At runtime you will get Warnings such as reported in [GEOT-6967](https://osgeo-org.atlassian.net/browse/GEOT-6967) (in this case with java 11 runtime started with https://github.com/locationtech/udig-platform) for ImageWorker. In our scenario we havent't seen any other issues but as @aaime stated: Only code thats loaded at runtime logs warnings. And we definitely do not use everything from GeoTools.

So to get feedback for the intergation tests, its required to compile and test with at least maven.target.level=11

aaime:
Ah, found it, Maven was still using java 11 because I overrode the PATH, but not JAVA_HOME.
Here is the output of a build with java 17 and source/target set to 11:

```
[INFO] Running org.geotools.image.ImageWorkerTest
[ERROR] Tests run: 69, Failures: 0, Errors: 2, Skipped: 0, Time elapsed: 2.215 s <<< FAILURE! - in org.geotools.image.ImageWorkerTest
[ERROR] testJPEGWrite(org.geotools.image.ImageWorkerTest)  Time elapsed: 0.005 s  <<< ERROR!
java.lang.IllegalStateException: Class 'Unable to find JDK JPEG Writer' is illegal. It must be '{1}' or a derivated class.
	at org.geotools.image.ImageWorkerTest.testJPEGWrite(ImageWorkerTest.java:517)

[ERROR] testOpacityAlphaRGBDirect(org.geotools.image.ImageWorkerTest)  Time elapsed: 0.002 s  <<< ERROR!
java.lang.IllegalAccessError: class javax.media.jai.RasterAccessor (in unnamed module @0x2a48d10f) cannot access class sun.awt.image.BytePackedRaster (in module java.desktop) because module java.desktop does not export sun.awt.image to unnamed module @0x2a48d10f
	at org.geotools.image.ImageWorkerTest.testAlphaRGB(ImageWorkerTest.java:1256)
	at org
```

We need to get that writer spi instance, just allowing free SPI usage is to be avoided, ``ImageWorker.writeJPEG`` allows to choose which one to use at runtime (native vs JDK). We'll go there only if there is no other way, as that will break API and also [the GeoServer image writing configurations](https://docs.geoserver.org/master/en/user/configuration/image_processing/index.html).

However, it should be possible to get to the writer in other ways, for example, looping over the results of ``ImageIO.getImageWritersByMIMEType``, checking the class names, and then grabbing the originating provider.

bjornharrtell:
@aaime ok still find it strange that it is not seen with Java 17 in JAVA_HOME but source/target left as is.

bjornharrtell:
@aaime revisiting this, I still don't understand why it is/would be needed to compile to target Java 11 to get the runtime problems. And in either case we want to target Java 8 to be able to do this in not too far future.

aaime:
@bjornharrtell I'm just as clueless as you are... internet searches are giving me no hints either. Maybe ask on stackoverflow?

bjornharrtell:
@aaime I think I found that it's `--illegal-access=debug` that suppresses the problem. That's also a bit surprising since I thought the illegal-access flag was fully obsoleted in Java 17.

bjornharrtell:
I suppose `--illegal-access=debug` should be removed and all instances of `setAccessible` needs rework.. and then the SPI stuff. So lots of work. :)

bjornharrtell:
For reference, the obsoletion of `illegal-access` is described at https://openjdk.java.net/jeps/403.

bjornharrtell:
@aaime I think I've gotten it to work using `getImageWritersByMIMEType`. But looks like it needs to be done for all SPI stuff (but seems not exercised by tests). But I would like you to have a look at https://github.com/geotools/geotools/pull/3612/commits/20116577fb69593016e4edbf10acb89edb871d63 before I make a more proper rework of all SPI usage.

dromagnoli:
> @aaime I think I've gotten it to work using `getImageWritersByMIMEType`. But looks like it needs to be done for all SPI stuff. But I would like you to have a look at [2011657](https://github.com/geotools/geotools/commit/20116577fb69593016e4edbf10acb89edb871d63) before I make a more proper rework of all SPI usage.

I think it looks good.

bjornharrtell:
Finished for ImageWriter for https://github.com/geotools/geotools/pull/3612/commits/24862d9432ec707996f2014e32b4ff645adafa20.

But then there are test failures for the other way round in imagemosaic. And I find that imagemosaic properties point directly to internal Spi classes.. that seems a bit unfortunate (fx https://github.com/geotools/geotools/blob/06d379a2fdbfdbf94641c16813c37ef02aa7f63c/modules/plugin/imagemosaic/src/test/resources/org/geotools/gce/imagemosaic/test-data/remote_test/remote_test.properties#L13)

bjornharrtell:
Found that I could use `ServiceRegistry.lookupProviders` to resolve ImageReaderSpi from classnames. I guess it could be used to resolve ImageWriterSpi too.. instead of using ImageIO.getImageWritersByMIMEType. Opinions?

bjornharrtell:
A blocker to complete this might be that we need to wait for a newer release of easymock. (https://github.com/easymock/easymock/issues/274)

bjornharrtell:
I've reworked the SPI stuff to be a less intrusive change.

bjornharrtell:
Downstream integration build fails because it also uses an illegal trick to override a static final field in testing. It needs to adapt to use the new public API introduced here, done with https://github.com/geoserver/geoserver/pull/5319.

bjornharrtell:
I actually this this is complete now but needs local verification and coordinated merge with geoserver. @aaime and/or @dromagnoli are you interested in doing that?

aaime:
Wow great job, all tests with GeoTools passing! The downstream issue might indicate a change in behavior vs image reader choice (just guessing, I'm currently in bed with fever, writing from my phone)

aaime:
Oh ok, now saw the other comments about the downstream build.

aaime:
Did a quick pass... I have a hesitation about all the "add opens" added for tests. Maybe they are all for test code that does reflection things, but at the same time, don't they cover any other java 17 runtime issue? Tests are the only place where the code is actually run....

bjornharrtell:
@aaime yes, ideally tests should not require those add opens to verify other potential runtime issues with illegal reflection. But I think the amount of work to refactor them into something that can work without that is very large/difficult.

aaime:
I'm smelling a trap here! But I also so much want to stick my nose into it!


Found this actual issue... JAI cannot work on Java 17 without an open:

```
java.lang.IllegalAccessError: class javax.media.jai.RasterAccessor (in unnamed module @0x9e89d68) cannot access class sun.awt.image.BytePackedRaster (in module java.desktop) because module java.desktop does not export sun.awt.image to unnamed module @0x9e89d68

	at javax.media.jai.RasterAccessor.<init>(RasterAccessor.java:802)
	at com.sun.media.jai.opimage.CopyOpImage.computeRect(CopyOpImage.java:68)
	at javax.media.jai.PointOpImage.computeTile(PointOpImage.java:914)
	at com.sun.media.jai.util.SunTileScheduler.scheduleTile(SunTileScheduler.java:904)
	at javax.media.jai.OpImage.getTile(OpImage.java:1129)
	at javax.media.jai.PointOpImage.computeTile(PointOpImage.java:911)
	at com.sun.media.jai.util.SunTileScheduler.scheduleTile(SunTileScheduler.java:904)
	at javax.media.jai.OpImage.getTile(OpImage.java:1129)
	at javax.media.jai.RenderedOp.getTile(RenderedOp.java:2257)
	at org.geotools.image.ImageWorkerTest.testAlphaRGB(ImageWorkerTest.java:1256)
	at org.geotools.image.ImageWorkerTest.testOpacityAlphaRGBDirect(ImageWorkerTest.java:1185)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:103)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:63)
	at org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:79)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:329)
	at org.junit.runners.ParentRunner.access$100(ParentRunner.java:66)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:293)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:413)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
	at com.intellij.junit4.JUnit4IdeaTestRunner.startRunnerWithArgs(JUnit4IdeaTestRunner.java:69)
	at com.intellij.rt.junit.IdeaTestRunner$Repeater.startRunnerWithArgs(IdeaTestRunner.java:33)
	at com.intellij.rt.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:235)
	at com.intellij.rt.junit.JUnitStarter.main(JUnitStarter.java:54)
```

Long term I guess we can migrate to ImageN, and fix the direct reference to BytePackedRaster there... but I'm guessing it's just going to be one of many others.

So that one justifies the export. About the opens, two are needed for EasyMock, one for Parboiled, one I guess for an old version of EHCache we're using in s3-geotiff. Pushed a commit that makes these open a bit more surgical, and comments on the origin of the need.

bjornharrtell:
@aaime, makes sense and also makes it a bit more clear what might be long term goals.

bjornharrtell:
@aaime, I'm actually right now running GeoServer through IntelliJ IDEA and the org.geoserver.web.Start entry point with Java 17 + local GeoTools compilation pointing at the release configuration and things actually seem to work including WMS, WFS 2.0.0 and raster mosiac sources.

modules/plugin/coverage-multidim/netcdf/src/main/java/org/geotools/imageio/netcdf/utilities/NetCDFCRSUtilities.java
```diff
@@ -129,7 +129,7 @@
     public static final String CONVERT_AXIS_KM_KEY =
             "org.geotools.coverage.io.netcdf.convertAxis.km";
 
-    private static final boolean CONVERT_AXIS_KM;
+    private static boolean convertAxisKm;
```
aaime, line 132: Just thinking out loud here... if the name is left as CONVERT_AXIS_KM then the existing code in GeoServer would not need to be changed right? This would allow starting to merge the GeoTools part without having to go and also have a GeoServer PR ready (the GeoServer PR for Java 17 would still use the new setter).

bjornharrtell: Not sure that would work because the removal of final

bjornharrtell: Don't have to finish/merge the full GeoServer Java 17 thing, could just merge this specific one in coordination - https://github.com/geoserver/geoserver/pull/5319.

aaime:
Re-build everything with Java 8, indeed both builds pass. Merging. Thanks!

PR #3623, opened 2021-08-31
[GEOT-6965] encoding a polygon as a multipolygon gives an unwanted du…

NielsCharlier:
[GEOT-6965](https://osgeo-org.atlassian.net/browse/GEOT-6965) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

…plicated srsname tag in polygon

  

# Checklist

### DIFF
--- a/modules/extension/xsd/xsd-gml3/src/test/java/org/geotools/gml3/bindings/MultiPolygonTypeBindingTest.java
+++ b/modules/extension/xsd/xsd-gml3/src/test/java/org/geotools/gml3/bindings/MultiPolygonTypeBindingTest.java
@@ -78,7 +78,7 @@ public class MultiPolygonTypeBindingTest extends GML3TestSupport {
         geometry.setUserData(CRS.decode("EPSG:4326"));
         GML3EncodingUtils.setID(geometry, "geometry");
         Document dom = encode(geometry, GML.MultiPolygon);
-        print(dom);
+        // print(dom);
         assertEquals(
                 "urn:x-ogc:def:crs:EPSG:4326",
                 dom.getElementsByTagNameNS(GML.NAMESPACE, "MultiPolygon")
--- a/modules/extension/xsd/xsd-gml3/src/test/java/org/geotools/gml3/bindings/MultiPolygonTypeBindingTest.java
+++ b/modules/extension/xsd/xsd-gml3/src/test/java/org/geotools/gml3/bindings/MultiPolygonTypeBindingTest.java
@@ -18,10 +18,12 @@ package org.geotools.gml3.bindings;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 
 import org.geotools.gml3.GML;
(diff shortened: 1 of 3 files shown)


modules/extension/xsd/xsd-gml3/src/test/java/org/geotools/gml3/bindings/MultiPolygonTypeBindingTest.java
```diff
@@ -69,4 +71,25 @@ public void testEncode() throws Exception {
...
         assertEquals("geometry.2", getID(children.item(1)));
     }
+
+    @Test
+    public void testEncodePolygon() throws Exception {
+        Geometry geometry = GML3MockData.polygon();
+        geometry.setUserData(CRS.decode("EPSG:4326"));
+        GML3EncodingUtils.setID(geometry, "geometry");
+        Document dom = encode(geometry, GML.MultiPolygon);
+        print(dom);
```
mprins, line 81: print should probably be commented
```suggestion
        // print(dom);
```

mprins (approved):
LGTM

modules/library/main/src/main/java/org/geotools/data/util/GeometryTypeConverterFactory.java
```diff
@@ -162,8 +162,16 @@ else if (source instanceof GeometryCollection)
                     Polygon[] polygons;
                     // NC - Empty Geometry Support
                     if (sourceGeometry.isEmpty()) polygons = new Polygon[0];
-                    else if (source instanceof Polygon) polygons = new Polygon[] {(Polygon) source};
-                    else if (source instanceof GeometryCollection)
+                    else if (source instanceof Polygon) {
+                        // copy polygon
+                        Polygon polygon = (Polygon) source;
```
aaime, line 167: I'm guessing the objective of this copy is to eliminate the srid from the polygon definition?
However the converter is in main, will affect all code using converters, not only GML encoding.
No test is failing, so apparently, either nothing is using that information, or the eventual code path is not tested.
It's not wrong, but I would have fled more comfortable seeing this in the GML encoding modules (limiting the scope of this change).

NielsCharlier: I understand your concern, but only the CRS in the user data of an entire geometry is ever used, never of subgeometries. If you take a multipolygon directly from database or WKT, the CRS is also only set at the root level, so you cannot rely on it being anywhere else either way. (Otherwise we'd see srsname tags pop-up everywhere in subgeometries.) I think it is consistent to always keep the CRS in the root level user data. (Conversely, if we convert from multipolygon to polygon, the converter will set user data in the returned value.)

Changing the encoder would be a fair bit harder because of the way it is implemented with bindings.

aaime: Since the issue I reported is hypothetical for the moment, let's merge and see how it goes.

aaime:
@NielsCharlier do you want to backport?

NielsCharlier:
No need, thanks

PR #3634, opened 2021-09-10
[GEOT-6948] MarkFactory performance optimization hints support

fernandor777:
[GEOT-6948](https://osgeo-org.atlassian.net/browse/GEOT-6948) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

This PR introduces support for filtering and ordering hints on mark factories during rendering.

JIRA ticket:
https://osgeo-org.atlassian.net/browse/GEOT-6948
Proposal:
https://github.com/geotools/geotools/wiki/Rendering-pre-process-Mark-Factories-Hint

# Checklist

### DIFF
--- a/modules/library/render/src/main/java/org/geotools/renderer/style/DynamicSymbolFactoryFinder.java
+++ b/modules/library/render/src/main/java/org/geotools/renderer/style/DynamicSymbolFactoryFinder.java
@@ -17,8 +17,11 @@
 package org.geotools.renderer.style;
 
 import java.util.Arrays;
+import java.util.Comparator;
 import java.util.Iterator;
+import java.util.function.Predicate;
 import java.util.logging.Logger;
+import java.util.stream.Stream;
 import org.geotools.util.factory.FactoryCreator;
 import org.geotools.util.factory.FactoryRegistry;
 import org.geotools.util.factory.Hints;
@@ -47,6 +50,17 @@ public final class DynamicSymbolFactoryFinder {
     protected static final Logger LOGGER =
             org.geotools.util.logging.Logging.getLogger(DynamicSymbolFactoryFinder.class);
 
+    /**
+     * Rendering hint key to store a {@link Comparator}<{@link MarkFactory}> instance to set the
+     * {@link MarkFactory} execution order.
+     */
+    public static final Hints.Key MARK_FACTORY_ORDER = new Hints.Key(Comparator.class);
+    /**
+     * Rendering hint key to store a {@link Predicate}<{@link MarkFactory}> instance to set the
+     * allowed {@link MarkFactory} instances to be evaluated.
+     */
+    public static final Hints.Key MARK_FACTORY_FILTER = new Hints.Key(Predicate.class);
+
     /** The service registry for this manager. Will be initialized only when first needed. */
     private static volatile FactoryRegistry registry;
 
@@ -62,6 +76,29 @@ public final class DynamicSymbolFactoryFinder {
      */
     public static synchronized Iterator<MarkFactory> getMarkFactories() {
         return getServiceRegistry().getFactories(MarkFactory.class, null, null).iterator();
+    }
+
+    /**
+     * Finds all implementations of {@link MarkFactory} which have registered using the services
+     * mechanism. Returns a filtered and ordered iterator based on the provided Hints input
+     * parameter.
+     *
+     * @param hints An optional map of hints for factory configuration, or {@code null} if none.
+     *     Allowed Hints are: {@link DynamicSymbolFactoryFinder.MARK_FACTORY_ORDER}, {@link
+     *     DynamicSymbolFactoryFinder.MARK_FACTORY_FILTER}
+     * @return A filtered and ordered iterator over all discovered datastores which have registered
+     *     factories, and whose available method returns true.
+     */
+    @SuppressWarnings("unchecked")
+    public static synchronized Iterator<MarkFactory> getMarkFactories(Hints hints) {
+        Comparator<MarkFactory> sort =
+                hints != null ? (Comparator<MarkFactory>) hints.get(MARK_FACTORY_ORDER) : null;
+        Predicate<MarkFactory> filter =
+                hints != null ? (Predicate<MarkFactory>) hints.get(MARK_FACTORY_FILTER) : null;
+        Stream<MarkFactory> factories =
+                getServiceRegistry().getFactories(MarkFactory.class, filter, hints);
+
+        return sort != null ? factories.sorted(sort).iterator() : factories.iterator();
     }
 
     /**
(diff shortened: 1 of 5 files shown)


modules/library/render/src/main/java/org/geotools/renderer/style/MarkFactoryListComparator.java
```diff
@@ -0,0 +1,46 @@
...
+    private final List<String> markFactoryList;
+
+    public MarkFactoryListComparator(List<String> markFactoryList) {
+        this.markFactoryList = markFactoryList;
+    }
+
+    @Override
+    public int compare(MarkFactory mf1, MarkFactory mf2) {
+        Integer index1 = getIndex(mf1);
+        Integer index2 = getIndex(mf2);
```
jodygarnett, line 34: No need for boxing, or am I missing something:
```java
int index1 = getIndex(mf1);
int index2 = getIndex(mf2);
return Integer.compare(index1, index2);
```

modules/library/render/src/main/java/org/geotools/renderer/style/MarkFactoryListPredicate.java
```diff
@@ -0,0 +1,36 @@
...
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ *    Lesser General Public License for more details.
+ */
+package org.geotools.renderer.style;
+
+import java.util.List;
+import java.util.function.Predicate;
+
+/** Mark factory filtering predicate based on a provided class name identifiers list. */
+public class MarkFactoryListPredicate implements Predicate<MarkFactory> {
```
jodygarnett, line 23: Excellent :)

I am not sure there is anything specific about MarkFactory in this class, since it is just checking classname against a list.

You may be able to set this up as  a generic `org.geotools.util.factory.FactoryListPredicate<? extends Factory>` since this is a nice understandable way of filtering.

jodygarnett (approved):
Change looks great, nice and clean with minimal impact.

Although it goes beyond your use-case you could consider making the two implementations available for use with other factories:

modules/library/render/src/main/java/org/geotools/renderer/style/MarkFactoryListComparator.java
fernandor777: Good point, fixing

modules/library/render/src/main/java/org/geotools/renderer/style/MarkFactoryListPredicate.java
fernandor777: The problem on making the filter class more generic is MarkFactory does not extend Factory or any other interface, so I need to make MarkFactory as type hierarchy root entity.
Also I tried to move the class to gt-metadata but since MarkFactory is a gt-render type the compilation fails on that type resolution.

jodygarnett: I am sure a generic `org.geotools.util.factory.FactoryListPredicate<T> implements Predicate<T> ` would work fine then.

I just assumed factory because of the name.

modules/library/render/src/main/java/org/geotools/renderer/style/DynamicSymbolFactoryFinder.java
```diff
@@ -47,6 +50,9 @@
     protected static final Logger LOGGER =
             org.geotools.util.logging.Logging.getLogger(DynamicSymbolFactoryFinder.class);
 
+    public static final Hints.Key MARK_FACTORY_ORDER = new Hints.Key(Comparator.class);
```
aaime, line 53: Bit of javadoc for the new public fields?

modules/library/render/src/main/java/org/geotools/renderer/style/DynamicSymbolFactoryFinder.java
```diff
@@ -64,6 +70,26 @@ private DynamicSymbolFactoryFinder() {}
         return getServiceRegistry().getFactories(MarkFactory.class, null, null).iterator();
     }
 
+    /**
+     * Finds all implementations of {@link MarkFactory} which have registered using the services
+     * mechanism.
+     *
+     * @param hints An optional map of hints for factory configuration, or {@code null} if none.
```
aaime, line 77: Tell the user which hints could be useful here.

modules/library/render/src/main/java/org/geotools/renderer/style/MarkFactoryListComparator.java
```diff
@@ -0,0 +1,44 @@
...
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ *    Lesser General Public License for more details.
+ */
+package org.geotools.renderer.style;
+
+import java.util.Comparator;
+import java.util.List;
+
+/** Mark factory comparator based on a simple class name identifiers list. */
+public class MarkFactoryListComparator implements Comparator<MarkFactory> {
```
aaime, line 23: Does it need to be public?

modules/library/render/src/main/java/org/geotools/renderer/style/MarkFactoryListPredicate.java
aaime: Does it need to be public?

aaime (changes requested):
Looks good, see feedback on minor issues.

modules/library/render/src/main/java/org/geotools/renderer/style/DynamicSymbolFactoryFinder.java
fernandor777: Makes sense, adding them.

modules/library/render/src/main/java/org/geotools/renderer/style/DynamicSymbolFactoryFinder.java
fernandor777: Added allowed hints.

modules/library/render/src/main/java/org/geotools/renderer/style/MarkFactoryListComparator.java
fernandor777: Yes, this class is used on GeoServer rendering package to inject the Comparator hint.

aaime (approved):

aaime:
Weird, I cannot find GEOT-6984:
image

At first, I thought it was an issue with a typo, but I cannot find GeoTools issues for markFactory at all. @fernandor777 do you know what is going on? :-D We need a ticket for the release notes.

PR #3642, opened 2021-09-20
[GEOS-10146] app-schema support for multiple geometries with differen…

NielsCharlier:
…t CRS

https://osgeo-org.atlassian.net/browse/GEOS-10146

Note: I first attempted to refactor ComplexFeatureRegistry to plug-in multiple geometries, as well as support the reprojection of multiple geometries in app-schema
The first problem is that in complex features, there are circular type references, which can be resolved by proxies. However, the more fundamental issues is that a certain complex types referenced from different paths can have different CRS on those different paths, and there is no way to resolve this except for taking the correct CRS from inside the feature data rather than attempting to keep it in the feature type. This lead to the approach I eventually followed which also required a lot less changes.
  

# Checklist


modules/extension/app-schema/app-schema/src/main/java/org/geotools/data/complex/filter/XPath.java
```diff
@@ -276,11 +276,8 @@ public Attribute set(
                                             new GeometryTypeImpl(
                                                     targetNodeType.getName(),
                                                     targetNodeType.getBinding(),
-                                                    crs != null
-                                                            ? crs
-                                                            : ((GeometryDescriptor)
-                                                                            actualDescriptor)
-                                                                    .getCoordinateReferenceSystem(),
+                                                    ((GeometryDescriptor) actualDescriptor)
```
NielsCharlier, line 279: Note: the CRS passed in the constructor of XPath is now only used as a back-up, when there is no CRS feature data. Otherwise, the CRS from the feature data will be used (see AttributeBuilder).

modules/library/main/src/main/java/org/geotools/feature/AttributeBuilder.java
```diff
@@ -214,15 +215,15 @@ public void setCRS(CoordinateReferenceSystem crs) {
...
-            Object userData = ((Geometry) geom).getUserData();
-            if (userData != null && userData instanceof CoordinateReferenceSystem) {
-                return (CoordinateReferenceSystem) userData;
+        if (geom != null && geom instanceof Geometry) {
+            // the geometry in the geometry itself is preferred
+            // to support multiple geometries/CRS
+            CoordinateReferenceSystem featureCrs = GeometryUtil.getCRS((Geometry) geom);
+            if (featureCrs != null) {
+                return featureCrs;
             }
```
NielsCharlier, line 224: Note: this is the main change in app-schema. The CRS in the data itself will be used if it is available.

taba90:
The pr looks in general good to me. However the changes are spreaded across different module, and I'm not comfortable in approving the whole pr.

modules/extension/xsd/xsd-gml2/src/main/java/org/geotools/gml2/bindings/GML2EncodingUtils.java
```diff
@@ -128,29 +128,6 @@ public static String toURI(CoordinateReferenceSystem crs, SrsSyntax srsSyntax) {
         return null;
     }
 
-    /**
-     * Determines the crs of the geometry by checking {@link Geometry#getUserData()}.
-     *
-     * <p>This method returns <code>null</code> when no crs can be found.
-     */
-    public static CoordinateReferenceSystem getCRS(Geometry g) {
```
aaime, line 136: Direct removal of public API is not acceptable, the process is:
- Provide an explanation as to why the method is problematic in the PR.
- Deprecate
- Eventually provide instructions in the upgrade guide

In this case, it seems there has been a centralization of the method to ``gt-main`` new ``GeometryUtil``, and in a similar manner, centralization of duplicated code to set the CRS in a Geometry. Which is a good thing!
In this case you should still keep the original method, deprecated it, and delegate to ``GeometryUtil`` instead.

Also wondering if the existing ``JTS`` class could not be a suitable target for that method, rather than creating a new utility class (it's already dealing with CRS objects for other reasons).
@jodygarnett might have an opinion on this.

modules/library/main/src/main/java/org/geotools/data/crs/ReprojectFeatureReader.java
```diff
@@ -109,18 +118,37 @@ public ReprojectFeatureReader(
...
+        this.schema = FeatureTypes.transform(originalType, cs);
+
         this.reader = reader;
-        transformer.setMathTransform(CRS.findMathTransform(original, cs, true));
+        this.transformers = new HashMap<>();
+        for (int i = 0; i < originalType.getDescriptors().size(); i++) {
+            if (originalType.getDescriptor(i) instanceof GeometryDescriptor) {
+                GeometryDescriptor descr = (GeometryDescriptor) originalType.getDescriptor(i);
+                CoordinateReferenceSystem original = descr.getCoordinateReferenceSystem();
+                if (CRS.isCompatible(cs, original)) {
```
aaime, line 136: And if not compatible... it does nothing? Shouldn't it be throwing an exception instead? Please discuss :-)

modules/library/main/src/main/java/org/geotools/data/crs/ReprojectFeatureReader.java
```diff
@@ -162,9 +190,19 @@ public SimpleFeature next()
         Object[] attributes = next.getAttributes().toArray();
 
         try {
-            for (int i = 0; i < attributes.length; i++) {
-                if (attributes[i] instanceof Geometry) {
-                    attributes[i] = transformer.transform((Geometry) attributes[i]);
+            for (int i = 0; i < schema.getDescriptors().size(); i++) {
+                if (schema.getDescriptor(i) instanceof GeometryDescriptor
+                        && attributes[i] != null) {
```
aaime, line 195: Since ``attributes[i]`` is cast to Geometry a few lines down, the check for the value being a Geometry, rather than not null, is safer (I do realize a geometry description should have a Geometry value).

modules/library/main/src/main/java/org/geotools/feature/AttributeBuilder.java
aaime: Does not seem like a big change in behavior, the old code was already picking the CRS from the Geometry value... The changed code introduces a variation, will pick the CRS also from a map in the userdata. Does not seem problematic.

modules/library/main/src/main/java/org/geotools/feature/AttributeBuilder.java
```diff
@@ -214,15 +215,15 @@ public void setCRS(CoordinateReferenceSystem crs) {
...
     /** @return The coordinate reference system of the feature, or null if not set. */
     public CoordinateReferenceSystem getCRS(Object geom) {
-        if (crs != null) {
-            return crs;
-        } else if (geom != null && geom instanceof Geometry) {
-            Object userData = ((Geometry) geom).getUserData();
-            if (userData != null && userData instanceof CoordinateReferenceSystem) {
-                return (CoordinateReferenceSystem) userData;
+        if (geom != null && geom instanceof Geometry) {
+            // the geometry in the geometry itself is preferred
```
aaime, line 219: "the geometry in the geometry itself"

Something seems to have gone wrong here :-D
You probably wanted to say, the "CRS in the geometry itself".

The statement "to support multiple geometries/CRS" also confuses me... this one builds a single attribute, and the PR is trying to allow different CRSs in different attributes of the same feature... not different CRSs inside the same column, or not? From the ticket "multiple geometries in features, particularly if they have different CRS."

Ah, another thing.. this is a GeoTools change, please have a GeoTools ticket describing it, and refer to it from the commit message. Not all use GeoTools through GeoServer, a ticket in the GeoTools tracker ends up in the release notes and informs the direct GeoTools users too.

modules/library/main/src/main/java/org/geotools/feature/FeatureTypes.java
```diff
@@ -344,7 +345,11 @@ public static SimpleFeatureType transform(
                 GeometryDescriptor geometryType = (GeometryDescriptor) attributeType;
 
                 tb.descriptor(geometryType);
-                if (!forceOnlyMissing || geometryType.getCoordinateReferenceSystem() == null) {
+
+                if (geometryType.getCoordinateReferenceSystem() == null
```
aaime, line 349: I don't understand this change... looking at it, it also feels like some parenthesis may be missing to get the desired execution order (or maybe not, but parenthesis could clarify intention).

modules/library/referencing/src/main/java/org/geotools/referencing/CRS.java
```diff
@@ -266,6 +266,25 @@ private static boolean isForcedLonLat() {
         return FORCED_LON_LAT.get();
     }
 
+    /**
+     * Determines if two coordinate reference systems are compatible with each other, i.e. can be
+     * converted from one to another in a meaningful way.
+     *
+     * <p>They must have the same dimension (see OGC 09-025r2).
```
aaime, line 273: Nope! I believe there are some code paths allowing it to go from a 2D+1 CRS down to its equivalent 2D (basically just stripping the Z) when the 2D part is common to the two. And while others are missing today, they could be added tomorrow or be present in specific usage scenarios (transformations are pluggable).
It's probably more robust to go and search for a math transform and catch an exception if not found... since catching an exception can be expensive, would likely be a good idea to first check if the two CRSs are not already ``equalsIgnoreMetadata``.

aaime (changes requested):

modules/library/main/src/main/java/org/geotools/feature/AttributeBuilder.java
NielsCharlier: It seems that most simple feature vector datastore implementations (at least all of the JDBC, as well as property files) encode the CRS from the data directly into the geometry. This mechanism supports different columns with each their own CRS.

 App-schema however assumes a single CRS and this is the place where it is written into each geometry field of the final type/descriptor structure that is served to the encoder. Previously, the CRS inside the geometry was only used as a fallback in case the single CRS attribute is empty  (which is in practice, never) . The order of preference is changed here so that the CRS attribute becomes the fall-back mechanism (in case there was no CRS inside the data), and in this way, the different CRS in the different geometry columns are passed along into the complex type.

modules/library/main/src/main/java/org/geotools/feature/AttributeBuilder.java
NielsCharlier: The big change in behaviour is the order of preference (see above) which prevents a single CRS to be used for each geometry.

modules/library/main/src/main/java/org/geotools/feature/FeatureTypes.java
NielsCharlier: it is correct but the second part of the disjunction could have optional parentheses. If geometryType.getCoordinateReferenceSystem() == null it cannot be checked if they are compatible (one is null).
If it isn't null, it should be checked if they are compatible but it should also be checked that the forceOnlyMissing flag is not on.

modules/extension/xsd/xsd-gml2/src/main/java/org/geotools/gml2/bindings/GML2EncodingUtils.java
NielsCharlier: Okay. JTS seems indeed be a good choice.

modules/library/main/src/main/java/org/geotools/data/crs/ReprojectFeatureReader.java
NielsCharlier: I was unsure myself but based this behaviour on the OGC standard. In "OGC® Web Feature Service 2.0 Interface Standard – With Corrigendum" 7.9.2.4.4 srsName parameter (https://docs.opengeospatial.org/is/09-025r2/09-025r2.html#95) 

"The optional srsName attribute may be used to assert a specific WFS-supported CRS transformation to be applied to the compatible (see 7.9.2.4.4.2) geometries of the features returned in a response document."

And then further in 7.9.2.4.4.2 Compatibility of coordinate reference systems:

"Coordinate reference systems that differ in dimensionality or that have no defined datum transformation between them may be incompatible for the purposes of geometry coordinate conversion. Where the request srsName references a CRS that is incompatible with the storage CRS of a geometry property, the service shall encode that geometry property using some other CRS identified with an srsName attribute on that property"

This seems to suggest to do nothing indeed for those geometries that are incompatible with the reprojection.

It seems that many "Inspire" schemas are requiring complex feature types with both a (primary) 2D and a (secondary) 3D geometry inside them. If an exception was thrown, any 2D reprojection applied to one of these schemas would fail. This is particularly problematic because in geoserver the implementations of some versions of WFS pass on a reprojection CRS in the Query always (even if no srsName was specified in the http request, it has something to do with the different ways different wfs versions handle crs encoding and axis order), so that all requests would fail.

modules/library/main/src/main/java/org/geotools/data/crs/ReprojectFeatureReader.java
NielsCharlier: Ok.

modules/library/referencing/src/main/java/org/geotools/referencing/CRS.java
NielsCharlier: Okay. The specs I referred to elsewhere was not very clear about compatibility, and mentioned only different dimensions as a clear example. I indeed wanted to avoid it being expensive. Especially because this method is called from three different places (reprojecting the type as well as the feature, and first even to check if a reprojectfeaturereader should be created at all). It would be ideal if we only attempted to create the mathtransform once and caught the exception once. I will look into it.

It should in any case then be made more clear then that the order of arguments in this method matters.

NielsCharlier:
Thanks for the reviews @taba90 and @aaime . I believe all the requested changes have been made (in a separate commit)

PR #3644, opened 2021-09-27
[GEOT-6988] Add process to provide a centre line to label irregular p…

ianturton:
…olygons

  

# Checklist

### DIFF
--- a/modules/unsupported/process-geometry/src/main/java/org/geotools/process/geometry/CentreLine.java
+++ b/modules/unsupported/process-geometry/src/main/java/org/geotools/process/geometry/CentreLine.java
@@ -1,0 +1,146 @@
+package org.geotools.process.geometry;
+
+import java.util.ArrayList;
+import java.util.List;
+import org.geotools.data.DataUtilities;
+import org.geotools.data.simple.SimpleFeatureCollection;
+import org.geotools.data.simple.SimpleFeatureIterator;
+import org.geotools.feature.simple.SimpleFeatureBuilder;
+import org.geotools.feature.simple.SimpleFeatureTypeBuilder;
+import org.geotools.geometry.jts.JTS;
+import org.geotools.graph.build.line.LineStringGraphGenerator;
+import org.geotools.graph.path.DijkstraShortestPathFinder;
+import org.geotools.graph.path.Path;
+import org.geotools.graph.structure.Edge;
+import org.geotools.graph.structure.Graph;
+import org.geotools.graph.structure.Node;
+import org.geotools.graph.traverse.standard.DijkstraIterator.EdgeWeighter;
+import org.locationtech.jts.geom.Geometry;
+import org.locationtech.jts.geom.GeometryFactory;
+import org.locationtech.jts.geom.LineString;
+import org.locationtech.jts.geom.MultiLineString;
+import org.locationtech.jts.simplify.TopologyPreservingSimplifier;
+import org.opengis.feature.simple.SimpleFeature;
+import org.opengis.feature.simple.SimpleFeatureType;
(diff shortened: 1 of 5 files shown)


modules/unsupported/process-geometry/src/main/java/org/geotools/process/geometry/CentreLine.java
```diff
@@ -0,0 +1,163 @@
...
+import org.locationtech.jts.geom.Geometry;
+import org.locationtech.jts.geom.GeometryFactory;
+import org.locationtech.jts.geom.LineString;
+import org.locationtech.jts.geom.MultiLineString;
+import org.locationtech.jts.simplify.TopologyPreservingSimplifier;
+import org.opengis.feature.simple.SimpleFeature;
+import org.opengis.feature.simple.SimpleFeatureType;
+import org.opengis.feature.type.AttributeDescriptor;
+
+public class CentreLine {
```
aaime, line 44: UK attack!
https://www.grammar.com/center_vs._centre

modules/unsupported/process-geometry/src/main/java/org/geotools/process/geometry/CentreLineProcess.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    @DescribeProcess(
+        title = "Centre Line with tolerance",
+        description = "Extract Centre Line of a Polygon"
+    )
+    @DescribeResult(
+        description = "A geometry that is the centre line (skeleton) of the input polygon"
+    )
+    public static Geometry centreLineWithTolerance(
```
aaime, line 63: Just thinking out loud... what about making tolerance a Double and mark it as optional instead?

modules/unsupported/process-geometry/src/main/java/org/geotools/process/geometry/Skeletonize.java
```diff
@@ -0,0 +1,123 @@
...
+import org.locationtech.jts.geom.LineString;
+import org.locationtech.jts.geom.LinearRing;
+import org.locationtech.jts.geom.MultiLineString;
+import org.locationtech.jts.geom.MultiPolygon;
+import org.locationtech.jts.geom.Polygon;
+import org.locationtech.jts.operation.linemerge.LineMerger;
+import org.locationtech.jts.simplify.TopologyPreservingSimplifier;
+import org.locationtech.jts.triangulate.VoronoiDiagramBuilder;
+
+public class Skeletonize {
```
aaime, line 39: Does it need to be public? Probably not since all methods are package private?

aaime (commented):
Interesting work, looking forward to try it out.

modules/unsupported/process-geometry/src/main/java/org/geotools/process/geometry/Skeletonize.java
ianturton: I am intending to add a process for this one too, some people might want the actual skeleton.

modules/unsupported/process-geometry/src/main/java/org/geotools/process/geometry/CentreLineProcess.java
ianturton: After 25 years I can no longer spell `color` in the British way but seems like `centre` is hardcoded into my fingers.

modules/unsupported/process-geometry/src/main/java/org/geotools/process/geometry/CentreLine.java
ianturton: After 25 years I can no longer spell `color` in the British way but seems like `centre` is hardcoded into my fingers.

PR #3646, opened 2021-09-29
[GEOT-6989] Replace method addTileToCache of TileService

roarbra:
[GEOT-6989](https://osgeo-org.atlassian.net/browse/GEOT-6989) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

The function addTileToCache isn't used properly. The purpose is to add the incoming tile to a cache if it's not there, and then return the tile that's in the cache. Which should be the one used thereafter. It doesn't seem to be used in such a way.

My suggestion is to use a different name, and use a tile as an argument, but the tile identifier which is actually what's being used in cache-lookup.

Tile addTileToCache(Tile tile)

replaced with

Tile obtainTile(TileIdentifier identifier)

To accomplish this I also added a new function identifyTileAtCoordinate, that resembles the function findTileAtCoordinate. The only difference being that findTileAtCoordinate finishes off with creating a Tile by the found TileIdentifier.

# Checklist

### DIFF
--- a/modules/extension/tile-client/src/main/java/org/geotools/tile/TileService.java
+++ b/modules/extension/tile-client/src/main/java/org/geotools/tile/TileService.java
@@ -270,11 +270,10 @@ public abstract class TileService implements ImageLoader {
         Set<Tile> tileList = new HashSet<>(100);
 
         // Let's get the first tile which covers the upper-left corner
-        Tile firstTile =
-                tileFactory.findTileAtCoordinate(
-                        extent.getMinX(), extent.getMaxY(), zoomLevel, this);
+        TileIdentifier identifier =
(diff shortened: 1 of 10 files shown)


modules/extension/tile-client/src/main/java/org/geotools/tile/TileService.java
```diff
@@ -349,6 +352,7 @@ public BufferedImage loadImageTileImage(Tile tile) throws IOException {
      * <p>Subclasses services (such as WMTS) may have some more hints about the tile TTL, so a more
      * controllable cache should be implemented in these cases.
      */
+    @Deprecated
```
ianturton, line 355: I think we can probably skip deprecation for a protected method - no one we don't know about is likely to be using it?

ianturton (approved):

PR #3647, opened 2021-10-01
[GEOT-6990] TWKB VarInt decoding for PostGIS is limited to 32-bit

guischulz:
Use 64-bit for reading double values from VarInts when parsing TWKB
https://osgeo-org.atlassian.net/browse/GEOT-6990

# Checklist


modules/plugin/jdbc/jdbc-postgis/src/test/java/org/geotools/data/postgis/TWKBReaderTest.java
```diff
@@ -55,6 +55,27 @@ public void testPointGeometries() throws ParseException, IOException {
...
+                "220002b89a9309d6b8f93ce40103",
+                "LINESTRING(959452.4 6390941.9,959463.8 6390941.7)",
+                2);
+
+        checkTWKBGeometry(
+                "420002c088c05be6b6bee104ea1127",
+                "LINESTRING(959452.48 6390941.95,959463.89 6390941.75)",
+                2);
+
+        checkTWKBGeometry(
```
aaime, line 73: In the end, there are few decimals and a small number of digits before the separator too.
The unscaled number ``959452483``` can be represented as a 32-bit int... would have this test actually failed with the old code? Seems at least an extra digit is needed to make it go beyond the reach of 32-bit integers.

guischulz: The old code fails for the unscaled number {{6390941959}} in the last assertion of the test.

PR #3651, opened 2021-10-08
[GEOT-6995] StreamingRenderer - Exception when shapefile features (3D points) get loaded into memory and displayed in a 2D coordinate system

ianturton:
[GEOT-6995](https://osgeo-org.atlassian.net/browse/GEOT-6995) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

  

# Checklist


aaime:
Having a look and... so much going off the rails:
* The CRS claims 2d data, but data returned is 3d + measure
* The feature source is given a FEATURE_2D hint, but it ends up being completely ignored because the map was fed a feature collection, which is not re-queriable (FeatureCollection.subCollection(Query) does not exist!), but just filterable... so the hints that would skip reading the Z and M are never getting to the data source. To add insult to the injury, it also means all DBF attributes are going to be read even if none is used while rendering.

All stems from giving the layer a feature collection instead of a feature source. Take aways:
- Can be fixed easily by using a feature source instead
- FeatureCollection is a pretty broken intermediate object, it could be fixed by adding ``subCollection(Query)``, and then properly implement it around in ContentFeatureCollection and all common wrappers (quite a bit of work yup)
- Given the information we have about the coordiantes is wrong, LiteCoordinateSequence should not trust its inputs and do less efficient code paths for sequences that are mis-matched from the upper level metadata.

aaime:
Pushed a tentative fix at the LiteCoordinateSequence level.
The class was designed in such a way that it would handle dimensions, but not measures... at some point, measures were brought into the picture, possibly by changes in JTS, but without really updating LiteCoordinateSequence to handle measures (just to make it compile). The change adds a little bit of direct support for measures while cloning data.

aaime:
@ianturton what do you think?

ianturton:
looks good to me.

PR #3654, opened 2021-10-16
[GEOT-7001] XmlComplexFeatureParser gives wrong name for ComplexAttribute

roarbra:
[GEOT-7001](https://osgeo-org.atlassian.net/browse/GEOT-7001) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

When accessing a WFS with a complex structure a single feature shouldn't have properties with names other than those reported by a call to getSchema().

That is the basic of my added test in KartverketStedsnavnDataStoreOnlineTest.

Unfortunately a lot of tests were based on the wrong assumption about how to name complex attributes.

# Checklist

### DIFF
--- a/docs/src/main/java/org/geotools/wfs/ReadWFSKommuner.java
+++ b/docs/src/main/java/org/geotools/wfs/ReadWFSKommuner.java
@@ -1,0 +1,149 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ *
+ *    (C) 2021, Open Source Geospatial Foundation (OSGeo)
+ *
+ *    This library is free software; you can redistribute it and/or
+ *    modify it under the terms of the GNU Lesser General Public
+ *    License as published by the Free Software Foundation;
+ *    version 2.1 of the License.
+ *
+ *    This library is distributed in the hope that it will be useful,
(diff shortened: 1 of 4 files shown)


roarbra:
I'm not that familiar with the GeoServer codebase, but it seems to ignore complex attributes regardless of my change.
I will have a look at the documentation, but I don't see how it fits in with my code change. However I do see that it violates quite a lot of the original test's, so I can see your concern.

aaime:
@roarbra using GeoServer was just a quick way to test the parsing and re-encoding, all the actual machinery is in GeoTools, but to try that out, one has to write some code to wire the bits together.

roarbra:
@aaime I've looked at a solution to write out the features using org.geotools.xsd.Encoder, but without any luck. It doesn't seem to like complex features either.

I've added an example to docs, and wonder if that could be useful to show how my fix makes the code more readable.

taba90:
@aaime The code modifications seems legit to me and are providing a fix for the parser. 
For what is seems to me before this fix, although the GML object property model was respected in the Feature structure, the parser was creating Feature attributes with the wrong name (equals to the Type name instead of using the descriptor name as should be done).

aaime:
@roarbra merged, feel free to backport as you see fit.

PR #3673, opened 2021-11-11
Update docker images and actions in github CI

mprins:
- [x] Bump Db2 version in CI from `11.5.5.0` to `11.5.6.0a`
- [x] Use a proper version tag for the Oracle docker image and fixup documentation
- [x] use `action/setup-java@v2` with `temurin` (`adopt` is deprecated)
- [x] use `actions/cache@v2`

## for discussion

- [ ] Run Oracle OnlineTest against both 18.4 as well as 21.3 Express editions? 
      I'm not sure if there are incompatibilities between these two and if it's worth the extra overhead (our Oracle Online tests take about 10 minutes, making them the slowest already, OTOH adding doesn't add 10 mins. as these builds run in parallel. Note that during the last 2 weeks we silently wen from 18.4.0 to 21.3.0 because we were using the `latest` tag in our workflow
  

# Checklist

### DIFF
--- a/.github/workflows/db2_online.yml
+++ b/.github/workflows/db2_online.yml
@@ -12,9 +12,10 @@ jobs:
 
     steps:
       - name: Set up JDK ${{ matrix.java }}
-        uses: actions/setup-java@v1
+        uses: actions/setup-java@v2
         with:
           java-version: ${{ matrix.java }}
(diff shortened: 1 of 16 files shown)


aaime:
Looks good to me, thanks for the updates!

PR #3674, opened 2021-11-11
[GEOT-7014] Move functionality for filtering tiles by the layer's tileMatrixLimits

roarbra:
[GEOT-7014](https://osgeo-org.atlassian.net/browse/GEOT-7014) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

Some functionality are moved from AbstractGetTileRequest to WMTSTileService. The functionality should be the same, but with less code, and a call for WMTSTileService.findTilesInExtent would potentially return tiles that wasn't within the limits of WMTSLayer.

Some functionality are moved from WMTSTileFactory to WMTSTileService. Leaving behind some deprecated functions.

I've added some tests that covers the change.

# Checklist

### DIFF
--- a/modules/extension/wmts/src/main/java/org/geotools/ows/wmts/client/WMTSTileFactory.java
+++ b/modules/extension/wmts/src/main/java/org/geotools/ows/wmts/client/WMTSTileFactory.java
@@ -69,6 +69,7 @@ public class WMTSTileFactory extends TileFactory {
     }
 
     /** Find the first valid Upper Left tile for the current layer. */
+    @Deprecated
     public Tile findUpperLeftTile(
             double lon, double lat, WMTSZoomLevel zoomLevel, WMTSTileService service) {
         // get the tile in the tilematrix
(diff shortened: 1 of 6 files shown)


modules/extension/wmts/src/main/java/org/geotools/ows/wmts/client/WMTSTileFactory.java
```diff
@@ -69,13 +69,15 @@ public Tile findTileAtCoordinate(
     }
 
     /** Find the first valid Upper Left tile for the current layer. */
+    @Deprecated
```
aaime, line 72: So these public methods are gone, and there is no replacement? I don't think they are widely used, but still... would be nice if there was a way to get the same functionality after the refactor. If there is, the javadoc should say what it is (and the upgrade rst file too).

roarbra: Sorry, I was a little quick there. I will take the deprecation with its own issue at a later stage.

roarbra: @aaime You've changed the status of the ticket in JIRA to Solved, but the PR isn't committed.

aaime: Whoops... easily fixed.

PR #3677, opened 2021-11-14
Improve MAVEN build Performance

hj987654321:
Maven can recompile only the classes that were affected by a change. This feature is the default. We can activate it by setting `<useIncrementalCompilation>true</useIncrementalCompilation>`.

=====================
If there are any inappropriate modifications in this PR, please give me a reply and I will change them.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1595,6 +1595,7 @@
           <compilerArgs>
             <arg>-Xlint:${lint}</arg>
           </compilerArgs>
+            	<useIncrementalCompilation>true</useIncrementalCompilation>
         </configuration>
       </plugin>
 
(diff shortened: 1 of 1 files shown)


aaime:
According to the documentation, the incremental compilation is turned on by the default:
https://maven.apache.org/plugins/maven-compiler-plugin/compile-mojo.html

If so, then this change adds noise in the pom.xml file, without actually adding any performance benefit. While not a strict rule, we only configure what deviates from the default values.

mprins:
I've reported this bot for opening the same useless PR on 30-plus projects

aaime:
Excellent way to waste every maven based project time I see. Is there any way to report this bot to Github and ban it?

mprins:
@aaime  there's a link to "Report content" under the ellipsis of each pr/comment
image

aaime:
Tried, but the list of categories is quite restrictive, not sure any of them applies, they are all for much more serious situations.

PR #3678, opened 2021-11-16
[GEOT-7022]: Referencing factories deadlock prevention

dromagnoli:
[GEOT-7022](https://osgeo-org.atlassian.net/browse/GEOT-7022) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

# Checklist

### DIFF
--- a/modules/library/referencing/src/main/java/org/geotools/referencing/factory/DeferredAuthorityFactory.java
+++ b/modules/library/referencing/src/main/java/org/geotools/referencing/factory/DeferredAuthorityFactory.java
@@ -27,13 +27,14 @@ import org.geotools.metadata.i18n.ErrorKeys;
 import org.geotools.metadata.i18n.Errors;
 import org.geotools.metadata.i18n.LoggingKeys;
 import org.geotools.metadata.i18n.Loggings;
+import org.geotools.referencing.ReferencingFactoryFinder;
 import org.geotools.util.factory.Hints;
 import org.geotools.util.factory.OptionalFactory;
 import org.opengis.referencing.FactoryException;
 
 /**
  * A buffered authority factory which will defer the {@linkplain #createBackingStore creation of a
- * backing store} until when first needed. This approach allow to etablish a connection to a
+ * backing store} until when first needed. This approach allow to establish a connection to a
  * database (for example) only when first needed. In addition, the backing store can be
  * automatically disposed after a timeout and recreated when needed again.
  *
@@ -109,7 +110,12 @@ public abstract class DeferredAuthorityFactory extends BufferedAuthorityFactory
     @Override
     protected final AbstractAuthorityFactory getBackingStore() throws FactoryException {
         if (backingStore == null) {
-            synchronized (this) {
+            // [GEOT-7022]
+            // Concurrent accesses to Authority Factories may result in one factory locking
+            // on "this" whilst the other factory goes through ReferencingFactoryFinder
+            // which may result into trying to lock on "this" too.
+            // So we are locking on ReferencingFactoryFinder to avoid deadlocks.
+            synchronized (ReferencingFactoryFinder.class) {
                 if (backingStore == null) {
                     backingStore = createBackingStore();
                     if (backingStore == null) {
--- a/modules/library/referencing/src/main/java/org/geotools/referencing/factory/epsg/ThreadedEpsgFactory.java
+++ b/modules/library/referencing/src/main/java/org/geotools/referencing/factory/epsg/ThreadedEpsgFactory.java
@@ -366,7 +366,11 @@ public class ThreadedEpsgFactory extends DeferredAuthorityFactory
      *     code.
      */
     private AbstractAuthorityFactory createBackingStore0() throws FactoryException, SQLException {
-        assert Thread.holdsLock(this);
+        /*
+         * We are locking on ReferencingFactoryFinder to avoid deadlocks.
+         * @see DeferredAuthorityFactory#getBackingStore()
+         */
+        assert Thread.holdsLock(ReferencingFactoryFinder.class);
         final Hints sourceHints = new Hints(hints);
         sourceHints.putAll(factories.getImplementationHints());
         if (datasource != null) {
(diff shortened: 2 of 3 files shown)


modules/library/referencing/src/main/java/org/geotools/referencing/factory/DeferredAuthorityFactory.java
```diff
@@ -109,7 +110,7 @@ public boolean isAvailable() {
     @Override
     protected final AbstractAuthorityFactory getBackingStore() throws FactoryException {
         if (backingStore == null) {
-            synchronized (this) {
+            synchronized (ReferencingFactoryFinder.class) {
```
aaime, line 113: I believe some javadoc should explain why the synchronization is done on another class rather than "this". The old code seems obvious, even if leading to a deadlock, this one should have an explanation.

modules/library/referencing/src/main/java/org/geotools/referencing/factory/epsg/ThreadedEpsgFactory.java
```diff
@@ -366,7 +366,7 @@ protected AbstractAuthorityFactory createBackingStore(final Hints hints) throws
      *     code.
      */
     private AbstractAuthorityFactory createBackingStore0() throws FactoryException, SQLException {
-        assert Thread.holdsLock(this);
+        assert Thread.holdsLock(ReferencingFactoryFinder.class);
```
aaime, line 369: Same as above.

modules/plugin/geotiff/src/test/java/org/geotools/gce/geotiff/GeoTiffCRSFactoryDeadlockTest.java
```diff
@@ -0,0 +1,90 @@
...
+public class GeoTiffCRSFactoryDeadlockTest {
+
+    @Test
+    public void testDeadlockOnCRSFactories() throws ExecutionException, InterruptedException {
+        boolean potentialDeadlock = false;
+        final long timeout = 5;
+
+        final boolean[] exceptionOccurred = {false};
+        ExecutorService executorService = Executors.newFixedThreadPool(2);
+        Future[] futures = new Future[2];
```
aaime, line 47: Did the test actually lock up frequently before the fix? 
Using just 2 calls is the lightest possible load, I was expecting to see a storm of tens of parallel executors all thrown together on a thread pool with 8-16 threads to make this lock up.

aaime (changes requested):

modules/plugin/geotiff/src/test/java/org/geotools/gce/geotiff/GeoTiffCRSFactoryDeadlockTest.java
dromagnoli: Before the fix, the test was failing very often (I would say 8 of 10 runs).
I was also testing it in a loop and with more threads but it didn't change very much.
I'll double check and retry.

dromagnoli: Done

dromagnoli:
The Oracle online test failure is unrelated. I see it failing on another couple of PRs too.

aaime:
Yeah it's a random failure, not sure what's causing it:

```
Error:  Failures: 
Error:    OracleJoinOnlineTest>OnlineTestCase.run:117->JDBCJoinOnlineTest.testSpatialJoin:491->JDBCJoinOnlineTest.doTestSpatialJoin:509 expected:<6> but was:<0>
```

aaime:
@mprins the Oracle image used to run tests has been changed recently (Nov 11th?), could it be a cause?
There is also a change in gt-jdbc, but that dates back to Nov 3rd. Cannot think of anything else, the test is one of those shared across several databases (same test code) and it's not failing in the others.

mprins:
ehmm... https://github.com/geotools/geotools/actions/workflows/oracle_online.yml shows several passed oracle builds after updating the image tag, however a new version of that tagged image was pushed 5 days ago (Nov 14 @02:09): https://hub.docker.com/r/gvenzl/oracle-xe/tags

It's unclear to what changed in the image from the commit log https://github.com/gvenzl/oci-oracle-xe/commits/main

I'll see if I can reproduce locally...

mprins:
I can reproduce on `main` with `gvenzl/oracle-xe:21.3.0`, now trying `gvenzl/oracle-xe:21.3.0-full` (which hasn't had the Java stored procedure package removed)

mprins:
I get successful builds with both `gvenzl/oracle-xe:18.4.0` as well as `gvenzl/oracle-xe:18.4.0-full`,  so I suggest we drop back to `gvenzl/oracle-xe:18.4.0`

```diff
diff --git a/build/ci/oracle/start-oracle.sh b/build/ci/oracle/start-oracle.sh
index 797bd2e60c..ff2d51035f 100755
--- a/build/ci/oracle/start-oracle.sh
+++ b/build/ci/oracle/start-oracle.sh
@@ -3,12 +3,12 @@ docker version
 
 # this docker image does not have Java installed, which means you can not use Java stored procedures
 # such as "SDO_GEOMETRY('POINT (1.0 2.0)', 4326))"
-docker pull gvenzl/oracle-xe:21.3.0
+docker pull gvenzl/oracle-xe:18.4.0
 
 # start the dockerized oracle-xe instance (the container will be destroyed/removed on stopping)
 # this container can be stopped using: docker stop geotools
 # start with user/credentials (user/password = system/oracle)
-docker run --rm -p 1521:1521 --name geotools -h geotools -e ORACLE_PASSWORD=oracle -d gvenzl/oracle-xe:21.3.0
+docker run --rm -p 1521:1521 --name geotools -h geotools -e ORACLE_PASSWORD=oracle -d gvenzl/oracle-xe:18.4.0
 
 printf "\n\nStarting Oracle XE container, this could take a few minutes..."
 printf "\nWaiting for Oracle XE database to start up.... "

```

aaime:
@dromagnoli merged

dromagnoli:
Thanks

PR #3679, opened 2021-11-17
[GEOT-7023] Fix cascade WMS layer wrong native CRS selection

fernandor777:
[GEOT-7023](https://osgeo-org.atlassian.net/browse/GEOT-7023) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

WMS cascade layer returns wrong native CRS to GeoServer. The returned SRS and bounding boxes are unordered making hard to GeoServer to pick the first native CRS to use during rendering when the source GetCapabilities document root layer reports more supported CRSs and bounding boxes.

This PR introduces a fix for this, making the original native CRSs available first and in order, to be prioritized.

JIRA issue:
https://osgeo-org.atlassian.net/browse/GEOT-7023

# Checklist


modules/extension/wms/src/main/java/org/geotools/ows/wms/map/WMSCoverageReader.java
```diff
@@ -207,6 +212,31 @@ void addLayer(Layer layer, String style) {
...
+                CRS.decode(srs);
+                return srs;
+            } catch (Exception e) {
+                // it's fine, we could not decode that code
+            }
+        }
+        return null;
+    }
+
+    private List<String> getPriorityCRSs(Layer layer) {
```
aaime, line 231: Naming is vague, "priority" might mean many things. What this does, is to return the CRSs with a matching bounding box definition. ``getCRSWithBoundingBoxes`` maybe?

modules/extension/wms/src/main/java/org/geotools/ows/wms/map/WMSCoverageReader.java
```diff
@@ -207,6 +212,31 @@ void addLayer(Layer layer, String style) {
         updateBounds();
     }
 
+    /** Returns the first CRS from layer srs list that matches with bounding boxes map keys. */
+    private String getPriorityCRS(Layer layer) {
+        List<String> priorityCRSs = getPriorityCRSs(layer);
+        if (!priorityCRSs.isEmpty()) {
+            String srs = priorityCRSs.get(0);
+            try {
+                // check it's valid, if not we crap out and move to the next
```
aaime, line 221: "move to the next" but there is no loop. Next what? Next layer?

aaime (changes requested):
Some parts are improved, the caps document has an order for SRS and bounding boxes, and it's good that the WMS client honors that order. 
Is this change coupled in any way with more GeoServer changes?

modules/extension/wms/src/main/java/org/geotools/ows/wms/Layer.java
```diff
@@ -213,7 +204,8 @@ public Layer(String title) {
                     allBoundingBoxesCache.put(bbox.getSRSName(), bbox);
                 }
                 parent = parent.getParent();
-            }
+            } // linked hash map to preserve order, add parent ones only if not available.
```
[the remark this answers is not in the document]
fernandor777: Indeed a leftover, fixed.

modules/extension/wms/src/main/java/org/geotools/ows/wms/Layer.java
```diff
@@ -213,7 +204,8 @@ public Layer(String title) {
                     allBoundingBoxesCache.put(bbox.getSRSName(), bbox);
```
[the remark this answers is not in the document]
fernandor777: Right, added a condition checking that, thanks!

modules/extension/wms/src/main/java/org/geotools/ows/wms/Layer.java
```diff
@@ -232,7 +224,18 @@ public void setBoundingBoxes(CRSEnvelope boundingBox) {
      */
     public void setBoundingBoxes(Map<String, CRSEnvelope> boundingBoxes) {
         this.boundingBoxes.clear();
-        this.boundingBoxes.addAll(boundingBoxes.values());
+        // prioritize CRS order available already on SRS set
+        LinkedHashMap<String, CRSEnvelope> bboxesInternal = new LinkedHashMap<>(boundingBoxes);
```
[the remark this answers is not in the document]
fernandor777: Right, but this new map is used as temp variable to mutate its state by removing the values already present on native layer CRSs to be appended to the end.  In this way we don't change items on the original map.

modules/extension/wms/src/main/java/org/geotools/ows/wms/map/WMSCoverageReader.java
fernandor777: Right, better name, fixed.

modules/extension/wms/src/main/java/org/geotools/ows/wms/map/WMSCoverageReader.java
fernandor777: Fixed, indeed we need to iterate CRSs with BBOXs in case the first ones result in CRS decode exception.

fernandor777:
Thanks @aaime for your review.  There is not a GeoServer counter part for this bug fix, once the layer returns bounding boxes in native CRSs former order the cascade WMS use the first one and rendering bbox computation works, fixing our target GeoServer bug.

aaime (approved):

aaime:
Built the GT and GS sides locally on my machine, full pass, merging both.

PR #3680, opened 2021-11-18
github release notes

jodygarnett:
Update developer guide procedure to use github release page to publish release notes publicly (as generated from Jira which is no longer public).

# Checklist

### DIFF
--- a/docs/developer/procedures/release.rst
+++ b/docs/developer/procedures/release.rst
@@ -11,7 +11,7 @@ Before you start
 Notify developer list
 ^^^^^^^^^^^^^^^^^^^^^
 
-It is good practice to notify the `GeoTools developer list <https://lists.sourceforge.net/lists/listinfo/geotools-devel>`_ of the intention to make the release a few days in advance, even though the release date has been agreed upon before hand. 
+It is good practice to notify the `GeoTools developer list <https://lists.sourceforge.net/lists/listinfo/geotools-devel>`__ of the intention to make the release a few days in advance, even though the release date has been agreed upon before hand. 
 
 Prerequisites
 -------------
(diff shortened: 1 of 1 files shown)


docs/developer/procedures/release.rst
```diff
@@ -187,29 +187,40 @@ Download and try out some of the artifacts from the above location and do a
...
-* Check the eclipse quickstart section on `geotools.version`, should reference the correct release tag and snapshot tag.
+4. Binary download:
+   
+   * Checking the README.html links go to the correct stable or maintenance user guide
+   
+   * Check library loads:
+     
+     .. code-block:: bash
+        
+        java -cp "lib/*" org.geotools.util.factory.GeoTools
```
aaime, line 211: Is this is also something we could automate, right after doing a build off an empty repo?

docs/developer/procedures/release.rst
```diff
@@ -187,29 +187,40 @@ Download and try out some of the artifacts from the above location and do a
...
+   
+   * Run quickstart:
+     
+     .. code-block:: bash
+     
+        mkdir bin
+        javac -cp "lib/*" -d bin src/org/geotools/tutorial/quickstart/Quickstart.java 
+        java -cp "lib/*:bin" org.geotools.tutorial.quickstart.Quickstart
```
aaime, line 219: I'm guessing this one starts a Swing GUI. Wondering if there is any other check that could be done automatically though (some example that does not run a GUI and whose results could be checked)

docs/developer/procedures/release.rst
aaime: And also in normal github actions for PRs, to avoid last minute surprises

aaime (commented):
Looks good, throwing in a couple of ideas

docs/developer/procedures/release.rst
jodygarnett: We can check this one, but not the Quickstart.

docs/developer/procedures/release.rst
jodygarnett: CRS has a main method and can be run, we could also produce the function list.

jodygarnett (commented):
I have answered questions; going to  merge now. 

Was surprised that release notes were not produced for the most recent releases.

PR #3691, opened 2021-11-29
[GEOT-7028] - GetTile request's in WMTS represents one single tile, not a set

roarbra:
[GEOT-7028](https://osgeo-org.atlassian.net/browse/GEOT-7028) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

I've added a function issueRequestForTileResponse to org.geotools.ows.wmts.WebMapTileServer. The name isn't in accordance with the usual OWS server instances, but that's because of the original issueRequest that returns a Set of Tile's. This is unfortunate, but before we can deprecate and change the return, we must find an alternative solution. This signature is also used within GeoServer.

In addition I've changed GetTileResponse so that it tries to read the image of the response.

# Checklist

### DIFF
--- a/modules/extension/wmts/src/main/java/org/geotools/ows/wmts/WebMapTileServer.java
+++ b/modules/extension/wmts/src/main/java/org/geotools/ows/wmts/WebMapTileServer.java
@@ -33,6 +33,7 @@ import org.geotools.ows.wms.CRSEnvelope;
 import org.geotools.ows.wms.Layer;
 import org.geotools.ows.wms.request.GetFeatureInfoRequest;
 import org.geotools.ows.wms.response.GetFeatureInfoResponse;
+import org.geotools.ows.wmts.WMTSSpecification.GetSingleTileRequest;
 import org.geotools.ows.wmts.model.WMTSCapabilities;
 import org.geotools.ows.wmts.model.WMTSServiceType;
 import org.geotools.ows.wmts.request.GetTileRequest;
@@ -190,18 +191,21 @@ public class WebMapTileServer extends AbstractOpenWebService<WMTSCapabilities, L
         specs[0] = new WMTSSpecification();
     }
 
-    /** */
+    /** @deprecated - change to tileRequest.getTiles() */
+    @Deprecated
     public Set<Tile> issueRequest(GetTileRequest tileRequest) throws ServiceException {
         return tileRequest.getTiles();
     }
 
-    /** The new issueRequest, which gives a GetTileResponse in response for a GetTileRequest. */
-    public GetTileResponse issueRequestForTileResponse(GetTileRequest tileRequest)
+    /**
+     * The new issueRequest, which gives a GetTileResponse in response for a GetSingleTileRequest.
+     */
+    public GetTileResponse issueRequest(GetSingleTileRequest tileRequest)
             throws IOException, ServiceException {
         return (GetTileResponse) internalIssueRequest(tileRequest);
     }
 
-    /** @return */
+    /** Creates a GetMultiTileRequest */
     public GetTileRequest createGetTileRequest() {
 
         URL url;
@@ -219,7 +223,33 @@ public class WebMapTileServer extends AbstractOpenWebService<WMTSCapabilities, L
         props.put("type", type.name());
         GetTileRequest request =
                 ((WMTSSpecification) specification)
-                        .createGetTileRequest(url, props, capabilities, this.getHTTPClient());
+                        .createGetMultiTileRequest(url, props, capabilities, this.getHTTPClient());
+
+        if (headers != null) {
+            request.getHeaders().putAll(headers);
+        }
+        return request;
+    }
+
+    /** Creates a GetSingleTileRequest */
+    public GetTileRequest createGetTileRequest(WMTSServiceType type) {
+
+        OperationType getTile = capabilities.getRequest().getGetTile();
+        if (getTile == null) {
+            type = WMTSServiceType.REST;
+        }
+
+        URL url;
+        if (WMTSServiceType.KVP.equals(type)) {
+            url = findURL(getTile);
+        } else {
+            type = WMTSServiceType.REST;
+            url = serverURL;
+        }
+
+        GetSingleTileRequest request =
+                ((WMTSSpecification) specification)
+                        .createGetTileRequest(url, type, this.getHTTPClient());
 
         if (headers != null) {
             request.getHeaders().putAll(headers);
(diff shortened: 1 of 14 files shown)


modules/extension/wmts/src/main/java/org/geotools/ows/wmts/WebMapTileServer.java
```diff
@@ -140,10 +141,19 @@ public WebMapTileServer(URL serverURL) throws IOException, ServiceException {
...
     public WebMapTileServer(WMTSCapabilities capabilities) throws ServiceException, IOException {
-        super(
-                capabilities.getRequest().getGetCapabilities().getGet(),
-                HTTPClientFinder.createClient(),
-                capabilities);
+        this(capabilities, HTTPClientFinder.createClient());
+    }
+
+    /**
+     * Set up the WebMapTileServer with the given capabilities. Using the default http client.
```
aaime, line 148: This one actually uses the provided HTTP client.

modules/extension/wmts/src/main/java/org/geotools/ows/wmts/request/AbstractGetTileRequest.java
```diff
@@ -62,9 +61,11 @@
 
     static WMTSTileFactory factory = new WMTSTileFactory();
 
-    public static final String LAYER = "Layer";
+    public static final String LAYER = "layer";
```
aaime, line 64: Keys in OWS GET requests are supposed to be case insensitive, so the change should not be necessary (unless you're trying to deal with a broken implementation). Also, keys before the change were all following the same case, now we have a mix of all lowercase and camelcase.

modules/extension/wmts/src/main/java/org/geotools/ows/wmts/response/GetTileResponse.java
```diff
@@ -16,23 +16,51 @@
...
 import org.geotools.http.HTTPResponse;
+import org.geotools.image.io.ImageIOExt;
 import org.geotools.ows.ServiceException;
 import org.geotools.ows.wmts.model.WMTSServiceType;
 
-/** @author ian */
+/**
+ * Represents the response of a tile request.
+ *
+ * <p>In most cases this is a image, but it can also be a file like kml with reference to an image.
```
aaime, line 31: Vector tiles are pretty common these days, they can also be returned, in that case, there is no image at all.

modules/extension/wmts/src/main/java/org/geotools/ows/wmts/WebMapTileServer.java
```diff
@@ -185,6 +195,12 @@ protected void setupSpecifications() {
         return tileRequest.getTiles();
     }
 
+    /** The new issueRequest, which gives a GetTileResponse in response for a GetTileRequest. */
+    public GetTileResponse issueRequestForTileResponse(GetTileRequest tileRequest)
```
aaime, line 199: I believe this new API is confusing, in the context of what GetTileRequest can do... the object has a generic bounding box and will fetch al the tiles in it. It also allows generic width/height. The request must be carefully built to just get one tile. It seems to me, if you wanted to fetch a single tile, then a new request/response pairs should be used instead, and make sure they can only do that (e.g., identify the tile by xyz rather than bbox/with/height).

modules/extension/wmts/src/main/java/org/geotools/ows/wmts/request/AbstractGetTileRequest.java
roarbra: It was a test that failed, but I've changed the assumption of that one instead.

modules/extension/wmts/src/main/java/org/geotools/ows/wmts/WebMapTileServer.java
roarbra: For me the old API have been confusing since I started to look into this. The GetTileResponse wasn't actually used and GetTileRequest was actually many http-calls without using the ordinary OWS-mechanism. So I doubt it would handle an error from the server, and surely not vector tiles.

I was thinking about adding the properties for tileset, row and col. Was just wondering how to do it. I could perhaps do it with a parallell interface, but that seems like a lot of work.

aaime: I agree the API is confusing (an operation that returns multiple tiles should be called GetTiles), but the fix does not make it better, since the parameters of GetTileRequest are still allowing for multiple tiles.

A potential way out of this without a solid API break can be:
- Rename GetTile to getMultiTile
- Create a deprecated subclass of it called GetTile
- Create a getSingleTile request whose parameters only allow to ask for a single tile (e.g. x,y,z)

Also, a note: asking for multiple tiles is not a weird operation, while it's not in the WMTS standard, it has been considered for addition for a long time, just did not happen due to lack of resources in OGC.
OGC API Tiles instead allows to request single tiles in core, while allowing for multiple tiles in [an optional extension](https://github.com/opengeospatial/ogcapi-tiles/blob/master/extensions/multitile/standard/clause_7_tile_multitiles.adoc).

The reason why multiple tiles request should be allowed is evident, it's more efficient, instead of paying latency N times, it will be paid only once.

roarbra: I've introduced GetSingleTileRequest, GetMultiTileRequest and also GetKVPTileRequest and GetRestTileRequest, and deprecated the old GetTileRequest.

I see the point of getting many tiles by one http call, but anyway there would have been a Response that wrapped the http response to extract those tiles. Not like today where it comes from a call to WMTSTileService.

modules/extension/wmts/src/main/java/org/geotools/ows/wmts/request/GetTileRequest.java
```diff
@@ -42,6 +42,14 @@
 
     void setFormat(String format);
 
+    void setTileMatrixSet(String tileMatrixSet);
```
aaime, line 45: This interface is not deprecated and can now play the role of both single and multi-tile request? Seems inconsistent, one  should probably either use xyz or bbox no?

aaime:
Love the direction this is going, just wondering about one interface change (see latest comment)

modules/extension/wmts/src/main/java/org/geotools/ows/wmts/request/GetTileRequest.java
roarbra: @aaime I was wondering about deprecating those properties (bbox) at a later stage. My intention is to work more on how WMTSCoverageReader fetches the tiles. I've noticed that someone else is working on that class with GEOT-7036, so I would wait with too much changes. After changing WMTSCoverageReader it will be a lot more easier to deprecate.

aaime:
Ok, in it goes. Let's not forge that deprecation, down the road :-D

PR #3692, opened 2021-11-29
Fix for ImageWorker

irisiflimsi:
A lot of my data is raster data with alpha-channel. About 2% fail when serving them through the tile server at some resolutions.  I can't produce a small, self-contained example, unfortunately. But I have tracked it down and it seems the alpha-channel is rendered separately from the rest of the image and is "allergic" to 0-size tile-width. This is the way I fixed it in 21.1 (verified in my environment). If necessary, I can give you the stack trace from the logs, the AWT class doesn't like 0.

# Checklist

I wasn't able to create a Jira issue, because the Attlassian license limit was reached, or so it said. The issue was discussed on the mailing list a few weeks ago, Jody Garnett was involved. If you need a signed contribution agreement, I can forward that, but I consider the issue minor enough not to do so at this point in time.

### DIFF
--- a/modules/library/coverage/src/main/java/org/geotools/image/ImageWorker.java
+++ b/modules/library/coverage/src/main/java/org/geotools/image/ImageWorker.java
@@ -2,7 +2,7 @@
  *    GeoTools - The Open Source Java GIS Toolkit
  *    http://geotools.org
  *
- *    (C) 2006-2016, Open Source Geospatial Foundation (OSGeo)
+ *    (C) 2006-2021, Open Source Geospatial Foundation (OSGeo)
  *
  *    This library is free software; you can redistribute it and/or
  *    modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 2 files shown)


dromagnoli (approved):
Looks good to me

aaime:
We need the issue in JIra and the commit message referring to it, before merge. You can force-push to update the PR.

irisiflimsi:
This mail wasn't addressed to me, but just to be clear: I can't do anything about it, I can't access your Jira.

Michael

aaime:
Yes, all comments to a PR are delivered to the PR author, and other participants, by mail.

Anyone should be able to create a Jira ticket, it just requires an account (which can also be freely created).
Did you create the account and it still does not work? Or did you have difficulties creating an account on Jira?
See also: https://osgeo-org.atlassian.net/jira/dashboards/last-visited

irisiflimsi:
See the mail for details. I have an account, which I can use to access atlassian "root", I can see anonymous information on your board but nothing reserved for authenticated members - or so it feels.

aaime:
For this time only, I've created the ticket in Jira and fixed the commit message to follow, by using squash-merge.
Next time please try to create a separate account, OSGeo is a different organization than the Atlassian root, so it's the likely cause (I don't remember getting report of similar issues).

The commit has been merged on main, will be part of the release in March 2022. If you want it to show up earlier, you can create backport PRS to 26.x and 25.x. Please cherry-pick the commit as merged 6ccf156, as it has the right commit message, allowing to cross-reference the commit to Jira.

PR #3701, opened 2021-12-10
[GEOT-7036] WMTS reader doesn't reproject result

taba90:
[GEOT-7036](https://osgeo-org.atlassian.net/browse/GEOT-7036) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

This pr:
- fixes reprojection to unsupported CRS in wmts.
- adds a GeneralParameterValue to pass down to the reader the sourceCRS from which to reproject in case the target CRS is not supported from server.
- fixes some problems in the retrivial of CRS from the capabilities document: the existing logic: for each srs added to the list of server supported srs all the possible definitions (eg the urn ogc if the actual def is EPSG).
 

# Checklist


taba90:
Downstream integration will be green when the gs side pr will be merged https://github.com/geoserver/geoserver/pull/5461

modules/extension/wmts/src/main/java/org/geotools/ows/wmts/map/WMTSCoverageReader.java
```diff
@@ -346,14 +338,23 @@ GridCoverage2D getMap(ReferencedEnvelope requestedEnvelope, int width, int heigh
      * @throws TransformException
      */
     private ReferencedEnvelope toEastNorthAxisOrder(ReferencedEnvelope envelope)
-            throws FactoryException, TransformException {
-        CoordinateReferenceSystem crsExtent = envelope.getCoordinateReferenceSystem();
-        if (CRS.getAxisOrder(crsExtent) == CRS.AxisOrder.NORTH_EAST) {
-            String srsExtent = CRS.toSRS(crsExtent);
-            crsExtent = CRS.decode(srsExtent, true);
-            envelope = envelope.transform(crsExtent, false);
+            throws FactoryException {
```
aaime, line 341: In the javadoc above, while you're at it, "eith" should be "with"

modules/extension/wmts/src/main/java/org/geotools/ows/wmts/map/WMTSCoverageReader.java
```diff
@@ -563,4 +564,141 @@ public String getRequestedTime() {
...
+        }
+    }
+
+    // get the CRS to be used to reproject to a CRS that the server doesn't support.
+    CoordinateReferenceSystem getBestSourceCRS(
+            CoordinateReferenceSystem targetCRS, WMTSReadParameters readParameters)
+            throws IOException {
+        try {
+            CoordinateReferenceSystem result = null;
+            String switched = switchDefinition(targetCRS);
```
aaime, line 676: I am confused... why try directly to switch to another style? Ah because by the time you call this method, you already know the targetCRS identifier is not among the supported ones?

modules/extension/wmts/src/main/java/org/geotools/ows/wmts/map/WMTSCoverageReader.java
```diff
@@ -563,4 +564,141 @@ public String getRequestedTime() {
...
+            if (switched != null) {
+                // check first by changing the srs definition
+                result = CRS.decode(switched);
+            } else if (readParameters.getSourceCRS() != null) {
+                // otherwise get the preferred one
+                result = readParameters.getSourceCRS();
+            } else if (!validSRS.isEmpty()) {
+                // last attempt is to retrieve the firs valid srs available
+                LOGGER.fine(() -> "Retrieving random srs to perform reprojection");
+                String srs = validSRS.iterator().next();
```
aaime, line 686: This could likely be more robust. What if the remote server first SRS cannot be decoded?
I'd loop with a try/catch inside, and try them all, before throwing the towel.

modules/extension/wmts/src/main/java/org/geotools/ows/wmts/map/WMTSCoverageReader.java
```diff
@@ -563,4 +564,141 @@ public String getRequestedTime() {
     public void setRequestedTime(String requestedTime) {
         this.requestedTime = requestedTime;
     }
+
+    boolean delegateToRemote(CoordinateReferenceSystem crs) {
+        String srs = CRS.toSRS(crs);
+        return srs != null && validSRS.contains(srs);
+    }
+
+    private String switchDefinition(CoordinateReferenceSystem crs) {
```
aaime, line 573: This could use a javadoc.

modules/extension/wmts/src/main/java/org/geotools/ows/wmts/map/WMTSCoverageReader.java
```diff
@@ -563,4 +564,141 @@ public String getRequestedTime() {
...
+            if (validSRS.contains("EPSG:3857")) {
+                switched = "EPSG:3857";
+            } else if (validSRS.contains("urn:ogc:def:crs:EPSG::3857")) {
+                switched = "urn:ogc:def:crs:EPSG::3857";
+            }
+        }
+        return switched;
+    }
+
+    private GridCoverage2D reprojectManually(
```
aaime, line 604:
```suggestion
    private GridCoverage2D reproject(
```

No hand gets dirty here

aaime (changes requested):
Lots of clarifying refactors, which I liked, but some of the new parts need better javadoc/comments, still a bit impenetrable :-D

aaime:
Ah, one thing that seems missing, to me, is checking that delegation happens to the remote server, whenever there is a match between the requested CRS and the remotely offered one (even when the identifiers are hard to match, try both ways), while projection only actually happens when all attempts at delegating failed.

modules/extension/wmts/src/test/java/org/geotools/ows/wmts/map/WMTSMapLayerTest.java
```diff
@@ -0,0 +1,38 @@
+package org.geotools.ows.wmts.map;
```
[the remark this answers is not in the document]
taba90: fixed

modules/extension/wmts/src/test/java/org/geotools/ows/wmts/online/LantmaterietWMTSServerOnlineTest.java
```diff
@@ -188,63 +171,90 @@ public void testWMTSCoverageReaderWithStylePlaceholder()
 
         // envelope to request tiles for 5th zoom level
         ReferencedEnvelope re1 = new ReferencedEnvelope(977650, 1913530, 5838030, 6281290, crs);
-        RenderedImage extent = getRenderImageResult(wmts, capabilities, re1, re1, "geolandbasemap");
+        RenderedImage ri = getRenderImageResult(wmts, capabilities, re1, re1, "geolandbasemap");
+        File file = new File("C://marco_workspace/resulttopoweb.png");
```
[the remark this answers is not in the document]
taba90: ops.... fixed

modules/extension/wmts/src/main/java/org/geotools/ows/wmts/map/WMTSReadParameters.java
```diff
@@ -0,0 +1,110 @@
...
+    /** Get the requestedEnvelope. */
+    Envelope getRequestedEnvelope() {
+        Envelope result = gridGeom.getEnvelope();
+        if (result == null) result = originalEnvelope;
+        return result;
+    }
+
+    /** Get the image width. */
+    int getWidth() {
+        int result = 640;
```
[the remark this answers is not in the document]
taba90: fixed

modules/extension/wmts/src/main/java/org/geotools/ows/wmts/map/WMTSCoverageReader.java
```diff
@@ -563,4 +564,141 @@ public String getRequestedTime() {
     public void setRequestedTime(String requestedTime) {
         this.requestedTime = requestedTime;
     }
+
+    boolean delegateToRemote(CoordinateReferenceSystem crs) {
```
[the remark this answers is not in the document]
taba90: fixed

modules/extension/wmts/src/main/java/org/geotools/ows/wmts/map/WMTSCoverageReader.java
taba90: fixed

modules/extension/wmts/src/main/java/org/geotools/ows/wmts/map/WMTSCoverageReader.java
taba90: yes the getBestSourceCRS should be called when we know the targetCRS is not among the supported ones. I've added a comment.

modules/extension/wmts/src/main/java/org/geotools/ows/wmts/map/WMTSCoverageReader.java
taba90: fixed

modules/extension/wmts/src/main/java/org/geotools/ows/wmts/map/WMTSCoverageReader.java
taba90: fixed

modules/extension/wmts/src/main/java/org/geotools/ows/wmts/map/WMTSCoverageReader.java
taba90: fixed

taba90:
> Ah, one thing that seems missing, to me, is checking that delegation happens to the remote server, whenever there is a match between the requested CRS and the remotely offered one (even when the identifiers are hard to match, try both ways), while projection only actually happens when all attempts at delegating failed.

If I got your point this should already happening:
the reader first try to see if the srs is among the supported ones. If not will try to switch it to another definition eg to the urn ogc one. If this one is among the supported ones the tiles will be requested with that CRS, in this case only the Axis order would differ and reprojection is not needed.

taba90:
thanks for your review @aaime your feedbacks have been applied. I've also added few more comments to clarify the various steps.

aaime:
> If I got your point this should already happening: the reader first try to see if the srs is among the supported ones. If not will try to switch it to another definition eg to the urn ogc one. If this one is among the supported ones the tiles will be requested with that CRS, in this case only the Axis order would differ and reprojection is not needed.

What I meant is, I don't see any test checking that the delegation actually happens, rather than picking data in some SRS and then reprojecting locally.

modules/extension/wmts/src/test/java/org/geotools/ows/wmts/online/WMTSOnlineTestCase.java
```diff
@@ -0,0 +1,89 @@
...
+import org.geotools.test.OnlineTestCase;
+import org.geotools.tile.Tile;
+import org.opengis.parameter.GeneralParameterValue;
+import org.opengis.referencing.FactoryException;
+import org.opengis.referencing.crs.CoordinateReferenceSystem;
+
+public abstract class WMTSOnlineTestCase extends OnlineTestCase {
+
+    @Override
+    protected String getFixtureId() {
```
roarbra, line 31: This resembles how it originally was. I would like to keep the separate getFixtureId for different infrastructures. That will keep the tests running at other infrastructures if there is a failure within a single infrastructure. And I would also have a WMTSTestSupport class rather than a super class.

aaime: Online tests in GeoTools commonly extend OnlineTestCase, please let's not add a new infrastructure for online test cases, many of us work across several modules, we cannot have each one roll their own. However, having a dedicated subclass for WMTS, with extra goodies specific for WMTS, may be a good idea.

taba90:
> > If I got your point this should already happening: the reader first try to see if the srs is among the supported ones. If not will try to switch it to another definition eg to the urn ogc one. If this one is among the supported ones the tiles will be requested with that CRS, in this case only the Axis order would differ and reprojection is not needed.
> 
> What I meant is, I don't see any test checking that the delegation actually happens, rather than picking data in some SRS and then reprojecting locally.

Thanks for the point @aaime. Added the required test.

modules/extension/wmts/src/test/java/org/geotools/ows/wmts/online/WMTSOnlineTestCase.java
aaime: BTW, the method is protected and overridable, so different tests can have different fixture files, if needs be.
Ideally, I'd prefer to see something more self-contained, using containers (e.g., see ElasticSearch test, using testcontainers.org), rather that external servers whose contents can change completely out of our control... but I understand that's quite a bit more work, the WMTS server containers [would have to be prepared ad-hoc](https://www.testcontainers.org/features/creating_images/).

PR #3708, opened 2021-12-14
[GEOT-7037] wfs-ng doesn't dispose reponses; connection leak

NielsCharlier:
[GEOT-7037](https://osgeo-org.atlassian.net/browse/GEOT-7037) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

  

# Checklist

### DIFF
--- a/modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/WFSFeatureReader.java
+++ b/modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/WFSFeatureReader.java
@@ -40,6 +40,12 @@ class WFSFeatureReader implements FeatureReader<SimpleFeatureType, SimpleFeature
 
     private GetFeatureResponse response;
 
+    /** @Deprecated (pass on response so it can be disposed.) */
+    @Deprecated
+    public WFSFeatureReader(final GetParser<SimpleFeature> parser) throws IOException {
+        this(parser, null);
+    }
+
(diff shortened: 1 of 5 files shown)


modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/WFSFeatureReader.java
```diff
@@ -37,7 +38,12 @@
 
     private SimpleFeatureType featureType;
 
-    public WFSFeatureReader(final GetParser<SimpleFeature> parser) throws IOException {
+    private GetFeatureResponse response;
+
+    public WFSFeatureReader(
```
mprins, line 43: This is an API change without a deprecation cycle, I think the old constructor can be deprecated and call this with `null` for the response

NielsCharlier: Okay, I'll change that (though it's a package private class never intended to be called directly, you never know I guess).

mprins: that skipped my attention, in that case it would be OK I guess

mprins (approved):

NielsCharlier:
Thank you for the review @mprins

PR #3722, opened 2021-12-27
[GEOT-7044] WMTSTileService should use WebMapTileServer to load images

roarbra:
[GEOT-7044](https://osgeo-org.atlassian.net/browse/GEOT-7044) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

WMTSTileService will use WebMapTileServer to load images. To make this work I've added some constructors in WMTSTileService and overridden the function loadImageTileImage.

I've also changed it so that in case of a RESTful url we call getTemplateUrl instead of getFinalUrl.

# Checklist


modules/extension/wmts/src/main/java/org/geotools/ows/wmts/WebMapTileServer.java
```diff
@@ -205,46 +212,142 @@ public GetTileResponse issueRequest(GetSingleTileRequest tileRequest)
         return (GetTileResponse) internalIssueRequest(tileRequest);
     }
 
-    /** Creates a GetMultiTileRequest, signature kept for now. */
+    /** Creates a GetMultiTileRequest. */
     public GetTileRequest createGetTileRequest() {
 
-        URL url;
         OperationType getTile = getCapabilities().getRequest().getGetTile();
-        if (getTile == null) {
```
ianturton, line 213: what happens now if `getTile` returns null because there is no getTile element in the capabilities?

ianturton (approved):

modules/extension/wmts/src/main/java/org/geotools/ows/wmts/WebMapTileServer.java
roarbra: @ianturton Good question. The function findGetTileURL will return null, and then we will have the same setting as before.
I would very much like to add unit tests to cover all possible configurations.

ianturton: I think `testGEOT6741` should cover this situation, so I guess we're good.

modules/extension/tile-client/src/main/java/org/geotools/tile/TileService.java
```diff
@@ -56,28 +54,55 @@
...
+    private final String name;
     private final HTTPClient client;
 
     /**
-     * Create a new TileService with a name and a base URL
+     * Creates a TileService
+     *
+     * <p>Client isn't set so you should override loadImageTileImage.
+     *
+     * @param name
```
jodygarnett, line 75: What is the name? Name of tileset? Javadoc is incomplete …

aaime:
Oddly enough, the JDK 11 build failed with an extlink not using the short syntax... but not the others.
Anyway, it's about NetCDF docs, unrelated to this PR.

aaime:
@jodygarnett I see you added a question but then marked it as resolved, was that intended?

roarbra:
@aaime I've pushed a commit that adresses Jody's comment.

PR #3723, opened 2021-12-28
Make GeoJSON store usable in all GeoServer modules

aaime:
This is a set of changes I've added to allow GeoJSONStore be used as a replacement for gt-geojson in GeoServer. 
It's really a follow up to "[GEOT-6887] Bring GeoJsonWriter/GeoJsonReader on par with gt-geojson", could use the same Jira ticket or a new one.

Also sneaked in a performance improvement and [started looking into a more significant one](https://github.com/FasterXML/jackson-core/issues/346) that will require some Jackson changes.

@ianturton could use your review.

# Checklist

### DIFF
--- a/modules/unsupported/elasticsearch/src/test/java/org/geotools/data/elasticsearch/ElasticFilterTest.java
+++ b/modules/unsupported/elasticsearch/src/test/java/org/geotools/data/elasticsearch/ElasticFilterTest.java
@@ -648,15 +648,15 @@ public class ElasticFilterTest {
 
     @Test
     public void testGeoShapeBboxFilter() {
-        BBOX filter = ff.bbox("geom", 0., 0., 1.1, 1.1, "EPSG:4326");
-        List<List<Double>> coords = new ArrayList<>();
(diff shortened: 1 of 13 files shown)


aaime:
ElasticSearch failures are legit, will have to adjust expectations, or have a minimum of one decimal again, and adjust GeoServer tests as a consequence.... personally I prefer more compact numbers, rather than having  ".0" in all coordinates expressed as integer numbers. Was there a rationale to force at least one decimal?

And oh, while I'm here, while 4 decimals are more than enough for common cartography, they are not enough for modern municipal cartography where centimeter precision is required. XKCD is fun, but this table is more informative:
https://en.wikipedia.org/wiki/Decimal_degrees

I'd say at least 6 digits are required (10cm precision), though the conversations I've had in recent times with municipalities were considering 1cm the tolerable error.

ianturton (approved):

aaime:
@ianturton any comment on the number of digits bit?

aaime:
Merging so that I can move on with the gitattributes work, but still interested in the decimal places bit :-p

PR #3737, opened 2022-01-12
[GEOT-7052] OperationsMetadata is optional for RESTful WMTS

roarbra:
[GEOT-7052](https://osgeo-org.atlassian.net/browse/GEOT-7052) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

Fixing a problem where OperationsMetadata isn't included in the capabilities document. In the original code this section were used to find a serverUrl. When this section is missing, there isn't actually anyway to get the serverUrl from the capabilities document.

I've introduced new constructors that takes serverUrl as arguments, and deprecated the old ones.

This is a fix for [GEOS-10354](https://osgeo-org.atlassian.net/browse/GEOS-10354). In addition the constructor for org.geoserver.security.decorators.SecuredWebMapTileServer should call `super(delegate);`.

# Checklist

### DIFF
--- a/modules/extension/wmts/src/main/java/org/geotools/ows/wmts/WebMapTileServer.java
+++ b/modules/extension/wmts/src/main/java/org/geotools/ows/wmts/WebMapTileServer.java
@@ -17,6 +17,7 @@
 package org.geotools.ows.wmts;
 
 import java.io.IOException;
+import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.Map;
 import java.util.Properties;
@@ -87,6 +88,21 @@ public class WebMapTileServer extends AbstractOpenWebService<WMTSCapabilities, L
 
     /**
      * Set up the WebMapTileServer by the given capabilities, with the given serverURL. Using the
+     * default httpClient.
+     *
+     * @param serverURL
+     * @param httpClient
+     * @param capabilities
+     * @throws ServiceException
+     * @throws IOException
+     */
+    public WebMapTileServer(URL serverURL, WMTSCapabilities capabilities)
+            throws ServiceException, IOException {
+        this(serverURL, HTTPClientFinder.createClient(), capabilities, null);
+    }
+
+    /**
+     * Set up the WebMapTileServer by the given capabilities, with the given serverURL. Using the
      * given httpClient.
      *
      * @param serverURL
@@ -97,8 +113,7 @@ public class WebMapTileServer extends AbstractOpenWebService<WMTSCapabilities, L
      */
     public WebMapTileServer(URL serverURL, HTTPClient httpClient, WMTSCapabilities capabilities)
             throws ServiceException, IOException {
-        super(serverURL, httpClient, capabilities);
-        setupType();
+        this(serverURL, httpClient, capabilities, null);
     }
 
     /**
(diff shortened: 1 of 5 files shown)


modules/extension/wmts/src/test/java/org/geotools/ows/wmts/model/WMTSCapabilitiesTest.java
```diff
@@ -300,6 +300,20 @@ public void testParseWithIncompleteAddressMetadata() throws Exception {
...
+     * argument for WebMapTileServer shouldn't make a NullPointerException.
+     */
+    @Test
+    public void testMissingOperationsMetadata() throws Exception {
+        WMTSCapabilities capabilities = createCapabilities("linz_basemaps_capabilities.xml");
+        WebMapTileServer tileServer =
+                new WebMapTileServer(
+                        new URL("https://basemaps.linz.govt.nz/v1/tiles/WMTSCapabilities.xml"),
+                        capabilities);
+        Assert.assertNotNull(tileServer);
```
ianturton, line 314: Is this enough to prove the fix?works, original bug report indicated the error is thrown on requesting actual tiles not creation

modules/extension/wmts/src/main/java/org/geotools/ows/wmts/WebMapTileServer.java
```diff
@@ -151,22 +168,37 @@ public WebMapTileServer(WMTSCapabilities capabilities) throws ServiceException,
...
      * @throws ServiceException
      * @throws IOException
+     * @deprecated Use constructor with serverUrl, capabilities and httpClient.
      */
+    @Deprecated
     public WebMapTileServer(WMTSCapabilities capabilities, HTTPClient httpClient)
             throws ServiceException, IOException {
-        super(capabilities.getRequest().getGetCapabilities().getGet(), httpClient, capabilities);
+        super(extractServerURL(capabilities), httpClient, capabilities);
         setupType();
```
ianturton, line 177: How does setuptype react if there is no getCapabilities object, might be best to go with a null and force the code to use the input url as the base, and make sure the type is set to REST

modules/extension/wmts/src/test/java/org/geotools/ows/wmts/model/WMTSCapabilitiesTest.java
roarbra: It's the constructor of WMTSCapabilities that fails, sending in serverURL=null. This test failed before I did my changes, but because of the deprecation I changed the constructor.
Not sure if that's the way to do it.

modules/extension/wmts/src/main/java/org/geotools/ows/wmts/WebMapTileServer.java
roarbra: I'm sending in the capabilities, so no change there for setupType.
When it comes to the url. The problem is that there are no useful url within the capabilities. What do you mean by input url?

ianturton: In the last resort we can fall back to the URL we got the getCapabilities from, not sure if we do currently,

aaime:
Looks good to me. The downstreams failure is unrelated.

PR #3740, opened 2022-01-14
[GEOT-7011] GeoPackage.add(FeatureEntry, SimpleFeatureCollection) forced to XY order

jodygarnett:
[GEOT-7011](https://osgeo-org.atlassian.net/browse/GEOT-7011) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

This is a port of a fix proposed here https://github.com/geoserver/geoserver/pull/5558 for the geopackage output format. 

Feedback from @groldan indicated the fix could be better applied directly in the GeoPackage API.

# Checklist

- [ ] There is an issue: [ in [GeoTools Jira](https://osgeo-org.atlassian.net/projects/GEOT) (except for changes not visible to end users). ](https://osgeo-org.atlassian.net/browse/GEOT-7011)

### DIFF
--- a/modules/plugin/geopkg/src/main/java/org/geotools/geopkg/GeoPackage.java
+++ b/modules/plugin/geopkg/src/main/java/org/geotools/geopkg/GeoPackage.java
@@ -56,6 +56,7 @@ import org.geotools.data.simple.SimpleFeatureIterator;
 import org.geotools.data.simple.SimpleFeatureReader;
 import org.geotools.data.simple.SimpleFeatureSource;
 import org.geotools.data.simple.SimpleFeatureWriter;
+import org.geotools.data.store.ReprojectingFeatureCollection;
 import org.geotools.filter.identity.FeatureIdImpl;
 import org.geotools.geometry.GeneralEnvelope;
 import org.geotools.geometry.jts.Geometries;
@@ -82,6 +83,7 @@ import org.opengis.feature.type.PropertyDescriptor;
 import org.opengis.filter.Filter;
 import org.opengis.filter.identity.Identifier;
 import org.opengis.referencing.FactoryException;
+import org.opengis.referencing.ReferenceIdentifier;
 import org.opengis.referencing.crs.CoordinateReferenceSystem;
 import org.sqlite.Function;
 import org.sqlite.SQLiteConfig;
@@ -796,6 +798,38 @@ public class GeoPackage implements Closeable {
     }
 
     /**
+     * According to GeoPKG spec, the coordinates MUST be in XY order. If this FC is in YX format, we
+     * reproject to the equivalent XY project.
+     *
+     * <p>If already XY, return the input FC.
+     *
+     * @param fc underlying feature collection
+     * @return feature collection which is has axis order in XY (NORTH_EAST)
+     */
+    public static SimpleFeatureCollection forceXY(SimpleFeatureCollection fc) {
+        CoordinateReferenceSystem sourceCRS = fc.getSchema().getCoordinateReferenceSystem();
+        if ((CRS.getAxisOrder(sourceCRS) == CRS.AxisOrder.EAST_NORTH)
+                || (CRS.getAxisOrder(sourceCRS) == CRS.AxisOrder.INAPPLICABLE)) {
+            return fc;
+        }
+
+        for (ReferenceIdentifier identifier : sourceCRS.getIdentifiers()) {
+            try {
+                String _identifier = identifier.toString();
+                CoordinateReferenceSystem flippedCRS = CRS.decode(_identifier, true);
+                if (CRS.getAxisOrder(flippedCRS) == CRS.AxisOrder.EAST_NORTH) {
+                    ReprojectingFeatureCollection result =
+                            new ReprojectingFeatureCollection(fc, flippedCRS);
+                    return result;
+                }
+            } catch (Exception e) {
+                // couldn't flip - try again
+            }
+        }
+        return fc;
+    }
+
+    /**
      * Adds a new feature dataset to the geopackage.
      *
      * @param entry Contains metadata about the feature entry.
@@ -806,6 +840,8 @@ public class GeoPackage implements Closeable {
     public void add(FeatureEntry entry, SimpleFeatureCollection collection) throws IOException {
         FeatureEntry e = new FeatureEntry();
         e.init(entry);
+
+        collection = forceXY(collection);
 
         if (e.getBounds() == null) {
             e.setBounds(collection.getBounds());
--- a/modules/plugin/geopkg/src/main/java/org/geotools/geopkg/GeoPackage.java
+++ b/modules/plugin/geopkg/src/main/java/org/geotools/geopkg/GeoPackage.java
@@ -806,7 +806,7 @@ public class GeoPackage implements Closeable {
      * @param fc underlying feature collection
      * @return feature collection which is has axis order in XY (NORTH_EAST)
      */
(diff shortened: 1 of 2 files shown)


modules/plugin/geopkg/src/main/java/org/geotools/geopkg/GeoPackage.java
```diff
@@ -807,6 +841,8 @@ public void add(FeatureEntry entry, SimpleFeatureCollection collection) throws I
         FeatureEntry e = new FeatureEntry();
         e.init(entry);
 
+        collection = forceXY(collection);
```
jodygarnett, line 844: The origional geoserver issue indicated this was a common problem for GeoPackage and CSV; if we see more instances of forceXY being needed to would be kind to add a utility method to DataUtilities.

modules/plugin/geopkg/src/test/java/org/geotools/geopkg/GeoPackageTest.java
```diff
@@ -1123,4 +1120,82 @@ private void createBugSites() throws Exception {
         geopkg.add(entry, shp.getFeatureSource(), null);
         assertTableExists("bugsites");
     }
+
+    // if the FC already is XY, then forceXY does nothing (should return same FC)
+    @Test
+    public void testForceXY_alreadyXY() throws Exception {
```
jodygarnett, line 1126: Unusual name (no java camelCase), I can see what you are trying to do. Is their a JUnit convention to follow?

modules/plugin/geopkg/src/test/java/org/geotools/geopkg/GeoPackageTest.java
```diff
@@ -1123,4 +1120,82 @@ private void createBugSites() throws Exception {
...
+        String wkt_xy =
+                "GEOGCS[\"WGS 84\", \n"
+                        + "  DATUM[\"World Geodetic System 1984\", \n"
+                        + "    SPHEROID[\"WGS 84\", 6378137.0, 298.257223563, AUTHORITY[\"EPSG\",\"7030\"]], \n"
+                        + "    AUTHORITY[\"EPSG\",\"6326\"]], \n"
+                        + "  PRIMEM[\"Greenwich\", 0.0, AUTHORITY[\"EPSG\",\"8901\"]], \n"
+                        + "  UNIT[\"degree\", 0.017453292519943295], \n"
+                        + "  AXIS[\"Geodetic latitude\", EAST], \n"
+                        + "  AXIS[\"Geodetic longitude\", NORTH], \n"
+                        + "  AUTHORITY[\"EPSG\",\"4326\"]]";
```
jodygarnett, line 1137: You may be able to directly use the `DefaultGeographicCRS.WGS84`; it is a constant defined with the above axis order.

jodygarnett (commented):
Thanks @davidblasby, some optional feedback below. I will a note about this to the geopackage docs.

modules/plugin/geopkg/src/main/java/org/geotools/geopkg/GeoPackage.java
davidblasby: That's certainly a possibility.  There are 2 reprojecting feature collections - one in GT and one in GS.  The one is GS operates a bit differently - it first uses the underlying Geometry's userdata (expecting to see a CRS) instead of the one in the Schema.

modules/plugin/geopkg/src/test/java/org/geotools/geopkg/GeoPackageTest.java
davidblasby: I can change this - but is common to see "_" in test case names.  I wanted to make the function name obvious as to what its doing.  I will change to `testForceXYAlreadyXY` and the same, below.

modules/plugin/geopkg/src/test/java/org/geotools/geopkg/GeoPackageTest.java
davidblasby: I'd like to be explicit (and obvious) and be consistent with the other test case, below.

jodygarnett:
> Thanks @davidblasby, some optional feedback below. I will a note about this to the geopackage docs.

David I found that the geopackage docs are quite out of date and will make a seperate pull request.

modules/plugin/geopkg/src/test/java/org/geotools/geopkg/GeoPackageTest.java
davidblasby: I added a comment to make this more obvious.

modules/plugin/geopkg/src/main/java/org/geotools/geopkg/GeoPackage.java
jodygarnett: Understood :)

I think if we do this for real as part of DataUtilities we would need to update the CoordianteReferenceSystem stored in the Geometry user data (for output formats such as GML that can benefit).

aaime:
The ticket title should not refer to WFS, there is no WFS server in GeoTools. This might be better:
"Force GeoPackage geometry ordinate order to XY, as required by specification"

modules/plugin/geopkg/src/main/java/org/geotools/geopkg/GeoPackage.java
```diff
@@ -795,6 +797,38 @@ public void create(FeatureEntry entry, SimpleFeatureType schema) throws IOExcept
...
+    /**
+     * According to GeoPKG spec, the coordinates MUST be in XY order. If this FC is in YX format, we
+     * reproject to the equivalent XY project.
+     *
+     * <p>If already XY, return the input FC.
+     *
+     * @param fc underlying feature collection
+     * @return feature collection which is has axis order in XY (NORTH_EAST)
+     */
+    public static SimpleFeatureCollection forceXY(SimpleFeatureCollection fc) {
```
aaime, line 809: This is a generic utility method (nothing specific to GPKG) and still allows to write ordinates in the wrong direction.
I was expecting to see a fix in the actual writing machinery (which I believe would require a small change in JDBCFeatureStore or JDBCDataStore along with a Dialect change to allow implementation-specific axis order forcing).

jodygarnett:
> The ticket title should not refer to WFS, there is no WFS server in GeoTools. This might be better: "Force GeoPackage geometry ordinate order to XY, as required by specification"

True, updated PR to be really clear about the scope of this fix.

modules/plugin/geopkg/src/main/java/org/geotools/geopkg/GeoPackage.java
jodygarnett: I would also like to see a fix to JDBCDataStore createSchema as a separate PR - for now I am calling it out as a user-beware in the documentation. You are correct that such a change would require an API change (require access to the CoordinateReferenceSystem rather than just the SRID number). If you are interested in outlining fix together @aaime I would be happy to work on it with you.

jodygarnett: @davidblasby could you reduce this method to package visibility (so test cases can run) without committing to a new public api here.

jodygarnett:
Going to rebase to pick up the documentation fixes

jodygarnett:
Thanks for the improvement @davidblasby, going to continue the discussion on JDBCDataStore Dialect change on the https://osgeo-org.atlassian.net/browse/GEOT-7011 ticket

PR #3744, opened 2022-01-17
[GEOT-7054] Like Query Addition to Geopackage

ydeliorman:
[GEOT-7054](https://osgeo-org.atlassian.net/browse/GEOT-7054) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

Like query was not supported in geopackage. With this PR requests using like query will be supported in geopackage.

# Checklist

### DIFF
--- a/modules/library/jdbc/src/main/java/org/geotools/data/jdbc/FilterToSQL.java
+++ b/modules/library/jdbc/src/main/java/org/geotools/data/jdbc/FilterToSQL.java
@@ -393,6 +393,7 @@ public class FilterToSQL implements FilterVisitor, ExpressionVisitor {
         capabilities.addAll(FilterCapabilities.SIMPLE_COMPARISONS_OPENGIS);
         capabilities.addType(PropertyIsNull.class);
         capabilities.addType(PropertyIsBetween.class);
+        capabilities.addType(PropertyIsLike.class);
         capabilities.addType(Id.class);
         capabilities.addType(IncludeFilter.class);
         capabilities.addType(ExcludeFilter.class);
--- a/modules/plugin/geopkg/src/test/java/org/geotools/geopkg/GeoPkgFilterToSQLTest.java
+++ b/modules/plugin/geopkg/src/test/java/org/geotools/geopkg/GeoPkgFilterToSQLTest.java
@@ -1,3 +1,19 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ *
+ *    (C) 2022, Open Source Geospatial Foundation (OSGeo)
+ *
+ *    This library is free software; you can redistribute it and/or
+ *    modify it under the terms of the GNU Lesser General Public
+ *    License as published by the Free Software Foundation;
+ *    version 2.1 of the License.
+ *
+ *    This library is distributed in the hope that it will be useful,
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ *    Lesser General Public License for more details.
+ */
 package org.geotools.geopkg;
 
 import static org.junit.Assert.assertEquals;
--- a/modules/plugin/geopkg/src/test/java/org/geotools/geopkg/GeoPkgFilterToSQLTest.java
+++ b/modules/plugin/geopkg/src/test/java/org/geotools/geopkg/GeoPkgFilterToSQLTest.java
@@ -1,0 +1,53 @@
+package org.geotools.geopkg;
+
+import static org.junit.Assert.assertEquals;
+
+import java.io.StringWriter;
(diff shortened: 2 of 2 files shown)


modules/plugin/geopkg/src/main/java/org/geotools/geopkg/GeoPkgFilterToSQL.java
```diff
@@ -237,4 +238,11 @@ protected Object visitBinarySpatialOperator(
 
         return extraData;
     }
+
+    @Override
+    protected FilterCapabilities createFilterCapabilities() {
```
aaime, line 243: This is worrysome, no clue as to why FilterToSQL does not claim support for PropertyIsLike, looks like an issue that might be affecting other databases.. I've done a quick check, PostGIS and Oracle setup the filter capabilities in a different way and end up having PropertyIsLike support, but for example MySQL is affected by this problem as well.

@ydeliorman could you try declaring support for PropertyIsLike in the base class instead, and see if any test breaks?

ydeliorman: @aaime after moving the logic should I be running the tests on the GeoTools side? Do I need to get a full run at GeoServer sider too?

aaime: Running all database tests on your machine is quite a bit of setup, but the PR checks do it for you so... just make the change and push it here, then see what the various database test runs tell you, e.g.:

image

modules/plugin/geopkg/src/test/java/org/geotools/geopkg/GeoPkgFilterToSQLTest.java
```diff
@@ -0,0 +1,53 @@
...
+        PropertyIsLike like =
+                ff.like(
+                        ff.function("strToLowerCase", ff.property("testString")),
+                        "a_literal",
+                        "%",
+                        "-",
+                        "\\",
+                        true);
+
+        filterToSql.encode(like);
```
aaime, line 49: So before the change in FilterToSQL this line would have thrown a "Filter type not supported" exception right?

aaime (changes requested):

modules/plugin/geopkg/src/test/java/org/geotools/geopkg/GeoPkgFilterToSQLTest.java
ydeliorman: Yes, it was throwing that exception.

modules/plugin/geopkg/src/test/java/org/geotools/geopkg/GeoPkgFilterToSQLTest.java
```diff
@@ -0,0 +1,53 @@
+package org.geotools.geopkg;
```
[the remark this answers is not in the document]
ydeliorman: added

ydeliorman:
@aaime changes are done.

modules/library/jdbc/src/main/java/org/geotools/data/jdbc/FilterToSQL.java
```diff
@@ -393,6 +393,7 @@ protected FilterCapabilities createFilterCapabilities() {
         capabilities.addAll(FilterCapabilities.SIMPLE_COMPARISONS_OPENGIS);
         capabilities.addType(PropertyIsNull.class);
         capabilities.addType(PropertyIsBetween.class);
+        capabilities.addType(PropertyIsLike.class);
```
jodygarnett, line 396: Not sure I get it, was this filter not encoded for any of the JDBCDataStore implementations previously?

Making the change here should require testing in all the datastore’s that use FilterToSQL …

jodygarnett: Reading previous comment it appears that PropertyIsLike was not being encoded for any of the JDBC datastore instances?

This should be a really good improvement then …

aaime: Wrong, it was encoded in most, but not all. Some ignore the setup of FilterToSQL and do their own, which includes PropertyIsLike, but GeoPackage and MySQL inherited the base config from FilterToSQL instead.

aaime: And yep, we have tests checking the result of PropertyIsLike, just not the encoding of it.

aaime (approved):

aaime:
Thankfully we have a good battery of DBMS tests here in the PR checks covering most databases... We should probably require a database to have a PR check in order to be in supported, since changes to the code DB machinery affect them all, but nobody can run all of the database types.

aaime:
@ydeliorman squash merged, backport as you see fit.

PR #3747, opened 2022-01-22
[GEOT-7055] Split gt-geojsonstore in two modules, gt-geojson-core and gt-geojson-store

aaime:
[GEOT-7055](https://osgeo-org.atlassian.net/browse/GEOT-7055) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

As per discussion during the last PSC meeting.
Also added jacoco support so that we can measure code coverage from maven (looks good).

One annoying bit is that the module changed name, which would break existing users.
On the other end, it makes the naming consistent with other cases of related modules, and it's the last good time to do such change, before the modules are proposed for supported status.

GeoServer follow up PR here:
https://github.com/geoserver/geoserver/pull/5549

# Checklist


aaime:
@ianturton thoughts?

modules/unsupported/geojson-core/src/test/java/org/geotools/data/geojson/GeoJSONTestSupport.java
```diff
@@ -0,0 +1,56 @@
...
+
+    private static String removeWhitespace(String actual) {
+        if (actual == null) return "";
+        StringBuffer crush = new StringBuffer(actual);
+        int ch = 0;
+        while (ch < crush.length()) {
+            char chracter = crush.charAt(ch);
+            if (Character.isWhitespace(chracter)) {
+                crush.deleteCharAt(ch);
+                continue;
```
ianturton, line 43: minor nitpick: would an `else` be more efficient here?

ianturton (approved):
Technically, I would say the Jacoco changes should go in a different PR but provided you don't mind I doubt anyone else will :-)

modules/unsupported/geojson-core/src/test/java/org/geotools/data/geojson/GeoJSONTestSupport.java
aaime: Not sure if it's going to be more efficient, but, updated it, and rebased the PR.
I'm going to run a full build on the geoserver side too, and if everything is green, merge.

aaime:
Here goes nothing!

PR #3798, opened 2022-02-18
[GEOT-7076] Enable to use curly braces out of env variable definition

sergitopo:
[GEOT-7076](https://osgeo-org.atlassian.net/browse/GEOT-7076) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

[GEOT-7076] Enable to use curly brackets out of env variable definitions in sql start/release scripts
  
There is a bug in stored start/stop session custom sql script. When the script has a closing bracket char '}' out of a env variable definition, the script is treated as invalid and then it cannot use the db connection at all.

This commit reverts such a limitation and adds some more test cases for valid/invalid sql start/stop connection scripts

# Checklist

### DIFF
--- a/modules/library/jdbc/src/main/java/org/geotools/jdbc/SessionCommandsListener.java
+++ b/modules/library/jdbc/src/main/java/org/geotools/jdbc/SessionCommandsListener.java
@@ -158,38 +158,34 @@ public class SessionCommandsListener implements ConnectionLifecycleListener {
                 // mark the beginning and skip the next character
                 inEnvVariable = true;
                 i++;
-            } else if (curr == '}') {
-                if (inEnvVariable) {
+            } else if (curr == '}' && inEnvVariable) {
+                if (sb.length() == 0)
+                    throw new IllegalArgumentException(
+                            "Invalid empty environment variable reference ${} at " + (i - 1));
 
-                    if (sb.length() == 0)
+                String name = sb.toString();
+                String defaultValue = null;
+                int idx = name.indexOf(',');
+                if (idx >= 0) {
+                    if (idx == 0) {
                         throw new IllegalArgumentException(
-                                "Invalid empty environment variable reference ${} at " + (i - 1));
-
-                    String name = sb.toString();
-                    String defaultValue = null;
-                    int idx = name.indexOf(',');
-                    if (idx >= 0) {
-                        if (idx == 0) {
-                            throw new IllegalArgumentException(
-                                    "There is no variable name before "
-                                            + "the comma, the valid format is '${name,defaultValue}'");
-                        } else if (idx < name.length() - 1) {
-                            defaultValue = name.substring(idx + 1);
-                            name = name.substring(0, idx);
-                        }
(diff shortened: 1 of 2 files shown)


modules/library/jdbc/src/main/java/org/geotools/jdbc/SessionCommandsListener.java
```diff
@@ -159,36 +159,37 @@ Expression expandEviromentVariables(String sql) {
...
                 i++;
             } else if (curr == '}') {
-                if (!inEnvVariable)
-                    throw new IllegalArgumentException(
-                            "Already found a ${ sequence before the one at " + (i + 1));
-
-                if (sb.length() == 0)
-                    throw new IllegalArgumentException(
-                            "Invalid empty environment variable reference ${} at " + (i - 1));
+                if (inEnvVariable) {
```
aaime, line 162: The fix looks good, looking at it now it seems that doing a``if (curr == '}' && inEnvVariable)`` would make it better (one less nesting).

modules/library/jdbc/src/test/java/org/geotools/jdbc/SessionCommandListenerTest.java
```diff
@@ -78,7 +78,48 @@ public void testOnlyRelease() throws Exception {
...
 
     @Test
-    public void testExpandVariables() throws Exception {
+    public void testTwoExpandVariablesAreReplaced() throws Exception {
+        SessionCommandsListener listener =
+                new SessionCommandsListener(
+                        "call startSession('select ${user}, ${version} from 1')", null);
+
+        // check borrow
+        EnvFunction.setLocalValue("user", "user1");
```
aaime, line 87: While I realize the issue was already there, it's probably best to add a @After method that clears out the EnvFunction local variables, as the same JVM is used to run other tests (which may cause them to fail unexpectedly).

modules/library/jdbc/src/test/java/org/geotools/jdbc/SessionCommandListenerTest.java
```diff
@@ -99,12 +140,10 @@ public void testExpandVariables() throws Exception {
...
-            new SessionCommandsListener("startSession('${user')", null);
-            fail("This should have failed, the syntax is not valid");
-        } catch (IllegalArgumentException e) {
-            // fine
-        }
+    public void testInvalidVariableDefinitionMissingClosingBracket() throws Exception {
+        Assert.assertThrows(
+                IllegalArgumentException.class,
+                () -> new SessionCommandsListener("select ${user from 1)", null));
+        conn.commands.clear();
```
aaime, line 147: If the test pass, this line is actually unreachable. Maybe also put this clean- up in the ``@After`` block?

aaime (changes requested):

aaime:
Thank you @sergitopo , you can backport as you see fit.

PR #3806, opened 2022-02-23
[GEOT-7081] Choose correct EPSG library

roarbra:
[GEOT-7081](https://osgeo-org.atlassian.net/browse/GEOT-7081) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

Fixes a problem with choosing correct EPSG-library when urn specifies a version.
The main solution is to give VERSION as a hint to each EPSG-library involved, in this case gt-epsg-wkt.

I've added some logging within AuthorityFactoryAdaptor which I thought were useful.

There were code within DefaultAuthorityFactory that I think made the whole process less intuitive. I think what that code tried to solve, are solved by the FallbackAuthorityFactory.

# Checklist

### DIFF
--- a/modules/library/referencing/src/main/java/org/geotools/referencing/DefaultAuthorityFactory.java
+++ b/modules/library/referencing/src/main/java/org/geotools/referencing/DefaultAuthorityFactory.java
@@ -21,6 +21,8 @@ import java.util.Collection;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 import org.geotools.metadata.iso.citation.Citations;
 import org.geotools.referencing.crs.DefaultGeographicCRS;
 import org.geotools.referencing.factory.AbstractAuthorityFactory;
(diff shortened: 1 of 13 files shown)


mprins (changes requested):
this is missing documentation update; docs state there must be only one EPSG jar

roarbra:
Removed the comment about throwing FactoryException, which is wrong. Replaced it with a comment about it causing conflict, which has been used earlier.

modules/library/referencing/src/main/java/org/geotools/referencing/DefaultAuthorityFactory.java
```diff
@@ -80,36 +84,35 @@ private static AbstractAuthorityFactory getBackingFactory(final boolean longitud
...
-                if (authorities.contains(authority)) {
-                    continue;
-                }
-                for (final Citation check : authorities) {
-                    if (Citations.identifierMatches(authority, check)) {
-                        continue searchNews;
-                    }
-                }
-                factories.add(factory);
+
```
aaime, line 94: So... this change is giving me nightmares. Not because I think it's wrong, per se, but because the referencing subsystem is known for reconfiguring itself dynamically, scanning for factories makes new factories pop up or re-order themselves (don't ask, I don't remember why this happens). That's why there was this loop, which would otherwise seem nonsensical.

That said, all tests are passing. On the other end, the whole ticket is addressing an IDE setup deficiency, not an actual runtime issue.
I'm going to wait until 27-RC is cut, and then merge when main becomes 28.x. This should give us six months to see f there are issues with the change that tests are not finding, and revert before this gets in the user's face.

roarbra:
Thanks!

PR #3807, opened 2022-02-28
[GEOT-7084] - [WFST] : Empty property generates braces during encoding

eropartz:
[GEOT-7084](https://osgeo-org.atlassian.net/browse/GEOT-7084) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

  

# Checklist

### DIFF
--- a/modules/extension/xsd/xsd-wfs/src/main/java/org/geotools/wfs/v2_0/bindings/PropertyTypeBinding.java
+++ b/modules/extension/xsd/xsd-wfs/src/main/java/org/geotools/wfs/v2_0/bindings/PropertyTypeBinding.java
@@ -55,7 +55,7 @@ public class PropertyTypeBinding extends AbstractComplexEMFBinding {
 
     @Override
     protected void setProperty(EObject eObject, String property, Object value, boolean lax) {
-        if ("Value".equals(property) && Map.class.isAssignableFrom(value.getClass())) {
+        if ("Value".equals(property) && value instanceof Map) {
             super.setProperty(eObject, property, null, lax);
         } else {
             super.setProperty(eObject, property, value, lax);
--- a/modules/extension/xsd/xsd-wfs/src/main/java/org/geotools/wfs/v2_0/bindings/PropertyTypeBinding.java
+++ b/modules/extension/xsd/xsd-wfs/src/main/java/org/geotools/wfs/v2_0/bindings/PropertyTypeBinding.java
@@ -16,17 +16,17 @@
  */
 package org.geotools.wfs.v2_0.bindings;
 
+import java.util.Map;
 import javax.xml.namespace.QName;
 import net.opengis.wfs20.PropertyType;
 import net.opengis.wfs20.Wfs20Factory;
+import org.eclipse.emf.ecore.EObject;
 import org.geotools.gml2.GML;
 import org.geotools.gml2.GMLConfiguration;
 import org.geotools.wfs.v2_0.WFS;
 import org.geotools.xsd.AbstractComplexEMFBinding;
-import org.geotools.xsd.ElementInstance;
 import org.geotools.xsd.Encoder;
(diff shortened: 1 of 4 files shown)


modules/extension/xsd/xsd-wfs/src/main/java/org/geotools/wfs/v2_0/bindings/PropertyTypeBinding.java
```diff
@@ -54,9 +54,12 @@ public Element encode(Object object, Document document, Element value) throws Ex
     }
 
     @Override
-    public Object parse(ElementInstance instance, Node node, Object value) throws Exception {
-        // TODO: implement and remove call to super
-        return super.parse(instance, node, value);
+    protected void setProperty(EObject eObject, String property, Object value, boolean lax) {
+        if ("Value".equals(property) && Map.class.isAssignableFrom(value.getClass())) {
```
aaime, line 58: I don't understand what's going on here, could you explain? How do you get from "I received a map" to "hence I should consider the input null"?

Regardless, this bit is NPE prone, the following should work better:

```suggestion
        if ("Value".equals(property) &&  value instanceof Map) {
```

aaime (changes requested):

modules/extension/xsd/xsd-wfs/src/main/java/org/geotools/wfs/v2_0/bindings/PropertyTypeBinding.java
eropartz: When a WFS property contains an empty value, this value is transformed has Map during the parsing

- WFS

`<wfs:Property xmlns:wfs="http://www.opengis.net/wfs/2.0" xmlns:fes="http://www.opengis.net/fes/2.0" xmlns:gml="http://www.opengis.net/gml/3.2" xmlns:ows="http://www.opengis.net/ows/1.1" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:xs="http://www.w3.org/2001/XMLSchema">
   <wfs:ValueReference action="replace">myproperty</wfs:ValueReference>
   <wfs:Value/>
</wfs:Property>`

- WFS Property Parsed

image

- WFS Property Encoded : `<wfs:Value>{}</wfs:Value>`
The Map is serialized with **{}**

aaime: I see... however, the approach currently looks like "it's solving my contingent problem" rather than "it's the right thing to do in general". 
What if the map was not empty for example? Of if the property data type was actually meant to be a map? WFS-T is currently not implemented for complex features, but the contents of Value could be in general a complex object, not necessarily a simple number/string/date.

Also, you say the issue happens during parsing, but the ticket title talks about encoding only... I guess I'm missing some bit of the whole story? A transaction is normally built programmatically, however here is seems you are reading a transaction and then encoding it back?

eropartz: OK, we have seen this error with Geoserver. Some WFST request generated error on our database : `UPDATE XX SET "YYY" = '{}'  WHERE ("code" = 'ZZZ)`

The ticket talk about encoding, because our problem appears during this operation. But, for me the cause is the parsing.
What do you prefer? A check to control if the map is empty?

aaime: So you have a GeoServer with a WFS store, that's cascading another WFS server, and it's sending over the WFS-T requests it's receiving? (that would involve both parsing and re-encoding in fact).

Ideally, I'd look at parsing and see that it does not emit an empty map when the input XML has a null value.

PR #3809, opened 2022-03-02
[GEOT-7086] Introduce Log42JLoggerFactory for integration with Log4j 2 API

jodygarnett:
[GEOT-7086](https://osgeo-org.atlassian.net/browse/GEOT-7086) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

This pull-request contributes:

- [GEOT-7104](https://osgeo-org.atlassian.net/browse/GEOT-7104) Clarify GeoTools.init() workflow and logging setup and configuration
- [GEOT-7086](https://osgeo-org.atlassian.net/browse/GEOT-7086) Introduces Log4J2LoggerFactory making use of LOG4J 2 api, making use of CONFIG and FINER using Log4J markers
- [GEOT-7099](https://osgeo-org.atlassian.net/browse/GEOT-7099) Introduces LogbackAdapterFactory (making use of SLF4J API, with ability to use setLevel if logback-classic used as a backend). Also supports use of CONFIG and FINER with logback markers.
- [GEOT-4248](https://osgeo-org.atlassian.net/browse/GEOT-4248) Support logging to SLF4J - see above
- [GEOT-7100](https://osgeo-org.atlassian.net/browse/GEOT-7100) Tests Log4JLoggerFactory against reload4J dependencies
- [GEOT-5625](https://osgeo-org.atlassian.net/browse/GEOT-5625) introducing Log4JLoggerFactory levels for CONFIG and FINEST

In addition:
- Update developers guide source code conventions on logging with some testing guidance, moving discussion of logging to user guide
- Provides some guidance in our user guide for logging integration, using live code examples from integration tests (Integration tests use CC0 header)
- Sort pom properties with comments for functional breaks
- Introduces integration tests to `metadata/src/it`to confirm functionality, and uses as documentation examples

Keep in mind:
- NetCDF plugin requires SLF4J API for one of its dependencies, set up to test with org.slf4j:slf4j-jdk14

Related:

* https://github.com/GeoWebCache/geowebcache/pull/1025
* https://github.com/geoserver/geoserver/pull/5734
  
# Checklist


jodygarnett:
I am quite excited by this result, only the downstream integration builds failing.

aaime:
Some downstreams projects could still be using Log4J 1.2.x, or have switched to [Reload4J](https://reload4j.qos.ch/).
Rather than switching the existing logger factory to 2.x, it may be better to:

* Create a new Log4J2Factory that depends on Log4J
* Deprecate Log4JFactory, move it to its own unsupported module, and make it use reload4j  (or just keep it there and make it depend on reload4j)

jodygarnett:
I was able to support reload4j alongside log4j and logback.

I have added integration tests to the metadata module (only way to test GeoTools.init() method which depends on class path). So far the result is "good" but complicated, I think everyone would prefer if these were each their own tutorial as per https://github.com/geotools/geotools/pull/3823

Still there is something to be said for making use of maven built-in facilities (`modules/library/metadata/src/it/log4j`), rather than a convention of building some examples(`tutorial/log4j`).

modules/library/metadata/pom.xml
```diff
@@ -134,6 +127,38 @@
...
+      <artifactId>log4j-api</artifactId>
+      <optional>true</optional>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.logging.log4j</groupId>
+      <artifactId>log4j-core</artifactId>
+      <optional>true</optional>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.logging.log4j</groupId>
```
aaime, line 148: Is this one actually needed in GeoTools?

modules/library/metadata/pom.xml
```diff
@@ -146,6 +171,37 @@
...
+            <goals>
+              <goal>install</goal>
+              <goal>run</goal>
+              <goal>verify</goal>
+            </goals>
+            <configuration>
+              <goals>
+                <goal>exec:exec</goal>
+              </goals>
+              <verifyHookScript>verify.bsh</verifyHookScript>
```
aaime, line 190: Oh... interesting use of integration tests with beanshell.

modules/plugin/coverage-multidim/netcdf/pom.xml
```diff
@@ -108,15 +108,8 @@
...
-      <groupId>org.slf4j</groupId>
-      <artifactId>slf4j-log4j12</artifactId>
-      <version>1.6.4</version>
-      <scope>runtime</scope>
-    </dependency>
-    <dependency>
-      <groupId>org.slf4j</groupId>
-      <artifactId>log4j-over-slf4j</artifactId>
-      <version>1.6.4</version>
+      <groupId>org.apache.logging.log4j</groupId>
```
aaime, line 111: I was just talking with @dromagnoli the other day about this dependency, in the context of the GeoServer NetCDF module... there is seems not to be required after all. Wondering if here it's the same... you could try kicking it away and see if the module builds anyways.

modules/library/metadata/pom.xml
jodygarnett: The choice given by invoker plugin is beanshell or groovy, I chose well.

The real annoying part is having to use `exec:exec` to run java; since use of `exec:java` runs in process and uses the already configured logging environment. To get an independent logging environment `exec:exec` is needed.

modules/plugin/coverage-multidim/netcdf/pom.xml
jodygarnett: That would be *so* much help.

I think it is just saying it needs sl4j api to be available (and hosting application should be able to choose what sl4j uses).

jodygarnett: Reading https://docs.unidata.ucar.edu/netcdf-java/current/userguide/using_netcdf_java_artifacts.html indicates they need at least one jar providing slf4j api. 

I think the least damaging approach would be to use the slf4j-jdk14 dependency for testing, and indicate in our docs that netCDF requires SL4J matching your environment.

docs/developer/conventions/code/log.rst
```diff
@@ -53,27 +61,38 @@ Finer      no unless configured             common when entering, returning, or
...
+
+.. code-block:: java
    
    LOGGER.info("There is a message of interest for ordinary user");
    
-
 Do not use the logging info level as a replacement of System.out.println for displaying debug information to the console.
    
-The info level is for end users. Use the fine, finer or finest levels for debug information, and setup yours $JAVA_HOME/jre/lib/logging.properties file accordingly (see Logging Configuration below).
+The INFO level is for end users. Use the FINE, FINER or FINEST levels for debug information, and setup yours :file:`logging.properties` file accordingly (see Logging Configuration below).
```
ianturton, line 72: You seem to have deleted the Logging Configuration section below that is referenced on this line. Is there still a section on configuration somewhere?

modules/library/metadata/src/main/java/org/geotools/util/logging/Log4JLogger.java
```diff
@@ -49,37 +82,37 @@ public Log4JLogger(final String name, final org.apache.log4j.Logger logger) {
...
-            case 4:
-                return org.apache.log4j.Level.TRACE; // FINER
+            case 5: // FINE
+                return org.apache.log4j.Level.DEBUG;
+            case 4: // FINER
+                return org.apache.log4j.Level.TRACE;
             case 3: // FINEST
+                return FINEST;
             case 2: // (not allocated)
             case 1: // (not allocated)
```
ianturton, line 101: are these two supposed to fall through to 0?

ianturton (approved):

modules/library/metadata/src/main/java/org/geotools/util/logging/Log4JLogger.java
jodygarnett: Yes, `ALL`.

modules/library/metadata/src/main/java/org/geotools/util/factory/GeoTools.java
```diff
@@ -619,19 +619,33 @@ public static Manifest getManifest(Class<?> type) {
...
-     * @param factory The logger factory to use.
+     * <p>If {@code null} is used, the Java logging {@linkplain java.util.logging.Formatter
+     * formatter} for console output is replaced by a {@linkplain
+     * org.geotools.util.logging.MonolineFormatter monoline formatter}.
+     *
+     * @param factory The logger factory to use, or null for native java util logging.
      * @see Logging#setLoggerFactory(LoggerFactory)
      * @since 2.4
      */
     public void setLoggerFactory(final LoggerFactory<?> factory) {
```
jodygarnett, line 637: My own feedback is that this method should be ``initLoggerFactory(factory)` as can only really be usefully called during start up.

Trying to use this for GeoWebCache I ended up just talking to ``Logging`` directly as there was no way to check state:
``` java
if(Logging.GEOTOOLS.getLoggerFactory() == null ){
    GeoTools.setLoggerFactory(Log4J2LoggerFactory.getInstance());
}
```

jodygarnett: oh wow, the method is not even static and cannot be called

jodygarnett:
This is frustrating:
```
[INFO] |  |  \- org.geotools:gt-render:jar:27-SNAPSHOT:compile
[INFO] |  |     +- org.geotools:gt-cql:jar:27-SNAPSHOT:compile
[INFO] |  |     \- com.conversantmedia:disruptor:jar:1.2.13:compile
[INFO] |  |        \- org.slf4j:slf4j-api:jar:1.7.13:compile
```

The dependency *com.conversantmedia:disruptor* above is pulling slf4j, breaking GeoTools init ability to determine what logging system user prefers.

jodygarnett:
Newer versions of ``com.conversantmedia:disruptor`` just use slf4j for test scope, reading https://github.com/conversant/disruptor/issues/21 this was @fgdrf noticing :)

I am going to exclude `slf4j` and see how we go, cannot upgrade as 1.2.19 is compiled with Java 11.

modules/library/metadata/src/main/java/org/geotools/util/factory/GeoTools.java
aaime: GeoServer uses ``Logging.ALL.setLoggerFactory(...)``

pom.xml
```diff
@@ -1294,6 +1359,11 @@
       </dependency>
 
       <!-- Third-party -->
+      <dependency>
+        <groupId>com.conversantmedia</groupId>
+        <artifactId>disruptor</artifactId>
+        <version>1.2.15</version>
```
aaime, line 1365: I see you went back a few versions in the last commit, but I'm confused, 1.2.19 depends on slf4j just like 1.2.15: https://mvnrepository.com/artifact/com.conversantmedia/disruptor/1.2.15
(even the older 1.2.13 had that dependency)

modules/library/metadata/src/it/log4j/src/main/java/org/geotools/tutorial/logging/Log4JIntegration.java
```diff
@@ -0,0 +1,92 @@
...
+        }
+        LOGGER.info("Configuration " + lookupConfiguration());
+
+        LOGGER.finest("Everything is finest...");
+        LOGGER.finer("Everything is finer...");
+        LOGGER.fine("Everything is fine...");
+        LOGGER.config("Everything is configured...");
+        LOGGER.info("Everything is okay.");
+        LOGGER.warning("Everything is alarming!");
+        LOGGER.severe("Everything is terrible!");
```
aaime, line 59: LOL!

aaime (commented):
I guess you might want to add a note in upgrade.rst, especially if there is any API change (seen at least one visiblity change) or just better practice on how to configure things compared to the past (GeoTools.setLogging vs Logging.ALL.setFactory)... or even just to tell users about the new possibilities (mention and link to more extensive docs).

pom.xml
jodygarnett: Yes I found the research from @fgdrf and eventually just added an exclude

aaime:
Still trying to understand... we have dependencies that use slf4j, so they at least need the api jar, or first usage will fail with a ClassNotFoundException. However, the API is not alone, we need to give it also an implementation bridge jar, or the API would not know what to redirect to. Since we default to Log4J, that would seem to be the implementation we put in the classpath, and we'll need to provide instructions to users that want to use another logging framework (to replace the slf4j to the bridge of the chosen logging framework).

jodygarnett:
> Still trying to understand... we have dependencies that use slf4j, so they at least need the api jar, or first usage will fail with a ClassNotFoundException.

Okay I can explain:

1. the `org.apache.logging.log4j:log4j-slf4j-impl* jar includes and implements the slf4j api in one go
2. This replaces the need to depend on the ``org.slf4j:slf4j-api``.
3. Therefore if you have both jars in the application you get a problem (a warning and probably some missing log messages).

aaime:
> 1. the `org.apache.logging.log4j:log4j-slf4j-impl* jar includes and implements the slf4j api in one go

Surprising.... Looking at https://mvnrepository.com/artifact/org.apache.logging.log4j/log4j-slf4j-impl/2.17.2 it seems like it depends on [slf4j-api](https://mvnrepository.com/artifact/org.slf4j/slf4j-api) and checking the jar contents it does not seem like it actually includes the API in the jar.

jodygarnett:
Nope I was wrong, it was a transitive dependency.

The following is also helpful https://docs.spring.io/spring-framework/docs/4.3.10.RELEASE/spring-framework-reference/html/overview.html#overview-logging-log4j

modules/library/metadata/src/main/java/org/geotools/util/factory/GeoTools.java
jodygarnett: Yes, the key functional difference appears to be using forceMonolineConsoleOutput() when passed in null.
Very old martin decisions, probably to keep test case output tidy.

PR #3811, opened 2022-03-04
[GEOT-7088] Fix shapefile resource leaks

sikeoka:
[GEOT-7088](https://osgeo-org.atlassian.net/browse/GEOT-7088) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

This PR fixes file descriptor leaks when reading shapefiles with invalid files for certain file formats.  Some debug logging was added for cases where exceptions from invalid files were being swallowed even when there was no leak.  This PR also fixes read locks on QIX files not being properly tracked so GeoTools wouldn't report the QIX files as being open even though they were still open.

  

# Checklist

### DIFF
--- a/modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/IndexManager.java
+++ b/modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/IndexManager.java
@@ -372,6 +372,7 @@ class IndexManager {
         if (!shpFiles.isLocal()) {
             return null;
         }
+        FileSystemIndexStore idxStore = null;
         URL treeURL = shpFiles.acquireRead(QIX, writer);
         try {
             File treeFile = URLs.urlToFile(treeURL);
@@ -380,14 +381,14 @@ class IndexManager {
                 return null;
             }
 
-            try {
-                FileSystemIndexStore idxStore = new FileSystemIndexStore(treeFile);
-                return idxStore.load(store.shpManager.openIndexFile(), store.isMemoryMapped());
-            } catch (IOException e) {
-                throw new StoreException(e);
-            }
+            idxStore = new FileSystemIndexStore(treeFile, shpFiles);
         } finally {
             shpFiles.unlockRead(treeURL, writer);
+        }
+        try {
+            return idxStore.load(store.shpManager.openIndexFile(), store.isMemoryMapped());
+        } catch (IOException e) {
+            throw new StoreException(e);
         }
     }
 
--- a/modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/ShapefileFeatureSource.java
+++ b/modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/ShapefileFeatureSource.java
@@ -502,6 +502,7 @@ class ShapefileFeatureSource extends ContentFeatureSource {
                     crs = prj.getCoordinateReferenceSystem();
                 }
             } catch (FactoryException fe) {
+                LOGGER.log(Level.FINE, "Error reading prj file", fe);
                 crs = null;
             }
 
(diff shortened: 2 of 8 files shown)


modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/IndexManager.java
```diff
@@ -372,6 +372,7 @@ protected QuadTree openQuadTree() throws StoreException {
         if (!shpFiles.isLocal()) {
             return null;
         }
+        FileSystemIndexStore idxStore = null;
```
aaime, line 375: I don't see what this change is actually accomplishing... by pushing the FileSystemIndexStore variable declaration out, I thought you were going for a cleanup attempt later, but I don't see it. 
So... what's the actual change? You need to unlock the treeURL before trying to load? That seems more of a regression, other processes could try to write on it while the read is happening....

modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/ShapefileFeatureSource.java
```diff
@@ -502,6 +502,7 @@ protected SimpleFeatureType buildFeatureType() throws IOException {
                     crs = prj.getCoordinateReferenceSystem();
                 }
             } catch (FactoryException fe) {
+                LOGGER.log(Level.FINE, "Error reading prj file", fe);
```
aaime, line 505: We have several users setting logging to "VERBOSE" and then complaining there are "errors" in the logs just because they see stack traces. Could you reword the message to at least state this is likely not fatal and could actually even be expected? (same for the other logs added in the PR).

modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/dbf/DbaseFileReader.java
```diff
@@ -181,6 +181,23 @@ private void init(
             final TimeZone timeZone)
             throws IOException {
         this.channel = dbfChannel;
+        boolean initialized = false;
```
aaime, line 184: I see you're trying to clean up in case the initialization fails. Per se this is fine.
However the approach is bothering me, what I see:
* Keep one field initialization in the original method (arbitrary)
* Create a new overloaded method with the same name and slightly different signature, which is actually doing the risky job and initializing some other fields

Could you make things clearer by:
* Creating a new method, calling it "initSafe" (or other name of your choice), do only the try/finally bits in there
* Call the the old init, un-modified

(same goes for the other changes following the same pattern).

modules/plugin/shapefile/src/test/java/org/geotools/data/shapefile/ShapefileDescriptorLeakTest.java
```diff
@@ -0,0 +1,184 @@
...
+import org.geotools.factory.CommonFactoryFinder;
+import org.geotools.util.URLs;
+import org.geotools.util.factory.GeoTools;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+import org.opengis.filter.FilterFactory;
+import org.opengis.filter.spatial.BBOX;
+
+public class ShapefileDescriptorLeakTest {
```
aaime, line 43: Nice testing!

modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/ShapefileSetManager.java
```diff
@@ -115,6 +115,7 @@ protected DbaseFileReader openDbfReader(boolean indexed) throws IOException {
                 }
             } catch (IOException e) {
                 // could happen if cpg file does not exist remotely
+                LOGGER.log(Level.FINE, "Error reading cpg file", e);
```
aaime, line 118: We have several users setting logging to "VERBOSE" and then complaining there are "errors" in the logs just because they see stack traces. Could you reword the message to at least state this is likely not fatal and could actually even be expected? (same for the other logs added in the PR, at least for files that are not required by the shapefile spec).

aaime (changes requested):

modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/ShapefileSetManager.java
sikeoka: I updated the logging.

PR #3821, opened 2022-03-15
[GEOT-7087] Followup to number_format change

roarbra:
[GEOT-7087](https://osgeo-org.atlassian.net/browse/GEOT-7087) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

I discovered some test failures in gt-wfs-ng as a consequence of the change of default locale in the number_format function. This will only be visible when using a locale with a different decimal separator.

# Checklist


roarbra:
@ianturton I saw your message, but didn't have the time to test it then.

aaime:
Hmm... but does that mean the change in the formatting function broke stored queries support? We cannot have random decimal separators in them, they have to be in english.... this code change makes the code pass, but would it work now on a real server?

ianturton:
It did use to be English by default but that caused unexpected results for users with non-english locales set - I would hope that tests in the stored query code would check if it was broken and you two at least would see failures there?

Also on a real server you could add  `-Duser.language=en` to the Java opts if necessary to maintain the old behaviour

PR #3822, opened 2022-03-16
[GEOT-7093] WMTSCoverageReader shouldn't have instance variables for single requests

roarbra:
[GEOT-7093](https://osgeo-org.atlassian.net/browse/GEOT-7093) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

A little rewrite of previous PR that introduced a lot of instance variables and public getters that are related to a single request. Even though this class would easily been used to serve several requests. Maybe even in parallel.

# Checklist

### DIFF
--- a/modules/extension/wmts/src/main/java/org/geotools/ows/wmts/map/WMTSCoverageReader.java
+++ b/modules/extension/wmts/src/main/java/org/geotools/ows/wmts/map/WMTSCoverageReader.java
@@ -71,22 +71,16 @@ public class WMTSCoverageReader extends AbstractGridCoverage2DReader {
     static GridCoverageFactory gcf = new GridCoverageFactory();
 
     /** The WMTS server */
-    WebMapTileServer wmts;
+    final WebMapTileServer wmts;
 
     /** The layer */
-    WMTSLayer layer = null;
+    private final WMTSLayer layer;
 
-    /** The chosen SRS name */
+    /** The default SRS name for this layer */
     String srsName;
 
     /** The format to use for requests */
     String format;
-
-    /** The last GetTile request */
-    private GetTileRequest tileRequest;
-
-    /** The last GetMap response */
-    GridCoverage2D grid;
 
     /** The set of SRS available for this layer */
     Set<String> validSRS;
@@ -94,28 +88,20 @@ public class WMTSCoverageReader extends AbstractGridCoverage2DReader {
     /** The cached layer bounds */
     ReferencedEnvelope bounds;
 
-    /** The last request envelope */
-    // ReferencedEnvelope requestedEnvelope;
+    /** Keep the last tileRequest that was handled by this Reader */
+    private GetTileRequest lastTileRequest;
 
-    /** Last request width */
-    int width;
-
-    /** Last request height */
-    int height;
-
-    /** Last request CRS (used for reprojected GetFeatureInfo) */
-    CoordinateReferenceSystem requestCRS;
-
-    String requestedTime;
+    private String requestedTime;
 
     public final boolean debug = System.getProperty("wmts.debug") != null;
 
     /** Builds a new WMS coverage reader */
     public WMTSCoverageReader(WebMapTileServer server, Layer layer) {
         this.wmts = server;
+        this.layer = (WMTSLayer) layer;
 
         // init the reader
-        setLayer(layer);
+        updateLayerBounds();
 
         // best guess at the format with a preference for PNG (since it's
         // normally transparent)
@@ -135,15 +121,15 @@ public class WMTSCoverageReader extends AbstractGridCoverage2DReader {
         }
     }
 
-    final void setLayer(Layer owsLayer) {
-        this.layer = (WMTSLayer) owsLayer;
+    /** Set the srsName, validSRS and bounds based on the setting in layer */
+    final void updateLayerBounds() {
 
         if (srsName == null) { // initialize from first (unique) layer
 
             // prefer 4326
             for (String preferred :
                     new String[] {"EPSG:4326", "WGS84", "CRS:84", "WGS 84", "WGS84(DD)"}) {
-                if (owsLayer.getSrs().contains(preferred)) {
+                if (layer.getSrs().contains(preferred)) {
                     srsName = preferred;
                     if (LOGGER.isLoggable(Level.INFO)) LOGGER.info("defaulting CRS to: " + srsName);
                 }
@@ -151,7 +137,7 @@ public class WMTSCoverageReader extends AbstractGridCoverage2DReader {
 
             // no 4326, let's see if the layer is offering something valid
             if (srsName == null) {
-                for (String srs : owsLayer.getSrs()) {
+                for (String srs : layer.getSrs()) {
                     try {
                         // check it's valid, if not we crap out and move to the
                         // next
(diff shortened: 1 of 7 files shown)


modules/extension/wmts/src/main/java/org/geotools/ows/wmts/map/WMTSMapLayer.java
```diff
@@ -123,12 +125,14 @@ public boolean isNativelySupported(CoordinateReferenceSystem crs) {
         return getReader().isNativelySupported(crs);
     }
 
+    @Deprecated
     public String getRawTime() {
         return rawTime;
     }
 
+    /** @deprecated - Should be passed in by parameters */
+    @Deprecated
```
aaime, line 134: This deprecation is going to break GeoServer build, it uses setRawTime in GetMap.java

modules/extension/wmts/src/main/java/org/geotools/ows/wmts/map/WMTSMapLayer.java
```diff
@@ -99,8 +99,9 @@ public synchronized ReferencedEnvelope getBounds() {
     }
 
     /** Returns the {@link WebMapTileServer} used by this layer */
+    @Deprecated
     public WebMapTileServer getWebMapServer() {
-        return getReader().wmts;
+        throw new UnsupportedOperationException();
```
aaime, line 104: No, this deprecates and removes the functionality at the same time... it's not how deprecation works.
Deprecated methods must keep on working while they are deprecated, then they can be broken right away.

modules/extension/wmts/src/main/java/org/geotools/ows/wmts/map/WMTSMapLayer.java
aaime: Btw, I cannot find how the time value is managed now, is there an example? I can only see a bit that sets a default value for the time, but no place where an actual time value is provided from outside.

modules/extension/wmts/src/main/java/org/geotools/ows/wmts/map/WMTSCoverageReader.java
```diff
@@ -259,33 +227,27 @@ GridCoverage2D getMap(
...
 
-    private GridCoverage2D getMap(ReferencedEnvelope envelope, ReferencedEnvelope gridEnvelope)
+    GridCoverage2D createTileMap(TileRequest request, ReferencedEnvelope tileEnvelope, String time)
             throws IOException {
         try {
-            if (LOGGER.isLoggable(Level.FINE)) {
-                LOGGER.log(
-                        Level.FINE,
-                        "Issuing request: " + getTileRequest().getFinalURL(),
-                        new RuntimeException("TRACE!"));
```
aaime, line 282: This exception bit was truly bad, but the logging of the request was useful. Is it happening anywhere else?

modules/extension/wmts/src/main/java/org/geotools/ows/wmts/map/WMTSCoverageReader.java
```diff
@@ -734,4 +680,42 @@ private CoordinateReferenceSystem randomSupportedCRS() {
...
+        int width;
+        int height;
+
+        ReferencedEnvelope envelope;
+
+        String time;
+
+        String tileSRS;
+
+        TileRequest(
```
aaime, line 695: So this class accumulates values, which are only used inside the class itself, in order to create a GetTileRequest object almost right after being created... Seems like what a method would do, but in a more complicated way.

If ``initTileRequest`` returned directly a GetTileRequest object, the code would be simpler.

modules/extension/wmts/src/main/java/org/geotools/ows/wmts/map/WMTSCoverageReader.java
```diff
@@ -549,21 +491,25 @@ public String getMetadataValue(String name) {
     }
 
     /** @return the mapRequest */
+    @Deprecated
     public GetTileRequest getTileRequest() {
-        return tileRequest;
+        throw new UnsupportedOperationException();
```
aaime, line 496: See other comments about deprecation, this is not how deprecation works, you might as well just remove the method. Keeping the method would defeat the purpose of this refactor though... maybe ask on the list about having a direct method removal, without deprecation cycle? It does not seem these methods are used in GT or GS, but still, they are public and usable by other clients....

aaime (changes requested):

modules/extension/wmts/src/main/java/org/geotools/ows/wmts/map/WMTSCoverageReader.java
roarbra: I've done it this way, because my intention is to avoid using GetTileRequest. I will go more directly on WMTSTileService.

modules/extension/wmts/src/main/java/org/geotools/ows/wmts/map/WMTSMapLayer.java
roarbra: Sorry about that one, I was a little bit too eager to get rid of it. I mean we should send this in by parameter, but then we should make it more generic. Having support for whatever dimension the server can offer. I will reintroduce everything regarding the requestedTime.

modules/extension/wmts/src/main/java/org/geotools/ows/wmts/map/WMTSCoverageReader.java
roarbra: I've kept the functionality, but added a deprecation.

modules/extension/wmts/src/main/java/org/geotools/ows/wmts/map/WMTSMapLayer.java
roarbra: I've reentered the old functionality.

modules/extension/xsd/xsd-wmts/src/main/java/org/geotools/wmts/bindings/TileMatrixLimitsBinding.java
```diff
@@ -131,7 +131,7 @@ public Object parse(ElementInstance instance, Node node, Object value) throws Ex
     }
 
     private static BigInteger toBigInt(Object o) {
-        return BigInteger.valueOf(((Integer) o).longValue());
+        return BigInteger.valueOf(o instanceof Integer ? ((Integer) o).longValue() : (Long) o);
```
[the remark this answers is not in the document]
roarbra: Done.

modules/extension/wmts/src/main/java/org/geotools/ows/wmts/map/WMTSCoverageReader.java
roarbra: I was just tired of that trace. Never considered the log message itself, but I've reintroduced it at the very end of the handling within WMTSCoverageReader.

aaime:
Hum... processed requests top to bottom and noticed another PR of yours was dealing with requested time as well. Is this one still valid?

roarbra:
Yes, it's still valid.
Not sure if it would be easier if I committed a merge on the PR, or if that would mess up everything?

aaime:
Easiest way to make sure everything is good, rebase your branch on main and force-push.

aaime (approved):

aaime:
Merged, thanks!

PR #3826, opened 2022-03-22
Fixing ImageMosaicCogOnlineTests due to S3 datasets no more available

dromagnoli:
Landsat 8 S3 datasets being referred by Some COG tests are no longer reachable, causing the test to fail.
Switching to use Sentinel2 datasets


modules/plugin/imagemosaic/src/test/java/org/geotools/gce/imagemosaic/ImageMosaicCogOnlineTest.java
```diff
@@ -247,19 +254,18 @@ public void testEmptyMosaic() throws Exception {
     /** Checking time get extracted from remote URL too. */
     @Test
     public void testTimeDimensionMosaic() throws Exception {
-
         final File workDir = prepareWorkingDir("emptycog.zip", "timeMosaic", "");
         try (FileWriter out =
                 new FileWriter(
                         new File(TestData.file(this, "."), "/timeMosaic/datastore.properties"))) {
-            out.write("database=cogmosaic\n");
+            out.write("database=cogtimemosaic\n");
```
aaime, line 261: Is it going to create the db during the test?

See also the setup of the online test:
https://github.com/geotools/geotools/blob/main/.github/workflows/postgis_online.yml

dromagnoli: Yep. it's just H2 db

ianturton (approved):

mprins:
should be backported to 26.x as well I think: https://github.com/geotools/geotools/runs/5633047158?check_suite_focus=true#step:8:506

dromagnoli:
> should be backported to 26.x as well I think: https://github.com/geotools/geotools/runs/5633047158?check_suite_focus=true#step:8:506

Done

PR #3842, opened 2022-04-02
outDimension support z

liujkang:
outDimension support z


aaime:
The checklist is not a suggestion, it's mandatory: I've added it back. Please make sure you worked through all the items in it.

liujkang:
this new WKBWriter() used org.locationtech.jts.io.WKBWriter,when param outputDimension doesn't set. it's use 
image

aaime:
I'm sorry, for the pull request to be considered you have to:
* Target the main branch, not 26.x (you can do that after the PR is merged on main)
* Have a JUnit test proving the fix is working (the test should fail without your changes, and pass with them).

We don't need an explanation of why the code change is working, you need to work on the two requests above, or the PR will be closed.

PR #3843, opened 2022-04-05
[GEOT-7110] Special characters in style names cause malformed GetTileRequests when cascading WMTS

singhdan:
[GEOT-7110](https://osgeo-org.atlassian.net/browse/GEOT-7110) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

Fixes to prevent double encoding of UTF-8 parameters in WMTS requests. The problem was where the parameter was encoded separately and then put into the request template and encoded again. This meant that any '%' signs in the encoded parameter were converted to %25 rather than being left alone. The change uses the Java URI class to perform the encoding into the template which does not do this.

One of the existing unit tests have been updated to include this type of scenario.

# Checklist

- [X] [Documentation](https://github.com/geotools/geotools/tree/main/docs) has been updated (if change is visible to end users).


aaime:
@roarbra this seems up your alley, have time to check this PR?

roarbra (changes requested):
The url encoding within this module is  painful. That's because the parameters could either be a part of the path, or the query string.

You write in the issue that the url are encoded twice. Looking at the code it seems like we're calling URLEncoder.encode both in replaceToken and usePercentEncodingForSpace. I think maybe we should avoid calling it in replaceToken. That would mean that we must call usePercentEncodingForSpace for the other parameters as well.

I would also suggest that you create your own new test cases specific for the problem that you're dealing with. Is that style name a real name?

roarbra:
Great! Sorry I missed the issue. This was a great example to check what I mentioned about KVP / REST behavior.

singhdan:
@roarbra I had to do a bit of refactoring to make sure things got encoded with KVP and REST scenarios so there are a few more changes than in your original comments but I think the code is a little clearer now

singhdan:
Although I'd changed styleName to private, I'd missed a place where styleName is set within the AbstractGetTileRequest itself.

singhdan:
I can't reproduce the Downstream integration build error. I've even created a new [branch](https://github.com/singhdan/geotools/pull/2) in my own repo so I could run the same github actions and it ran through fine. Any ideas what to do?

roarbra:
The failure is due to some NetCDF-error, so I can't see how it relates to this change.

roarbra:
@singhdan I see that you have started to encode the properties within the set-method's of AbstractGetTileRequest. That is causing some of our online test cases to fail. And I also think that we could have other situations were this way of doing it, will cause some trouble. When calling for instance getStyle we do want to have the real style name and not the encoded name.
The encoding should only be done were it's necessary, and that is in combination with generation of the REST url's.

singhdan:
@roarbra Understood. I've changed it to only encode when URLs are being created

roarbra:
Thanks, it looks good now.

aaime:
Squash-merged on main. If you need it on the stable branch too please issue a backport PR against 26.x (using the squashed commit).

roarbra:
Had a little look, and backporting seems like a bigger task. There's a lot of previous PR that isn't backported.

PR #3858, opened 2022-04-21
[GEOT-7120] Implement jsonArrayContains function for Oracle and Postgis

fostash:
JIRA Ticket https://osgeo-org.atlassian.net/browse/GEOT-7120

Implement jsonArrayContains function for Oracle and Postgis
  

# Checklist


modules/library/main/src/main/java/org/geotools/filter/function/JsonArrayContainsFunction.java
```diff
@@ -0,0 +1,78 @@
...
+import com.fasterxml.jackson.core.JsonPointer;
+import com.fasterxml.jackson.core.JsonToken;
+import com.fasterxml.jackson.core.JsonTokenId;
+import java.io.IOException;
+import java.io.StringWriter;
+import org.geotools.filter.FunctionExpressionImpl;
+import org.geotools.filter.capability.FunctionNameImpl;
+import org.opengis.filter.capability.FunctionName;
+
+public class JsonArrayContainsFunction extends FunctionExpressionImpl implements JsonFunctionUtils {
```
nmco, line 33: I would add some JavaDoc providing some info about what this function does.

modules/library/main/src/main/java/org/geotools/filter/function/JsonFunctionUtils.java
```diff
@@ -0,0 +1,94 @@
...
+ */
+package org.geotools.filter.function;
+
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.JsonToken;
+import com.fasterxml.jackson.core.JsonTokenId;
+import java.io.IOException;
+
+public interface JsonFunctionUtils {
```
nmco, line 25: I would make this interface package protected, and add some JavaDoc.

modules/library/main/src/test/java/org/geotools/filter/function/JsonArrayContainsFunctionTest.java
```diff
@@ -0,0 +1,62 @@
+/*
```
nmco, line 1: Thx for caring for code duplication.

modules/library/main/src/main/java/org/geotools/filter/function/JsonArrayContainsFunction.java
```diff
@@ -0,0 +1,78 @@
...
+
+    private static final JsonToken END_OF_STREAM = null;
+    private final JsonFactory factory;
+
+    public static FunctionName NAME =
+            new FunctionNameImpl(
+                    "jsonArrayContains",
+                    Boolean.class,
+                    parameter("column", String.class),
```
nmco, line 42: JSON property or JSON maybe a more abstract name!

modules/library/main/src/main/java/org/geotools/filter/function/JsonArrayContainsFunction.java
```diff
@@ -0,0 +1,78 @@
...
+
+    @Override
+    public Object evaluate(Object object) {
+        final String json = getExpression(0).evaluate(object, String.class);
+        final String pointerSpec = getExpression(1).evaluate(object, String.class);
+        final String expected = getExpression(2).evaluate(object, String.class);
+        boolean found = false;
+
+        if (json != null) {
+            final JsonPointer expectedPointer = JsonPointer.compile(pointerSpec);
```
nmco, line 59: The compilation of expressions are usually quite expensive anyway we can optimize this?

modules/library/main/src/main/java/org/geotools/filter/function/JsonArrayContainsFunction.java
```diff
@@ -0,0 +1,78 @@
...
+                            && parser.currentTokenId() == JsonTokenId.ID_START_ARRAY) {
+                        StringWriter writer = new StringWriter();
+                        try (final JsonGenerator generator = factory.createGenerator(writer)) {
+                            serializeArray(parser, generator);
+                        }
+                        found = writer.toString().contains(expected);
+                    }
+                }
+            } catch (IOException e) {
+                throw new RuntimeException(e);
```
nmco, line 73: Here I would add a bit more context, for example, gin the JSON that makes the parsing fail.

```
LOGGER.severe("Error when parsing the JSON: " + json, exception)
```

modules/library/main/src/test/java/org/geotools/filter/function/JsonArrayContainsFunctionTest.java
```diff
@@ -0,0 +1,62 @@
...
+                        + "    \"arrayDate\" : [],\n" //
+                        + "    \"arrayDouble\" : [3.14, 4.6],\n" //
+                        + "    \"onNestedArrayStr\" : { \"nestedArray\": [\"bar\", \"baz\"] }," //
+                        + "    \"onNestedArrayInt\" : { \"nestedArray\": [4, 87] }" //
+                        + "}";
+
+        assertTrue((boolean) containsFunction(json, "/arrayString", "bar").evaluate(null));
+        assertFalse((boolean) containsFunction(json, "/arrayString", "notpresent").evaluate(null));
+        assertFalse((boolean) containsFunction(json, "/arrayString", 1).evaluate(null));
+        assertTrue((boolean) containsFunction(json, "/arrayInt", 1).evaluate(null));
```
nmco, line 46: I would add a test where we have the number **111** doesn't the array and check if contains **1** and it should return false.
I would add a test where we have the number **aaa** doesn't the array and check if contains **a** and it should return false.

modules/plugin/jdbc/jdbc-oracle/src/test/java/org/geotools/data/oracle/OracleJsonArrayContainsOnlineTest.java
```diff
@@ -0,0 +1,131 @@
...
+import org.geotools.data.store.ContentFeatureSource;
+import org.geotools.factory.CommonFactoryFinder;
+import org.geotools.jdbc.JDBCDelegatingTestSetup;
+import org.geotools.jdbc.JDBCTestSupport;
+import org.opengis.feature.simple.SimpleFeature;
+import org.opengis.filter.Filter;
+import org.opengis.filter.FilterFactory2;
+import org.opengis.filter.expression.Function;
+
+public class OracleJsonArrayContainsOnlineTest extends JDBCTestSupport {
```
nmco, line 31: The tests should be made abstract, you can check another online test for examples, so if someone implements support for this function in another datastore, for example, SQLServer, they can reuse the tests (and they should!)  and they just need to define the data setup in SLQServer.

nmco:
Overall looks good and looks like a useful enhancement to deal with JSON, I just made a couple of comments. I would add to general comments:

- More JavaDoc \ code comments should be added.
- This function should be documented [here](https://docs.geoserver.org/stable/en/user/filter/function_reference.html), I would add a new section for JSON functions.

modules/library/main/src/main/java/org/geotools/filter/function/JsonArrayContainsFunction.java
```diff
@@ -0,0 +1,122 @@
...
+        if (json != null) {
+            JsonPointer expectedPointer = jsonPointersCache.get().get(pointerSpec);
+            if (expectedPointer == null) {
+                expectedPointer = JsonPointer.compile(pointerSpec);
+                jsonPointersCache.get().put(pointerSpec, expectedPointer);
+            }
+            try (JsonParser parser = factory.createParser(json)) {
+                // Iterate over the JSON util reach the END or found the first occurrence of the
+                // expected element
+                while (parser.nextToken() != END_OF_STREAM && !found) {
```
fernandor777, line 86: Please, let's break this while content into a new private method, to make this code more clean.

modules/library/main/src/main/java/org/geotools/filter/function/JsonFunctionUtils.java
```diff
@@ -0,0 +1,95 @@
...
+package org.geotools.filter.function;
+
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.JsonToken;
+import com.fasterxml.jackson.core.JsonTokenId;
+import java.io.IOException;
+
+/** Utility methods for serialize JSON array and object as String */
+interface JsonFunctionUtils {
```
fernandor777, line 26: Making these utils methods on an interface makes the default methods public.  Looks like the methods are internal tooling, maybe could be better to use an abstract class with protected methods or a package protected class and call it as dependency?

modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/PostGISDialect.java
```diff
@@ -1510,7 +1511,8 @@ public Filter[] splitFilter(Filter filter, SimpleFeatureType schema) {
 
                     @Override
                     public Object visit(Function expression, Object notUsed) {
-                        if (expression instanceof JsonPointerFunction) {
+                        if (expression instanceof JsonPointerFunction
+                                || expression instanceof JsonArrayContainsFunction) {
```
fernandor777, line 1515: We need to rely on the function register on the dialect and remove this if clause.

fernandor777 (changes requested):
Overall looks good and works on GeoServer runtime.  Some feedback provided.

fernandor777:
Also some build issues, need to check the build log.

modules/library/main/src/main/java/org/geotools/filter/function/JsonFunctionUtils.java
nmco: Well, this is an interesting one :scream::
- As someone who did Scala, I understand the functional approach, 
- as a Java developer, indeed the abstract class or Utils class with static methods is the way to go.

I think Java wins here.

fernandor777:
Thanks for the fixes, looks good to me.

PR #3861, opened 2022-04-21
[GEOT-7126] Keep custom query parameters in url

roarbra:
[GEOT-7126](https://osgeo-org.atlassian.net/browse/GEOT-7126) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

Some WMTS servers use custom query parameters. By specifying them in the constructor for WebMapTileServer, this solution will make sure they are used in all GetTile requests.

The original functionality in WebMapTileServer for getting the url for KVP requests isn't necessary.

# Checklist

### DIFF
--- a/modules/extension/wmts/src/main/java/org/geotools/ows/wmts/WebMapTileServer.java
+++ b/modules/extension/wmts/src/main/java/org/geotools/ows/wmts/WebMapTileServer.java
@@ -26,7 +26,6 @@ import java.util.logging.Level;
 import org.geotools.data.ResourceInfo;
 import org.geotools.data.ServiceInfo;
 import org.geotools.data.ows.AbstractOpenWebService;
-import org.geotools.data.ows.OperationType;
 import org.geotools.geometry.GeneralEnvelope;
 import org.geotools.http.HTTPClient;
 import org.geotools.http.HTTPClientFinder;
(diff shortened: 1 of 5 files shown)


modules/extension/wmts/src/main/java/org/geotools/ows/wmts/WebMapTileServer.java
```diff
@@ -245,22 +244,13 @@ public GetTileResponse issueRequest(GetSingleTileRequest tileRequest)
...
-        }
-
         Properties props = new Properties();
         props.put("type", type.name());
 
         GetTileRequest request =
                 getSpecification()
                         .createGetMultiTileRequest(
-                                url, props, getCapabilities(), this.getHTTPClient());
+                                serverURL, props, getCapabilities(), this.getHTTPClient());
```
aaime, line 253: I'm not sure about this change... the WMTS specification allows a server to have the capability URL on one host/port/path and the GetTile on an entirely different combination (yes, on a different host)... and other calls or RESTful endpoints can also be spread at different locations. 

A compliant client should always use the values provided in the operations section of the Capabilities document.

roarbra: Ok, it was an easy way to pass on extra parameter's, and it will only work for KVP.
@aaime What if I check that the operations capabilities are respected?

aaime: What do you mean?

roarbra: What I was thinking of was a function that checked if the serverURL have the same host / path as the one specified within the operations section of the capabilities.

aaime: Ok but... by specification you should not really use the server URL (there is no such a thing, it's just the capabilities URL), once you have the capabilities document, all you need to follow are either the backlinks in the Operations section (if KVP) or the Resource URLs (if RESTful).

aaime: Looking at the code seems to confirm my suspicion, the URL provided in the constructor will be used as the base URL to build the final GetTile URL (becomes "onlineResource" in the request classes). As said, this breaks compliance, it would make the changed code work only against WMTS server that co-locate all services on the same machine and path, while the spec allows freedom to put single request on different hosts and path, and in the case of REST requests, to have specific endpoints for different layers (e.g., a bit dynamic, a bit static on S3, and so on).

roarbra:
I will make another try after my new PR has been accepted.

PR #3864, opened 2022-04-26
[GEOT-7129] Allow StreamingParser to set the entity resolver

sikeoka:
[GEOT-7129](https://osgeo-org.atlassian.net/browse/GEOT-7129) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

This PR allows the EntityResolver to be set on a StreamingParser instance.

  

# Checklist

### DIFF
--- a/modules/extension/xsd/xsd-core/src/main/java/org/geotools/xsd/StreamingParser.java
+++ b/modules/extension/xsd/xsd-core/src/main/java/org/geotools/xsd/StreamingParser.java
@@ -25,6 +25,7 @@ import javax.xml.parsers.SAXParserFactory;
 import org.geotools.xsd.impl.ElementNameStreamingParserHandler;
 import org.geotools.xsd.impl.StreamingParserHandler;
 import org.geotools.xsd.impl.TypeStreamingParserHandler;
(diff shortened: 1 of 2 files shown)


modules/extension/xsd/xsd-core/src/main/java/org/geotools/xsd/StreamingParser.java
```diff
@@ -214,6 +215,10 @@ protected StreamingParser(
         this.input = input;
     }
 
+    public void setEntityResolver(EntityResolver entityResolver) {
```
aaime, line 218: Looks good to me... while you're at it, do you want to add a similar method to the PullParser as well? (it's used by the WFS store).

sikeoka: The PullParser uses a different XML API so I can completely disable external entities there but not set an entity resolver.  This shouldn't be a problem for the WFS store since its other feature parser that doesn't use PullParser already disables external entities.

aaime:
Merged, backport as you see fit.

PR #3865, opened 2022-04-27
[GEOT-7130] Automated tests for the SAP HANA data store

stefanuhrig:
[GEOT-7130](https://osgeo-org.atlassian.net/browse/GEOT-7130) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

Data stores with online tests will be required to have automated tests in future.

This pull request adds a GitHub workflow that triggers the HANA data store tests. Unlike other automated data store tests, the HANA tests will connect against a HANA test instance running in the cloud, because there are no up-to-date HANA docker container images. Furthermore, HANA docker containers require too much memory to be able to run on a GitHub workflow instance.

Using a shared HANA test instance requires a configurable database test schema so that tests running in parallel don’t conflict with each other.

# Checklist


stefanuhrig:
Well, good news is that the HANA tests work. Bad news is that it takes very long. That's because the tests are connecting to a database instance in the cloud, so the roundtrip times per request are longer.

I'll try to reduce the number of requests and re-connects...

aaime:
Is there no "developer edition" that could be put in a docker image? Besides timing for tests, I'd guess that instance would be basically open to anyone that can read the code... which would include tampering with it.

stefanuhrig:
There was a community edition once, but it is not maintained anymore. It took at least 8GB RAM to start up, so we could not use it in a GitHub hosted runner anyway.

Of course, anyone can log in using the credentials of the test user and can interfere with tests that are currently running. But I think there is no real incentive to do so. We are using the same testing approach in QGIS and GDAL.

If it becomes a problem, we might switch to an approach where a test run requests an isolated temporary account. Then nobody can interfere with the test run.

aaime:
Ok. The last concern I have is with timing, at 37 minutes it increases the wait for a full green build by 10 minutes, or  37% longer (the second slowest check takes 27 minutes, it's the QA build). Is there any way to make this build run faster? Probably not (if due to latency) but maybe I'm missing some creative approach....

stefanuhrig:
I will do some experimenting. I guess we can reduce the runtime further.

One idea is using the same connection pool for the whole HANA test suite. Currently, a new connection pool is created during test setup, then the pool opens and closes a connection for verification and then the test opens another connection. During test tear down, the pool is disposed and the connection is closed. So, using only one pool should already get rid of over 900 connects and disconnects with all the SSL handshaking and initializing.

Another idea is reducing the number of roundtrips caused by test setup. Currently, the test data is inserted row by row in auto-commit mode. So the client has to wait for confirmation from the database after each row. Using batch-inserts in non-auto-commit mode might help bringing down the runtime further.

Finally, we could run test cases in parallel, each test case using its own schema to not interfere with each other.

My goal is to get the runtime below the longest running test...

stefanuhrig:
Well, drat! There seems to be a bug in Indriya. Although concurrent running threads use their very own local instance of `UnitConverter`, the call of `#convert()` resulted in a `ConcurrentModificationException`.

Issue seems to be here: https://github.com/unitsofmeasurement/indriya/blob/b8e58ca04913c858038217e7c32bc9d26798518f/src/main/java/tech/units/indriya/function/Calculus.java#L150. `piCache` is a static `HashMap` and therefore not synchronized...

I'll open an issue at the Indriya project.

I'll be on vacation for 3 weeks now and will continue work afterwards.

stefanuhrig:
@aaime Test execution time looks good now (4 minutes). A review would be appreciated, at least of the non-HANA specific code changes.

The HANA integration test suite seems to be the first one that uses parallel test execution. So, it might happen that it uncovers/runs into some concurrency issues (like https://github.com/unitsofmeasurement/indriya/issues/371). But I think that's a good thing because those issues are then likely caused by some concurrency bugs.

If there are any failures, please reach out to me. I'll have a look then.

mprins (approved):
LGTM

modules/plugin/jdbc/jdbc-hana/src/test/java/org/geotools/data/hana/ci/CIBase.java
```diff
@@ -0,0 +1,116 @@
+package org.geotools.data.hana.ci;
```
aaime, line 1: Lacks copyright header

modules/plugin/jdbc/jdbc-hana/src/test/java/org/geotools/data/hana/ci/CISetup.java
```diff
@@ -0,0 +1,103 @@
+package org.geotools.data.hana.ci;
```
aaime, line 1: Lacks copyright header

modules/plugin/jdbc/jdbc-hana/src/test/java/org/geotools/data/hana/ci/CITeardown.java
```diff
@@ -0,0 +1,49 @@
+package org.geotools.data.hana.ci;
```
aaime, line 1: Lacks copyright header

modules/plugin/jdbc/jdbc-hana/src/test/java/org/geotools/data/hana/HanaTestSetupBase.java
```diff
@@ -46,6 +77,14 @@ public void setFixture(Properties fixture) {
         String database = fixture.getProperty("database");
         String useSsl = fixture.getProperty("use ssl");
 
+        if ((fixture.getProperty("schemabase") != null)
+                && (fixture.getProperty("schema") == null)) {
+            String schemaBase = fixture.getProperty("schemabase");
+            int counter = schemaCounter.getAndIncrement();
+            String schema = schemaBase + "_" + counter;
```
aaime, line 84: Oh interesting, so each test class is using the same connection but working in a different schema, to avoid stepping on each other toes. And the setup class clears all the schemas before starting up the tests, right?

aaime (changes requested):

modules/plugin/jdbc/jdbc-hana/src/test/java/org/geotools/data/hana/HanaTestSetupBase.java
stefanuhrig: That's the basic idea, the details are slightly different:

- The `CISetup` class generates a unique schema name prefix which is used by that CI run (this enables parallel CI runs)
- Each test gets its own unique schema name to avoid interference with other tests
- Each test gets its own connection(s) from the connection pool that is shared in the CI run (there are too many potential issues in sharing the same connection between two or more threads concurrently)
- The `CITeardown` class deletes the schemas created by the tests
- Because a CI run might be aborted (and the teardown is then not executed), the `CISetup` class will additionally delete geotools test schemas that are older than 24 hours to get rid of leftovers.

modules/plugin/jdbc/jdbc-hana/src/test/java/org/geotools/data/hana/ci/CITeardown.java
stefanuhrig: Done

aaime (approved):

aaime:
Good, merging. Let's see if the backport bot manages to create backpor PRs too.

stefanuhrig:
> Good, merging. Let's see if the backport bot manages to create backpor PRs too.

Most likely not because this pull request depends on #3880. Should I attempt to backport #3880?

aaime:
Never mind, having it on main is good enough IMHO.

PR #3867, opened 2022-04-28
[GEOT-6927] OSM usage of pre-generalized can cause exception

ahmedababnehgeo:
[GEOT-6927](https://osgeo-org.atlassian.net/browse/GEOT-6927) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

  
This PR contains a fix for the problem where OSM usage of pre-generalized can cause " DataSource not available after calling dispose() or before being set".

JIRA ticket is here: https://osgeo-org.atlassian.net/browse/GEOT-6927

# Checklist

### DIFF
--- a/modules/plugin/feature-pregeneralized/pom.xml
+++ b/modules/plugin/feature-pregeneralized/pom.xml
@@ -84,6 +84,11 @@
       <version>${project.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+      <scope>test</scope>
+    </dependency>
 
   </dependencies>
 
--- a/modules/plugin/feature-pregeneralized/src/test/java/org/geotools/data/gen/PreGeneralizedFeatureSourceNotCachedTest.java
+++ b/modules/plugin/feature-pregeneralized/src/test/java/org/geotools/data/gen/PreGeneralizedFeatureSourceNotCachedTest.java
@@ -1,0 +1,96 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ *
+ *    (C) 2002-2022, Open Source Geospatial Foundation (OSGeo)
+ *
+ *    This library is free software; you can redistribute it and/or
+ *    modify it under the terms of the GNU Lesser General Public
+ *    License as published by the Free Software Foundation;
+ *    version 2.1 of the License.
+ *
+ *    This library is distributed in the hope that it will be useful,
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ *    Lesser General Public License for more details.
+ */
+
+package org.geotools.data.gen;
+
+import static org.mockito.ArgumentMatchers.any;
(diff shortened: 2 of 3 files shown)


ahmedababnehgeo:
@aaime Ready for review

modules/plugin/feature-pregeneralized/pom.xml
```diff
@@ -84,6 +84,12 @@
       <version>${project.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+      <version>4.5.1</version>
```
aaime, line 90: Do not track library version numbers in the module poms, they should be managed centrally in the top level library pom instead... which it is happening already for mockito, but it's using version 3.x

If you absolutely need to use 5.x for some reason, the whole library needs to be upgraded, including handling eventual API deprecations (no deprecated API usage tolerated, QA check will fail the build).

Oh, just verified it works also with the current Mockito version.

modules/plugin/feature-pregeneralized/src/test/java/org/geotools/data/gen/PreGeneralizedFeatureSourceNotCachedTest.java
```diff
@@ -0,0 +1,75 @@
+package org.geotools.data.gen;
```
aaime, line 1: Copyright header missing (grab one from other tests, change year to 2022)

modules/plugin/feature-pregeneralized/src/test/java/org/geotools/data/gen/PreGeneralizedFeatureSourceNotCachedTest.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    @Test
+    public void getBoundsHitsRepositoryDataStore() {
+        try {
+            PreGeneralizedDataStore ds = getDataStore(ConfigName);
+            PreGeneralizedFeatureSource fs =
+                    (PreGeneralizedFeatureSource) ds.getFeatureSource("GenStreams");
+
+            fs.getBounds();
```
aaime, line 34: Can you run the getBounds once more, after this call, and verify this will also make the store fetch from the repository?

aaime (changes requested):

modules/plugin/feature-pregeneralized/pom.xml
ahmedababnehgeo: Yeah ... got your point .. version numbers are handled in the top level POM ... removing the version from here suffices

modules/plugin/feature-pregeneralized/src/test/java/org/geotools/data/gen/PreGeneralizedFeatureSourceNotCachedTest.java
ahmedababnehgeo: We can do that ... Will call it again

ahmedababnehgeo:
@aaime Comments have been fixed

aaime:
Thanks @ahmedababnehgeo, as a GeoSolutions collaborator you're covered by the company's CLA. Looks good, I'm merging, backport as you see fit.

PR #3870, opened 2022-05-03
[GEOT-7086] Better map CONFIG, FINEST, FATAL and OPERATION levels

jodygarnett:
[GEOT-7086](https://osgeo-org.atlassian.net/browse/GEOT-7086) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

This change gathers the custom levels in GeoTools (FATAL and OPERATION) to the Logging class for easy reference.

Visibility for custom levels and markers defined logger by logger is also reviewed ensuring they can be accessed by downstream code. This addresses some feedback on https://github.com/geoserver/geoserver/pull/5823 (so custom levels CONFIG and FINEST no longer need to be defined in each configuration file exactly as specified in GeoTools).

# Checklist


aaime:
Builds are failing due to a packaging issue, [this PR attempts to fix it](https://github.com/geotools/geotools/pull/3871).

jodygarnett:
> Builds are failing due to a packaging issue, [this PR attempts to fix it](https://github.com/geotools/geotools/pull/3871).

Thanks Andrea, rebased the PR to try again.

modules/library/metadata/src/main/java/org/geotools/util/logging/Log4J2Logger.java
```diff
@@ -140,20 +143,20 @@ private static Level toJavaLevel(final org.apache.logging.log4j.Level level) {
...
-            return Level.SEVERE;
+            return Logging.FATAL;
         } else if (n <= StandardLevel.ERROR.intLevel()) {
             return Level.SEVERE;
         } else if (n <= StandardLevel.WARN.intLevel()) {
             return Level.WARNING;
         } else if (n <= StandardLevel.INFO.intLevel()) {
             return Level.INFO;
-        } else if (n < StandardLevel.DEBUG.intLevel()) {
+        } else if (n <= CONFIG.intLevel()) {
```
aaime, line 153: I don't know where the need for setting up ``operation`` and ``fatal`` came from, but this seems to be the change that would make config messages appear on the DEFAULT_LOGGING profile in GeoServer?

jodygarnett: I adjusted default logging profile to show config messages in geoserver (it now says level "config" rather than level "info").

jodygarnett: I skipped over mapping FATAL previously as I did not think we had any custom java levels, but then I noticed OPERATION go by during debugging.

PR #3880, opened 2022-05-10
[GEOT-7139] Migrate JDBC tests from JUnit3 to JUnit4

stefanuhrig:
[GEOT-7139](https://osgeo-org.atlassian.net/browse/GEOT-7139) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

JUnit4 offers some advantages over JUnit3 like
 - support for parallel test execution
 - more flexibility via annotations
 - assumptions for skipping tests

# Checklist


stefanuhrig:
``JDBCDataStoreAPIOnlineTest#testGetFeatureWriterConcurrency()`` might block for some providers as I removed a PostGIS specific skip check based on database version from the base class. Let's see... Might require some further adaption in the actual providers...

stefanuhrig:
Will continue tomorrow...

stefanuhrig:
Looks good to me. The number of executed online tests is identical to before for each plugin (it's even one more for GeoPackage because there was a test-method not starting with ``test``).

Some tests are reported as skipped now, because I introduced assumptions in tests of the style

```java
public void test() {
    if (!cond) return;
    // test code
}
```

which look now like this

```java
public void test() {
    assumeTrue(cond);
    // test code
}
```

modules/library/jdbc/src/test/java/org/geotools/jdbc/JDBCGeometryOnlineTest.java
```diff
@@ -41,40 +45,49 @@ public abstract class JDBCGeometryOnlineTest extends JDBCTestSupport {
     @Override
     protected abstract JDBCGeometryTestSetup createTestSetup();
 
+    @Test
```
aaime, line 48: The ``@SuppressWarnings("PMD.JUnit4TestShouldUseTestAnnotation")`` annotation can be removed in this class too.

modules/plugin/jdbc/jdbc-oracle/src/test/java/org/geotools/data/oracle/sdo/SDOOnlineTest.java
```diff
@@ -68,54 +75,65 @@ protected void disconnect() throws Exception {
         super.disconnect();
     }
 
+    @Test
     public final void testGType() throws SQLException {
-        assertEquals(2003, SDO.gType(fixture.rectangle));
+        Assert.assertEquals(2003, SDO.gType(fixture.rectangle));
```
aaime, line 80: Why not sure a static import for this one as well?

aaime (changes requested):
Wow, epic PR!

Looks good to me, found a couple of minor issues, see below.

modules/library/jdbc/src/test/java/org/geotools/jdbc/JDBCGeometryOnlineTest.java
stefanuhrig: Done

modules/plugin/jdbc/jdbc-oracle/src/test/java/org/geotools/data/oracle/sdo/SDOOnlineTest.java
stefanuhrig: The static import would be shadowed by the `assertEquals()` member method (https://stackoverflow.com/questions/6595980/import-static-does-not-work-when-the-class-has-methods-with-the-same-name-as-the).

aaime (approved):

PR #3908, opened 2022-05-26
[GEOT-7150] Name of type in AttributeDescriptor are wrong

roarbra:
[GEOT-7150](https://osgeo-org.atlassian.net/browse/GEOT-7150) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

Removed the setName call. Now the name of AttributeType would either be explicitly set, or it will be the name of the binding class.

The functions buildDescriptor shouldn't been a part of AttributeTypeBuilder rather they should have their own AttributeDescriptorBuilder.

There were just two tests that was written based on wrong type names.

# Checklist

### DIFF
--- a/modules/library/sample-data/src/main/java/org/geotools/test/OnlineTestCase.java
+++ b/modules/library/sample-data/src/main/java/org/geotools/test/OnlineTestCase.java
@@ -25,6 +25,7 @@ import java.util.Map;
 import java.util.Properties;
 import junit.framework.TestCase; // NOPMD
 import junit.framework.TestResult;
+import org.junit.Assume;
 
 /**
  * Test support for test cases that require an "online" resource, such as an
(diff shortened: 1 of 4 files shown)


modules/library/sample-data/src/main/java/org/geotools/test/OnlineTestCase.java
```diff
@@ -243,7 +244,13 @@ protected final void setUp() throws Exception {
                 // disable the test
                 fixture = null;
                 // leave some trace of the swallowed exception
-                java.util.logging.Logger.getGlobal().log(java.util.logging.Level.INFO, "", e);
+                java.util.logging.Logger.getGlobal()
```
aaime, line 247: While you're at it, can you add imports and avoid fully qualified class references?
I've guessing it would avoid most of the line wrapping here.

aaime (changes requested):
Looks good, just added a comment on an annoying wrapping (yes it has been done by the automatic formatter, but we can try to collaborate with it to get something conforming to standard and good looking at the same time).

modules/library/sample-data/src/main/java/org/geotools/test/OnlineTestCase.java
roarbra: Done

PR #3914, opened 2022-05-31
[GEOT-7152] Use table spatial_ref_sys to create CRS

roarbra:
[GEOT-7152](https://osgeo-org.atlassian.net/browse/GEOT-7152) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

The title could be misleading. We're actually using the spatial_ref_sys to get the authority og srid.
Just in cases were our own plugins doesn't give a cos, we will try with the provided WKT in spatial_ref_sys.

# Checklist

### DIFF
--- a/modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCFeatureSource.java
+++ b/modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCFeatureSource.java
@@ -184,6 +184,9 @@ public class JDBCFeatureSource extends ContentFeatureSource {
         // grab the dialect
         SQLDialect dialect = getDataStore().getSQLDialect();
 
+        // logger
+        final Logger storeLogger = getDataStore().getLogger();
+
         // get metadata about columns from database
         try {
             DatabaseMetaData metaData = cx.getMetaData();
@@ -228,7 +231,7 @@ public class JDBCFeatureSource extends ContentFeatureSource {
                         relationships = ((PreparedStatement) st).executeQuery();
                     } else {
                         String sql = getDataStore().selectRelationshipSQL(tableName, name);
-                        getDataStore().getLogger().fine(sql);
+                        storeLogger.fine(sql);
 
                         st = cx.createStatement();
                         relationships = st.executeQuery(sql);
@@ -262,12 +265,10 @@ public class JDBCFeatureSource extends ContentFeatureSource {
 
                 // if still not found, ignore the column we don't know about
                 if (binding == null) {
-                    getDataStore()
-                            .getLogger()
-                            .warning(
-                                    "Could not find mapping for '"
-                                            + name
-                                            + "', ignoring the column and setting the feature type read only");
+                    storeLogger.warning(
+                            "Could not find mapping for '"
+                                    + name
+                                    + "', ignoring the column and setting the feature type read only");
                     readOnly = true;
                     continue;
                 }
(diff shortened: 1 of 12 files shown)


modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/PostGISDialect.java
```diff
@@ -687,6 +688,56 @@ public void handleUserDefinedType(
...
+    /** Using PostGIS table spatial_ref_sys to determine authority name and code of srid */
+    @Override
+    public CoordinateReferenceSystem createCRS(int srid, Connection cx) throws SQLException {
+        if (srid <= 0) {
+            return null;
+        }
+        try (Statement statement = cx.createStatement()) {
+            String sqlStatement =
+                    "SELECT AUTH_NAME, AUTH_SRID, SRTEXT FROM SPATIAL_REF_SYS WHERE SRID = "
+                            + Integer.toString(srid);
```
aaime, line 700: Unnecessary toString, that's what string concatenation does already. Also, please create the sqlStatement outside of the try, and then you can create both the statement and the resultSet in the same try-with-resources block, reducing nesting.

modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/PostGISDialect.java
```diff
@@ -687,6 +688,56 @@ public void handleUserDefinedType(
...
+            try (ResultSet result = statement.executeQuery(sqlStatement)) {
+                if (result.next()) {
```
aaime, line 702: One more level of nesting can be avoided by flipping the condition:

```java
if (!result.next()) {
LOGGER.warning(
                            "SPATIAL_REF_SYS didn't have a row for srid: "
                                    + Integer.toString(srid));
return null;
}
```

Generally speaking, having only 100 columns, try to keep nesting to a minimum and eventually factoring out deeply nested bits to separate methods.

aaime (changes requested):
Generally looks good, see comments below.

modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCFeatureSource.java
```diff
@@ -299,7 +299,23 @@ protected SimpleFeatureType buildFeatureType() throws IOException {
                         } else {
                             srid = dialect.getGeometrySRID(databaseSchema, tableName, name, cx);
                         }
-                        if (srid != null) crs = dialect.createCRS(srid, cx);
+                        if (srid != null) {
+                            crs = dialect.createCRS(srid, cx);
+                            if (crs == null) {
+                                getDataStore()
+                                        .getLogger()
```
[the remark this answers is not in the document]
roarbra: Personally I would've liked to use the LOGGER-variable of JDBCFeatureSource, but it might have been done like this throughout gt-jdbc.

modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/PostGISDialect.java
roarbra: Done

PR #3931, opened 2022-06-17
[GEOT-7165] Failing to find CRS factory for ESRI:102017 with gt-epsg-extension at classpath

roarbra:
[GEOT-7165](https://osgeo-org.atlassian.net/browse/GEOT-7165) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

The main change is within the function createFallbacks where I split Citations with more than one Identifier (Like ESRI, EPSG) into singular Citations. That will make them reachable for both ESRI and EPSG.

Added some logging to dispose which crs is being used in each situation.

Fixed an assertion within a test that I mean is counterintuitive.

# Checklist

### DIFF
--- a/modules/library/referencing/src/main/java/org/geotools/referencing/factory/ManyAuthoritiesFactory.java
+++ b/modules/library/referencing/src/main/java/org/geotools/referencing/factory/ManyAuthoritiesFactory.java
@@ -29,11 +29,13 @@ import org.geotools.metadata.i18n.ErrorKeys;
 import org.geotools.metadata.i18n.Errors;
 import org.geotools.metadata.i18n.Vocabulary;
 import org.geotools.metadata.i18n.VocabularyKeys;
+import org.geotools.metadata.iso.citation.CitationImpl;
 import org.geotools.metadata.iso.citation.Citations;
 import org.geotools.referencing.ReferencingFactoryFinder;
 import org.geotools.util.GenericName;
 import org.geotools.util.factory.Factory;
 import org.geotools.util.factory.FactoryRegistryException;
+import org.opengis.metadata.Identifier;
 import org.opengis.metadata.citation.Citation;
 import org.opengis.referencing.AuthorityFactory;
 import org.opengis.referencing.FactoryException;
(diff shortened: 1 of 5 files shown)


modules/library/referencing/src/main/java/org/geotools/referencing/factory/ManyAuthoritiesFactory.java
```diff
@@ -166,33 +168,45 @@ private static Collection<AuthorityFactory> createFallbacks(
          * the collection of factories for each authority.
          */
         int authorityCount = 0;
-        final Citation[] authorities = new Citation[factories.size()];
-        @SuppressWarnings("unchecked")
-        final List<AuthorityFactory>[] factoriesByAuthority = new List[authorities.length];
+        final List<Citation> authorities = new ArrayList<>(factories.size());
```
aaime, line 171: My my this method is painful to read... what would be needed here is a MultiMap with a custom equality, but there is no multimap in the classpath.

Anyhow the changes seem legit...

modules/library/referencing/src/main/java/org/geotools/referencing/factory/ThreadedAuthorityFactory.java
```diff
@@ -665,7 +665,11 @@ public synchronized CoordinateReferenceSystem createCoordinateReferenceSystem(fi
             crs = (CoordinateReferenceSystem) cached;
         } else {
             crs = getBackingStore().createCoordinateReferenceSystem(code);
+            if (LOGGER.isLoggable(Level.FINER)) {
```
aaime, line 668: Why checked logging here, and usage of a lambda a line below? Both legit, just trying to figure out the logic.

modules/library/referencing/src/main/java/org/geotools/referencing/factory/ManyAuthoritiesFactory.java
roarbra: Yes, the method is painful. Haven't looked at MultiMap, but I do believe it could be written more compact.

modules/library/referencing/src/main/java/org/geotools/referencing/factory/ThreadedAuthorityFactory.java
roarbra: I don't know. Just started to use lambda.

PR #3941, opened 2022-06-22
[GEOT-7167] Upgrade JTS to 1.19.0

ianturton:
[GEOT-7167](https://osgeo-org.atlassian.net/browse/GEOT-7167) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

  

# Checklist

### DIFF
--- a/modules/extension/xsd/xsd-gml3/src/test/java/org/geotools/gml3/bindings/LineStringTypeBindingTest.java
+++ b/modules/extension/xsd/xsd-gml3/src/test/java/org/geotools/gml3/bindings/LineStringTypeBindingTest.java
@@ -82,8 +82,9 @@ public class LineStringTypeBindingTest extends GML3TestSupport {
         LineString line = GML3MockData.lineString();
         Document doc = encode(line, GML.LineString);
 
-        checkDimension(doc, GML.LineString.getLocalPart(), 2);
-        checkPosListOrdinates(doc, 2 * line.getNumPoints());
+        int expectedDim = 2;
+		checkDimension(doc, GML.LineString.getLocalPart(), expectedDim);
+        checkPosListOrdinates(doc, expectedDim * line.getNumPoints());
     }
 
     @Test
@@ -91,9 +92,9 @@ public class LineStringTypeBindingTest extends GML3TestSupport {
         LineString line = GML3MockData.lineStringLite2D();
         Document doc = encode(line, GML.LineString);
 
-        int dimension = CoordinateSequences.coordinateDimension(line.getCoordinateSequence());
-        checkDimension(doc, GML.LineString.getLocalPart(), dimension);
-        checkPosListOrdinates(doc, dimension * line.getNumPoints());
+        int expectedDim = 2;
+		checkDimension(doc, GML.LineString.getLocalPart(), expectedDim);
+        checkPosListOrdinates(doc, expectedDim * line.getNumPoints());
     }
 
     @Test
@@ -101,8 +102,9 @@ public class LineStringTypeBindingTest extends GML3TestSupport {
         LineString line = GML3MockData.lineStringLite3D();
         Document doc = encode(line, GML.LineString);
 
-        checkDimension(doc, GML.LineString.getLocalPart(), 3);
-        checkPosListOrdinates(doc, 3 * line.getNumPoints());
+        int expectedDim = 3;
+		checkDimension(doc, GML.LineString.getLocalPart(), expectedDim);
+        checkPosListOrdinates(doc, expectedDim * line.getNumPoints());
     }
 
     /**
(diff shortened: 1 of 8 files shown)


modules/plugin/jdbc/jdbc-postgis/src/test/java/org/geotools/data/postgis/TWKBReaderTest.java
```diff
@@ -144,7 +144,7 @@ public void testTWKB() throws ParseException, IOException {
 
     @Test
     public void testTWKBM() throws ParseException, IOException {
-        checkTWKBGeometry("a11802", "POINT M EMPTY", 3);
+        // checkTWKBGeometry("a11802", "POINT M EMPTY", 3);
```
jodygarnett, line 147:
```suggestion
        checkTWKBGeometry("a11802", "POINT M EMPTY", 3);
```

jodygarnett (commented):
This is a nice update, it I believe opens up possibility of using JTS TWKB also.

Does LiteCoordinateSequence continue to be useful, or can we make use of a similar class from JTS now?

jodygarnett:
Thank you for the fixes @ianturton, I am a bit too close to the JTS update and would like to invite someone else to approve this PR.

modules/library/main/src/main/java/org/geotools/geometry/jts/coordinatesequence/CoordinateSequences.java
```diff
@@ -177,7 +178,7 @@ public static int coordinateDimension(CoordinateSequence seq) {
         // hack to handle issue that CoordinateArraySequence always reports
         // dimension = 3
         // check if a Z value is NaN - if so, assume dim is 2
-        if (seq instanceof CoordinateArraySequence) {
+        if (seq instanceof CoordinateArraySequence || seq instanceof LiteCoordinateSequence) {
```
aaime, line 181: Not sure about this change, could you elaborate? Maybe related to the removal of the dimension guess in LiteCoordinateSequence?

modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/TWKBReader.java
```diff
@@ -220,18 +220,19 @@ private TWKBMetadata readMetadata() throws IOException {
         metadata.setHeader(header);
 
         int dims = 2;
+        // accordingto https://github.com/TWKB/Specification/blob/master/twkb.md it is Z then M
```
aaime, line 223:
```suggestion
        // according to https://github.com/TWKB/Specification/blob/master/twkb.md it is Z then M
```

So did you compare the JTS and GeoTools implementation? Does the GeoTools one have any improvement related to lower memory allocation or performance in general?

Uh, I tried to have a quick look, the difference seems to be pretty significant in general... in the loop that reads the coordianates I see the JTS code is computing a Math.pow for each and every ordinate value, that does not look so good performance-wise (the GT version computes those once and then the loops reuses them instead).
The JTS code however keeps track of previous values as longs, before the double expansion, which is likely more precise.

I'd say the merger of the two would require some work...

modules/library/main/src/main/java/org/geotools/geometry/jts/coordinatesequence/CoordinateSequences.java
ianturton: Yes, otherwise `LiteCoordinateSequeces` return a dimension of 3 and the GML encoder exports the `Z` values as `NaN`

modules/library/main/src/main/java/org/geotools/geometry/jts/LiteCoordinateSequence.java
```diff
@@ -2,7 +2,7 @@
  *    GeoTools - The Open Source Java GIS Toolkit
  *    http://geotools.org
  *
- *    (C) 2004-2015, Open Source Geospatial Foundation (OSGeo)
+ *    (C) 2004-2022, Open Source Geospatial Foundation (OSGeo)
```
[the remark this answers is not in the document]
ianturton: My formatter messed it up so badly I had to copy and paste it from another file (I forgot I was using a new version of eclipse with out the usual settings)

modules/extension/xsd/xsd-gml3/src/test/java/org/geotools/gml3/bindings/LineStringTypeBindingTest.java
```diff
@@ -90,8 +91,9 @@ public void testEncodeLite2D() throws Exception {
         LineString line = GML3MockData.lineStringLite2D();
         Document doc = encode(line, GML.LineString);
 
-        checkDimension(doc, GML.LineString.getLocalPart(), 2);
-        checkPosListOrdinates(doc, 2 * line.getNumPoints());
+        int dimension = CoordinateSequences.coordinateDimension(line.getCoordinateSequence());
```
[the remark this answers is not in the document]
ianturton: OK, I've reverted those changes but made the "magic" number a variable to make it clearer what is being tested

ianturton:
I'm not sure if we need to update anything else, theoretically this upgrade doesn't break the API anywhere, just adds new objects/methods and handles ZM better/differently - Downstream GWC/GS tests are passing

ianturton:
well those were the tests that threw up the dimensioning issues

ianturton:
Looks like GWC & GeoServer work fine with JTS 1.19.0 

https://github.com/GeoWebCache/geowebcache/pull/1064
https://github.com/geoserver/geoserver/pull/6000

PR #3956, opened 2022-07-07
[GEOT-7175] Add native band selection support to the GeoTIFF reader

aaime:
Adds also a fix in the JIffle process, that became evident while working on the native GeoTIFF band selection process.

# Checklist


aaime:
Marked as a draft for early review, as imageio-ext needs to be released with the band selection improvements before we can upgrade the dependency.

modules/plugin/imagemosaic/src/test/java/org/geotools/gce/imagemosaic/ImageMosaicReaderTest.java
```diff
@@ -5038,6 +5042,41 @@ public void testOverviewSupportFiles() throws Exception {
...
+
+                            @Override
+                            public boolean errorOccurred(
+                                    String message,
+                                    Throwable thrown,
+                                    Object where,
+                                    boolean isRetryable)
+                                    throws RuntimeException {
+                                thrown.printStackTrace(System.out);
+                                return false;
```
dromagnoli, line 5058: I'm not sure I have understood this part.
Is it redirecting the exception stack trace to system out? Is it intentional? (or something there for previous debugging?)

pom.xml
```diff
@@ -479,7 +479,7 @@
     <db2.jdbc.version>11.5.6.0</db2.jdbc.version>
     <eclipse.emf.version>2.15.0</eclipse.emf.version>
     <elasticsearch.version>7.14.0</elasticsearch.version>
-    <imageio.ext.version>1.4.5</imageio.ext.version>
+    <imageio.ext.version>1.4-SNAPSHOT</imageio.ext.version>
```
dromagnoli, line 482: Right. you will need to do a new imageio-ext release to suppor that.

aaime:
The downstream build failures are due to a few tests that assume GeoTIFF cannot do band selection, plus one that just seemed buggy (verified the behavior with QGIS). They are going to be fixed by the GS downstreams PR.

PR #3958, opened 2022-07-08
[GEOT-7177] Remove randomSupportedCRS from WMTSCoverageReader

roarbra:
[GEOT-7177](https://osgeo-org.atlassian.net/browse/GEOT-7177) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

A little change to make it more clear which tileset being used when reprojecting a WMTS tile request.

# Checklist

### DIFF
--- a/modules/extension/wmts/src/main/java/org/geotools/ows/wmts/map/WMTSCoverageReader.java
+++ b/modules/extension/wmts/src/main/java/org/geotools/ows/wmts/map/WMTSCoverageReader.java
@@ -48,6 +48,7 @@ import org.geotools.renderer.lite.RendererUtilities;
 import org.geotools.renderer.lite.gridcoverage2d.GridCoverageRendererUtilities;
 import org.geotools.tile.Tile;
 import org.geotools.util.factory.Hints;
+import org.geotools.util.logging.Logging;
 import org.opengis.coverage.grid.Format;
 import org.opengis.geometry.Envelope;
 import org.opengis.parameter.GeneralParameterValue;
@@ -65,8 +66,7 @@ import org.opengis.referencing.operation.TransformException;
 public class WMTSCoverageReader extends AbstractGridCoverage2DReader {
 
     /** The logger for the map module. */
-    public static final Logger LOGGER =
-            org.geotools.util.logging.Logging.getLogger(WMTSCoverageReader.class);
+    public static final Logger LOGGER = Logging.getLogger(WMTSCoverageReader.class);
 
     static GridCoverageFactory gcf = new GridCoverageFactory();
 
@@ -195,8 +195,15 @@ public class WMTSCoverageReader extends AbstractGridCoverage2DReader {
         try {
             crs = CRS.decode(srsName);
         } catch (Exception e) {
-            if (LOGGER.isLoggable(Level.INFO))
-                LOGGER.log(Level.INFO, "Bounds unavailable for layer" + layer);
+            LOGGER.log(
+                    Level.INFO,
+                    () ->
+                            "Default srs ("
+                                    + srsName
+                                    + ") for layer ("
+                                    + layer
+                                    + ") couldn't be decoded. "
+                                    + e.getMessage());
         }
         this.crs = crs;
         updateBounds();
@@ -461,6 +468,12 @@ public class WMTSCoverageReader extends AbstractGridCoverage2DReader {
 
     /** Updates the coverage bounds based on layer and crs */
     public void updateBounds() {
+        if (crs == null) {
+
+            this.bounds = null;
+            this.originalEnvelope = null;
+            return;
+        }
         if (LOGGER.isLoggable(Level.FINER)) LOGGER.entering("WMTSCoverage", "updatingBounds");
         GeneralEnvelope envelope = layer.getEnvelope(crs);
         ReferencedEnvelope result = reference(envelope);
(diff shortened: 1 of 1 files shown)


modules/extension/wmts/src/main/java/org/geotools/ows/wmts/map/WMTSCoverageReader.java
```diff
@@ -195,8 +195,15 @@ final void updateLayerBounds() {
...
-                LOGGER.log(Level.INFO, "Bounds unavailable for layer" + layer);
+            LOGGER.log(
+                    Level.INFO,
+                    () ->
+                            "Default srs ("
+                                    + srsName
+                                    + ") for layer ("
+                                    + layer
+                                    + ") couldn't be decoded. "
+                                    + e.getMessage());
```
aaime, line 206: Do not eat stack traces:

```suggestion
            LOGGER.log(
                    Level.INFO,
                    () ->
                            "Default srs ("
                                    + srsName
                                    + ") for layer ("
                                    + layer
                                    + ") couldn't be decoded. "
                                    , e);
```

modules/extension/wmts/src/main/java/org/geotools/ows/wmts/map/WMTSCoverageReader.java
```diff
@@ -664,48 +677,31 @@ private GridCoverage2D getMapReproject(
...
             CoordinateReferenceSystem targetCRS, WMTSReadParameters readParameters)
             throws IOException {
-        try {
-            CoordinateReferenceSystem result = null;
-            String switched = switchDefinition(targetCRS);
-            if (switched != null) {
-                // check first by changing the srs definition
-                result = CRS.decode(switched);
-                LOGGER.fine(() -> "Will request tiles with server supported crs " + switched);
-            } else if (readParameters.getSourceCRS() != null) {
```
aaime, line 674: Removing support for user-provided CRS? 
It has been added on purpose to deal with poorly configured WMTS services, and to allow the WMTS client to decide which default CRS they prefer to use. It has nothing to do with removing a "random supported CRS", this is not random, it's a user decision.

modules/extension/wmts/src/main/java/org/geotools/ows/wmts/map/WMTSCoverageReader.java
```diff
@@ -664,48 +677,31 @@ private GridCoverage2D getMapReproject(
     CoordinateReferenceSystem getBestSourceCRS(
             CoordinateReferenceSystem targetCRS, WMTSReadParameters readParameters)
             throws IOException {
-        try {
-            CoordinateReferenceSystem result = null;
-            String switched = switchDefinition(targetCRS);
```
aaime, line 669: Why also remove support for switched axis match? This is also not random.

modules/extension/wmts/src/main/java/org/geotools/ows/wmts/map/WMTSCoverageReader.java
```diff
@@ -664,48 +677,31 @@ private GridCoverage2D getMapReproject(
...
-                // check first by changing the srs definition
-                result = CRS.decode(switched);
-                LOGGER.fine(() -> "Will request tiles with server supported crs " + switched);
-            } else if (readParameters.getSourceCRS() != null) {
-                // otherwise get the preferred one
-                result = readParameters.getSourceCRS();
-            } else if (!validSRS.isEmpty()) {
-                // last attempt is to retrieve the firs valid srs available
-                LOGGER.fine(() -> "Retrieving random srs to perform reprojection");
-                result = randomSupportedCRS();
```
aaime, line 680: This is the part that should indeed go... as far as I can see, there is already a loop trying to decode all supported CRSs while initializing the srsName, doing it again here does not seem like it would help.
@taba90 can you confirm?

aaime (changes requested):

modules/extension/wmts/src/main/java/org/geotools/ows/wmts/map/WMTSCoverageReader.java
roarbra: Nothing else than getRandomCrs are removed. The display og the change is somewhat misleading because I moved a try-catch. With this change I mean to respect the user's crs setting. That was my problem with the getRandomCrs.

aaime: You are right. Can you address the other feedback?

modules/extension/wmts/src/main/java/org/geotools/ows/wmts/map/WMTSCoverageReader.java
aaime: I disagree: the reason why the code could not be decoded can depend on the set of EPSG plugins available and how they interact with their environment, which may vary by the classpath, network and filesystem conditions (to mention the obvious sources). Removing the stack trace means one would have to exactly reproduce the deployment conditions, leading to a debugging nightmare.

roarbra: In most cases I will agree with you, but for gt-wmts this encode / decode has become somewhat turned around. It isn't the SRS specified by the remote server that are decoded in this function. That decoding is handled by this [function](https://github.com/geotools/geotools/blob/c7b6ac7d1ed6cb68be0497d592d1cbb7c4c3ae88/modules/extension/wmts/src/main/java/org/geotools/ows/wmts/model/TileMatrixSet.java#L66). Instead what is decoded by this function is the result of this [function](https://github.com/geotools/geotools/blob/c7b6ac7d1ed6cb68be0497d592d1cbb7c4c3ae88/modules/extension/wmts/src/main/java/org/geotools/ows/wmts/model/WMTSLayer.java#L105).
Since we're using the CRS.toSRS, I doubt we will have any of those problems you refer to, but sure this is a debugging nightmare.

roarbra:
@aaime I've implemented the logging as you asked for, and @taba90 doesn't seem to have anything objections.

modules/extension/wmts/src/main/java/org/geotools/ows/wmts/map/WMTSCoverageReader.java
taba90: @roarbra @aaime my apologies for the delay on this. 
Yes @aaime it is doing an unecessary loop over the srs as it has been already done before.

aaime (approved):

PR #3972, opened 2022-07-20
[GEOT-7181] GeoJSONReader should handle paged GeoJSON collections

aaime:
[GEOT-7181](https://osgeo-org.atlassian.net/browse/GEOT-7181) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

See ticket.

# Checklist

### DIFF
--- a/modules/unsupported/geojson-core/src/main/java/org/geotools/data/geojson/GeoJSONReader.java
+++ b/modules/unsupported/geojson-core/src/main/java/org/geotools/data/geojson/GeoJSONReader.java
@@ -221,14 +221,18 @@ public class GeoJSONReader implements AutoCloseable {
         }
         ObjectMapper mapper = ObjectMapperFactory.getDefaultMapper();
         List<SimpleFeature> features = new ArrayList<>();
+        URL next = null;
+        Integer matched = null;
         builder = null;
         while (!parser.isClosed()) {
             JsonToken token = parser.nextToken();
             if (token == null) {
                 break;
             }
-            if (JsonToken.FIELD_NAME.equals(token)
-                    && "features".equalsIgnoreCase(parser.currentName())) {
+            if (!JsonToken.FIELD_NAME.equals(token)) continue;
+            //
+            String name = parser.currentName();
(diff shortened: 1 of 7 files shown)


aaime:
Open questions: should this behavior controlled by a flag? Should it be off by default?

modules/unsupported/geojson-core/src/main/java/org/geotools/data/geojson/GeoJSONReader.java
```diff
@@ -240,6 +244,33 @@ public SimpleFeatureCollection getFeatures() throws IOException {
...
+                        JsonNode href = node.get("href");
+                        if (href != null) {
+                            next = new URL(href.textValue());
+                            break;
+                        }
+                    }
+                }
+            }
+            // checking matched attributes from STAC/Features to optimize out size() if possible
+            else if ("context".equals(name) && parser.nextToken() == JsonToken.START_OBJECT) {
```
ianturton, line 266: everywhere else we've used `equalsignorecase` rather than just `equals` - which may be paranoia or justified, but we should probably do that here, or just do a `toLower()` on name when creating it and using `equals` through out.

ianturton (approved):
looks good to me

modules/unsupported/geojson-core/src/main/java/org/geotools/data/geojson/GeoJSONReader.java
aaime: JSON is case sensitive but... why not. I'll make the change.

aaime:
Now tested against a few online services, seems to work. Optimized count when possible, scrolling through a million items in chunks of 10 to get a count, when there is one provided by the protocol, was pretty silly :-D

modules/unsupported/geojson-core/src/main/java/org/geotools/data/geojson/GeoJSONReader.java
aaime: Done, and merging.

PR #3976, opened 2022-07-25
[GEOT-7180] GeoJSONReader ignores feature "id" value.

turingtestfail:
[GEOT-7180](https://osgeo-org.atlassian.net/browse/GEOT-7180) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

cleanup

  

# Checklist


modules/unsupported/geojson-core/src/test/java/org/geotools/data/geojson/GeoJSONReaderTest.java
```diff
@@ -177,7 +177,7 @@ public void testFeatureCollectionParser() throws Exception {
         f = DataUtilities.first(features);
 
         assertNotNull(f);
-        assertEquals("features.0", f.getID());
+        assertEquals("feature.0", f.getID());
```
ianturton, line 180: I don't like that these have changed, is there a reason for the loss of the `s`?

modules/unsupported/geojson-core/src/main/java/org/geotools/data/geojson/GeoJSONReader.java
```diff
@@ -436,6 +451,24 @@ private SimpleFeature getNextFeature(ObjectNode node) throws IOException {
         return feature;
     }
 
+    private String getId(ObjectNode node) {
+        String newId = null;
+        if (node.has(ID)) {
+            if (idPrefix != null) {
+                newId = idPrefix + "." + node.get(ID).asText();
+            } else {
+                newId = node.get(ID).asText();
```
ianturton, line 460: I think this should still have the `basename` prefixing the `ID` attribute value

modules/unsupported/geojson-core/src/main/java/org/geotools/data/geojson/GeoJSONReader.java
```diff
@@ -118,6 +122,18 @@ public GeoJSONReader(URL url) throws IOException {
...
+    /**
+     * Builds a GeoJSON parser from a GeoJSON source, located at the specified URL.
+     *
+     * @param url String representation of the URL
+     * @param idPrefix prefix to use for feature ids
+     * @throws IOException if the URL is not valid
+     */
+    public GeoJSONReader(URL url, String idPrefix) throws IOException {
+        this(url);
+        this.idPrefix = idPrefix;
```
ianturton, line 134: don't you need to call `this(url)` here or initialize the `parser`?

modules/unsupported/geojson-store/src/main/java/org/geotools/data/geojson/store/GeoJSONDataStore.java
```diff
@@ -61,8 +62,8 @@ public GeoJSONDataStore(File f) {
     }
 
     GeoJSONReader read() throws IOException {
-        GeoJSONReader reader = new GeoJSONReader(getUrl());
-
+        GeoJSONReader reader =
+                new GeoJSONReader(getUrl(), FilenameUtils.getBaseName(url.getPath()));
```
ianturton, line 66: Why is this necessary?

modules/unsupported/geojson-core/src/main/java/org/geotools/data/geojson/GeoJSONReader.java
ianturton, line 460: What happens if the `ID` attributes are not unique?

modules/unsupported/geojson-core/src/main/java/org/geotools/data/geojson/GeoJSONReader.java
```diff
@@ -86,6 +86,8 @@ public class GeoJSONReader implements AutoCloseable {
 
     private SimpleFeatureBuilder builder;
 
+    private static final String ID = "id";
```
ianturton, line 89: The user should be able to choose this value, it could be `fid` or `gid` etc

modules/unsupported/geojson-core/src/main/java/org/geotools/data/geojson/GeoJSONReader.java
```diff
@@ -411,8 +427,7 @@ private SimpleFeature getNextFeature(ObjectNode node) throws IOException {
                 }
             }
             builder.set(GEOMETRY_NAME, g);
-
-            String newId = baseName + "." + nextID++;
+            String newId = getId(node);
```
ianturton, line 430: If we are using the `ID` attribute then it should be excluded from the feature that is created.

ianturton (changes requested):

modules/unsupported/geojson-core/src/main/java/org/geotools/data/geojson/GeoJSONReader.java
aaime: I'd argue a low-level GeoJSONReader should be truthful to the GeoJSON content, and be allowed to return exactly the id found in the document. Having an option to use the prefix, for datastore usage, allows for compatibilty.

modules/unsupported/geojson-store/src/main/java/org/geotools/data/geojson/store/GeoJSONDataStore.java
aaime: To make it prefix the ids with the file name, and retain the same behavior.

ianturton: but that is what the `new GeoJSONReader(getUrl());` call did any way - so I don't see any need for this

modules/unsupported/geojson-core/src/main/java/org/geotools/data/geojson/GeoJSONReader.java
aaime: My first reaction would have been "then the GeoJSON is broken". But it turns out the specification even allows the id to be missing and does not talk about uniqueness.

However this can be restricted by specifications that build on top of GeoJSON. For example, the [STAC-item specification](https://github.com/radiantearth/stac-spec/blob/master/item-spec/item-spec.md#item-fields) mandates "The ID should be unique within the [Collection](https://github.com/radiantearth/stac-spec/blob/master/collection-spec/collection-spec.md) that contains the Item."

So it seems we likely need some id generation strategies that a caller can use:

- Existing one, file name + progressive, for random GeoJSON files that use the full liberty of the spec.
- Using exactly the id found in the features, for contexts where uniqueness is guaranteed
- Also keep the file name + id, to make the ids "more globally unique" (needed for the STAC datastore, to make it behave like other GeoServer stores).

ianturton: OK, that makes sense

modules/unsupported/geojson-store/src/main/java/org/geotools/data/geojson/store/GeoJSONDataStore.java
aaime: Ah, did not notice, sorry.

turingtestfail: Yeah, I added the second argument as a means to seperate the featurebuilder name from the prefix.  If those should always be the same I can remove the extra constructor argument.

modules/unsupported/geojson-core/src/main/java/org/geotools/data/geojson/GeoJSONReader.java
turingtestfail: The standard is pretty specific about this:
If a Feature has a commonly used identifier, that identifier
      SHOULD be included as a member of the Feature object with the name
      "id", and the value of this member is either a JSON string or
      number.

modules/unsupported/geojson-core/src/test/java/org/geotools/data/geojson/GeoJSONReaderTest.java
turingtestfail: Originally changed to match GeoJSON in the tests.  I have gone ahead and added the s to JSON and reverted the asserts.

modules/unsupported/geojson-core/src/main/java/org/geotools/data/geojson/GeoJSONReader.java
turingtestfail: See comment below about the featurebuilder name and the prefix.  If a seperate prefix is not needed, this new constructor will also go away.

modules/unsupported/geojson-core/src/main/java/org/geotools/data/geojson/GeoJSONReader.java
turingtestfail: Yup, ambiguous method name here.  Changed to reflect that it either gets the ID or generates one if not found.

modules/unsupported/geojson-core/src/main/java/org/geotools/data/geojson/GeoJSONReader.java
aaime: Yes, but SHOULD is not mandatory, see the RFC titled "[Key words for use in RFCs to Indicate Requirement Levels](https://datatracker.ietf.org/doc/html/rfc2119)", quoting:

```
SHOULD   This word, or the adjective "RECOMMENDED", mean that there
   may exist valid reasons in particular circumstances to ignore a
   particular item, but the full implications must be understood and
   carefully weighed before choosing a different course.
```

So strong recommendation, but not an absolute requirement. Also, the wording talks about identifier, but does not say anything about its uniqueness and in which context (e.g., in the STAC spec the id must be unique only inside that collection). To make another example, I identify as Andrea, but it's too common, Andrea Aime is better, it's (accidentaly) unique in the province where I live, but if I want uniqueness at the country level, I have to switch to the SSN.

modules/unsupported/geojson-store/src/main/java/org/geotools/data/geojson/store/GeoJSONDataStore.java
ianturton: The idea of a 2nd argument is fine, it just isn't needed here. as if you don't supply the 2nd argument the other constructor will make a baseName call.

modules/unsupported/geojson-core/src/main/java/org/geotools/data/geojson/GeoJSONReader.java
ianturton: I was saying that if we use the `id` attribute as the feature ID then we should not return it as an attribute too

modules/unsupported/geojson-store/src/main/java/org/geotools/data/geojson/store/GeoJSONDataStore.java
turingtestfail: Ahh, gotcha.  It is now needed because without the second argument, if there is an existing id field, no prefix is added. In cases where there is no existing id field, it behaves as before and defaults to adding the derived prefix.

modules/unsupported/geojson-core/src/main/java/org/geotools/data/geojson/GeoJSONReader.java
turingtestfail: Gotcha.  Updated to add constructor that allows specification of an id field name.

modules/unsupported/geojson-core/src/main/java/org/geotools/data/geojson/GeoJSONReader.java
turingtestfail: Yup, my mistake.  Fixed to exclude id field from general attribute map.

modules/unsupported/geojson-store/src/main/java/org/geotools/data/geojson/store/GeoJSONDataStore.java
ianturton: As I said, I understand why you have provided a constructor with 2 arguments - my question is why in this exact line is it necessary to pass in the basename as the 2nd argument when the single argument constructor will do exactly the same thing?

aaime:
@turingtestfail there is a small conflict.

modules/unsupported/geojson-store/src/main/java/org/geotools/data/geojson/store/GeoJSONDataStore.java
turingtestfail: Now changed so the default setting mean that this is no longer needed, as requested.

modules/unsupported/geojson-core/src/main/java/org/geotools/data/geojson/GeoJSONReader.java
turingtestfail: Implemented, as described.

aaime:
Seems reasonable now, @ianturton what do you think?

ianturton (approved):

PR #3981, opened 2022-07-29
Improve usablity of STAC store for GetMap usage

aaime:
See tickets:
* https://osgeo-org.atlassian.net/browse/GEOT-7190
* https://osgeo-org.atlassian.net/browse/GEOT-7191

I think the first one could be of general interest, also for an eventual GeoJSONDataStore improvement, if the GeoJSON were to be placed online, or as a pattern to reuse on OGC clients fetching lots of data.

# Checklist

### DIFF
--- a/modules/unsupported/stac-store/src/main/java/org/geotools/stac/store/STACDataStoreFactory.java
+++ b/modules/unsupported/stac-store/src/main/java/org/geotools/stac/store/STACDataStoreFactory.java
@@ -69,7 +69,16 @@ public class STACDataStoreFactory implements DataStoreFactorySpi {
                     Integer.class,
                     "How many items to fetch in a single request (just a suggestion to the server)",
                     false,
-                    1000,
+                    STACDataStore.DEFAULT_FETCH_SIZE,
+                    Collections.singletonMap(Parameter.MIN, 1));
+
+    public static final Param HARD_LIMIT =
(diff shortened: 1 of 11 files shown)


modules/unsupported/stac-store/src/main/java/org/geotools/stac/store/STACDataStoreFactory.java
```diff
@@ -72,6 +72,15 @@ public class STACDataStoreFactory implements DataStoreFactorySpi {
                     1000,
                     Collections.singletonMap(Parameter.MIN, 1));
 
+    public static final Param HARD_LIMIT =
+            new Param(
+                    "hardLimit",
+                    Integer.class,
+                    "How many items to fetch from a collection, tops, even while paging. Set to zero or negative to disable.",
+                    false,
+                    100000,
```
ianturton, line 81: can't we refer to the static variable here?

aaime: You mean STACDataStore.DEFAULT_HARD_LIMIT? Oh yeah, it's actually possible, good point.

aaime: Fixed

PR #3998, opened 2022-08-29
[GEOT-7202] Vector Mosaicing Data Store

turingtestfail:
[GEOT-7202](https://osgeo-org.atlassian.net/browse/GEOT-7202) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

started on store, factory and unit tests

missing copyright

factory clean up and tests

mosaic store scan all delegate types for matches

finished store with test and started on source

added in preferred spi param

started on featuresource

able to get granules from shapefiles

more granule grabbing

started on buildfeaturetype

count and bounds

feature reader

some cleanup

pmd cleanup

  

# Checklist


modules/unsupported/vector-mosaic/src/main/java/org/geotools/vectormosaic/VectorMosaicStoreFactory.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    @Override
+    public String getDescription() {
+        return DESCRIPTION;
+    }
+
+    @Override
+    public Param[] getParametersInfo() {
+        return new Param[] {REPOSITORY_PARAM, NAMESPACE, DELEGATE_STORE_NAME};
```
aaime, line 94: Does not contain CONNECTION_PARAMETER_KEY, I believe as a result it won't show up in the GeoServer GUI. Did you do some interactive tests in GeoServer?

modules/unsupported/vector-mosaic/src/main/java/org/geotools/vectormosaic/VectorMosaicStore.java
```diff
@@ -0,0 +1,484 @@
...
+    @SuppressWarnings("PMD.UnusedPrivateField") // field is used in constructor
+    private final Repository repository;
+
+    DataStore delegateDataStore;
+    boolean tolerant;
+
+    public static final String PREFERRED_SPI_DEFAULT =
+            "org.geotools.data.shapefile.ShapefileDataStoreFactory";
+
+    String preferredSPI = PREFERRED_SPI_DEFAULT;
```
aaime, line 68: IntelliJ calls out a "redundant initializer", because the sole constructor will always override this value. Meaning the Shapefile is not really used.
Make it the sample value of the factory parameter instead.

modules/unsupported/vector-mosaic/src/main/java/org/geotools/vectormosaic/VectorMosaicStore.java
```diff
@@ -0,0 +1,484 @@
...
+            Repository repository,
+            String preferredDataStoreSPI,
+            String connectionParameterKey) {
+        this.repository = repository;
+
+        this.preferredSPI = preferredDataStoreSPI;
+
+        if (connectionParameterKey != null) {
+            this.connectionParameterKey = connectionParameterKey;
+        }
```
aaime, line 125: Please use a constructor argument for parameters that are mandatory for the class to work, and a bean property for those that have defaults and might be omitted.

modules/unsupported/vector-mosaic/src/main/java/org/geotools/vectormosaic/VectorMosaicStore.java
```diff
@@ -0,0 +1,484 @@
...
+import org.opengis.feature.type.Name;
+
+/** {@link DataStore} for a vector mosaic. */
+public class VectorMosaicStore extends ContentDataStore {
+    static final Logger LOGGER = Logging.getLogger(VectorMosaicStore.class);
+
+    @SuppressWarnings("PMD.UnusedPrivateField") // field is used in constructor
+    private final Repository repository;
+
+    DataStore delegateDataStore;
```
aaime, line 62: No, we discussed this, the delegate store must not be stored, it should always be retrieved from the repository (it's a cheap operation). Rationale: you don't manage the lifecycle of the store, the repository (e.g. GeoServer does), and may close the store and create a new one for you to use.

modules/unsupported/vector-mosaic/src/main/java/org/geotools/vectormosaic/VectorMosaicStore.java
```diff
@@ -0,0 +1,484 @@
...
+
+        this.preferredSPI = preferredDataStoreSPI;
+
+        if (connectionParameterKey != null) {
+            this.connectionParameterKey = connectionParameterKey;
+        }
+        Name name = buildName(delegateStoreName);
+        delegateDataStore = repository.dataStore(name);
+        try {
+            delegateTypes =
```
aaime, line 129: Same as with the store, you should not store the list of delegate type names, because they might be changing over time, you risk not seeing new tables made available after object construction.
ContentDataStore already caches feature types, don't go beyond that, when asked for the list of all type names, compute it on the fly.

If we find it's too slow, we can add a cache later down the road, but it would have to be checked with the list of type names in the delegate store, and the cache be rebuilt if the list of delegate type names changdd (so, it should be done in the getTypeNames anyways).

modules/unsupported/vector-mosaic/src/main/java/org/geotools/vectormosaic/VectorMosaicStore.java
```diff
@@ -0,0 +1,484 @@
...
+                    .equals(REQUIRED_VECTOR_MOSAIC_FIELDS.get(field))) {
+                LOGGER.log(
+                        Level.INFO,
+                        "Delegate store "
+                                + delegateStoreName
+                                + " schema with type name "
+                                + typeName
+                                + " has the wrong field type for field "
+                                + field);
+                return false;
```
aaime, line 239: Iterate on the map entries instead, put the descriptor in a local variable, to reduce the verboseness of this method.

modules/unsupported/vector-mosaic/src/main/java/org/geotools/vectormosaic/VectorMosaicStore.java
```diff
@@ -0,0 +1,484 @@
...
+        } else {
+            String ns = name.substring(0, idx);
+            String local = name.substring(idx + 1);
+            return new NameImpl(ns, local);
+        }
+    }
+
+    @Override
+    protected List<Name> createTypeNames() throws IOException {
+        List<Name> result = new ArrayList<>();
```
aaime, line 265: Here it should grab the delegate store and compute the type names list.

modules/unsupported/vector-mosaic/src/main/java/org/geotools/vectormosaic/VectorMosaicStore.java
```diff
@@ -0,0 +1,484 @@
...
+            String connectionParameterKey) {
+        this.repository = repository;
+
+        this.preferredSPI = preferredDataStoreSPI;
+
+        if (connectionParameterKey != null) {
+            this.connectionParameterKey = connectionParameterKey;
+        }
+        Name name = buildName(delegateStoreName);
+        delegateDataStore = repository.dataStore(name);
```
aaime, line 127: I think the repository would have worked also with simple strings, did you try it out in GeoServer?

modules/unsupported/vector-mosaic/src/main/java/org/geotools/vectormosaic/VectorMosaicStore.java
```diff
@@ -0,0 +1,484 @@
...
+            DataStoreFactorySpi spi = it.next();
+            if (spi.getClass().getTypeName().equals(preferredSPI)) {
+                dataStoreFactorySpi = spi;
+                break;
+            }
+        }
+        return dataStoreFactorySpi;
+    }
+
+    public boolean isTolerant() {
```
aaime, line 477: Delaying the calculation of feature types should also remove the need for this property.

modules/unsupported/vector-mosaic/src/main/java/org/geotools/vectormosaic/VectorMosaicStore.java
```diff
@@ -0,0 +1,484 @@
...
+            "PMD.UseTryWithResources") // PMD complains about FeatureIterator, but it is not ready
+    // to be instantiated yet
+    private void addTypeName(String delegateTypeName) {
+        SimpleFeatureIterator iterator = null;
+        try {
+            SimpleFeatureSource featureSource =
+                    delegateDataStore.getFeatureSource(delegateTypeName);
+            SimpleFeatureCollection features = featureSource.getFeatures();
+            iterator = features.features();
+            while (iterator.hasNext()) {
```
aaime, line 176: Oh no, keeping the granules in memory? This is a major issue:
- Invites OOMs, as the granules might be many
- Makes the system nonresponsive to granules being removed or added from the delegate store

Each time the list of granules is needed, it should be queried, taking into account the filter that's coming from the client (so pre-filter based on bbox and other properties contained in the index table, and then apply a post filter once the features in the target store are loaded).

modules/unsupported/vector-mosaic/src/main/java/org/geotools/vectormosaic/VectorMosaicFeatureSource.java
```diff
@@ -0,0 +1,340 @@
...
+     * @return The FeatureStore for the feature source.
+     */
+    public VectorMosaicStore getStore() {
+        return (VectorMosaicStore) getEntry().getDataStore();
+    }
+
+    @Override
+    protected ReferencedEnvelope getBoundsInternal(Query query) throws IOException {
+        List<ReferencedEnvelope> allBounds = new ArrayList<>();
+        for (VectorMosaicGranule granule : config.getGranules()) {
```
aaime, line 80: Code in this class is mostly broken because of this pattern: it should split the filter based on the attributes that are available in the index, query the index only hitting the granules that are matching, and only after go and open the potential candidate granules...

modules/unsupported/vector-mosaic/src/main/java/org/geotools/vectormosaic/VectorMosaicFeatureSource.java
```diff
@@ -0,0 +1,340 @@
...
+            return null;
+        } else {
+            return result;
+        }
+    }
+
+    @Override
+    protected int getCountInternal(Query query) throws IOException {
+        List<Long> counts = new ArrayList<>();
+        for (VectorMosaicGranule granule : config.getGranules()) {
```
aaime, line 125: Same as above...

modules/unsupported/vector-mosaic/src/main/java/org/geotools/vectormosaic/VectorMosaicFeatureSource.java
```diff
@@ -0,0 +1,340 @@
...
+            // get the first store
+            VectorMosaicGranule firstGranule = config.getGranules().get(0);
+            Query psQuery = new Query(query);
+            psQuery.setTypeName(firstGranule.getGranuleTypeName());
+            DataStore ps = firstGranule.getDataStore();
+            SimpleFeatureType target = DataUtilities.createView(ps, psQuery).getSchema();
+            target = retypeNameSchema(target);
+
+            FeatureQueue queue = new FeatureQueue();
+            for (VectorMosaicGranule vectorMosaicGranule : config.getGranules()) {
```
aaime, line 169: Same as above. We'll have thousands of granules, we cannot go and query each one of them...

modules/unsupported/vector-mosaic/src/main/java/org/geotools/vectormosaic/VectorMosaicFeatureSource.java
```diff
@@ -0,0 +1,340 @@
...
+            q.setSortBy(null);
+            Filter originalFilter = q.getFilter();
+            if (originalFilter != null && !Filter.INCLUDE.equals(originalFilter)) {
+                // eliminate the extra attribute the delegate source does not know about
+                FilterAttributeExtractor extractor = new FilterAttributeExtractor();
+                originalFilter.accept(extractor, null);
+                Set<String> filterNames = extractor.getAttributeNameSet();
+                Set<String> sourceNames = getSourceAttributes(source);
+                if (!sourceNames.containsAll(filterNames)) {
+                    MissingPropertiesEraser eraser = new MissingPropertiesEraser(sourceNames);
```
aaime, line 208: Nope... shoudl have been implemented as a filter splitter subclass...

aaime (changes requested):
Nope, there are serious issues here...

modules/unsupported/vector-mosaic/src/main/java/org/geotools/vectormosaic/VectorMosaicFeatureSource.java
```diff
@@ -0,0 +1,340 @@
...
+            }
+            // get the first store
+            VectorMosaicGranule firstGranule = config.getGranules().get(0);
+            Query psQuery = new Query(query);
+            psQuery.setTypeName(firstGranule.getGranuleTypeName());
+            DataStore ps = firstGranule.getDataStore();
+            SimpleFeatureType target = DataUtilities.createView(ps, psQuery).getSchema();
+            target = retypeNameSchema(target);
+
+            FeatureQueue queue = new FeatureQueue();
```
aaime, line 168: Nope, don't accumulate features in memory, it's gonna cause a OOM (this store is for large datasets).
Instead, create a reader that has two delegate readers (or feature iterators):
* One over the index table, pointing at the current granule
* One over the  current granule
*  When the current granule iterator is done, close, skip to the next granule, open a new iterator, and move on.
This keeps at most one feature in memory. Similar (but not same) to this one:
https://github.com/geoserver/geoserver/blob/9e40cbbccb9626b8d08516c8eacaf5a8f168189b/src/extension/csw/api/src/main/java/org/geoserver/csw/feature/CompositeIterator.java

dromagnoli:
Hi @turingtestfail . Please, let me know when you are done with the changes so I can take a look at the code and do some testing.
Did you manage to do that test mentioned by @aaime about having a big collection and performing the query on a reduced country to demonstrate/verify that the request is fast being restricted to that reduced part?

modules/unsupported/vector-mosaic/src/main/java/org/geotools/vectormosaic/VectorMosaicStoreFactory.java
turingtestfail: Fixed

modules/unsupported/vector-mosaic/src/main/java/org/geotools/vectormosaic/VectorMosaicStore.java
turingtestfail: Fixed

modules/unsupported/vector-mosaic/src/main/java/org/geotools/vectormosaic/VectorMosaicStore.java
turingtestfail: Fixed

modules/unsupported/vector-mosaic/src/main/java/org/geotools/vectormosaic/VectorMosaicStore.java
turingtestfail: Fixed

modules/unsupported/vector-mosaic/src/main/java/org/geotools/vectormosaic/VectorMosaicStore.java
turingtestfail: Fixed

modules/unsupported/vector-mosaic/src/main/java/org/geotools/vectormosaic/VectorMosaicStore.java
turingtestfail: Fixed

modules/unsupported/vector-mosaic/src/main/java/org/geotools/vectormosaic/VectorMosaicStore.java
turingtestfail: Fixed

modules/unsupported/vector-mosaic/src/main/java/org/geotools/vectormosaic/VectorMosaicStore.java
turingtestfail: Fixed

modules/unsupported/vector-mosaic/src/main/java/org/geotools/vectormosaic/VectorMosaicStore.java
turingtestfail: Fixed

modules/unsupported/vector-mosaic/src/main/java/org/geotools/vectormosaic/VectorMosaicFeatureSource.java
turingtestfail: Fixed

modules/unsupported/vector-mosaic/src/main/java/org/geotools/vectormosaic/VectorMosaicFeatureSource.java
turingtestfail: Fixed

modules/unsupported/vector-mosaic/src/main/java/org/geotools/vectormosaic/VectorMosaicFeatureSource.java
turingtestfail: Fixed

modules/unsupported/vector-mosaic/src/main/java/org/geotools/vectormosaic/VectorMosaicFeatureSource.java
turingtestfail: Fixed

modules/unsupported/vector-mosaic/src/main/java/org/geotools/vectormosaic/VectorMosaicFeatureSource.java
turingtestfail: Fixed

modules/unsupported/vector-mosaic/src/main/java/org/geotools/vectormosaic/VectorMosaicGranule.java
```diff
@@ -0,0 +1,88 @@
...
+    public static final String CONNECTION_PARAMETERS_DELEGATE_FIELD_DEFAULT = "params";
+    public static final String GRANULE_TYPE_NAME = "type";
+    public static final String GRANULE_ID_FIELD = "id";
+
+    /** The feature type name */
+    String name;
+
+    String granuleTypeName;
+    String params;
+    DataStore dataStore;
```
aaime, line 36: Since it's dealing with a resource that needs to be closed, it should be a Closable, and should close the eventual datastore on close.

modules/unsupported/vector-mosaic/src/main/java/org/geotools/vectormosaic/VectorMosaicFeatureReader.java
```diff
@@ -0,0 +1,218 @@
...
+            if (fi.hasNext()) {
+                firstDelegateFeature = fi.next();
+            }
+            if (firstDelegateFeature == null) {
+                throw new IllegalArgumentException("No delegate features found");
+            }
+            VectorMosaicGranule granule =
+                    VectorMosaicGranule.fromDelegateFeature(firstDelegateFeature);
+            source.initGranule(granule, true);
+            return granule.getDataStore().getSchema(granule.getGranuleTypeName());
```
aaime, line 87: Here a datastore is created but not disposed of: making it closeable should make sure that you don't lose bits along the road.

modules/unsupported/vector-mosaic/src/main/java/org/geotools/vectormosaic/VectorMosaicFeatureReader.java
```diff
@@ -0,0 +1,218 @@
...
+                                    delegateFeature.getAttribute(attributeDescriptor.getName()));
+                        }); // then add all delegate attributes except mandatory index types
+
+        return SimpleFeatureBuilder.build(
+                mergedType, attributes, peekGranule.getIdentifier().getID());
+    }
+
+    @Override
+    public void close() throws IOException {
+        if (delegateIterator != null) {
```
aaime, line 211: If the iterator is closed before reaching the end, I believe the delegate store won't be closed.

aaime (changes requested):
Tested it with a large mosaic, it seems to work fine, but found places where stores can be leaked. Reported them, could you make sure that for every chance of creating a granule store, there is a guarantee that same store gets closed?

modules/unsupported/vector-mosaic/src/main/java/org/geotools/vectormosaic/VectorMosaicGranule.java
turingtestfail: Fixed

modules/unsupported/vector-mosaic/src/main/java/org/geotools/vectormosaic/VectorMosaicFeatureReader.java
turingtestfail: Fixed

modules/unsupported/vector-mosaic/src/main/java/org/geotools/vectormosaic/VectorMosaicFeatureReader.java
turingtestfail: Fixed

PR #4015, opened 2022-09-13
[GEOT-7220] Support datetime2 and datetimeoffset in JDBC SQL Server

bjornharrtell:
[GEOT-7220](https://osgeo-org.atlassian.net/browse/GEOT-7220) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

### DIFF
--- a/modules/plugin/jdbc/jdbc-sqlserver/src/main/java/org/geotools/data/sqlserver/SQLServerDialect.java
+++ b/modules/plugin/jdbc/jdbc-sqlserver/src/main/java/org/geotools/data/sqlserver/SQLServerDialect.java
@@ -24,7 +24,9 @@ import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
 import java.sql.Time;
+import java.sql.Timestamp;
 import java.sql.Types;
+import java.time.OffsetDateTime;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -165,6 +167,9 @@ public class SQLServerDialect extends BasicSQLDialect {
         mappings.put("uniqueidentifier", UUID.class);
         mappings.put("time", Time.class);
         mappings.put("date", Date.class);
+        mappings.put("datetime", Timestamp.class);
+        mappings.put("datetime2", Timestamp.class);
+        mappings.put("datetimeoffset", OffsetDateTime.class);
     }
 
     @Override
--- a/modules/plugin/jdbc/jdbc-sqlserver/src/test/java/org/geotools/data/sqlserver/SQLServerDateOnlineTest.java
+++ b/modules/plugin/jdbc/jdbc-sqlserver/src/test/java/org/geotools/data/sqlserver/SQLServerDateOnlineTest.java
@@ -16,13 +16,80 @@
  */
 package org.geotools.data.sqlserver;
 
+import static org.junit.Assert.assertEquals;
+
+import java.io.IOException;
+import java.sql.Date;
+import java.sql.Time;
+import java.sql.Timestamp;
+import java.util.List;
+import microsoft.sql.DateTimeOffset;
+import org.geotools.data.DataUtilities;
+import org.geotools.data.FeatureStore;
+import org.geotools.data.simple.SimpleFeatureIterator;
+import org.geotools.data.store.ContentFeatureCollection;
 import org.geotools.jdbc.JDBCDateOnlineTest;
 import org.geotools.jdbc.JDBCDateTestSetup;
+import org.junit.Test;
+import org.opengis.feature.simple.SimpleFeature;
+import org.opengis.feature.simple.SimpleFeatureType;
+import org.opengis.filter.identity.FeatureId;
 
 public class SQLServerDateOnlineTest extends JDBCDateOnlineTest {
 
     @Override
     protected JDBCDateTestSetup createTestSetup() {
         return new SQLServerDateTestSetup();
+    }
+
+    @Test
+    public void testInsert() throws Exception {
+        dataStore.setExposePrimaryKeyColumns(true);
+        SimpleFeatureType ft = dataStore.getSchema(tname("dates"));
+        SimpleFeature datesFeat =
+                DataUtilities.createFeature(
+                        ft,
+                        "id=4|2009-09-29|2009-09-29T17:54:23|2009-09-29T17:54:23|2009-09-29T17:54:23+02:00|17:54:23");
+        FeatureStore<SimpleFeatureType, SimpleFeature> fs =
+                (FeatureStore<SimpleFeatureType, SimpleFeature>)
+                        (dataStore.getFeatureSource("dates"));
+        List<FeatureId> featIds = fs.addFeatures(DataUtilities.collection(datesFeat));
+        assertEquals(1, featIds.size());
+    }
+
+    @Test
+    public void testQuery() throws IOException {
+        ContentFeatureCollection fc = dataStore.getFeatureSource("dates").getFeatures();
+        try (SimpleFeatureIterator it = fc.features()) {
+            SimpleFeature f1 = it.next();
+            Object a1 = f1.getAttribute(0);
+            assertEquals(Date.class, a1.getClass());
+            assertEquals("2009-06-28", ((Date) a1).toString());
+            Object a2 = f1.getAttribute(1);
+            assertEquals(Timestamp.class, a2.getClass());
+            assertEquals("2009-06-28 15:12:41.0", ((Timestamp) a2).toString());
+            Object a3 = f1.getAttribute(2);
+            assertEquals(Timestamp.class, a3.getClass());
+            assertEquals("2009-06-28 15:12:41.0", ((Timestamp) a3).toString());
+            Object a4 = f1.getAttribute(3);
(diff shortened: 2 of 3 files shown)


bjornharrtell:
I've tried this locally against an SQL Server datasource with GeoServer 2.21.0 and it appears to be working.

modules/plugin/jdbc/jdbc-sqlserver/src/main/java/org/geotools/data/sqlserver/SQLServerDialect.java
```diff
@@ -165,6 +166,9 @@ public void registerSqlTypeNameToClassMappings(Map<String, Class<?>> mappings) {
         mappings.put("uniqueidentifier", UUID.class);
         mappings.put("time", Time.class);
         mappings.put("date", Date.class);
+        mappings.put("datetime", Timestamp.class);
```
mprins, line 169: We have an existing testcase that checks a `datetime` column in SQLServerDateOnlineTest ⬅️ JDBCDateOnlineTest

https://github.com/geotools/geotools/blob/01e27991725d1394b683439a774c4827493d6f95/modules/library/jdbc/src/test/java/org/geotools/jdbc/JDBCDateOnlineTest.java#L24-L29bc/src/test/java/org/geotools/jdbc/JDBCDateOnlineTest.java#L24-L29

but no tests for datetime2 or datetimeoffset

mprins (commented):
Wondering if this also works for writing; there is no testcase

aaime:
Yup, no testcase, no merge.

bjornharrtell:
Found a bug in this patch. It doesn't handle null values.

bjornharrtell:
Hmm looking closer I don't see how null values would work for other types too which seems very strange... attempting to get proper tests done.

bjornharrtell:
Tests added/expanded and also verifies null values are handled properly debunking suspect case.

bjornharrtell:
@mprins and @aaime I've added test cases now. Please re-review.

modules/plugin/jdbc/jdbc-sqlserver/src/main/java/org/geotools/data/sqlserver/SQLServerDialect.java
```diff
@@ -165,6 +166,9 @@ public void registerSqlTypeNameToClassMappings(Map<String, Class<?>> mappings) {
...
+        mappings.put("datetime2", Timestamp.class);
+        mappings.put("datetimeoffset", Timestamp.class);
```
mprins, line 171: This conversion looses the timezone offset, not sure if that is what you expect to happen

mprins (commented):
I don't think the timezone offset part of DATETIMEOFFSET is handled or tested, all test data seems to be using the local timezone and not have an offset like `12-10-25 12:32:10 +02:00`

I'm fine with adding these types (datetime2 is actually the replacement of deprecated datetime) but some unexpected results can be expected for offset DATETIMEOFFSET I think.

modules/plugin/jdbc/jdbc-sqlserver/src/main/java/org/geotools/data/sqlserver/SQLServerDialect.java
bjornharrtell: Thanks, didn't consider that. I will look into it.

bjornharrtell: I've switched so it maps to driver specific type `microsoft.sql.DateTimeOffset` to ensure proper serialization. I've also added more test details. Does it seem sensible to you now @mprins?

aaime: @mprins will follow up too, but sharing my 2 cents... DateTimeOffset ensure no semantic is lost, but at the same time, it's not part of the "java.util.Date" inheritance hierarchy, so the rest of GeoTools (and GeoServer output formats) might not know what to do with it, it might toString it if you're lucky, or throw an error due to unrecognized type.

Ideally it could be mapped to one of Java 8 date primitives, but we'd be basically back to square one again, as the rest of GeoTools also does not know how to use them.

I see @ianturton added a JTDSDateConverterFactory to handle the JTDS driver dates, @ianturton would a dedicated converter form DateTimeOffset help in this case, and how is the wiring done?

bjornharrtell: Well.. at least I can confirm it works well as WFS datastore as used in GeoServer. I carry this patch as hotfix for several GeoServer deployments. The behaviour without this patch is that the columns are silently omitted in GeoServer which IMHO is worse than crashing.

aaime: With ``java.sql.Timestamp`` it's going to work but you might lose information (which could also be handled via documentation). 
Did you test it with ``DateTimeOffset`` instead? I assume the patch you have been carrying around for several deployments was the original one you submitted, using Timestamp only... or not?

bjornharrtell: The initial patch worked but lost timezone. The latest patch as is in this PR also works and retains timezone in WFS output.

bjornharrtell: Noticed some bad copy paste for the testInsert method which I've now fixed.

bjornharrtell: Found that there is an official `java.time.OffsetDateTime` that seems to work as well as the microsoft specific type. Switching to it.

bjornharrtell: `java.time.OffsetDateTime` does not inherit from `java.util.Date` so it won't help for the potential issues mentioned by @aaime but I would think it's the more appropriate and portable choice here.

mprins: I agree with @aaime; using a type (`java.time.OffsetDateTime`) that does not cast to `java.util.Date` will probably cause more grief than getting results that are a timezone worth of offset off; I seem to recall that can be fixed using a connection parameter, not sure about that.

As I said before I would be fine adding `datetimeoffset` as a `Timestamp.class` and live with the consequences. 

FWIW postgis uses `java.sql.Timestamp` for timezoned `timestamptz` data as well
https://github.com/geotools/geotools/blob/4b968cfe0559d131c24cffa841122442fe51c466/modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/PostGISDialect.java#L1109

ps. the Oracle driver does some fancy conversion of oracle time stamps with a converter too, as the oracle things cannot be cast to java.util classes

mprins: if `java.time.OffsetDateTime` works for your WFS's I'm happy with that

bjornharrtell: I'm not happy to revert this and lose the timezone again, because it's actually providing the correct behavior at the end user here now.

Does this mean I have to provide "fancy conversion" somehow in this PR to get it accepted? Or even possibly a complete migration to java.time.* (but I suppose that is a pretty large undertaking...)?

bjornharrtell: Verified working end to end GeoServer WFS also after switching to `java.time.OffsetDateTime`.

bjornharrtell: @aaime considering the above is this acceptable as is to you?

aaime:
Sure go with it

Il Mar 1 Nov 2022, 16:10 Björn Harrtell ***@***.***> ha
scritto:

bjornharrtell:
@aaime I lack merge permissions :D

mprins:
Does this PR need to be backported?

bjornharrtell:
@mprins personally I have no need for backport (if I'm correct in assuming it will be part of GeoTools 29 and GeoServer 2.22)

mprins:
GeoTools 29: yes
GeoServer 2.22: no

(GeoTools 29 will be in GeoServer 2.23, GeoServer 2.22 relies on GeoTools 28)

bjornharrtell:
Ah I see GeoServer 2.22.x is already branched away from master...

bjornharrtell:
Assuming backport PR is welcome, done at https://github.com/geotools/geotools/pull/4095.

PR #4023, opened 2022-09-23
[GEOT-7225] Assertion in ShpFilesTestStream depends on environment

roarbra:
[GEOT-7225](https://osgeo-org.atlassian.net/browse/GEOT-7225) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

The fails on JDK 11, because of the assertion that isLocal will return false. For JDK 8 and earlier that was true, because the url would be jar:file... That has changed and the url will now be file:/.......jar!shapes/statepop.shp.

I don't see the need for that check, so I'm removing it.

I'm also renaming the file, so that the suffix of the class will be Test.

# Checklist

### DIFF
--- a/modules/plugin/shapefile/src/test/java/org/geotools/data/shapefile/ShpFilesStreamTest.java
+++ b/modules/plugin/shapefile/src/test/java/org/geotools/data/shapefile/ShpFilesStreamTest.java
@@ -44,7 +44,7 @@ import org.geotools.data.shapefile.files.StorageFile;
 import org.junit.Before;
 import org.junit.Test;
 
-public class ShpFilesTestStream implements org.geotools.data.shapefile.files.FileWriter {
+public class ShpFilesStreamTest implements org.geotools.data.shapefile.files.FileWriter {
 
     private String typeName;
     private Map<ShpFileType, File> map;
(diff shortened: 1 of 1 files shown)


aaime:
I don't think it's JDK related, but more IDE related: maven will run the tests against the jars, while IDEs often do that against unpacked resources folders. The test is failing if I run it with JDK 8 from the IDE too.

roarbra:
Yes, it might be the IDE as well.
I've installed it with a fresh JDK and Eclipse, so there could be one or the other.

modules/plugin/shapefile/src/test/java/org/geotools/data/shapefile/ShpFilesStreamTest.java
```diff
@@ -182,13 +182,10 @@ public void testGetReadChannelFileChannel() throws IOException {
     }
 
     @Test
-    @SuppressWarnings("PMD.UnusedVariable") // really want to just open and close
```
aaime, line 185: Oh interesting, PMD is no longer complaining that the "read" variable is not used? Nice find. Did you just try or have a way to find unnecessary PMD annotations?

aaime (commented):
I'm good with removing the assert. Interested about the annotation removal, see below.

modules/plugin/shapefile/src/test/java/org/geotools/data/shapefile/ShpFilesStreamTest.java
roarbra: Just trying. Seemed wrong to say the variable was unused.

PR #4029, opened 2022-10-02
GEOT-7231: Z values shapefiles don't handle NaN M values correctly

ianturton:
[GEOT-7231](https://osgeo-org.atlassian.net/browse/GEOT-7231) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

  Page 2 of the [Shapefile spec](https://support.esri.com/en/white-paper/279) says:

>  Floating point numbers must be numeric values. Positive infinity, negative infinity, and Not-a-Number (NaN) values are not allowed in shapefiles. Nevertheless, shapefiles support the concept of "no data" values, but they are currently used only for measures.  Any floating point number smaller than –10^38 is considered by a shapefile reader to represent a "no data" value.

Combined with the note on page 14:

> Shapes of this type have an additional coordinateM. Note that "no data" value can be specified as a value for M (see Numeric Types on page 2).

Means that we need to check `M` values before adding them to the `CoordinateSequence` and take account of `NaN` in the `M` coordinate when checking for dimensions.

# Checklist

### DIFF
--- a/modules/library/main/src/main/java/org/geotools/geometry/jts/coordinatesequence/CoordinateSequences.java
+++ b/modules/library/main/src/main/java/org/geotools/geometry/jts/coordinatesequence/CoordinateSequences.java
@@ -173,7 +173,7 @@ public class CoordinateSequences extends org.locationtech.jts.geom.CoordinateSeq
         if (seq == null) return 3;
 
         int dim = seq.getDimension();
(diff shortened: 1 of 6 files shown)


ianturton:
https://github.com/geoserver/geoserver/pull/6209

PR #4038, opened 2022-10-03
[GEOT-7234] Implement `#getOptimizedBounds(...)` in SQLServerDialect

mprins:
[GEOT-7234](https://osgeo-org.atlassian.net/browse/GEOT-7234) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

  

# Checklist

- [x] ~~New unit tests have been added covering the changes.~~ covered by existing tests that call `JDBCDatastore#getBounds()` https://github.com/geotools/geotools/blob/e56300f8fd5dbc964a19f3aca19170a4e27d90ce/modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java#L1258-L1273 

### DIFF
--- a/modules/plugin/jdbc/jdbc-sqlserver/src/main/java/org/geotools/data/sqlserver/SQLServerDialect.java
+++ b/modules/plugin/jdbc/jdbc-sqlserver/src/main/java/org/geotools/data/sqlserver/SQLServerDialect.java
@@ -30,6 +30,7 @@ import java.time.OffsetDateTime;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
@@ -1041,5 +1042,53 @@ public class SQLServerDialect extends BasicSQLDialect {
             }
             return null;
         }
+    }
+
+    @Override
+    public List<ReferencedEnvelope> getOptimizedBounds(
+            String schema, SimpleFeatureType featureType, Connection cx) throws SQLException {
+
+        if (null != dataStore.getVirtualTables().get(featureType.getTypeName())) {
+            return null;
+        }
+
+        // select the x/y of opposing corners of the aggregate envelope
+        final StringBuffer sql = new StringBuffer("SELECT geometry::EnvelopeAggregate(");
+        encodeColumnName(null, featureType.getGeometryDescriptor().getLocalName(), sql);
+        sql.append(").STPointN(1).STX, geometry::EnvelopeAggregate(");
+        encodeColumnName(null, featureType.getGeometryDescriptor().getLocalName(), sql);
+        sql.append(").STPointN(1).STY, geometry::EnvelopeAggregate(");
+        encodeColumnName(null, featureType.getGeometryDescriptor().getLocalName(), sql);
+        sql.append(").STPointN(3).STX, geometry::EnvelopeAggregate(");
+        encodeColumnName(null, featureType.getGeometryDescriptor().getLocalName(), sql);
+        sql.append(").STPointN(3).STY FROM ");
+        encodeTableName(featureType.getTypeName(), sql);
+
+        LOGGER.log(Level.FINE, "Optimized bounds SQL: " + sql);
+        try (Statement st =
+                        cx.createStatement(
+                                ResultSet.TYPE_FORWARD_ONLY, ResultSet.CONCUR_READ_ONLY);
+                ResultSet rs = st.executeQuery(sql.toString())) {
+            if (rs.next()) {
+                // getDouble returns 0 if the value is null, so we need to check for nulls
+                // explicitly
+                if (null == rs.getObject(1)
+                        || null == rs.getObject(2)
+                        || null == rs.getObject(3)
+                        || null == rs.getObject(4)) {
+                    return null;
+                }
+                return Collections.singletonList(
+                        new ReferencedEnvelope(
+                                rs.getDouble(1),
+                                rs.getDouble(3),
+                                rs.getDouble(2),
+                                rs.getDouble(4),
+                                featureType
+                                        .getGeometryDescriptor()
+                                        .getCoordinateReferenceSystem()));
+            }
+        }
+        return null;
     }
 }


modules/plugin/jdbc/jdbc-sqlserver/src/main/java/org/geotools/data/sqlserver/SQLServerDialect.java
```diff
@@ -1017,4 +1018,52 @@ public boolean canGroupOnGeometry() {
         // The type "geometry" is not comparable. It cannot be used in the GROUP BY clause.
         return false;
     }
+
+    @Override
+    public List<ReferencedEnvelope> getOptimizedBounds(
```
aaime, line 1023: Hmm... not sure this is the right place, but maybe I'm misunderstanding.

A bit of context, [here is the code using optimized bounds in JDBCDataStore](https://github.com/geotools/geotools/blob/4ef74f630509d029c09b85af2d585f8f9acf8453/modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java#L1271):

* Calls onto getOptimizedBounds
* If that returns nothing, calls onto Envelope aggregation
* In case an envelope aggregation was not available, the code is also looping over the results and accumulating bounds in memory

At the same time, the Javadoc for ``getOptimizedBounds`` states:

```
     * Returns the bounds of all geometry columns in the layer using any approach that proves to be
     * faster than the plain bounds aggregation (e.g., better than the "plain select extent(geom)
     * from table" on PostGIS), or null if none exists or the fast method has not been enabled
     * (e.g., if the fast method is just an estimate of the bounds you probably want the user to
     * enable it manually)
```

What I'm seeing here, and maybe I'm misunderstanding, is simple aggregation, which should be going in ``encodeGeometryEnvelope`` instead, while the current code in the SQL Server dialect is currently just returning the envelope of every single geometry, basically using the fall back path in JDCBDataStore.

I'm also guessing that ``EnvelopeAggregate`` was not available in the SQLServer versions we were targeting back when the store was published, but I don't remember having a list of supported SQL Server versions either so... it's probably ok.

mprins: `EnvelopeAggregate` is available in (Microsoft) supported versions (2017 and up) I'm  not sure  about 2008 (the is a [hint](https://social.technet.microsoft.com/wiki/contents/articles/4136.new-spatial-features-in-sql-server-code-named-denali.aspx) it is, but that is no longer a supported product.

As far as I know `EnvelopeAggregate` provides an optimized, database-side aggregation of a set of geometries (in this case all of a column of a table); for ref. https://learn.microsoft.com/en-us/sql/t-sql/spatial-geometry/envelopeaggregate-geometry-data-type?view=sql-server-linux-ver15 
This should IMO be that fastest way to get the bounds of a SQL Server featuretype (but it would depend on unknown database internals such as spatial index use) 

To avoid parsing the SQL Server geometry and not parse WKT (when using toString()) I ask for the X and Y of opposing corners (points 1 and 3) of the envelope/geometry (we already know the SRID):

```
SELECT
    geometry::EnvelopeAggregate(myGeomCol).STPointN(1).STX,
    geometry::EnvelopeAggregate(myGeomCol.STPointN(1).STY,
    geometry::EnvelopeAggregate(myGeomCol.STPointN(3).STX,
    geometry::EnvelopeAggregate(myGeomCol.STPointN(3).STY
FROM myTable;
```

I could have done: `SELECT geometry::EnvelopeAggregate(myGeomCol) FROM myTable;`  and then returned  `#decodeGeometryEnvelope(ResultSet rs, int column, Connection cx)` 


I've been wondering about the javadoc of `getOptimizedBounds` as all of the implementations seem to return a list with 1 envelope (not an envelope for each geometry column of a table; just the default geometry column) and that is what this new code does, a list with 1 ReferencedEnvelope that is the envelope of the (default) geometry column of the table.

perhaps the javadoc should instead be something like:
```
     * Returns the bounds of all geometries of a column in the layer using any approach ...
```

Not sure why `encodeGeometryEnvelope` should be part of this game... it does not execute a query

mprins: So... I could (should) execute the same SQL in a loop for all other geometry columns of the featuretype to make it match the javadoc?...

modules/plugin/jdbc/jdbc-sqlserver/src/main/java/org/geotools/data/sqlserver/SQLServerDialect.java
```diff
@@ -1042,4 +1043,52 @@ public Object getNextSequenceValue(String schemaName, String sequenceName, Conne
...
+        final StringBuffer sql = new StringBuffer("SELECT geometry::EnvelopeAggregate(");
+        encodeColumnName(null, featureType.getGeometryDescriptor().getLocalName(), sql);
+        sql.append(").STPointN(1).STX, geometry::EnvelopeAggregate(");
+        encodeColumnName(null, featureType.getGeometryDescriptor().getLocalName(), sql);
+        sql.append(").STPointN(1).STY, geometry::EnvelopeAggregate(");
+        encodeColumnName(null, featureType.getGeometryDescriptor().getLocalName(), sql);
+        sql.append(").STPointN(3).STX, geometry::EnvelopeAggregate(");
+        encodeColumnName(null, featureType.getGeometryDescriptor().getLocalName(), sql);
+        sql.append(").STPointN(3).STY FROM ");
+        encodeTableName(featureType.getTypeName(), sql);
```
aaime, line 1065: Just reading the text, it would seem SQLServer might end up computing the aggregate four times. I guess you verified it's not the case?
The alternative would be to read the geometry produced by EnvelopeAggregate, the dialect is able to do so anyways.

mprins: will need to verify, tied up right now so I will mark this as draft

PR #4049, opened 2022-10-14
Revert "Bump aws-java-sdk-s3 from 1.11.36 to 1.12.261 in /modules/unsupported/s3-geotiff"

mprins:
Reverts geotools/geotools#3966

because the original PR did not comply with the QA: https://github.com/geotools/geotools/actions/runs/2679353957/jobs/4176462298

/cc @ianturton @aaime


ianturton:
Sorry, that was my fault - I assumed a change to the pom couldn't cause a QA fail and that it must be a bad jar or something.

PR #4066, opened 2022-10-18
[GEOT-7210] YSLD does not encode layer name

jodygarnett:
[GEOT-7210](https://osgeo-org.atlassian.net/browse/GEOT-7210) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

Previously the encoder relied on utility method to extract the first style and did not have any referenced to named layer to user layer. I am preserving the limitation of only encoding one style (as YSLD is more accurately a representation of Symbology Encoding and has not previously had the concept of SLD NamedLayer.

This pull request completes the work started by https://github.com/geotools/geotools/pull/4000 

# Checklist

### DIFF
--- a/docs/user/extension/ysld.rst
+++ b/docs/user/extension/ysld.rst
@@ -218,12 +218,28 @@ Style definition:
 .. code-block:: yaml
 
     # style definition 
-    layer-name: <text>
     name: <text>
     title: <text>
     abstract: <text>
     feature-styles:
     - <feature style>
+
+Optional top-level sld and layer definition:
+
+.. code-block:: yaml
+   
+   # sld definition
+   sld-name: <text>
+   sld-title: <text>
+   sld-abstract: <text>
+   
+   # named layer definition
+   layer-name: <text>
+
+   # user layer definition
+   user-name: <text>
+   user-remote: <text>
+   user-service: <text>
 
 Feature style definition:
 
--- a/modules/extension/ysld/src/main/java/org/geotools/ysld/encode/RootEncoder.java
+++ b/modules/extension/ysld/src/main/java/org/geotools/ysld/encode/RootEncoder.java
@@ -18,14 +18,23 @@
 package org.geotools.ysld.encode;
 
 import java.util.Collections;
+import java.util.List;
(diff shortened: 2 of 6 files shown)


jodygarnett:
@sikeoka can you have a look at this please? I did a separate PR rather than hold up #4000

jodygarnett:
@aaime if you are available to review; this is a minor change to match #4000 
Has a test case, should be quite straight forward

sikeoka:
`SLD.defaultStyle(sld)` actually returns the first style with the IsDefault element set to true and then returns the first style if there are no default styles in the SLD. Something like the following would handle that case:
```
    protected Pair<String, Style> findFirst(StyledLayerDescriptor sld) {
        Pair<String, Style> first = null;
        for (StyledLayer layer : sld.getStyledLayers()) {
            List<Style> styles = Collections.emptyList();
            if (layer instanceof UserLayer) {
                styles = ((UserLayer) layer).userStyles();
            } else if (layer instanceof NamedLayer) {
                styles = ((NamedLayer) layer).styles();
            }
            for (Style style : styles) {
                if (style.isDefault()) {
                    return Pair.of(layer.getName(), style);
                } else if (first == null) {
                    first = Pair.of(layer.getName(), style);
                }
            }
        }
        return first;
    }
```

jodygarnett:
> `SLD.defaultStyle(sld)` actually returns the first style with the IsDefault element set to true and then returns the first style if there are no default styles in the SLD. 

I see; starting to look more like option 2 was the correct choice :)

Let me make your suggested change; the trouble is I need to find the Layer that belongs to that default style ..

aaime:
@jodygarnett not that interested in YSLD... I would still work on it in paid time, but not considering it for spare time involvement.

jodygarnett:
While I think Option 2 would of been a better choice; there is no harm in handling this common case:

* ``sld-name``
* ``sld-title``
* ``sld-abstract``

User Layer:

* ``user-name``
* ``user-remote``
* ``user-service``

NamedLayer

* ``layer-name``

PR #4068, opened 2022-10-18
Fix the issue that function postgis_lib_version() does not exist

ChengWuOne:
# Checklist

### DIFF
--- a/modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/PostGISDialect.java
+++ b/modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/PostGISDialect.java
@@ -1491,7 +1491,7 @@ public class PostGISDialect extends BasicSQLDialect {
             ResultSet rs = null;
             try {
                 st = conn.createStatement();
-                rs = st.executeQuery("select PostGIS_Lib_Version()");
+                rs = st.executeQuery("select PostGIS_Lib_Version()");// If the PostGIS extension is in other schema (not public schema), an exception is thrown. 
                 if (rs.next()) {
                     version = new Version(rs.getString(1));
                 }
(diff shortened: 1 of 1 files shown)


ChengWuOne:
If the PostGIS extension is in other schema (not public schema), an exception is thrown. So I think we need to get PostgreSQL server version in the here. Do you agree with me? Thanks.
The following is the exception information:
"Caused by: org.postgresql.util.PSQLException: ERROR: function postgis_lib_version() does not exist".

modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/PostGISDialect.java
```diff
@@ -1491,7 +1491,8 @@ public Version getVersion(Connection conn) throws SQLException {
             ResultSet rs = null;
             try {
                 st = conn.createStatement();
-                rs = st.executeQuery("select PostGIS_Lib_Version()");
+                //rs = st.executeQuery("select PostGIS_Lib_Version()");// If the PostGIS extension is in other schema (not public schema), an exception is thrown. 
+                rs = st.executeQuery("SHOW server_version");
```
mprins, line 1495: this is wrong, it does not provide the version of the postgis extension that is installed in the schema

mprins (changes requested):
probably your database setup is incorrect.

**note** that this is not a support forum, there are other means for that, see eg. https://geotools.org/getinvolved.html

modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/PostGISDialect.java
ianturton: If you have installed postgis somewhere other than in public (why?) it is your responsibility to make sure the search path includes that location so that the postgis functions can be found. Otherwise not finding the version will be the least of your issues.

PR #4136, opened 2022-12-02
method ((Geometry) geom).WkbSize(); not up to date

WhiteSte:
GDAL changed the return of ((Geometry) geom).WkbSize(); from int to long. Here a workaround

  

# Checklist

### DIFF
--- a/modules/unsupported/ogr/ogr-jni/src/main/java/org/geotools/data/ogr/jni/JniOGR.java
+++ b/modules/unsupported/ogr/ogr-jni/src/main/java/org/geotools/data/ogr/jni/JniOGR.java
@@ -692,7 +692,8 @@ public class JniOGR implements OGR {
 
     @Override
     public int GeometryGetWkbSize(Object geom) {
-        return ((Geometry) geom).WkbSize();
+        long result = ((Geometry) geom).WkbSize();
+        return (int) result;
     }
 
     @Override


WhiteSte:
That's the first pull request of my life, i'm quite newbie but I found a bug and also a solution into ogr plugin

WhiteSte:
Thank you for the answer, that's clear. Do you think is feasible to add in the doc that if one wants to use the ogr plugin must use a GDAL version < 3.2? Otherwise java run into some weird exceptions. That's just to warn who comes after me. 

Also, do you think there's a way to ask imageio-ext mantainers to switch a newer version of GDAL java bindings? GDAL >3.2 introduced filtering with spatial index in gdb and that should speed up really a lot the access to the data. Now i'm forced to used a gdal version lower than 3.2 and i'm missing that important feature

aaime:
@WhiteSte you can create a PR to add the relevant documentation.

About the upgrade, that normally takes a few days of work, testing out a variety of raster data sources across 3 different operating systems, so I would not hold my breath on it. Generally speaking, if you want some change to happen, see this guide (valid also for imageio-ext, the same group of developers works on it as well as on GT and GS):
https://github.com/geoserver/geoserver/wiki/Successfully-requesting-and-integrating-fixes,-improvements-and-new-features-in-GeoServer

PR #4164, opened 2023-01-11
[GEOT-7285] JDBCDataStore.getConnection(Transaction t) throws NullPoi…

roarbra:
[GEOT-7285](https://osgeo-org.atlassian.net/browse/GEOT-7285) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

Changed the Javadoc to point out that a transaction must be provided.

# Checklist

### DIFF
--- a/modules/library/jdbc/src/test/java/org/geotools/jdbc/JDBCDataStoreTest.java
+++ b/modules/library/jdbc/src/test/java/org/geotools/jdbc/JDBCDataStoreTest.java
@@ -228,26 +228,25 @@ public class JDBCDataStoreTest {
 
     @Test
     public void testGetConnectionAutocommit() throws Exception {
-        JDBCDataStore store = new JDBCDataStore();
+
         JDBCMockObjectFactory jdbcMock = new JDBCMockObjectFactory();
         MockDataSource dataSource = jdbcMock.getMockDataSource();
         dataSource.setupConnection(jdbcMock.getMockConnection());
+        JDBCDataStore store = new JDBCDataStore();
         store.setSQLDialect(mock(BasicSQLDialect.class));
         store.setDataSource(dataSource);
 
-        Connection conn = store.getConnection(Transaction.AUTO_COMMIT);
-        Assert.assertEquals(Boolean.TRUE, conn.getAutoCommit());
-        conn.close();
+        try (Connection conn = store.getConnection(Transaction.AUTO_COMMIT)) {
+            Assert.assertEquals(Boolean.TRUE, conn.getAutoCommit());
+        }
 
-        Transaction transaction = new DefaultTransaction();
-        conn = store.getConnection(transaction);
-        Assert.assertEquals(Boolean.FALSE, conn.getAutoCommit());
-        transaction.close();
-        conn.close();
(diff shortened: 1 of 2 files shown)


mprins (approved):
LGTM

modules/library/jdbc/src/test/java/org/geotools/jdbc/JDBCDataStoreTest.java
```diff
@@ -221,4 +225,31 @@ public JDBCReaderCallback createReaderCallback() {
...
+                Connection conn2 = store.getConnection(transaction)) {
+            Assert.assertEquals(Boolean.FALSE, conn2.getAutoCommit());
+        }
+
+        Transaction transaction = null;
+        try (Connection conn = store.getConnection(transaction)) {
+            Assert.fail("Transaction should be set.");
+        } catch (Exception e) {
+            // should fail
+        }
```
aaime, line 253:
```suggestion
       assertThrows(Exception.class, () -> store.getConnection(null));
```

aaime (commented):
Looks good, just a minor thing: the check for exception being thrown can be more succint using [assertThrows](https://junit.org/junit4/javadoc/latest/org/junit/Assert.html#assertThrows(java.lang.Class,%20org.junit.function.ThrowingRunnable))

modules/library/jdbc/src/test/java/org/geotools/jdbc/JDBCDataStoreTest.java
```diff
@@ -221,4 +225,27 @@ public JDBCReaderCallback createReaderCallback() {
...
+        try (Connection conn = store.getConnection(Transaction.AUTO_COMMIT)) {
+            Assert.assertEquals(Boolean.TRUE, conn.getAutoCommit());
+        }
+
+        try (Transaction transaction = new DefaultTransaction();
+                Connection conn2 = store.getConnection(transaction)) {
+            Assert.assertEquals(Boolean.FALSE, conn2.getAutoCommit());
+        }
+
+        Transaction transaction = null;
```
aaime, line 248: PMD is a little dense here:

```
2023-01-14T15:44:59.3686060Z [INFO] PMD Failure: org.geotools.jdbc.JDBCDataStoreTest:248 Rule:CloseResource Priority:3 Ensure that resources like this Transaction object are closed after use.
```

Probably easier to just pass "null" to the getConnection method without holding a variable...

roarbra: Passing null will not compile, because there is two getConnection method's. It would be getConnection((Transaction)null).
The PMD is complaining about a missing close, so I'll fix it with a try-with-resources. Maybe that PMD would allow a SuppressWarnings attribute as well.

aaime: Yes, PMD allows for a SuppressWarning indeed, it's explained in the QA page, linked also from the checklist above:
https://docs.geotools.org/latest/developer/conventions/code/qa.html

PR #4174, opened 2023-01-20
[GEOT-7300] Filter capabilities does not support PropertyIsNil filters

sikeoka:
[GEOT-7300](https://osgeo-org.atlassian.net/browse/GEOT-7300) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

This PR adds PropertyIsNil filters to the filter Capabilities and fixes a ClassCastException when PropertyIsNil filters are supported.

  

# Checklist

### DIFF
--- a/modules/library/main/src/main/java/org/geotools/filter/Capabilities.java
+++ b/modules/library/main/src/main/java/org/geotools/filter/Capabilities.java
@@ -42,6 +42,7 @@ import org.opengis.filter.PropertyIsGreaterThanOrEqualTo;
 import org.opengis.filter.PropertyIsLessThan;
 import org.opengis.filter.PropertyIsLessThanOrEqualTo;
 import org.opengis.filter.PropertyIsLike;
+import org.opengis.filter.PropertyIsNil;
 import org.opengis.filter.PropertyIsNotEqualTo;
 import org.opengis.filter.PropertyIsNull;
 import org.opengis.filter.capability.FilterCapabilities;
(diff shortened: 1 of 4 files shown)


sikeoka:
The downstream build had an unrelated Java heap space error twice.

aaime:
Hum... I'm unsure about the side effects of this change. PropertyIsNil is not well supported across GeoTools.
I guess that using Capabilites/CapabilitiesFilterSplitter isolates the changes to the few parts of GeoTools actually using those classes (the bulk is using FilterCapabilities/PostPreProcessFilterSplittingVisitor instead, and throws Exceptions when PropertyIsNil is provided to the visitors).

That said, for example, ElasticSearch is likely going to be affected, as it uses CapabilitiesFilterSplitter but FilterToElastic throws an exception is PropertyIsNil is provided. The other bits that might be affected are wfs-ng (maybe in a good way!) and jdbcconfig (also in a good way, although I don't see PropertyIsNil being used, Predicates works with PropertyIsNull instead).

sikeoka:
There shouldn't be any side effects unless a module is adding PropertyIsNil to its filter capabilities even though it isn't supposed to be supporting it and even then, there won't be any side effects except when a PropertyIsNil filter is passed to the module.  This change is intended for jdbcconfig since it does implement support for PropertyIsNil.  I noticed that it had a unit test that said it was testing PropertyIsNil but was actually testing PropertyIsNull and changing the test to PropertyIsNil ran into this problem.

aaime:
@sikeoka makes sense. Wondering what's putting PropertyIsNil in the path to JDBCConfig though? 
As indicated above, Predicates (used to build catalog filters) seems to be using PropertyIsNull instead.

sikeoka:
It looks like PropertyIsNil isn't actually used anywhere at all outside of unit tests.  This PR shouldn't break anything and would be necessary if you ever decided to start using PropertyIsNil more but I could just remove the PropertyIsNil test from jdbcconfig for now if you prefer to do that.

PR #4202, opened 2023-02-17
[GEOT-7310] Update Oracle JDBC driver to 19.18.0.0

mprins:
[GEOT-7310](https://osgeo-org.atlassian.net/browse/GEOT-7310) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

Upgrade to the latest patch version.

Also update the CI workflow to pull from the Github registry instead of the Docker registry.

  

# Checklist

### DIFF
--- a/build/ci/oracle/start-oracle.sh
+++ b/build/ci/oracle/start-oracle.sh
@@ -3,12 +3,12 @@ docker version
 
 # this docker image does not have Java installed, which means you can not use Java stored procedures
 # such as "SDO_GEOMETRY('POINT (1.0 2.0)', 4326))"
-docker pull gvenzl/oracle-xe:18.4.0
+docker pull ghcr.io/gvenzl/oracle-xe:18.4.0
 
 # start the dockerized oracle-xe instance (the container will be destroyed/removed on stopping)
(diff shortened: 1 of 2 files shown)


mprins:
Well... perhaps not. Seems there is a bug in the driver wrt `return`, see also:
- https://stackoverflow.com/questions/72717261/autokeyinfo-is-not-initialized-when-trying-to-batch-insert-into-table-with-gen
- https://github.com/schauder/stackoverflow/tree/main/jdbc/oracle-autokeyinfo

The advice is to use the LTS version 19.x (which exists as ojdbc10, eg ojdbc10-19.18.0.0)

aaime:
Strange... wondering if some topological rule changed and the driver has some flag to enable/disable the different behavior? Just thinking out loud...

mprins:
the only thing I could think of is that a spatial index is required... but with 3 and 4 records in those tables...  ¯\_(ツ)_/¯

As any XE is an unsupported product and I don't have access to the real licensed thing or a community I'm just gonna leave it at this for now

PR #4205, opened 2023-03-03
Fix broken image links

VirtualTim:


VirtualTim:
I was mistaken about this. I thought the docs were for viewing on github, I now realise they're for viewing on https://docs.geotools.org/stable/userguide/welcome/architecture.html

PR #4241, opened 2023-03-17
remove bedatadriven JTS module and use 53N jar instead

ianturton:
# Checklist


ianturton:
Need to work out how to handle Empty Geometries in both reader and writer, and fix up GeoServer before merging.

aaime:
Interesting... is that a separate implementation from bedatadriven? Is it more complete/better? What's happening with emtpy geometries? Do we need to maybe do a contribution on their repo?

ianturton:
Yes, it's a bit more up to date than bedatadriven one which seems little
abandoned. Also it's a published jar rather than dumped in to our tree.

The current deserializer and serialisation code doesn't handle them, I need
to go back to the spec to check they are valid and the I might offer them a
pr. I can't remember why we added tests for it, is it an OGC thing?

Ian

aaime:
Was it this ticket? https://osgeo-org.atlassian.net/browse/GEOT-5192

modules/unsupported/geojson-core/pom.xml
mprins, line 1: Alternatively you could use the graphhopper fork that has been kept up to date instead of a whole different implementation
https://github.com/graphhopper/jackson-datatype-jts/

modules/unsupported/geojson-core/src/main/java/org/geotools/data/geojson/ObjectMapperFactory.java
```diff
@@ -43,4 +43,11 @@ class ObjectMapperFactory {
     public static ObjectMapper getDefaultMapper() {
         return DEFAULT_MAPPER;
     }
+
+    public static ObjectMapper getDefaultMapper(int maxDecimals) {
+        // I can't work out how to deregister a module from the mapper so make a new one.
```
mprins, line 48: It's not possible to un-/ deregister modules completely, creating a new mapper is the only way

aaime:
Thinking out loud... what happens when we upgrade JTS? We'll need a "friendly" implementation that lets us upgrade the JTS dependency and push a release. That is only going to be necessary when JTS breaks binary compatibility or behavior but... that has happened a few times already in last few years.

PR #4244, opened 2023-03-26
[GEOT-7333] The result of linearization after creating circulararc is wrong

stefanuhrig:
We improve the precision of the center computed from three control points in the CircularArc class.

The improvement considers the matrix condition number and selects the best 2 out of 3 equations for finding the center.

We reduce the error further by solving the equation system with an LU- decomposition with pivoting.

# Checklist


modules/library/main/src/main/java/org/geotools/geometry/jts/CircularArc.java
```diff
@@ -282,24 +285,151 @@ private void initializeCenterRadius() {
...
                 centerX = sx + (mx - sx) / 2.0;
                 centerY = sy + (my - sy) / 2.0;
-
-                radius = sqrt((centerX - sx) * (centerX - sx) + (centerY - sy) * (centerY - sy));
             } else {
-                temp = mx * mx + my * my;
-                bc = (sx * sx + sy * sy - temp) / 2.0;
-                cd = (temp - ex * ex - ey * ey) / 2.0;
-                determinate = (sx - mx) * (my - ey) - (mx - ex) * (sy - my);
+                // Let <a, b> be the dot product of the two-dimensional vectors a and b. Given three
```
aaime, line 289: This looks like some serious math. Do you have online references for it?

stefanuhrig: No, not really.

In most online resources, the methods/formulas for finding the circle defined by three points are correct, but numerically unstable.

We suffered numerical stability issues with circular arcs in our spatial engine as well some years ago and put some effort in finding numerically stable approaches. This was the result.

The system of equations this method starts with are the ones you get in step 3 of the method described in https://math.stackexchange.com/questions/213658/get-the-equation-of-a-circle-when-given-3-points.

The nice property of that approach is that the entries of the matrix are just differences of coordinates. There are no squares or products of coordinates that will cripple the matrix condition.

Furthermore, we also consider the equation that results from subtracting the second from the third equation and then select the two equations that result in the best-conditioned system.

Finding the formula for the condition number of an arbitrary 2x2 matrix was rather tedious. The derivation is lengthy, so I did not put it in the comments, but it can be tested by plugging in some numbers and comparing the result with online calculators like https://www.omnicalculator.com/math/condition-number.

The condition number threshold above which the points are considered collinear is of course debatable. It was the result of empiric data. However, it is much better than using the determinant as the determinant does not really tell you anything about how close a matrix is to being singular. This is discussed for example at https://www.physicsforums.com/threads/condition-number-of-a-matrix-is-1-but-determinant-is-almost-0.520860/.

Maybe you know someone in the GeoTools/GeoServer community with good math skills who can review the change?

aaime: Personally, changes look legit and no test has failed, so I'm going to go with it. I see you requested review to @ianturton as well, lets hear from him too :smile:

ianturton: The maths is over my head I'm afraid. But provided the existing tests don't break and the new ones pass I'm happy.

ianturton (approved):

aaime:
In it goes!

PR #4263, opened 2023-04-15
[GEOT-7340] Avoid PAMParser init causing ImageMosaic to fail

roarbra:
[GEOT-7340](https://osgeo-org.atlassian.net/browse/GEOT-7340) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

Fixes an annoying exception in static part of class initialization when using ImageMosaic.

Related to class loading of javax.xml.bind.JAXBException, and how this is treated in it.geosolutions.imageio.pam.PAMParser.

Not sure how to write a test case for this one.

# Checklist

### DIFF
--- a/modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/GranuleDescriptor.java
+++ b/modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/GranuleDescriptor.java
@@ -154,8 +154,7 @@ public class GranuleDescriptor {
                     new VectorBinarizeRIF(),
                     Registry.JAI_TOOLS_PRODUCT);
         } catch (Throwable e) {
-            LOGGER.log(Level.SEVERE, "Error setting up GranuleDescriptor", e);
-            throw new RuntimeException("Error setting up GranuleDescriptor.", e);
+            LOGGER.log(Level.SEVERE, "Error when registering RIF for GranuleDescriptor.", e);
         }
     }
 
--- a/modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/GranuleDescriptor.java
+++ b/modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/GranuleDescriptor.java
@@ -153,10 +153,9 @@ public class GranuleDescriptor {
                     new VectorBinarizeDescriptor(),
                     new VectorBinarizeRIF(),
                     Registry.JAI_TOOLS_PRODUCT);
-        } catch (Exception e) {
-            if (LOGGER.isLoggable(Level.FINE)) {
-                LOGGER.log(Level.FINE, e.getLocalizedMessage());
-            }
+        } catch (Throwable e) {
+            LOGGER.log(Level.SEVERE, "Error setting up GranuleDescriptor", e);
+            throw new RuntimeException("Error setting up GranuleDescriptor.", e);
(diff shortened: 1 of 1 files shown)


modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/GranuleDescriptor.java
```diff
@@ -153,10 +153,9 @@ public URL resolve(String granuleLocation) {
...
                     new VectorBinarizeRIF(),
                     Registry.JAI_TOOLS_PRODUCT);
-        } catch (Exception e) {
-            if (LOGGER.isLoggable(Level.FINE)) {
-                LOGGER.log(Level.FINE, e.getLocalizedMessage());
-            }
+        } catch (Throwable e) {
+            LOGGER.log(Level.SEVERE, "Error setting up GranuleDescriptor", e);
+            throw new RuntimeException("Error setting up GranuleDescriptor.", e);
         }
```
aaime, line 159: From ignored exception to throwing? That's not backwards compatible... also, what is the relationship with the issue at hand?

aaime (commented):
From the ticket it seems that there were classpath setup issues or incompatible libraries loaded in the classpath, causing the issue.

PAM files are sidecars that can be used with all sorts of input files, they are not images, but statistical summaries or the raster contents. 

The comment "... PAMParser has no meaning. Since the mosaic doesn’t consist of that kind of files." seems to have been written in a moment of distress, with no understanding of what a PAM file is (no mosaic can "consist of that kind of files" indeed).

An error such as " java.lang.ClassNotFoundException: javax.xml.bind.JAXBException" 
seems to simply point at a missing dependency.

It seems to me what you want is to have the JAXB dependency be made optional.

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/GranuleDescriptor.java
roarbra: I'm willing to skip the throwing, but I would keep the logging.

roarbra:
> The comment "... PAMParser has no meaning. Since the mosaic doesn’t consist of that kind of files." seems to have been written in a moment of distress, with no understanding of what a PAM file is (no mosaic can "consist of that kind of files" indeed).

That sentence could be interpreted in all kind of ways, and doesn't necessarily mean that I say it out of distress. Just that I have looked up what a PAM file is and see that the image files of my mosaic doesn't have such files.

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/GranuleDescriptor.java
```diff
@@ -641,7 +639,16 @@ private void checkPamDataset() throws IOException {
...
+        if (auxFile.exists()) {
+            if (pamParser == null) {
+                try {
+                    pamParser = PAMParser.getInstance();
+                } catch (Throwable e) {
+                    throw new RuntimeException("Couldn't initialize PAM parser.", e);
+                }
+            }
+            pamDataset = pamParser.parsePAM(auxFile);
+        }
```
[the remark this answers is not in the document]
dromagnoli: I agree. Nothing special to report here.

PR #4268, opened 2023-04-15
[GEOT-7341] Update Spotbugs to 4.7.3

aaime:
[GEOT-7341](https://osgeo-org.atlassian.net/browse/GEOT-7341) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

See ticket

# Checklist

### DIFF
--- a/modules/library/referencing/src/main/java/org/geotools/referencing/crs/DefaultProjectedCRS.java
+++ b/modules/library/referencing/src/main/java/org/geotools/referencing/crs/DefaultProjectedCRS.java
@@ -27,6 +27,7 @@ import javax.measure.quantity.Length;
 import org.geotools.referencing.operation.DefaultOperationMethod;
 import org.geotools.referencing.operation.DefiningConversion;
 import org.geotools.referencing.wkt.Formatter;
+import org.geotools.util.SuppressFBWarnings;
 import org.opengis.geometry.MismatchedDimensionException;
 import org.opengis.parameter.GeneralParameterDescriptor;
 import org.opengis.parameter.GeneralParameterValue;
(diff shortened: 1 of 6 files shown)


jodygarnett (approved):
Thanks for the upgrade!

modules/unsupported/process-feature/src/main/java/org/geotools/process/vector/Contours.java
```diff
@@ -45,6 +45,11 @@
 
 public class Contours {
 
+    private enum PointType {
+        Point,
+        MultiPoint
```
jodygarnett, line 50: Is this change related? Or off topic ...

sikeoka: I think this is related to a rule that looks for using == to compare strings instead of the equals method.  It seems like a simple boolean flag could be used here instead of strings or an enum.

modules/library/referencing/src/main/java/org/geotools/referencing/crs/DefaultProjectedCRS.java
```diff
@@ -250,6 +251,7 @@ public int hashCode() {
      * @return The name of the WKT element type, which is {@code "PROJCS"}.
      */
     @Override
+    @SuppressFBWarnings("ES_COMPARING_STRINGS_WITH_EQ")
```
sikeoka, line 254: Wouldn't it be better to change x==y to x.equals(y) to fix the warning even if == is okay here instead of suppressing the warning?

jodygarnett: Using ``equals`` would be a performance difference; not sure how performance sensitive this is.

modules/unsupported/process-feature/src/main/java/org/geotools/process/vector/Contours.java
aaime: Yep, it's to avoid string comparison with identity equality.

modules/library/referencing/src/main/java/org/geotools/referencing/crs/DefaultProjectedCRS.java
aaime: I left it with == since the original author went to the trouble of writing a comment explaining that identity comparison was ok in that case. Also, code is convoluted (name is assigned during a if and result passed into a method as a side effect... horrible...)

PR #4275, opened 2023-04-27
[GEOT-7345] WFS Capabilities from local file

roarbra:
[GEOT-7345](https://osgeo-org.atlassian.net/browse/GEOT-7345) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

The goal is to let WFSDataAccessFactory handle GetCapabilities stored as a local file.

To do this I've done a little change to MultithreadedHttpClient, so that it behaves in the same way as SimpleHttpClient for file url's. This is done by making them both subclasses og AbstractHttpClient and have functionality there for making connection against file url's.

I did also a refactor of the class SimpleHttpClient.SimpleHTTPResponse -> DefaultHttpResponse.

# Checklist

### DIFF
--- a/modules/library/http/src/main/java/org/geotools/http/DefaultHttpResponse.java
+++ b/modules/library/http/src/main/java/org/geotools/http/DefaultHttpResponse.java
@@ -1,0 +1,106 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ *
+ *    (C) 2023, Open Source Geospatial Foundation (OSGeo)
+ *
+ *    This library is free software; you can redistribute it and/or
+ *    modify it under the terms of the GNU Lesser General Public
+ *    License as published by the Free Software Foundation;
+ *    version 2.1 of the License.
+ *
+ *    This library is distributed in the hope that it will be useful,
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ *    Lesser General Public License for more details.
+ */
+package org.geotools.http;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.HttpURLConnection;
+import java.net.URLConnection;
+import java.util.zip.GZIPInputStream;
+
+/**
+ * HTTPResponse that uses URLConnection as its basis.
+ *
+ * @author Roar Brænden
+ */
+public class DefaultHttpResponse implements HTTPResponse {
+
+    private URLConnection connection;
+
+    private InputStream responseStream;
+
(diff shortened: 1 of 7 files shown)


modules/library/http/src/main/java/org/geotools/http/DefaultHttpResponse.java
```diff
@@ -0,0 +1,106 @@
...
+package org.geotools.http;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.HttpURLConnection;
+import java.net.URLConnection;
+import java.util.zip.GZIPInputStream;
+
+/**
+ * HTTPResponse that uses URLConnection as it's basis.
```
aaime, line 26:
```suggestion
 * HTTPResponse that uses URLConnection as its basis.
```

modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/impl/WFSDataAccessFactory.java
```diff
@@ -599,19 +598,30 @@ protected boolean canProcess(final Map<String, ?> params, int maximumGmlComplian
...
-                    // SimpleHTTPClient class
-                }
+            if (url == null) {
+                return false;
             }
-        } catch (Exception e) {
+        } catch (IOException e) {
             return false;
         }
+        /*
```
aaime, line 610: Avoid commented-out code sections.

modules/unsupported/wfs-ng/src/test/java/org/geotools/data/wfs/WFSDataStoreFactoryTest.java
```diff
@@ -108,10 +115,10 @@ public void testCreateDataStore() throws IOException {
      */
     private WFSDataStore testCreateDataStore(
             final String capabilitiesFile, final Version expectedVersion) throws IOException {
-
-        Map<String, Serializable> params = new HashMap<>();
-        params.put("TESTING", Boolean.TRUE);
-
+        /*
```
aaime, line 118: Avoid commented out code sections

modules/library/http/src/main/java/org/geotools/http/SimpleHttpClient.java
```diff
@@ -195,6 +151,7 @@ private URLConnection openConnection(URL finalURL) throws IOException {
         return connection;
     }
 
+    @Deprecated
     public static class SimpleHTTPResponse implements HTTPResponse {
```
aaime, line 155: Should it extend DefaultHTTPResponse? I'm guessing this is where the new DefaultHTTPResponse came from, when we do this we normally use inheritance to avoid duplication and deprecate the subclass.

roarbra:
I have made changes according to the comments. The test failure doesn't seem to be related to my update.

PR #4291, opened 2023-05-19
[GEOT-7332] Resolve incompatibility between LGPL-2.1 and LGPL-3+

sebastian-orlowski:
Before this change, the file GeometrySmoother.java was licensed both with "LGPL-2.1 only" and "LGPL-3.0 or any later version", which is impossible to fulfill at the same time -- which license could be used for this file?

By changing "LGPL-2.1"to "LGPL-2.1 or any later version", it is possible to license this file under LGPL-3.0, hereby resolving this conflict.

# Checklist

- [x ] I have read the [contribution guidelines](https://github.com/geotools/geotools/blob/main/CONTRIBUTING.md).
- [ ] I have sent a [Contribution Licence Agreement](https://docs.geotools.org/latest/developer/procedures/contribution_license.html) (not required for small changes, e.g., fixing typos in documentation). - Not needed, small change.

- [ ] New unit tests have been added covering the changes. - Not needed, only documentation change.

### DIFF
--- a/modules/library/main/src/main/java/org/geotools/geometry/jts/GeometrySmoother.java
+++ b/modules/library/main/src/main/java/org/geotools/geometry/jts/GeometrySmoother.java
@@ -7,7 +7,7 @@
  *    This library is free software; you can redistribute it and/or
  *    modify it under the terms of the GNU Lesser General Public
  *    License as published by the Free Software Foundation;
- *    version 2.1 of the License.
+ *    version 2.1 or any later versiun of the License.
  *
  *    This library is distributed in the hope that it will be useful,
  *    but WITHOUT ANY WARRANTY; without even the implied warranty of


ianturton:
I don't see any evidence that this file is also under LGPL-3.0

sebastian-orlowski:
Hi Ian,

the copyright notice states that the content of this class is basically composed of code from other classes from the jai-tools, which have been licensed under LGPL-3 or later, see the excerpt below. Doesn't this mean that this code is licensed with LGPL-3+? How does that fit together with the LGPL-2.1 license requirement stated in the same class? Which would be the license to choose (and why) if we distribute this class as a library used in our software (without any modifications)? We do not see how we could fulfill both requirements at the same time.

Thanks and best regards,
Sebastian

/* 
 * [...]
 * This class is composed of code taken from JAI-tools version 1.1.0
 * http://code.google.com/p/jai-tools/
 *
 * It combines the JAI-tools classes AbstractSmoother, LineSmoother,
 * PolygonSmoother and SmootherControl. All are covered by the following
 * copyright and license:
 *
 * Copyright 2010-2011 Michael Bedward
 *
 * This file is part of jai-tools.
 *
 * jai-tools is free software: you can redistribute it and/or modify
 * it under the terms of the GNU Lesser General Public License as
 * published by the Free Software Foundation, either version 3 of the
 * License, or (at your option) any later version.
 * [...]
 */

PR #4337, opened 2023-06-12
[GEOT-6512] URLCheckers normalize location

jodygarnett:
[GEOT-6512](https://osgeo-org.atlassian.net/browse/GEOT-6512) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

This is a follow up to https://github.com/geoserver/geoserver/pull/6916 which highlighted the need to account for normalization of locations prior to checking.

URLChecker.normalize(location) is introduced to:
- remove any redundant current directory or parent directory references.
- Converts ``file:/path`` locations to ``file:///path`` representation.

# Checklist

### DIFF
--- a/modules/library/main/src/test/java/org/geotools/data/ows/URLCheckersTest.java
+++ b/modules/library/main/src/test/java/org/geotools/data/ows/URLCheckersTest.java
@@ -17,8 +17,10 @@
 package org.geotools.data.ows;
 
 import static org.hamcrest.MatcherAssert.assertThat;
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertThrows;
 
+import java.io.File;
 import org.hamcrest.Matchers;
 import org.junit.After;
 import org.junit.Test;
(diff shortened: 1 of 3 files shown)


jodygarnett:
Okay I am not enjoying this one at all ... I really wish to use the Java built-in URI and Path normalization rather than roll my own as I do not wish any possibility of error/inconsistency in a facility setup to address security vulnerabilities.

Glitches:

1. File URI / URL handling has two supported representations for ``file:/path`` and ``file:///path`` both of which are equally valid. The ``uri.normalize().toString()`` workflow produces inconsistant output depending on how the URI was constructed. When creating a URI using parameters or normalize() this will create ``file:/path`` (no host location), although when URI is created using a String it is willing to preserve the string resulting in ``file:///path`` (empty host location).

2. Windows ``\\`` vs Linux ``/`` will cause path.normalize().toString() to change ``/path/file`` into ``\\path\\file`` which may or may not be what people expect to check against?

modules/library/main/src/test/java/org/geotools/data/ows/URLCheckersTest.java
```diff
@@ -72,18 +74,99 @@ public void testMultipleCheckers() {
...
+                "https://localhost:8080/geoserver/ne/ows?SERVICE=WFS&REQUEST=GetCapabilities&VERSION=1.1.0");
+
+        // valid normalization
+        URLCheckers.confirm("file:///data/../data/dem.tif");
+        URLCheckers.confirm("file:/data/../data/dem.tif");
+        URLCheckers.confirm("file:///data/./dem.tif");
+        URLCheckers.confirm("file:/data/./dem.tif");
+
+        // invalid normalization
+        assertThrows(URLCheckerException.class, () -> URLCheckers.confirm("file:///tmp"));
```
aaime, line 127: Maybe also add one to confirm path traversal has been neutralized, e.g. ``file:///data/../tmp``

modules/library/main/src/test/java/org/geotools/data/ows/URLCheckersTest.java
```diff
@@ -72,18 +74,99 @@ public void testMultipleCheckers() {
...
+    }
+
+    @Test
+    public void testNormalization() {
+        Assert.assertEquals(
```
aaime, line 132: static import for assertEquals?

aaime (changes requested):
Looks good to me. I see the normalization is a bit painful, but I would not be able to suggest a better approach.

jodygarnett:
This PR must be merged at the same time as https://github.com/geoserver/geoserver/pull/6916 for the "Downstream integration build (GeoWebCache and GeoServer) / build (pull_request) " test to pass.

jodygarnett:
Thanks @aaime feedback addressed, let me push up the changes to https://github.com/geoserver/geoserver/pull/6916 and we can stage all of this stuff for backport.

modules/library/main/src/test/java/org/geotools/data/ows/URLCheckersTest.java
jodygarnett: Added

ianturton (approved):

PR #4349, opened 2023-06-16
[GEOT-7381] FlatGeobufDataStore FID filtering

bjornharrtell:
[GEOT-7381](https://osgeo-org.atlassian.net/browse/GEOT-7381) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

### DIFF
--- a/modules/unsupported/flatgeobuf/src/main/java/org/geotools/data/flatgeobuf/FlatGeobufFeatureReader.java
+++ b/modules/unsupported/flatgeobuf/src/main/java/org/geotools/data/flatgeobuf/FlatGeobufFeatureReader.java
@@ -20,6 +20,7 @@ import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URL;
+import java.util.Arrays;
 import java.util.Iterator;
 import java.util.NoSuchElementException;
 import org.geotools.data.DataStore;
@@ -33,6 +34,7 @@ import org.opengis.feature.simple.SimpleFeature;
 import org.opengis.feature.simple.SimpleFeatureType;
 import org.opengis.filter.Filter;
 import org.opengis.filter.Id;
+import org.opengis.filter.identity.Identifier;
 import org.wololo.flatgeobuf.HeaderMeta;
 
 public class FlatGeobufFeatureReader implements FeatureReader<SimpleFeatureType, SimpleFeature> {
@@ -103,14 +105,30 @@ public class FlatGeobufFeatureReader implements FeatureReader<SimpleFeatureType,
                                     inputStream, headerMeta, featureType, bbox)
                             .iterator();
         else if (id != null) {
+            long[] fids = extractSortedFids(id);
             it =
-                    FeatureCollectionConversions.deserialize(inputStream, headerMeta, featureType)
+                    FeatureCollectionConversions.deserialize(
+                                    inputStream, headerMeta, featureType, fids)
                             .iterator();
         } else {
             it =
                     FeatureCollectionConversions.deserialize(inputStream, headerMeta, featureType)
                             .iterator();
         }
+    }
+
+    private static long[] extractSortedFids(Id id) {
+        long[] fids = id.getIdentifiers().stream().mapToLong(i -> extractFid(i)).toArray();
+        Arrays.sort(fids);
+        return fids;
+    }
+
+    private static long extractFid(Identifier i) {
+        String idStr = i.getID().toString();
+        int dotIndex = idStr.indexOf(".", 0);
+        String idPart = idStr.substring(dotIndex + 1);
+        long fid = Long.parseLong(idPart);
+        return fid;
     }
 
     public static void skipNBytes(InputStream stream, long skip) throws IOException {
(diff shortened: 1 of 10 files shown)


bjornharrtell:
Self review makes me realize this is not complete yet and lacks a test for the FlatGeobufFeatureReader.

aaime:
All hail the self review! :+1:

modules/unsupported/flatgeobuf/src/main/java/org/geotools/data/flatgeobuf/FlatGeobufFeatureReader.java
```diff
@@ -82,28 +86,55 @@ public class FlatGeobufFeatureReader implements FeatureReader<SimpleFeatureType,
...
+    private static long[] extractSortedFids(Id id) {
+        long[] fids = id.getIdentifiers().stream().mapToLong(i -> extractFid(i)).toArray();
+        Arrays.sort(fids);
+        return fids;
+    }
 
-        it =
-                FeatureCollectionConversions.deserialize(inputStream, headerMeta, featureType, bbox)
-                        .iterator();
+    private static long extractFid(Identifier i) {
```
bjornharrtell, line 126: Native feature id in FlatGeobuf are longs so I have to extract it from Identifer and I found no other way to get rid of the namespace prefix. Is this ok @aaime or do you know of a better way?

bjornharrtell:
I think this is ready now. 🙂

modules/unsupported/flatgeobuf/src/main/java/org/geotools/data/flatgeobuf/FlatGeobufFeatureReader.java
bjornharrtell: Good points. Added tests for unexpected id and out of bounds id. I think the exceptions thrown make sense.

modules/unsupported/flatgeobuf/src/test/java/org/geotools/data/flatgeobuf/FlatGeobufDataStoreTest.java
```diff
@@ -881,6 +883,36 @@ public void readCountriesFids() throws IOException {
...
+    public void readCountriesFidsInvalid() throws IOException {
+        SimpleFeatureSource featureSource = getFeatureSource("countries");
+        SimpleFeatureType schema = featureSource.getSchema();
+        FilterFactory2 ff = CommonFactoryFinder.getFilterFactory2();
+        Filter filter = ff.id(ff.featureId("countries.noreallong"));
+        Query query = new Query(schema.getTypeName(), filter);
+        SimpleFeatureCollection featureCollection = featureSource.getFeatures(query);
+        try (SimpleFeatureIterator it = featureCollection.features()) {
+            while (it.hasNext()) {
+                it.next();
```
aaime, line 896: If it reaches here, the test has actually failed, it matched a feature while the result should have been empty... or not? Something like this (may not pass QA, not sure if indentation is correct):


```suggestion
                SimpleFeature f = it.next();
                fail("Unexpected feature match" + f);
```

aaime: And no, it should not throw a number format, you should be getting nothing. is.hasNext() returns false and the loop exists.

modules/unsupported/flatgeobuf/src/test/java/org/geotools/data/flatgeobuf/FlatGeobufDataStoreTest.java
```diff
@@ -863,6 +865,54 @@ public void readCountriesBbox() throws IOException {
...
+            }
+        }
+    }
+
+    @Test(expected = NoSuchElementException.class)
```
aaime, line 901: Same here, it.hasNext should just return false and you get nothing back, including no exception.

modules/unsupported/flatgeobuf/src/test/java/org/geotools/data/flatgeobuf/FlatGeobufDataStoreTest.java
bjornharrtell: Ah.. yes, I should filter them out.

modules/unsupported/flatgeobuf/src/main/java/org/geotools/data/flatgeobuf/FlatGeobufFeatureReader.java
bjornharrtell: Reworked to filter out instead of throw.

bjornharrtell:
I have trouble understanding the Downstream integration build test output and if it's related or not to this PR.

aaime:
Yep, we have seen that elsewhere too. @NielsCharlier we have another failure in the gs-metadata module here (a GT PR, the downstream GS build is failing).

bjornharrtell:
@aaime time to merge?

aaime:
The downstream build should be in good shape now, trying it again.

bjornharrtell:
@aaime looks green :)

aaime:
If you need backports, please add the backport labels, the bot will try to make the PRs for you (if you can add them at least, otherwise let me know).

bjornharrtell:
Yes would be good with backports. Doesn't look like I can add labels.

PR #4364, opened 2023-06-29
[GEOT-7377] Corrected off-by-one-error to find upper left tile

gillesB:
[GEOT-7377](https://osgeo-org.atlassian.net/browse/GEOT-7377) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

There was a off-by-one error in WMTSTileService and WMTSTileFactory whose result was that tiles were found which do not intersect the requested envelope.

An existing test was extended to check the intersections.

Fixes: https://osgeo-org.atlassian.net/browse/GEOT-7377

  

# Checklist

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.0:compile (default-compile) on project gt-wmts: Fatal error compiling: Illegal/unsupported escape sequence near index 3
[ERROR] C:\Users\gilles.baatz\git\gilles\geotools\modules\extension\wmts\target/generated-sources/.*
```

### DIFF
--- a/modules/extension/wmts/src/main/java/org/geotools/ows/wmts/client/WMTSTileFactory.java
+++ b/modules/extension/wmts/src/main/java/org/geotools/ows/wmts/client/WMTSTileFactory.java
@@ -110,8 +110,8 @@ public class WMTSTileFactory extends TileFactory {
         long xTile = origxTile;
         long yTile = origyTile;
 
-        if (xTile >= limits.getMaxcol()) xTile = limits.getMaxcol() - 1;
-        if (yTile >= limits.getMaxrow()) yTile = limits.getMaxrow() - 1;
+        if (xTile >= limits.getMaxcol()) xTile = limits.getMaxcol();
+        if (yTile >= limits.getMaxrow()) yTile = limits.getMaxrow();
 
         if (xTile < limits.getMincol()) xTile = limits.getMincol();
         if (yTile < limits.getMinrow()) yTile = limits.getMinrow();
--- a/modules/extension/wmts/src/main/java/org/geotools/ows/wmts/client/WMTSTileService.java
+++ b/modules/extension/wmts/src/main/java/org/geotools/ows/wmts/client/WMTSTileService.java
@@ -630,8 +630,8 @@ public class WMTSTileService extends TileService {
         long xTile = origxTile;
         long yTile = origyTile;
 
-        if (xTile >= limits.getMaxcol()) xTile = limits.getMaxcol() - 1;
-        if (yTile >= limits.getMaxrow()) yTile = limits.getMaxrow() - 1;
+        if (xTile >= limits.getMaxcol()) xTile = limits.getMaxcol();
+        if (yTile >= limits.getMaxrow()) yTile = limits.getMaxrow();
 
         if (xTile < limits.getMincol()) xTile = limits.getMincol();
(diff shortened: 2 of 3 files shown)


gillesB:
To explain the changed unit test a bit, I created a visualization.

image


In the first part of `WMTSTileServiceTest.findTilesInExtentUsingLimits()` the bigger yellow envelope is requested.
And the two tiles are returned.
In the second part the smaller yellow envelope is requested. Before the change the same two tiles were returned.
After the change only `wmts_2_1_2` is returned.

(The green envlope is the boundix box of the tile matrix set. The red envelope is the limit of the tile matix set link.)

gillesB:
I have not realized that I broke `WMTSTileFactory4326Test.testGetTileFromCoordinate()` as there was already one broken test in `main` (`WebMapTileServerTest.testGetTile())`.

aaime:
We're looking at the image and are a bit confused. The red "envelope" is... a reprojected rectangle? Maybe towards Lambert or Polar? And why are the top and bottom parts of that envelope not covered by tiles? We would expect to see two more tiles at the top and two at the bottom?

gillesB:
To avoid mistakes from my side I describe in the following in detail how I created the picture.

* I opened QGIS and then dragged the OSM Layer to the active layers.
* Then I switched the QGIS project CRS to EPSG:32633 
  * This is the reason why the OSM layer is "rounded"  
* Afterwards I created `Temporary Scratch Layers` to draw the different polygons in their actual CRS
  * This means the the red and yellow envelops were reprojected by QGIS
  * To create the different polygons I simply copy/pasted the coordinates from the source code of `WMTSTileServiceTest.findTilesInExtentUsingLimits()` respectively from the created Java objects while debugging.

The green layer is:

* `tileMatrixSet.setBbox(new CRSEnvelope("EPSG:32633", -300000, 6400000, 1300000, 8000000))`

The yellow layer is:

* `ReferencedEnvelope requestEnvelope =
                new ReferencedEnvelope(7.1, 12.1, 61.4, 64.6, CRS.decode("EPSG:4326", true));`
* `ReferencedEnvelope requestEnvelope2 =
                new ReferencedEnvelope(12.5, 15.3, 64.3, 66.0, CRS.decode("EPSG:4326", true))`

The red layer is:

```
createLayerWithLimits(
        limits,
        new CRSEnvelope(
                new ReferencedEnvelope(
                        4.4, 14.7, 57.8, 65.3, CRS.decode("CRS:84"))))
```

As for the tiles: 

I just draw `wmts_2_0_2` and `wmts_2_1_2` because those are the only tiles which are returned by both calls to `service.findTilesInExtent()` in the `main` branch.

I hope this clears things up a bit.

ianturton:
I'll check more thoroughly tomorrow but I suspect that this is due to
mixing projections and is probably due to curvature of the sides of your
box.

Ideally you need to provide a single box based request for us to check,
along with a list of expected tiles and observered tiles.

gillesB:
I think the SSCCE in the original Jira Issue is a better example. It uses a real world layer and also only one projection.

prusswan:
I may have encountered a similar issue (through geoserver), from the logs `wmts.client` *thinks* the zoom 0 tile is bounded by the yellow outline (when the actual tile is labelled as 0,0,-1):

image

PR #4383, opened 2023-07-06
GeoTools docbuild - support maintenance (again?) (28.x)

tbarsballe:
So, for some reason, geotools-28.x-docs [fails on the new version of  update_site.sh](https://build.geoserver.org/view/geotools/job/geotools-28.x-docs/92/console): 

```
+ ssh -oStrictHostKeyChecking=no -p 2223 geotools@geo-docs.geoserver.org cd  /var/www/geotools/docs/maintenance/../ && ./update_site.sh maintenance
Usage ./update_site.sh <stable|latest>
```

However, the code that appears to be the culprit hasn't been changed since at least 2011.

The old maintenance builds were [working fine passing in maintenance](https://build.geoserver.org/view/geotools/job/geotools-28.x-docs/91/console), so I can only surmise we've been using an even older version of the script, or one with uncommitted changes, for the past several years. 

Does anyone have more insight into this? If there's another version of update_site.sh lurking around, I'd like to make sure there aren't any other missing features hiding in it...

Strictly speaking this should be opened against master before being backported to 28.x, but this is the only branch that is actually affected right now - I can forwardport this instead after merging.

In the interim, I've also manually extracted the maintenance docs on the new server, so that they'll actually be there when we switch over servers.

### DIFF
--- a/docs/site/update_site.sh
+++ b/docs/site/update_site.sh
@@ -1,12 +1,7 @@
 #!/bin/bash
-#
-# Last update 2011-04-10 MB: 
-#   The zip files are now uploaded to ~/stable or ~/latest.
-#   Move xxx.zip to xxx.old.zip after unpacking.
-#
 
-if [ "$1" != "stable" ] && [ "$1" != "latest" ]; then
-  echo "Usage $0 <stable|latest>"
+if [ "$1" != "maintenance" ] && [ "$1" != "stable" ] && [ "$1" != "latest" ]; then
+  echo "Usage $0 <maintenance|stable|latest>"
   exit 1
 fi
 


docs/site/update_site.sh
```diff
@@ -1,12 +1,7 @@
...
-#
-# Last update 2011-04-10 MB: 
-#   The zip files are now uploaded to ~/stable or ~/latest.
-#   Move xxx.zip to xxx.old.zip after unpacking.
-#
 
-if [ "$1" != "stable" ] && [ "$1" != "latest" ]; then
-  echo "Usage $0 <stable|latest>"
+if [ "$1" != "maintenance" ] && [ "$1" != "stable" ] && [ "$1" != "latest" ]; then
+  echo "Usage $0 <maintenance|stable|latest>"
```
tbarsballe, line 4: Removing this whole block would also be a viable fix, in case we ever want to support other paths in the future

tbarsballe:
I've done a bit more digging, and found that indeed, the update_site.sh version in svn is different than the original one copied over to the geotools git, and there's a commit in there that dates to just after the geotools svn:git import, along with some uncommited chages.
```
------------------------------------------------------------------------
r37048 | jdeolive | 2011-05-02 17:32:46 -0700 (Mon, 02 May 2011) | 1 line

moving update_site.sh to site directory
------------------------------------------------------------------------
```

The actual changes is basically the same as what I've changed here, adding maintenance as a supported path.
```
tbarsballe@osgeo6:/home/geotools$ svn diff
Index: update_site.sh
===================================================================
--- update_site.sh	(revision 38852)
+++ update_site.sh	(working copy)
@@ -5,12 +5,12 @@
 #   Move xxx.zip to xxx.old.zip after unpacking.
 #
 
-if [ "$1" != "stable" ] && [ "$1" != "latest" ]; then
-  echo "Usage $0 <stable|latest>"
+if [ "$1" != "stable" ] && [ "$1" != "latest" ] && [ "$1" != "maintenance" ]; then
+  echo "Usage $0 <stable|maintenance|latest>"
   exit 1
 fi
 
-WWW=/osgeo/geotools/htdocs
+WWW=/var/www/geotools
 pushd $WWW/docs/$1 &&
 
 if [ -e ~/$1/userguide.zip ]; then
```

PR #4392, opened 2023-07-11
WIP: NetCDFImageReader: fix manual joining causing StringIndexOutOfBoundsException

ClundXIII:
In case of coverages being zero length, the following call would be made: `typeNames.substring(0, -1);`, causing out of bounds error.

Joining strings is typically handled by a StringJoiner which takes care of such edge cases.

# Checklist

### DIFF
--- a/modules/plugin/coverage-multidim/netcdf/src/main/java/org/geotools/imageio/netcdf/NetCDFImageReader.java
+++ b/modules/plugin/coverage-multidim/netcdf/src/main/java/org/geotools/imageio/netcdf/NetCDFImageReader.java
@@ -51,6 +51,7 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.StringJoiner;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.imageio.IIOException;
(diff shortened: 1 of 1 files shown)


aaime:
Unfortunately, we cannot consider the PR, unless the mandatory items are checked, rather than crossed out.
For a small change we can do without the CLA, but not without a Jira ticket and a test.

ClundXIII:
> Unfortunately, we cannot consider the PR, unless the mandatory items are checked, rather than crossed out.
> For a small change we can do without the CLA, but not without a Jira ticket and a test.

Hi,

sorry, for some reason I assumed the PR did no affect "core and extension modules". I blame the heat. If someone else wants to fix this on their behalf, I wouldn't mind to be honest, you can use the patch as you want without referencing me. The fix is very minimal (and also not complete, I'll add a WIP:).

ClundXIII:
The issue occurred inside https://geoserver.org/ , I traced it back to this dependency.

ClundXIII:
I can implement a short test using an example file if required.

aaime:
It is indeed required. Smallest possible file that can reproduce the issue (use one of the NetCDF utilties to cut down the grid to a very small one, e.g., 10x10 pixels, 1 time).

jodygarnett:
> I can implement a short test using an example file if required.

I am looking at open PRs ahead of 30.0 release; and this one is still missing test case.

ClundXIII:
I don't think I will find the time

PR #4400, opened 2023-07-19
Upgrade flatgeobuf to  3.26.1

bjornharrtell:
This patch release upgrade fixes a performance issue in spatial index search, see https://github.com/flatgeobuf/flatgeobuf/issues/274.


bjornharrtell:
@aaime would be great if label for backporting can be added here. :) And thanks for noticing this problem. I think I've experienced it too but didn't release there was something wrong in the code.

PR #4405, opened 2023-07-22
[GEOT-7404] Add SQLite Connection Timeout in case of hung GeoPackage connections

turingtestfail:
[GEOT-7404](https://osgeo-org.atlassian.net/browse/GEOT-7404) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

# Checklist

### DIFF
--- a/modules/plugin/geopkg/src/test/java/org/geotools/geopkg/GeoPkgDataStoreFactoryTest.java
+++ b/modules/plugin/geopkg/src/test/java/org/geotools/geopkg/GeoPkgDataStoreFactoryTest.java
@@ -16,13 +16,22 @@
  */
 package org.geotools.geopkg;
 
+import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 
 import java.io.File;
 import java.io.IOException;
 import java.io.Serializable;
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.PreparedStatement;
+import java.sql.SQLException;
+import java.sql.Statement;
 import java.util.HashMap;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


modules/plugin/geopkg/src/test/java/org/geotools/geopkg/GeoPkgDataStoreFactoryTest.java
```diff
@@ -54,7 +137,6 @@ public void testBaseDirectory() throws IOException {
         DataStore data = factory.createDataStore(map);
         data.createSchema(b.buildFeatureType());
         data.dispose();
-
-        assertTrue(new File(tmp.getRoot(), "foo.gpkg").exists());
+        return data;
```
aaime, line 140: "data" here is dead, it has been disposed of. It's also not used anywhere I can see.
So I'd switch the method to void.

aaime (changes requested):
I would rename the ticket title, this is just a GeoPackage change, even if we noticed the trouble in the context of vector mosaic. See also the comment below.

modules/plugin/geopkg/src/test/java/org/geotools/geopkg/GeoPkgDataStoreFactoryTest.java
turingtestfail: fixed

turingtestfail:
> I would rename the ticket title, this is just a GeoPackage change, even if we noticed the trouble in the context of vector mosaic. See also the comment below.

Fixed

aaime (approved):

PR #4407, opened 2023-07-24
[GEOT-7405] Add option to disable GetFeatureInfo transforming raster layers

sikeoka:
[GEOT-7405](https://osgeo-org.atlassian.net/browse/GEOT-7405) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

This PR adds a vendor option to SLD FeatureTypeStyle elements to disable applying rendering transformations to raster data for WMS GetFeatureInfo requests.

I would like to get this into the 28.5 release if possible since I am supporting customers still using Java 8.

  

# Checklist

### DIFF
--- a/modules/library/main/src/main/java/org/geotools/styling/FeatureTypeStyle.java
+++ b/modules/library/main/src/main/java/org/geotools/styling/FeatureTypeStyle.java
@@ -137,6 +137,16 @@ public interface FeatureTypeStyle extends org.opengis.style.FeatureTypeStyle {
     public static String SORT_BY_GROUP = "sortByGroup";
 
     /**
+     * Boolean value allowing to control whether rendering transformations should be evaluated when
+     * performing WMS GetFeatureInfo requests. This option applies only to transformations with a
+     * raster source (i.e., raster-to-raster and raster-to-vector). The default value can be
+     * configured by administrators in the global and workspace-specific WMS service settings
+     * (transformations will be evaluated by default) and this SLD vendor option can be used to
+     * override the service setting for a specific FeatureTypeStyle element within the SLD document.
+     */
+    public static String TRANSFORM_FEATURES = "transformFeatures";
+
+    /**
      * String value allowing to control whether an SLD element should be included when applying the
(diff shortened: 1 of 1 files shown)


sikeoka:
The failed check seems to be an unrelated connection error.

modules/library/main/src/main/java/org/geotools/styling/FeatureTypeStyle.java
```diff
@@ -136,6 +136,16 @@ public interface FeatureTypeStyle extends org.opengis.style.FeatureTypeStyle {
      */
     public static String SORT_BY_GROUP = "sortByGroup";
 
+    /**
```
aaime, line 139: The flag and its documentation make sense for GeoServer, but in the context of GeoTools they do not (and it's not used anywhere). Maybe best to move this bit to the GeoServer code. Also, features in GeoTools/GeoServer are vector data. Finally, the flag would activate on feature info only. Maybe TRANSFORM_FEATURE_INFO would be a better name.

aaime (changes requested):

modules/library/main/src/main/java/org/geotools/styling/FeatureTypeStyle.java
aaime: I believe all vendor options can be used in applications built on top of GeoTools, without having to be used in GeoServer to make sense (e.g. a desktop client), and have GeoTools code using them.

This one is a mere constant, with no usage, does not seem to belong here.

sikeoka: Would it be better to skip a GeoTools PR and do GeoServer only?

aaime: Yes, that's what I mean by "Maybe best to move this bit to the GeoServer code"

sikeoka:
Closing this PR. Everything is now in the GeoServer PR: https://github.com/geoserver/geoserver/pull/7041

PR #4413, opened 2023-08-03
Speed up evaluation of symbolizer properties during rendering

aaime:
See the two tickets for details:
* [Avoid needless type conversions in InterpolateFunction](https://osgeo-org.atlassian.net/browse/GEOT-7413) (pure optimization, no change in behavior, cannot add new tests, existing ones guard against regression)
* [Speed up style evaluation during rendering, by simplifying its expressions before execution](https://osgeo-org.atlassian.net/browse/GEOT-7414) (this one is also an optimization, but has testable changes in simplifier visitor)

### DIFF
--- a/modules/library/main/src/main/java/org/geotools/filter/function/IDFunction.java
+++ b/modules/library/main/src/main/java/org/geotools/filter/function/IDFunction.java
@@ -22,6 +22,7 @@ import org.geotools.filter.capability.FunctionNameImpl;
 import org.opengis.feature.Attribute;
 import org.opengis.feature.simple.SimpleFeature;
 import org.opengis.filter.capability.FunctionName;
+import org.opengis.filter.expression.VolatileFunction;
 
 /**
  * Allow access to the value of Feature.getID() as an expression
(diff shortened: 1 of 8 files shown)


modules/library/main/src/main/java/org/geotools/filter/visitor/DuplicatingFilterVisitor.java
```diff
@@ -339,14 +339,22 @@ public Object visit(NilExpression expression, Object extraData) {
     public Object visit(Add expression, Object extraData) {
         Expression expr1 = visit(expression.getExpression1(), extraData);
         Expression expr2 = visit(expression.getExpression2(), extraData);
-        return getFactory(extraData).add(expr1, expr2);
+        Add result = getFactory(extraData).add(expr1, expr2);
+        if (expr1 instanceof Literal && expr2 instanceof Literal) {
+            return getFactory(extraData).literal(result.evaluate(null));
```
ianturton, line 344: should a **duplicating** visitor be changing the expression? I'm happy for the simplifying visitor to do this but I think I'd expect to get back an identical filter with a duplication

aaime: Whoops, you're right, this could should have been in the SimplifyingFilterVisitor. I guess that one did not have an ovverride and the IDE brought me to this class instead. Thanks for spotting it, fixing.

dromagnoli (approved):

PR #4442, opened 2023-08-23
org.opengis removal: DirectPosition, GeneralEnvelope, and BoundingBox cleanup

jodygarnett:
This feeds into https://github.com/geotools/geotools/pull/4367 as a separate QA activity check


jodygarnett:
This work was combined into https://github.com/geotools/geotools/pull/4443 in order to give styling API Change a clear run at merging.

PR #4448, opened 2023-08-24
[GEOT-7287] Remove opengis: Refactor DirectPostion and Envelope2D classes

jodygarnett:
[GEOT-7287](https://osgeo-org.atlassian.net/browse/GEOT-7287) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

Feeds into https://github.com/geotools/geotools/pull/4367


ianturton:
OK, I've fixed the BinaryOp tests - someone messed up creating the bbox of the original test examples and Jody misunderstood what they had done.

jodygarnett:
I am having trouble with difference between main and remove-opengis-bounds after the change to ReferencedEnvelope2D:

- Images are different sizes, found 400x450 size, while expecting 400x451 size.
- Adding logging shows crop operations working with slightly different values:
- There was some complicated recursive code that is presently commented out, it may of been designed to detect when not enough change occurred after reproduction 

BEFORE:
```
,GridCoverage2D["world_0_360", GeneralEnvelope[(-180.9, -90.00000000000001), (-99.89999999999998, 89.99999999999999)], DefaultGeographicCRS["WGS 84"]]
│   RenderedSampleDimension("RED_BAND":[0.0 ... 0.0])
│     ‣ Category("No data":[0.0 ... 0.0])
│   RenderedSampleDimension("GREEN_BAND":[0.0 ... 0.0])
│     ‣ Category("No data":[0.0 ... 0.0])
│   RenderedSampleDimension("BLUE_BAND":[0.0 ... 0.0])
│     ‣ Category("No data":[0.0 ... 0.0])
└ Image=RenderedOp["Crop"]
```

AFTER:
```
,GridCoverage2D["world_0_360", GeneralBounds[(-180.9, -90.00000000000001), (-99.89999999999998, 90.0)], DefaultGeographicCRS["WGS 84"]]
│   RenderedSampleDimension("RED_BAND":[0.0 ... 0.0])
│     ‣ Category("No data":[0.0 ... 0.0])
│   RenderedSampleDimension("GREEN_BAND":[0.0 ... 0.0])
│     ‣ Category("No data":[0.0 ... 0.0])
│   RenderedSampleDimension("BLUE_BAND":[0.0 ... 0.0])
│     ‣ Category("No data":[0.0 ... 0.0])
└ Image=RenderedOp["Crop"]
```

aaime:
Definitely do not comment out the "complicated recursive code", it's most likely "carefully crafted to handle difficult cases".

aaime:
Looking at the downstream compile errors, it would seem some of the styling migration bits are missing

.github/workflows/db2_online.yml
```diff
@@ -1,6 +1,13 @@
 name: Db2 online tests
 
 on: [pull_request]
+  # trigger on PR, but only on main (dev), stable, and maintenance branches
```
aaime, line 4: We'll want to back out these changes before merging (either in remove-opengis, or when that is merged into main). We don't want to have one more bit to maintain at every major release.

modules/library/main/src/main/java/org/geotools/geometry/jts/ReferencedEnvelope.java
```diff
@@ -822,8 +968,25 @@ public static ReferencedEnvelope create(
...
+     * @param x
+     * @param y
+     * @param width
+     * @param height
+     * @return ReferencedEnvelope WGS84
+     */
+    public static ReferencedEnvelope envelope2D(
+            CoordinateReferenceSystem crs, double x, double y, double width, double height) {
+        return new ReferencedEnvelope(x, x + width, y, y + height, crs);
     }
```
aaime, line 988: Wondering, should this method be inlined and removed, or made into a constructor leveraging the different order of arguments? Seeing "ReferencedEnvelope.evenvelope2D" around the codebase is weird... and the name is long too.  Thinking out loud, if you want to keep the static method, what about "ReferencedEnvelope.rect(x,y,w,h,crs)`` since it's using the rectangle convention?

ianturton:
OK, I've fixed the BinaryOp tests - someone messed up creating the bbox of the original test examples and Jody misunderstood what they had done.

> Looking at the downstream compile errors, it would seem some of the styling migration bits are missing

If you comment out the pattern in the `moveClass` macro fileset it works - not sure why it isn't matching

.github/workflows/db2_online.yml
jodygarnett: Yes of course; I just needed faster feedback and dislike wasting resources

modules/library/main/src/main/java/org/geotools/geometry/jts/ReferencedEnvelope.java
jodygarnett: ReferencedEnvelope has both constructors (which are easy to misuse combing 2D data with 3D CRS) and factory methods which create ReferencedEnvelope or ReferencedEnvelope3D based on CRS provided. 

I will rename to `rect(crs,x,y,w,h)` - the order was provided so there is an "easy" search and replace to perform when migrating from Envelope2D. I am happy to deprecate the method, or if it is not useful at the end of the day it can be removed.

jodygarnett:
> Looking at the downstream compile errors, it would seem some of the styling migration bits are missing

I am having some trouble with the macro's not making their own file set identifiers. All the styling was based on `org.opengis.style` rather than individual things like `org.opengis.style.StyleLayer`. I am not sure why I cannot get it to work yet but am testing locally.

ianturton:
> > Looking at the downstream compile errors, it would seem some of the styling migration bits are missing
> 
> I am having some trouble with the macro's not making their own file set identifiers. All the styling was based on `org.opengis.style` rather than individual things like `org.opengis.style.StyleLayer`. I am not sure why I cannot get it to work yet but am testing locally.

I've fixed that I think

jodygarnett:
> I've fixed that I think

I fixed this also :)

Indeed  it looks like you pushed up a change to remove-opengis.xml which has some xml formatting changes I will see if I can avoid losing your work.

jodygarnett:
I am stopping for the day, The class GWC.java contains a lot of JTS Envelope use which causes conflicts with the refactoring script.

jodygarnett:
Resuming, @ianturton let us try to avoid conflicts I we can on this branch.

PR #4454, opened 2023-08-25
[GEOT-7428] Support of jsonPointer function for Oracle

Mike7311:
[GEOT-7428](https://osgeo-org.atlassian.net/browse/GEOT-7428) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

Added delegation of jsonPointer function unto Oracle’s JSON_VALUE function

# Checklist


mprins (approved):
lgtm

modules/library/main/src/main/java/org/geotools/filter/visitor/JsonPointerFilterSplittingVisitor.java
```diff
@@ -0,0 +1,76 @@
...
+            DuplicatingFilterVisitor duplicating = new DuplicatingFilterVisitor();
+            Function duplicated = (Function) expression.accept(duplicating, null);
+            // if constant can encode
+            Object result = param.evaluate(null);
+            FilterFactory2 ff = CommonFactoryFinder.getFilterFactory2();
+            // setting constant expression evaluated to literal
+            duplicated.getParameters().set(paramIdx, ff.literal(result));
+            return duplicated;
+        }
+        return expression;
```
aaime, line 74: The code looks inconsistent to me. ``constantParameterToLiteral`` is more general, checks if the first parameter is equivalent to a static expression, e.g., it's able to recognize that a ``env`` function is equivalent to a constant for the current request. ``supports`` instead is checking for a straight literal, would claim ``env`` is not supported and thus move the JSON function to the post filter, right?

modules/plugin/jdbc/jdbc-oracle/src/main/java/org/geotools/data/oracle/OracleFilterToSQL.java
```diff
@@ -235,11 +243,42 @@ public Object visit(Function function, Object extraData) {
...
+            }
         }
 
         return super.visit(function, extraData);
     }
 
+    private String encodeJsonPointer(Function jsonPointer, Object extraData) {
+        Expression json = getParameter(jsonPointer, 0, true);
+        Expression pointer = getParameter(jsonPointer, 1, true);
+        if (json instanceof PropertyName && pointer instanceof Literal) {
```
aaime, line 261: Here is also assumes it has to be a straight literal. Ah, another important thing, is when FilterSimplifier is applied in the JDBC code. If it's applied before the splitting, then you'll receive a Literal anyways, as the ``env(...)`` has already been replaced by a Literal in the simplifier.

modules/library/main/src/main/java/org/geotools/filter/visitor/JsonPointerFilterSplittingVisitor.java
aaime: Also, I suggest adding a test for the splitter, with supported and non-supported cases.

aaime (changes requested):
The PR needs to be rebased on top of main.
More feedback below.

modules/library/main/src/main/java/org/geotools/filter/visitor/JsonPointerFilterSplittingVisitor.java
Mike7311: > The code looks inconsistent to me. `constantParameterToLiteral` is more general, checks if the first parameter is equivalent to a static expression, e.g., it's able to recognize that a `env` function is equivalent to a constant for the current request. `supports` instead is checking for a straight literal, would claim `env` is not supported and thus move the JSON function to the post filter, right?

constantParameterToLiteral is called from the if block, which is checking the expression to be JsonPointerFunction in [JsonPointerFilterSplittingVisitor#visit](https://github.com/Mike7311/geotools/blob/7c9eb32894310b8322fb8e425480d7e6bcc64187/modules/library/main/src/main/java/org/geotools/filter/visitor/JsonPointerFilterSplittingVisitor.java#L40). So env will not pass this check.

I will add tests for supported/non-supported cases

Mike7311: Tests added

modules/plugin/jdbc/jdbc-oracle/src/main/java/org/geotools/data/oracle/OracleFilterToSQL.java
Mike7311: I think the reply to the previous comment also applies here.

PR #4465, opened 2023-08-31
[GEOT-7431] Support custom HTTP headers for WFS-Client and HTTP-Client

awaterme:
[GEOT-7431](https://osgeo-org.atlassian.net/browse/GEOT-7431) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

From [GEOT-7431](https://osgeo-org.atlassian.net/browse/GEOT-7431): 

GeoTools WFS-Client should support sending custom HTTP request headers for GET and POST requests.

This is useful for various scenarios, for example for sending custom authentication and authorization headers (bearer tokens etc).

This includes the following tasks:

    WFSClient inherits an headers field already, but there is no way to init -> provide further constructor

    WFSClient.headers must be passed on the HTTPClient. Done for GET requests -> implement for POST

    HTTPClient provides an API for additional HTTP headers for GET only -> implement accordingly for POST.

 

# Checklist

### DIFF
--- a/modules/library/main/src/main/java/org/geotools/data/ows/AbstractOpenWebService.java
+++ b/modules/library/main/src/main/java/org/geotools/data/ows/AbstractOpenWebService.java
@@ -360,7 +360,7 @@ public abstract class AbstractOpenWebService<C extends Capabilities, R extends O
         if (exception != null) {
             IOException e =
                     new IOException(
-                            "Could not establish version neogitation: " + exception.getMessage(),
+                            "Could not establish version negotiation: " + exception.getMessage(),
                             exception);
             throw e;
         } else {
@@ -448,7 +448,11 @@ public abstract class AbstractOpenWebService<C extends Capabilities, R extends O
                 ByteArrayOutputStream out = new ByteArrayOutputStream();
                 request.performPostOutput(out);
                 try (InputStream in = getStream(out)) {
-                    httpResponse = httpClient.post(finalURL, in, postContentType);
+                    if (headers == null) {
+                        httpResponse = httpClient.post(finalURL, in, postContentType);
+                    } else {
+                        httpResponse = httpClient.post(finalURL, in, postContentType, headers);
+                    }
                 }
             } else {
                 if (headers == null) {
--- a/modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/impl/WFSDataAccessFactory.java
+++ b/modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/impl/WFSDataAccessFactory.java
@@ -127,7 +127,7 @@ public class WFSDataAccessFactory implements DataAccessFactory {
     }
 
     /** Access with {@link WFSDataStoreFactory#getParametersInfo()  */
-    private static final WFSFactoryParam<?>[] parametersInfo = new WFSFactoryParam[23];
+    private static final WFSFactoryParam<?>[] parametersInfo = new WFSFactoryParam[24];
 
     private static final int GMLComplianceLevel = 2;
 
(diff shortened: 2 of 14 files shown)


modules/library/main/src/main/java/org/geotools/data/ows/AbstractOpenWebService.java
```diff
@@ -448,7 +448,11 @@ protected Response internalIssueRequest(Request request) throws IOException, Ser
                 ByteArrayOutputStream out = new ByteArrayOutputStream();
                 request.performPostOutput(out);
                 try (InputStream in = getStream(out)) {
-                    httpResponse = httpClient.post(finalURL, in, postContentType);
+                	if(headers == null) {
```
mprins, line 451: this is causing a QA failure because you used tabs, run `mvn spotless:apply` to fix

awaterme: Hi Mark,
yes, I actually fixed that yesterday already but it seems my second commit, pushed with "-force" did not update the PR. 
I'll have a look shortly. 
Thank you for jumping in!
Andreas

awaterme: Ok, fixed now...

mprins (approved):
LGTM, but I think a [Contribution Licence Agreement](https://docs.geotools.org/latest/developer/procedures/contribution_license.html) is required? unless @awaterme 's work is covered by some corporate licence agreement

awaterme:
So far my contributions were considered small enough for being accepted without. You dont think this is the case here?

mprins:
this PR changes 14 files; according to the https://github.com/geotools/geotools/blob/main/CONTRIBUTING.md guideline this is no longer a small contribution/change. 
Ultimately this decision is not up to me, but the PMC/PSC

aaime:
Definitely big enough that requires a CLA.

Could also use a review from  @roarbra

modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/impl/WFSDataAccessFactory.java
```diff
@@ -576,6 +576,22 @@ public T lookUp(final Map params) throws IOException {
...
+    static {
+        String name = "WFSDataStoreFactory:ADDTIONAL_HEADERS";
+        String title = "Set additional HTTP request headers.";
+        String description =
+                "The given HTTP headers will be set on HTTP requests in addition to "
+                        + "the regular GeoTools managed headers (like Accept-Enconding, Content-type, "
+                        + "User-Agent, Authorization). Provided values must not conflict with GeoTools "
+                        + "managed headers. Parameters must use String keys and values.";
+        parametersInfo[23] =
+                ADDTIONAL_HEADERS =
```
roarbra, line 591: Typo here

roarbra (commented):
I see the point of having it the same way as for the get function, but I would prefer to have this as as property on HTTPClient and deprecate the get function.

It would also be interesting to discuss what should happen if both username/password is set, and an Authorization is provided by the additional headers.

awaterme:
> Definitely big enough that requires a CLA. If we accepted by mistake PRs as large as this one we'll have to either get a CLA, or hunt them down and revert them I'm afraid.
> 
> Could also use a review from @roarbra

Hi Andrea, the CLA is on the way, you should have received a copy already.

awaterme:
Hi Roar,

thank you for taking time!

> I see the point of having it the same way as for the get function, but I would prefer to have this as as property on HTTPClient and deprecate the get function.

I think the current solutions' advantage (per-request headers) is flexibility. For example in an scenario where frontend users autorization tokens have to be passed to a cascaded WFS this would be much easier with the current approach, which only one HttpClient instance.
Looking at Apache Commons HTTP Client as an experienced and mature role model they support both: Default headers bound to a client instance and additionally headers on a per-request basis. Furthermore it might be the case that an HTTPClient should send these headers only to certain endpoints, which really require confidental token.
In conclusion I would say: For the low-level HTTPClient per-instance headers would be too limited. Additional default headers on (per-client) might be useful, but as they are not yet required I would prefer to keep it simple.

> It would also be interesting to discuss what should happen if both username/password is set, and an Authorization is provided by the additional headers.

Yes, I agree. In my opinion, it is up to the user to avoid such conflicts. This is outlined in the commend for WFSDataAccessFactory#ADDTIONAL_HEADERS already:

The given HTTP headers will be set on HTTP requests in addition to 
the regular GeoTools managed headers (like Accept-Enconding, Content-type, 
User-Agent, Authorization). Provided values must not conflict with GeoTools 
managed headers.

Additionally I can imaging to log a warning message in case of conflicts. But I decided to let GeoTools managed headers to win because I assume the GT knows better what to send. Also as the name implies it's intended for *additional* headers.

Best regards, Andreas

modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/impl/WFSDataAccessFactory.java
awaterme: Ups, thank you for finding!

awaterme: Fixed now. Dont know if I am expected to resolve this now...

roarbra:
Then it's fine by me.

It's very good that it will fix this discrepancy between GET and POST request.

roarbra (approved):

aaime:
Confirming CLA received, and merging.

PR #4134, opened 2022-11-30
[GEOT-7288] upgrade javax.measure version to 2.1.3

AlbanSeurat:
[GEOT-7288](https://osgeo-org.atlassian.net/browse/GEOT-7288) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

Upgrade the javax.measure library
  

# Checklist

- [X] I have read the [contribution guidelines](https://github.com/geotools/geotools/blob/main/CONTRIBUTING.md).
- [ ] I have sent a [Contribution Licence Agreement](https://docs.geotools.org/latest/developer/procedures/contribution_license.html) (not required for small changes, e.g., fixing typos in documentation).
- [X] First PR targets the `main` branch (backports managed later; ignore for branch specific issues).
- [X] Avoid [Java 9+ split packages](http://tutorials.jenkov.com/java/modules.html#split-packages-not-allowed).
- [x] All the build checks are green ([see automated QA checks](https://docs.geotools.org/latest/developer/conventions/code/qa.html)).

For core and extension modules:

- [X] New unit tests have been added covering the changes.
- [x] [Documentation](https://github.com/geotools/geotools/tree/main/docs) has been updated (if change is visible to end users).
- [ ] There is an issue in [GeoTools Jira](https://osgeo-org.atlassian.net/projects/GEOT) (except for changes not visible to end users). 
- [ ] Commit message(s) must be in the form ``[GEOT-XYZW] Title of the Jira ticket``.
- [X] Bug fixes and small new features are presented as a single commit.
- [X] The commit targets a single objective (if multiple focuses cannot be avoided, each one is in its own commit, and has a separate ticket describing it).

PR #3573, opened 2021-07-08
[GEOT-6709] Fix crash caused by reflection use violating Java module system rules

soc:
[GEOT-6709](https://osgeo-org.atlassian.net/browse/GEOT-6709) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

GeoTools v24 fails with ExceptionInInitializerError caused by an InaccessibleObjectException trying to access tech.units.indriya internals

- [x] Confirm you have read the [contribution guidelines](https://github.com/geotools/geotools/blob/master/CONTRIBUTING.md) 
- [x] You have sent a Contribution Licence Agreement (CLA) as necessary (not required for small changes, e.g., fixing typos in documentation)
- [x] Make sure the first PR targets the master branch, eventual backports will be managed later. This can be ignored if the PR is fixing an issue that only happens in a specific branch, but not in newer ones.
- [ ] The changes are not breaking the build in downstream projects using SNAPSHOT dependencies, GeoWebCache and GeoServer (there is an automatic PR check verifying this, check this when it turns green).

The following are required only for core and extension modules (they are welcomed, but not required, for unsupported modules):
- [x] There is an issue in [Jira](https://osgeo-org.atlassian.net/projects/GEOT) describing the bug/task/new feature (a notable exemptions is, changes not visible to end users). The ticket is for the GeoTools project, if the issue was found elsewhere it's a good practice to link to the origin ticket/issue.
- [x] The pull request contains changes related to a single objective. If multiple focuses cannot be avoided, each one is in its own commit and has a separate ticket describing it.
- [x] PR for bug fixes and small new features are presented as a single commit
- [x] Commit message(s) must be in the form "[GEOT-XYZW] Title of the Jira ticket"
- [x] New unit tests have been added covering the changes
- [x] This PR passes all existing unit tests (test results will be reported by travis-ci after opening this PR)
- [x] This PR passes the [QA checks](https://docs.geotools.org/latest/developer/conventions/code/qa.html) (QA checks results will be reported by travis-ci after opening this PR)
- [x] Documentation has been updated accordingly.

PR #4467, opened 2023-09-05
GEOT-7422 Update javax.measure.unit to version 2.2 and replace Metric…

neilireson:
[GEOT-7422](https://osgeo-org.atlassian.net/browse/GEOT-7422) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

PR for GEOT-7422

Update javax.measure.unit to version 2.2 and replace MetricPrefix.DEKA (American spelling) with MetricPrefix.DECA (International spelling)

### DIFF
--- a/modules/library/metadata/src/main/java/org/geotools/measure/PrefixDefinitions.java
+++ b/modules/library/metadata/src/main/java/org/geotools/measure/PrefixDefinitions.java
@@ -20,8 +20,8 @@ import static java.util.Arrays.asList;
 import static java.util.Collections.unmodifiableList;
 import static javax.measure.MetricPrefix.ATTO;
 import static javax.measure.MetricPrefix.CENTI;
+import static javax.measure.MetricPrefix.DECA;
 import static javax.measure.MetricPrefix.DECI;
-import static javax.measure.MetricPrefix.DEKA;
 import static javax.measure.MetricPrefix.EXA;
 import static javax.measure.MetricPrefix.FEMTO;
 import static javax.measure.MetricPrefix.GIGA;
@@ -59,7 +59,7 @@ public final class PrefixDefinitions {
                             PrefixDefinition.of(MEGA),
                             PrefixDefinition.of(KILO),
                             PrefixDefinition.of(HECTO),
-                            PrefixDefinition.of(DEKA),
+                            PrefixDefinition.of(DECA),
                             PrefixDefinition.of(DECI),
                             PrefixDefinition.of(CENTI),
                             PrefixDefinition.of(MILLI),
@@ -89,7 +89,7 @@ public final class PrefixDefinitions {
                             PrefixDefinition.of(MEGA),
                             // no kilo
                             PrefixDefinition.of(HECTO),
-                            PrefixDefinition.of(DEKA),
+                            PrefixDefinition.of(DECA),
                             PrefixDefinition.of(DECI),
                             PrefixDefinition.of(CENTI),
                             PrefixDefinition.of(MILLI),
--- a/pom.xml
+++ b/pom.xml
@@ -573,7 +573,7 @@
       <dependency>
         <groupId>javax.measure</groupId>
         <artifactId>unit-api</artifactId>
-        <version>2.1.3</version>
+        <version>2.2</version>
       </dependency>
 
       <!-- Java Advanced Imaging (JAI) -->


neilireson:
The PR fails some UnitFormatterTest tests. However it’s unfathomable to me what these tests are doing. Something to do with an indriya bug, however there’s no documentation of what it’s doing. Perhaps using javax.measure.unit 2.2 changes/fixes this bug in some way?

neilireson:
I'll close the request until the failing tests are understood/fixed

ianturton:
Closing the PR will lead to it being forgotten. From a quick look at the git blame it looks like this is related to 
"GeoTools fails with ExceptionInInitializerError caused by an InaccessibleObjectException trying to access tech.units.indriya internals" - and it looks like it is checking that none of those internals are available. There now seem to be more units than expected so may be we can relax this test to just check that all the units in `indriyaBug` are in `unitsOnlyInOldWithoutBugs` rather than looking for equality?

neilireson:
OK, that makes sense. I did notice that "unitsOnlyInOld" had gone from 11 (when using javax.measure.unit 2.1.2) to 131 (using 2.2).

aaime:
@soc can you help us here? The tests in question have been added in this PR: https://github.com/geotools/geotools/pull/3573

soc:
See https://github.com/geotools/geotools/pull/4134#issuecomment-1519139420 and https://github.com/geotools/geotools/pull/4134#issuecomment-1519910360.

The test in question was changed in https://github.com/geotools/geotools/pull/4134/files#diff-c4d4bf8cddf97853a8ae37bb2abf14615946fc3d681f52748e6012ca90bdb835 and I have the impression this was done "to make it work" without trying to understand what it did.

soc:
I don't want to make your day worse than necessary, but I also don't think that https://github.com/geotools/geotools/pull/4134/files#diff-35c8f0438eede0e71cd5384ee2055daf02b91a7e51ea0b2f37a8598369f3e5dfR183 is sufficient according to the BSD-3 license.

jodygarnett:
> I don't want to make your day worse than necessary, but I also don't think that https://github.com/geotools/geotools/pull/4134/files#diff-35c8f0438eede0e71cd5384ee2055daf02b91a7e51ea0b2f37a8598369f3e5dfR183 is sufficient according to the BSD-3 license.

Not hard to fix, our policy is to stick the header at the top of the file: https://docs.geotools.org/latest/developer/conventions/code/style.html

gtatarsky-IB:
Hello—does this PR need help? I am trying to use GeoTools with the stable version 30.2 and javax.measure with 2.2—this leads to the problem, DEKA vs DECA, in the geoTools.measure.prefixdefinitions. I want to keep using GeoTools, so I'm just wondering why this went stale and was closed? I'd be happy to contribute.

soc:
@gtatarsky-IB Solving this will require making some decisions that nobody wants to make, because that would require work that nobody wants to do:

a) forking/reimplementing more parts of unit of measure libraries
b) giving up on certain features that GeoTools provided for a long time

In the end we all have better things to do than dealing with problems that have no reason to exist in the first place.

Though putting down the minimal amount of effort to keep things going for another minor release is what put GeoTools into this situation.

neilireson:
> Hello—does this PR need help? I am trying to use GeoTools with the stable version 30.2 and javax.measure with 2.2—this leads to the problem, DEKA vs DECA, in the geoTools.measure.prefixdefinitions. I want to keep using GeoTools, so I'm just wondering why this went stale and was closed? I'd be happy to contribute.

I'm not a developer or contributor to Geotools, just a user. However, I am a developer and very much a pragmatist, For me this issue is an annoyance but not a major one. My niaive understanding is this...

Updating to version 2.2 causes a test to fail but no-one (fully) understands what the test is for and whether it's valid.

There are three solutions:
1) Do nothing - accept that the upgrade to 2.2 is not possible due to the test failure and people (like you and I) who are impacted will have to find a work around, in my case maven dependency management.
  PROS - requires no work 
  CONS - the issue will not go away and will potentially impact an increasing number of people in the future
2) Remove the test and see if this has any negative impact
  PROS - requires virtually no work
  CONS - I have know idea, as I don't know why the test exists or the reprecussions of it breaking
3) Fix the test so it passes with version 2.2, as suggested by @ianturton 
  PROS - this is the generally accepted course of action in these cases
  CONS - just a pragmatic ("sticking plaster") fix and not addressing the underlying issue (@soc)

Personally I would say if the Geotool team think 3) is an acceptible solution and you have the desire and skills to do it then go at it and submit a fix, and I for one will be grateful.

soc:
Look, I really don't want to argue with you and rather spend my time on holiday without having to deal with this; and to clarify, I'm also not a member of the GeoTools project.

None of the three things you propose are even close to solutions, all of that is

> putting down the minimal amount of effort to keep things going for another minor release

I described above. That's not your fault.

It's the outcome of wanting to make "numbers go up" in terms of dependency management without understanding why people wrote the code they wrote and breaking things in the progress.

It annoys me personally, because [I put the time in to make things work correctly](https://github.com/geotools/geotools/pull/3573), not because I'm a perfectionist or anything, but because sitting down and fixing it was the only way forward I saw from a pragmatic perspective.

In the end I lament the enormous amount of wasted time for everyone involved.  
The author of the unit library is solely responsible for this.

soc:
For all I care feel free to delete the tests, they lost any meaning with the changes in https://github.com/geotools/geotools/pull/4134 anyway.

And if nobody complained since then, maybe nobody cares about this feature anyway.

gtatarsky-IB:
Thanks Soc and Neil for the replies! I honestly wasn't expecting to hear back so quickly if at all and was prepared to fork–I am going to dive into this PR, the intention of the test, and attempt to create a meaningful and functional commit to allow geotools to be able to manage this dependency again. I am not one for "making the numbers go up," agreed, Soc–that's shortsighted–but in my use case, as a microservice engineer for the national weather service, we have other libraries loading the newer version of measure and would like to avoid multiple-version dependency issues (and as we do not maintain these other libraries - they are a different federal team's problem - we cannot just downgrade). So, I'm going to see what I can do here, but avoid a quick fix, and try to understand that test or alternatively create a more meaningful and documented test. Geotools is a great library and it is incredibly important to us!

gtatarsky-IB:
@soc I am also going to review the Pr's you've submitted and attached in your comment so I can make sure to account for your intentions and work as well.

soc:
Though I'd rather do anything else than deal with this topic ever again, let's try to get you up to speed:

- GeoTools needs multiple distinct unit formatting instances for reading in various file formats. 
- The unit library *never* supported this use-case:
  - there was only one formatting instance that had various amounts of units registered, depending on which class initializers had already run at the point of inspection
  - most importantly `unit.toString()` depends on that hidden, mutable and global state
- GeoTools' previous workaround (copious amounts of reflection) stopped working under the Java Module system (and was never that correct to start with, due to the problematic `toString()` calls spread over the code-base)
  - that's what https://osgeo-org.atlassian.net/browse/GEOT-6709 was about
- my approach to fix this was to re-implement unit formatter initialization and to coax a *fresh* unit formatter instance out of the library through copious amounts of subclassing and forwarding (see code)
  - *fresh* in this case means an instance that hasn't been configured by whatever JAR file ended up on the classpath
  - that's what https://github.com/geotools/geotools/pull/3573/files was about
  - the `UnitFormatterTest` at this point documented the difference between the unit library's spagetti code used to initialize the instance vs. my correct and consistent implementation
- then v2.1.3 came around and it turned out that the author of the unit library had incompatibly removed the parts of the library that allowed GeoTools to get its own unit formatter instances out of it
  - the update in https://github.com/geotools/geotools/pull/4134 then made things worse by using unit library methods that did not return a *fresh* unit formatter instance, but an *already initialized* instance
  - attempts of addressing this with the unit library author failed (see https://github.com/geotools/geotools/pull/4134)

There might be some things I forgot, some inaccuracies, some missing details; please feel free to ask.

The fundamental issue is that people are using this `javax.` stuff under the assumption that it is some kind of official, professionally developed and maintained library, and the reality is anything but that.

gtatarsky-IB:
This is very helpful, and I appreciate you entering this world of unit pain again to help me out. I have started diving into the code and am seeing exactly what you're talking about; these `UnitFormatterTest` tests are very clever and useful as a check/balance against indriya and javax (the problem is immediately apparent with the initialized format in the current geotools vs yours in 2021 ), and I will be using a similar structure in my project as we have the same unfortunate need to use these unit libs.

The changes made to the DefaultFormat and the requirement of using init and getting the already init formatter is terribly irritating. Reading into the core code of the javax stuff, I am kinda shocked at how...bleh...a lot of the code is.

I will try to limit any questions I have, but I appreciate the offer and the thorough explanation. I've started with a rollback to your original code and am going to pursue an upgrade chain myself so I don't have to work with the code merged in 7288

soc:
@gtatarsky-IB One approach that you could try is to copy over the formatting code from `SimpleUnitFormatter` and combine it with the already existing initialization code.

As long as you never call `Unit#toString` you should be largely safe from that insanity, while avoiding having to copy/fork that stuff too.

soc:
@gtatarsky-IB Have a look at https://github.com/geotools/geotools/compare/main...soc:geotools:topic/unit-formatter.

I can't really check how well this works due to project import failing with

```
Could not find artifact org.geotools:docs:pom:32-SNAPSHOT in osgeo-snapshot (https://repo.osgeo.org/repository/snapshot/)
```

and compilation failing with

```
java: cannot find symbol
  symbol:   class Token
  location: package org.geotools.filter.text.cql_2.parsers
```

in IntelliJ when operating from a clean checkout. :shrug:

soc:
unit


Going this far, it probably makes sense to also
- make `NetCDFUnitFormat` less of a special case
- drop the mutating methods (`label`, `alias`) on `BaseUnitFormatter`

jodygarnett:
I would love to see this updated!

soc:
@jodygarnett The changes reduce the surface area to unit library stuff.  
Though there are bits and pieces left, they do not provide much attack area for unit lib authors to mess with GeoTools things.

soc:
@neilireson @gtatarsky-IB https://github.com/geotools/geotools/pull/4690

soc:
```
 Error:  /home/runner/geoserver/src/extension/netcdf/src/main/java/org/geoserver/netcdf/NetCDFUnitsConfigurator.java:[43,33] non-static method setAliases(java.util.Map<java.lang.String,java.lang.String>) cannot be referenced from a static context
Error:  /home/runner/geoserver/src/extension/netcdf/src/main/java/org/geoserver/netcdf/NetCDFUnitsConfigurator.java:[54,33] non-static method setReplacements(java.util.Map<java.lang.String,java.lang.String>) cannot be referenced from a static context
Error:  /home/runner/geoserver/src/extension/netcdf/src/main/java/org/geoserver/netcdf/NetCDFUnitsConfigurator.java:[71,40] loadPropertiesOrdered(java.io.InputStream) has private access in org.geotools.imageio.netcdf.NetCDFUnitFormat
```

is kinda interesting, because it's a first case to check whether the API changes to `NetCDFUnitFormat` are workable in practive.

PR #4475, opened 2023-09-16
[GEOT-7437] Avoid use of yield keyword for JDK 17 compatibility

bradh:
[GEOT-7437](https://osgeo-org.atlassian.net/browse/GEOT-7437) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

JEP 361 (Switch expression - https://openjdk.org/jeps/361) makes `yield` into a restricted identifier.

The JEP says:

>  If there is a unary method yield in scope, then the expression yield(x) would be ambiguous (could be either a method call, or a yield statement whose operand is a parenthesized expression), and this ambiguity is resolved in favor of the yield statement. If the method invocation is preferred then the method should be qualified, with this for an instance method or the class name for a static method.

There are two places (single class) where that causes a problem in GeoTools for yield() , and this PR adds the class name for the static method.

# Checklist

- [N/A] New unit tests have been added covering the changes.
- [N/A] [Documentation](https://github.com/geotools/geotools/tree/main/docs) has been updated (if change is visible to end users).

### DIFF
--- a/modules/library/xml/src/main/java/org/geotools/xml/gml/FCBuffer.java
+++ b/modules/library/xml/src/main/java/org/geotools/xml/gml/FCBuffer.java
@@ -176,7 +176,7 @@ public class FCBuffer extends Thread implements FeatureReader<SimpleFeatureType,
         Date d = new Date(Calendar.getInstance().getTimeInMillis() + timeout);
 
         while ((ft == null) && ((state != FINISH) && (state != STOP))) {
-            yield(); // let the parser run ... this is being called from
+            Thread.yield(); // let the parser run ... this is being called from
 
             if (d.before(Calendar.getInstance().getTime())) {
                 exception = new SAXException("Timeout");
@@ -316,7 +316,7 @@ public class FCBuffer extends Thread implements FeatureReader<SimpleFeatureType,
         } catch (SAXException e) {
             exception = e;
             state = STOP;
-            yield();
+            Thread.yield();
         }
     }
 


mprins (approved):

aaime:
Looks good to me. I have a question: we have been building with JDK 17 for a while, and I've just verified the build is actually using it (one never knows), but it's not breaking.

Ah... answering myself, there is a warning in the build output:

```
2023-09-15T20:15:59.0521146Z [WARNING] system modules path not set in conjunction with -source 11
2023-09-15T20:15:59.0524475Z [WARNING] /home/runner/work/geotools/geotools/modules/library/xml/src/main/java/org/geotools/xml/gml/FCBuffer.java:[179,13] 'yield' may become a restricted identifier in a future release
2023-09-15T20:15:59.0527260Z   (to invoke a method called yield, qualify the yield with a receiver or type name)
2023-09-15T20:15:59.0530263Z [WARNING] /home/runner/work/geotools/geotools/modules/library/xml/src/main/java/org/geotools/xml/gml/FCBuffer.java:[319,13] 'yield' may become a restricted identifier in a future release
2023-09-15T20:15:59.0533350Z   (to invoke a method called yield, qualify the yield with a receiver or type name)
```

bradh:
It becomes an error if using source=17 (or maybe source=17 and target=17, I only check that case).

PR #4497, opened 2023-09-29
[GEOT-7458] JDBCDataStore: Unique visitor not always compatible with …

turingtestfail:
…sortBy

  

# Checklist


mprins (approved):

aaime:
The downstream build issue should have been resolved, could you rebase this one? Changes in JDBC could legitimately affect some downstream tests (e..g, WFS ones based on H2)

modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
```diff
@@ -1430,6 +1430,14 @@ protected Object getAggregateValue(
             }
         }
 
+        // SortBy is incompatible with UniqueVisitor in JDBC due to issues when sort
+        // attribute does not match the unique attribute
```
aaime, line 1434:
```suggestion
        // In the SQL standard distinct and order by can work only if all order by attributes also show up in the select
        // Given the Unique visitor makes no promise regarding sorting, it's easier to just remove the sort
```

modules/library/jdbc/src/test/java/org/geotools/jdbc/JDBCAggregateFunctionOnlineTest.java
```diff
@@ -415,6 +415,24 @@ public void testUniqueWithLimitOnVisitor() throws Exception {
...
+
+        FilterFactory ff = dataStore.getFilterFactory();
+        PropertyName p = ff.property(aname("stringProperty"));
+
+        UniqueVisitor v = new MyUniqueVisitor(p);
+        v.setStartIndex(0);
+        Query q = new Query(tname("ft1"));
+        PropertyName pNonMatching = ff.property(aname("doubleProperty"));
+        q.setSortBy(new SortByImpl(pNonMatching, ASCENDING));
+        dataStore.getFeatureSource(tname("ft1")).accepts(q, v, null);
```
aaime, line 430:
```suggestion
        // used to fail with an exception, because stringProperty was in order by, but not in select
        dataStore.getFeatureSource(tname("ft1")).accepts(q, v, null);
```

modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
```diff
@@ -3153,6 +3161,7 @@ protected String selectGeometryAssociationSQL(String fid, String gid, String gna
 
         return sql.toString();
     }
+
```
aaime, line 3164: What's happening with these single line changes? Maybe newlines difference...

aaime (approved):
Looks good, needs a rebase, and see suggestions for comments.

modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java
turingtestfail: Fixed

modules/library/jdbc/src/test/java/org/geotools/jdbc/JDBCAggregateFunctionOnlineTest.java
turingtestfail: Fixed

PR #4509, opened 2023-10-02
[GEOT-7467] ESRI Shapefile read DBF-only data

bestrauss:
[GEOT-7467](https://osgeo-org.atlassian.net/browse/GEOT-7467) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

Some ESRI Shapefile data providers provide data without geometries. `*.dbf` files may come without corresponding `*.shp` etc. files. One such provider is [HERE/NavStreets](https:/www.here.com/platform/map-data). Support reading of data which comes without geometries, furnishing features without geometries.

# Checklist


aaime:
Thinking out loud... if there are no geometries, why pretend having an empty one?
JDBC data stores in the same situation just report the alphanumeric attributes, without any extra geometry.

bestrauss:
@aaime 
Whether empty geometries, zero geometries, or no geometries at all, all alternatives are fine for me. It's just been more easy and less changes to use empty geometries or zero geometries. I see your point, given the behaviour of JDBC data stores. Do you want me to copy that behaviour?

bestrauss:
@aaime 
I changed the code to have the DataStore furnish features without geometry-attribute and without geometries.

bestrauss:
@aaime
No geometries, not even empty or null ones, Checklist all ok, all checks passed, no conflicts. I'd say this is ready to be merged. What do you say?

modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/ShapefileFeatureSource.java
```diff
@@ -345,7 +346,17 @@ protected FeatureReader<SimpleFeatureType, SimpleFeature> getReaderInternal(Quer
...
+        ShapefileReader shapeReader = null;
+        try {
+            @SuppressWarnings("PMD.CloseResource") // managed as a field of the return value
+            final ShapefileReader sr =
+                    shpManager.openShapeReader(geometryFactory, goodRecs != null);
+            shapeReader = sr;
+        } catch (final FileNotFoundException e) {
+            final String format = "Ignoring missing shp-file and moving on: %s";
+            LOGGER.fine(() -> String.format(format, e.getMessage()));
+        }
```
aaime, line 358: Ouch, relying on exceptions for predictable situations is a bad habit. Is there no way to check if the shapefile is present instead?

bestrauss: @aaime 
In case the files are local, the existence of the shp-file can be tested. I implemented this test. If the files are not local, there is no way to avoid catching the FileNotFoundException.

aaime:
Looks good, merging.

PR #4513, opened 2023-10-07
[GEOT-7434] ElasticSearch throws repeated point exceptions on small bbox filters

turingtestfail:
[GEOT-7434](https://osgeo-org.atlassian.net/browse/GEOT-7434) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

…box filters

  

# Checklist

### DIFF
--- a/modules/unsupported/elasticsearch/src/main/java/org/geotools/data/elasticsearch/FilterToElastic.java
+++ b/modules/unsupported/elasticsearch/src/main/java/org/geotools/data/elasticsearch/FilterToElastic.java
@@ -19,6 +19,7 @@ package org.geotools.data.elasticsearch;
 import static org.geotools.process.elasticsearch.ElasticBucketVisitor.ES_AGGREGATE_BUCKET;
 import static org.geotools.util.factory.Hints.VIRTUAL_TABLE_PARAMETERS;
 
+import com.bedatadriven.jackson.datatype.jts.JtsModule;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.ObjectReader;
 import com.google.common.collect.ImmutableList;
@@ -31,6 +32,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.logging.Level;
 import java.util.logging.Logger;
 import org.geotools.api.data.Query;
 import org.geotools.api.feature.simple.SimpleFeatureType;
@@ -107,6 +109,7 @@ import org.geotools.util.Converters;
 import org.geotools.util.factory.Hints;
 import org.geotools.util.logging.Logging;
 import org.locationtech.jts.geom.CoordinateSequence;
+import org.locationtech.jts.geom.Envelope;
 import org.locationtech.jts.geom.Geometry;
 import org.locationtech.jts.geom.GeometryFactory;
 import org.locationtech.jts.geom.LinearRing;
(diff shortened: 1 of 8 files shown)


jodygarnett:
@turingtestfail not sure how to get the actions to re-runt

modules/unsupported/elasticsearch/src/main/java/org/geotools/data/elasticsearch/FilterToElastic.java
```diff
@@ -1217,11 +1220,26 @@ private void visitLiteralGeometry(Literal expression) throws IOException {
...
+    private int getMaxDecimalsForGeometry(Geometry currentGeometry) {
+        Envelope envelope = currentGeometry.getEnvelopeInternal();
+        double min = Math.min(Math.abs(envelope.getWidth()), Math.abs(envelope.getHeight()));
+        if (min == 0) {
+            LOGGER.log(
+                    Level.WARNING,
+                    "BBox Geometry has no width or height, it is either a point or a line.");
+            return 0;
+        } else if (min >= 0.000001) {
+            return JtsModule.DEFAULT_MAX_DECIMALS;
```
aaime, line 1237: Did you try zooming in a lot and doing a GetFeatureInfo? Eventually using only 6 decimals will not be enough and the request will fail. The idea was to adjust the decimals so that one would not get an error, at least until we don't reach the full precision at which we can represent doubles (16 decimals)

modules/unsupported/elasticsearch/src/main/java/org/geotools/data/elasticsearch/FilterToElastic.java
```diff
@@ -1217,11 +1220,26 @@ private void visitLiteralGeometry(Literal expression) throws IOException {
...
+        } else {
+            return Math.abs((int) Math.floor(Math.log10(min)) + 1);
+        }
```
aaime, line 1240: When one zooms out a lot the numbers start to get silly in the other direction, e.g., longitudes at "10000", for which this algorithm would state something like... 4 decimals? If the log is negative, use it as decimals, but if it's positive, I'd say settle on zero decimals?

This method could be taking an Envelope, be static, and be tested against various situations in JUnit, making sure it works well in a variety of situations, since it's the core of the fix.

modules/unsupported/geojson-core/src/main/java/org/geotools/data/geojson/GeoJSONWriter.java
```diff
@@ -80,7 +80,7 @@ public class GeoJSONWriter implements AutoCloseable {
             FastDateFormat.getInstance(DEFAULT_DATE_FORMAT, DEFAULT_TIME_ZONE);
 
     /** Maximum number of decimal places (see https://xkcd.com/2170/ before changing it) */
-    private int maxDecimals = 4;
+    private int maxDecimals = 6;
```
aaime, line 83: While we're at it, can we use JtsModule.DEFAULT_MAX_DECIMAL here?

aaime (changes requested):
Overall the change is what I was looking forward to. The logic to determine the number of decimals could use a few tweaks, see below.

modules/unsupported/elasticsearch/src/main/java/org/geotools/data/elasticsearch/FilterToElastic.java
turingtestfail: fixed

modules/unsupported/geojson-core/src/main/java/org/geotools/data/geojson/GeoJSONWriter.java
turingtestfail: fixed

jodygarnett:
@aaime flagged this fix as important for 30.0 release, labeling for backport.

PR #4557, opened 2023-11-17
Upgrade flatgeobuf to 3.26.2

bjornharrtell:
This patch release upgrade fixes an issue that spatial index search on FlatGeobuf produces a result with wrong feature ids.  It also adds a test cases that reproduces the issue and verifies the fix.

See https://github.com/flatgeobuf/flatgeobuf/pull/334.

### DIFF
--- a/modules/unsupported/flatgeobuf/pom.xml
+++ b/modules/unsupported/flatgeobuf/pom.xml
@@ -25,7 +25,7 @@
     <dependency>
       <groupId>org.wololo</groupId>
       <artifactId>flatgeobuf</artifactId>
-      <version>3.26.1</version>
+      <version>3.26.2</version>
     </dependency>
     <dependency>
       <groupId>org.geotools</groupId>
--- a/modules/unsupported/flatgeobuf/src/test/java/org/geotools/data/flatgeobuf/FlatGeobufDataStoreTest.java
+++ b/modules/unsupported/flatgeobuf/src/test/java/org/geotools/data/flatgeobuf/FlatGeobufDataStoreTest.java
@@ -881,7 +881,7 @@ public class FlatGeobufDataStoreTest {
             SimpleFeature f2 = it.next();
             assertEquals("countries.48", f2.getID());
             boolean hasNext = it.hasNext();
-            assertEquals(false, hasNext);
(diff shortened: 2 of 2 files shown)


bjornharrtell:
@aaime this is unfortunately a pretty serious issue and was due to bad port of C++ to Java since way back but never noticed because data is otherwise correct so if feature id is not used it might not be noticed. Anyway it was recently found in the wild in use of GeoServer WFS with FlatGeobuf datasource causing complete confusion when trying to use the feature ids from a spatial filtered GetFeature request fx. in a later feature id filter request.

So I would suggest this to be backported to stable release so that it can reach an eventual GeoServer patch release.

aaime:
Yes, let's merge as soon as possible. 

Currently, I see a bunch of Github action failures:

* The QA build fails on the flatgeobuf module:

```
2023-11-17T21:37:05.4058061Z [INFO] PMD Failure: org.geotools.data.flatgeobuf.FlatGeobufDataStoreTest:884 Rule:SimplifiableTestAssertion Priority:3 Assertion may be simplified using assertFalse.
```

* The assembly fails to grab the dependency, and so does the downstream build (has it been published yet?):

```
  cast to org.geotools.api.filter.sort.SortBy[] for a non-varargs call and to suppress this warning
Error:  Failed to execute goal on project gt-flatgeobuf: Could not resolve dependencies for project org.geotools:gt-flatgeobuf:jar:31-SNAPSHOT: Could not find artifact org.wololo:flatgeobuf:jar:3.26.2 in osgeo (https://repo.osgeo.org/repository/release/) -> [Help 1]
```

```
Error:  Failed to execute goal on project gt-flatgeobuf: Could not resolve dependencies for project org.geotools:gt-flatgeobuf:jar:31-SNAPSHOT: Could not find artifact org.wololo:flatgeobuf:jar:3.26.2 in osgeo (https://repo.osgeo.org/repository/release/) -> [Help 1]
```

bjornharrtell:
I was a bit quick, flatgeobuf dependency wasn't propagated to maven repos but should be now.

aaime:
Ok, fix the QA issue, that will restart all builds and we can merge and backport.

PR #4579, opened 2023-11-30
Added a utility method to test for JAXP 1.5 support.

sikeoka:
This PR adds a utility method to check whether the TransformerFactory and SchemaFactory implementations support JAXP 1.5.  This is intended to allow programmatically fixing an error caused by the recent Hazelcast upgrade in GeoServer/GeoWebcache that breaks the affected modules because Hazelcast is internally using JAXP 1.5 properties without having multiple copies of this code around.  There is a chance that this could potentially be useful with a lot of major dependency upgrades planned for next year.  There are no unit tests because the check depends on external XML libraries being on the classpath.  GeoTools does not add the Xalan or Xerces libraries and GeoServer uses the Xalan library while a couple of GeoServer extensions use the Xerces library.  Depending on how GeoServer is being installed, external XML libraries may be included in the classpath by the web application server.

If this PR is approved, it must be backported to 30.x to be able to fix the Hazelcast modules in GeoServer 2.24.x.

  

# Checklist

### DIFF
--- a/modules/library/xml/src/main/java/org/geotools/xml/XMLUtils.java
+++ b/modules/library/xml/src/main/java/org/geotools/xml/XMLUtils.java
@@ -16,7 +16,13 @@
  */
 package org.geotools.xml;
 
+import java.util.ArrayList;
+import java.util.List;
+import javax.xml.XMLConstants;
 import javax.xml.namespace.QName;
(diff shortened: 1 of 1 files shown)


mprins (approved):
lgtm

PR #4590, opened 2023-12-07
[GEOT-7502] Optimize FlatGeobuf DataStore for raw Query with startIndex

bjornharrtell:
[GEOT-7502](https://osgeo-org.atlassian.net/browse/GEOT-7502) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

  

# Checklist

### DIFF
--- a/modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCFeatureSource.java
+++ b/modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCFeatureSource.java
@@ -492,10 +492,10 @@ public class JDBCFeatureSource extends ContentFeatureSource {
         try {
 
             if ((postFilter != null) && (postFilter != Filter.INCLUDE)
-                    || (query.getMaxFeatures() < Integer.MAX_VALUE && !canLimit())
+                    || (query.getMaxFeatures() < Integer.MAX_VALUE && !canLimit(query))
                     || (query.getStartIndex() != null
                             && query.getStartIndex() > 0
-                            && !canOffset())) {
+                            && !canOffset(query))) {
                 // calculate manually, don't use datastore optimization
                 getDataStore().getLogger().fine("Calculating bounds manually");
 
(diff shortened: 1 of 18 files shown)


bjornharrtell:
@aaime when you got time I would appreciate if you can sanity check this.

modules/unsupported/flatgeobuf/src/test/java/org/geotools/data/flatgeobuf/FlatGeobufDataStoreTest.java
```diff
@@ -903,6 +903,24 @@ public void readCountriesFids() throws IOException {
...
+        query.setMaxFeatures(2);
+        query.setStartIndex(10);
+        SimpleFeatureCollection featureCollection = featureSource.getFeatures(query);
+        try (SimpleFeatureIterator it = featureCollection.features()) {
+            SimpleFeature f1 = it.next();
+            assertEquals("countries.107", f1.getID());
+            SimpleFeature f2 = it.next();
+            assertEquals("countries.108", f2.getID());
+            boolean hasNext = it.hasNext();
+            assertFalse(hasNext);
```
aaime, line 920: Against a database one would need to use a order by to get a predictable output sequence, but it's very easy to imagine a file read to have a predictable order instead. Just trying to make sure.

aaime (changes requested):
What's missing here is the implementation of "canOffset -> true" in the FlatGeoBufFeatureSource.

modules/unsupported/flatgeobuf/src/test/java/org/geotools/data/flatgeobuf/FlatGeobufDataStoreTest.java
bjornharrtell: Hmm yes, and FlatGeobuf doesn't allow for any mutation so ordering can be assumed to be static.

bjornharrtell:
> What's missing here is the implementation of "canOffset -> true" in the FlatGeoBufFeatureSource.

Hmm ok. I just have be sure something doesn't think it can do more than it can - that is, a FlatGeobuf can only be read with an offset if it has an index and there is no other (spatial) filter.

bjornharrtell:
@aaime I think I see what you mean... I've tried to get my head around it but as of yet I'm failing to understand how to implement. It seems a bit deep of a rabbit hole (to me) because it works togheter with sorting (which does make sense in the general case) but I'm not sure how to special case / override that.

bjornharrtell:
The current implementation isn't effective. It is not possible to implement in a straight forward was due to inability to optimize this case for specific queries, see https://github.com/geotools/geotools/pull/4593.

Added a failing test what I think is what should be the correct result. The fail seems to have something to do with the default sorting that is applied in the non native case and should be looked into.

bjornharrtell:
Merged with https://github.com/geotools/geotools/pull/4593 and https://github.com/geotools/geotools/pull/4594 and reenabled the optimization and looks like it working.

aaime:
With https://github.com/geotools/geotools/pull/4593 merged, would you update this one? Or close?

jodygarnett:
@bjornharrtell I opened up this ticket https://osgeo-org.atlassian.net/browse/GEOT-7509 to cover doc change missed by this PR.

PR #4593, opened 2023-12-11
[GEOT-7509] Introduce query as context to ContentFeatureSource capabilities

bjornharrtell:
[GEOT-7509](https://osgeo-org.atlassian.net/browse/GEOT-7509) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

The intent is to allow for data stores that can optimize for certain query capabilities based on the actual query. The need for this was discovered in https://github.com/geotools/geotools/pull/4590.

This PR implements this by adding new capability methods that has query as parameter and deprecates the only parameter-less capability methods.
  

# Checklist

### DIFF
--- a/modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCFeatureSource.java
+++ b/modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCFeatureSource.java
@@ -492,10 +492,10 @@ public class JDBCFeatureSource extends ContentFeatureSource {
         try {
 
             if ((postFilter != null) && (postFilter != Filter.INCLUDE)
-                    || (query.getMaxFeatures() < Integer.MAX_VALUE && !canLimit())
+                    || (query.getMaxFeatures() < Integer.MAX_VALUE && !canLimit(query))
                     || (query.getStartIndex() != null
                             && query.getStartIndex() > 0
-                            && !canOffset())) {
+                            && !canOffset(query))) {
                 // calculate manually, don't use datastore optimization
                 getDataStore().getLogger().fine("Calculating bounds manually");
 
@@ -538,27 +538,25 @@ public class JDBCFeatureSource extends ContentFeatureSource {
     }
 
     @Override
-    protected boolean canFilter() {
+    protected boolean canFilter(Query query) { return true; }
+
+    @Override
+    protected boolean canSort(Query query) {
         return true;
     }
 
     @Override
-    protected boolean canSort() {
+    protected boolean canRetype(Query query) {
         return true;
     }
 
     @Override
-    protected boolean canRetype() {
-        return true;
-    }
-
-    @Override
-    protected boolean canLimit() {
+    protected boolean canLimit(Query query) {
         return getDataStore().getSQLDialect().isLimitOffsetSupported();
     }
 
     @Override
-    protected boolean canOffset() {
+    protected boolean canOffset(Query query) {
         return getDataStore().getSQLDialect().isLimitOffsetSupported();
     }
 
--- a/modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCFeatureSource.java
+++ b/modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCFeatureSource.java
@@ -538,7 +538,9 @@ public class JDBCFeatureSource extends ContentFeatureSource {
     }
 
     @Override
-    protected boolean canFilter(Query query) { return true; }
+    protected boolean canFilter(Query query) {
+        return true;
+    }
 
     @Override
     protected boolean canSort(Query query) {
--- a/modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCFeatureStore.java
+++ b/modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCFeatureStore.java
@@ -183,28 +183,28 @@ public final class JDBCFeatureStore extends ContentFeatureStore {
     }
 
     @Override
-    protected boolean canFilter() {
-        return delegate.canFilter();
+    protected boolean canFilter(Query query) {
+        return delegate.canFilter(query);
     }
 
     @Override
-    protected boolean canSort() {
-        return delegate.canSort();
+    protected boolean canSort(Query query) {
+        return delegate.canSort(query);
     }
 
     @Override
-    protected boolean canRetype() {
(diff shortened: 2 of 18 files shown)


modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCFeatureSource.java
```diff
@@ -538,12 +538,12 @@ protected ReferencedEnvelope getBoundsInternal(Query query) throws IOException {
     }
 
     @Override
-    protected boolean canFilter() {
+    protected boolean canFilter(Query query) {
```
jodygarnett, line 541: Not quite what I had in mind, the API change should be additive:
```
protected boolean canFilter() { return true; }
protected boolean canFilter(Query query) { return canFilter(); }
```

bjornharrtell: Agreed, reworked.

jodygarnett:
My initial feedback was addressed.

aside: If you deprecate canFilter() then the default implementation of canFilter(Query) would not be able to call it without producing a warning.

bjornharrtell:
Yes I had the feeling this should deprecate the old methods.. and I suppose the new can just import the logic as is instead of calling to avoid the calling deprecated issue.

aaime:
> aside: If you deprecate canFilter() then the default implementation of canFilter(Query) would not be able to call it without producing a warning.

That's what ``@suppressWarnings`` is for :rofl: 
That said, I'm the only one that finds the new setup flexible, but confusing? What do you propose? Carefully documenting it instead?

aaime:
> Yes I had the feeling this should deprecate the old methods.. and I suppose the new can just import the logic as is instead of calling to avoid the calling deprecated issue.

Not really, store implementations exists also out there, if you stop calling the old methods, they will break.

bjornharrtell:
I'm thinking it might as well be a breaking API change..?

jodygarnett:
We do our best to avoid breaking API change (last release cycle as an exception). 

I think the methods are confusing, although they read okay when used in an if statement. Do you have a suggestion for an alternative name? `isFilterOptmized(Query)` or `optimizeFilter(Query)`.

jodygarnett:
There is also something missing with the check being simple / true false.  Andrea talked about pre-filtering before in gitter chat. 

We kind of want to know two independent things:

a) is their an pre-processing available to run for this Query? Even it does not 100% address the problem?

b) is any post-processing required afterwords? To make it to 100%?

Not sure what to do with that insight however.

bjornharrtell:
Not sure I understand. Was not the whole purpose of these methods to indicate that the source can natively pre-process the capability to skip the GeoTools generic implementation? And the parameter query just adds the context of a query that implictly was all queries before?

While I agree this could probably be more fine grained than a boolean state but that is a different matter than what this PR attempts to do, and I think it might be over complicating things.

I don't quite see the naming issue.

jodygarnett:
I guess the naming issue is this `canFilter()` makes sense as true or false because it is just a chance to indicate if an optimization is available; or if the superclass will need to do all the work.

Introducing `canFilter(Query)` opens up the possibility of doing 1/2 of the work. If they query had a BBOX and an attribute filter many implementations with a spatial index would be able to optimize the BBOX part ( `canFilter(Query)` returning true) but then need post-processing for the attribute filter.

bjornharrtell:
The query param only makes the path per query, which is was implicitly anyway? The responsibility is still that a data store needs to take full reponsibility if answering true and this is clear from the docs?

jodygarnett:
I am somewhat speculating on how Andrea found the names confusing.

I am also keeping in mind this code:

```
        if (!canFilter(query)) {
            if (query.getFilter() != null && query.getFilter() != Filter.INCLUDE) {
                reader = new FilteringFeatureReader<>(reader, query.getFilter());
            }
        }
```

See how it is used as part of a block to determine if post-processing using FilteringFeatureReader is required?

Q: Is each case where these methods are called using the `!` test? That may be an indication of why the method names are confusing... is there a way to rename the method so that the answer matches the use?

Q: Is the case where the query splits (with both an optimization and post-processing required) worth while to address as part of the superclass here?

aaime:
The source of confusion is simple, canFilter() vs canfilter(Query), an existing implementation out of our control may not know which one to implement and how --> update docs please (including the data store tutorial).

The pre-filtering is already done in a lot of places, simple rules:
* If one can only partially execute the filter, ``canFilter -> false``, and the ContentFeatureSource will run the entire filter after the first pass. The first pass is often rough and incomplete, useful to trim down the set of features to check, but not exact.
* If one can fully and correctly execute the filter, then ``canFilter -> true`` and he's responsible for correct filter execution in all cases.

Using the query as a parameter does not change the above, it just makes it contextual to the query. It's an improvement over the same pattern.

Mind, filters are the easy case, because if implemented correctly, one can apply the full filter again without changing the results. It's offset that is not following that rule, one cannot offset twice, it has to be done either in the implementation, or in the base class, but not in both.

I read somewhere comments about the interaction between sort and offset. Indeed if a feature source can offset but cannot sort, we have a problem, ContentFeatureSource does not seem to handle that (in case of sort, it should refuse to pass down the startIndex to the implementation, and handle both in post processing instead).

aaime:
One more note: I agree this approach will solve your issue more cleanly. But it's also a change in one of the principal interfaces, a backport of this PR may not be allowed, we normally keep these on main only (I can't remember this kind of breakage being backported, when it was affecting a core extension point, at the very least).

bjornharrtell:
@aaime agreed but I currently feel I don't have any hack in mind that could solve it in a way I can live with the current APIs so the long road is the only plan I have at this time. However, as this has now been reworked with adding the new API as overloads and old APIs "just" deprecated isn't not really a breaking change right?

I'm not sure I can process the feedback correctly with regard to the naming/usage etc. so I would rather ask if the conclusion is that this PR in its current form is something we can go with or do we need to go back to drawing board?

bjornharrtell:
I've merged this into https://github.com/geotools/geotools/pull/4590 and using it for that with what looks like success.

bjornharrtell:
I'm fine with this target next major only i.e no backport, if in agreement on that this is ready for merge.

aaime:
I'm good with merging only on main. @jodygarnett what do you think?

PR #4594, opened 2023-12-11
Fix ShapefileDataStore feature ordering not respected with offset

bjornharrtell:
This PR currently only demonstrates unexpected behaviour that ordering is not the same if using an offset to get features in ShapefileDataStore. AFAIK this is due to sorting happening using alphabetic sort on the stringified typename + numeric id.
  

# Checklist

### DIFF
--- a/modules/library/main/src/main/java/org/geotools/data/sort/AlphanumComparator.java
+++ b/modules/library/main/src/main/java/org/geotools/data/sort/AlphanumComparator.java
@@ -1,0 +1,116 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ *
+ *    (C) 2004-2008, Open Source Geospatial Foundation (OSGeo)
+ *
+ *    This library is free software; you can redistribute it and/or
+ *    modify it under the terms of the GNU Lesser General Public
+ *    License as published by the Free Software Foundation;
+ *    version 2.1 of the License.
+ *
+ *    This library is distributed in the hope that it will be useful,
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ *    Lesser General Public License for more details.
+ */
+package org.geotools.data.sort;
+
+/*
+ * Copyright © 2021 Dave Koelle www.davekoelle.com/
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the “Software”), to deal
+ * in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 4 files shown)


bjornharrtell:
I think this also implies that the ShapefileDataStore is not optimizing offsetted access in the same way as I found is the case for FlatGeobuf (https://osgeo-org.atlassian.net/browse/GEOT-7502).

bjornharrtell:
@aaime one naive way to fix this I guess could be to do alphanumeric / natural comparison in FidComparator. Sensible though? Not sure... then again, geotools doesn't seem to be aware of anything else than feature id as strings. Perhaps it should internally store the parts that made up the string feature id  (typename + some id possibly ordered integer) and sort/compare by that but seems like a large or perhaps undesirable undertaking.

bjornharrtell:
Added implementation to solve this using alphanumeric comparison.

aaime:
I don't see a clear explanation of what the problem is. Here is the expected behavior:

- If the query has no paging and no sorting, the store is free to return features in whatever order it wants
- If there is paging, the store is supposed to do whatever it can to support stable paging, which is implemented by sorting over fid, unless the store has a different way to do that.

Long story short, unless there is explicit sorting required by the user, the store is free to do whatever it feels is best, as long as paging ensures the same feature does not appear on more than one page, and minimizes the chance of multiple features showing up in pages, if editing occurs (of course in that case, it's not possible to ensure false reads, duplicate features, in the absolute sense).

bjornharrtell:
@aaime I suppose/guess there is no critical bug here, the wierd ordering will "work" (?). But more generally I don't think it's sensible that geotools cannot order sequential feature ids correctly. That said, fixing that by alphanumeric sort may also not be sensible (complex, costly). A more IMHO proper fix would require very invasive changes in geotools that is not wished for in short term perhaps.

So I guess this PR is more of a discourse in what is sensible or not and a possible starting point for a discussion if geotools should handle fids better/differently.

aaime:
Seems like a lot of work for an aesthetic advantage, I would not go for it. Stores that know they have numeric ids can implement their own native sorting if they so like. Just having a quick look at the comparator makes me wonder how it would behave if the ids are actual strings that happen to sometimes contain a numeric part in them.

But it's only my opinion, with the changes you're proposing lately (including the ContentFeatureSource) you're entering squarely in "format proposal" territory. We normally recommend contributors to set down a discussion before starting to code, but I understand that some use prototypes to clarify their direction, before going formal.

See here: https://docs.geotools.org/latest/developer/procedures/proposal.html

bjornharrtell:
@aaime I agree that this change deserves a formal process, if wanted. I also more or less agree it's a bit aesthetic and considering that it might not be worth the effort. I will leave this open for a while, but likely it can/will be closed as a curiosity.

Note that this is not related or required for https://github.com/geotools/geotools/pull/4593 or https://github.com/geotools/geotools/pull/4590.

PR #4603, opened 2023-12-19
WIP Support meters in dwithin filter

elahrvivaz:
This PR attempts to get DWITHIN filters working correctly in meters, vs in the native cartesian units. It's currently just a draft for gathering comments.

Based on the geoserver mailing list thread [here](https://sourceforge.net/p/geoserver/mailman/geoserver-users/thread/CALHQADUEnqdJdW2-yEQ5fn4Tk6GHsJEcnTqh-pMdxfo5Zc5T%3DA%40mail.gmail.com/#msg58713421).
  

# Checklist

### DIFF
--- a/modules/library/main/src/main/java/org/geotools/filter/spatial/DWithinImpl.java
+++ b/modules/library/main/src/main/java/org/geotools/filter/spatial/DWithinImpl.java
@@ -16,13 +16,33 @@
  */
 package org.geotools.filter.spatial;
 
+import java.util.logging.Logger;
+import org.apache.commons.lang3.tuple.ImmutablePair;
+import org.apache.commons.lang3.tuple.Pair;
 import org.geotools.api.filter.FilterVisitor;
 import org.geotools.api.filter.expression.Expression;
 import org.geotools.api.filter.spatial.DWithin;
+import org.geotools.api.geometry.Position;
+import org.geotools.api.referencing.FactoryException;
+import org.geotools.data.util.DistanceBufferUtil;
 import org.geotools.filter.CartesianDistanceFilter;
+import org.geotools.geometry.Position2D;
+import org.geotools.referencing.CRS;
+import org.geotools.referencing.GeodeticCalculator;
+import org.locationtech.jts.geom.Coordinate;
+import org.locationtech.jts.geom.Envelope;
 import org.locationtech.jts.geom.Geometry;
+import org.locationtech.jts.operation.distance.DistanceOp;
 
 public class DWithinImpl extends CartesianDistanceFilter implements DWithin {
+
+    private static final Logger LOGGER =
+            org.geotools.util.logging.Logging.getLogger(DWithinImpl.class);
+
+    private static final ThreadLocal<Pair<Integer, GeodeticCalculator>> CALCULATORS =
+            new ThreadLocal<>();
+
+    private double distanceInMeters;
 
     public DWithinImpl(Expression e1, Expression e2) {
         super(e1, e2);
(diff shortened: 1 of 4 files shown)


modules/library/main/src/main/java/org/geotools/filter/spatial/DWithinImpl.java
```diff
@@ -32,13 +52,89 @@ public DWithinImpl(Expression e1, Expression e2, MatchAction matchAction) {
...
+        setDistanceInMeters();
+    }
+
+    private void setDistanceInMeters() {
+        distanceInMeters = DistanceBufferUtil.getDistanceInMeters(this);
+    }
+
     @Override
     public boolean evaluateInternal(Geometry left, Geometry right) {
+        if (left.getSRID() != 0 && left.getSRID() == right.getSRID()) {
```
elahrvivaz, line 73: it seems like the SRID doesn't actually get set in the geometry, and I can't figure out the correct CRS without access to the feature type. Without that, this doesn't seem like it's feasible to implement.

aaime: Implementing here would be ideal, it's completely store-independent, but as you say, also hard.
And it's also true that the SRID is not always populated, and now that we support multiple authorities, it's also not useful in general (EPSG:4326 is not necessarily the same as IAU:4326 or ESRI:4326).

What you can try to do is to reach out for the expressions, and do something like this (mind, might not compile, just take it as a "fill in the blanks" example):

```
if (feature instancef Feature) {
  FeatureType schema = ((Feature) feature).getFeatureType();
  AttributeDescriptor ad = getExpression1().evaluate(schema, AttributeDescriptor.class);
}
```

If the expression is a property name, it will return an attribute descriptor from which you can grab the CRS. 
Downside, you'll have to rewrite the evaluate method of the superclass entirely.

I would have taken a different route: replace the filter inside ContentFeatureSource, where you have full access to the schema structure, using a DuplicatingFilterVisitor class. And that action could be subject to a ``canHandleDistanceUnits`` method just like the other ``canLimit``, ``canFilter`` and so on, realdy available in that class.

elahrvivaz: Do all the geotools datastores extend ContentDataStore at this point? It seems like with that approach, we'd probably want to set `canHandleDistanceUnits` to default to true, so that we keep the same current behavior and don't break anything. but that means any non-geotools datastores would have to opt-in to get the fix.

aaime: Almost all of them do, here is a search for a store that implements DataStore directly:

```
$ git grep "implements DataStore "
modules/extension/transform/src/main/java/org/geotools/data/transform/SingleFeatureSourceDataStore.java:public class SingleFeatureSourceDataStore implements DataStore {
modules/library/main/src/main/java/org/geotools/data/DataStoreAdaptor.java:final class DataStoreAdaptor implements DataStore {
modules/library/main/src/main/java/org/geotools/data/directory/DirectoryDataStore.java:public class DirectoryDataStore implements DataStore {
modules/library/main/src/main/java/org/geotools/data/store/ContentDataStore.java:public abstract class ContentDataStore implements DataStore {
modules/library/main/src/main/java/org/geotools/data/store/DecoratingDataStore.java:        implements DataStore {
modules/plugin/feature-pregeneralized/src/main/java/org/geotools/data/gen/PreGeneralizedDataStore.java:public class PreGeneralizedDataStore implements DataStore {
modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/catalog/DataAccessStoreWrapper.java:class DataAccessStoreWrapper implements DataStore {
modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/catalog/oracle/DataStoreWrapper.java:public abstract class DataStoreWrapper implements DataStore {
```

All of them look like wrappers, PreGeneralizedDataStore is also one, it delegates to secondary stores.
It's also been a while [since our documentation taught people to use ContentDataStore](https://docs.geotools.org/stable/userguide/tutorial/datastore/index.html) so the majority of stores out there should also be using it.

elahrvivaz: thanks!

elahrvivaz:
Going to close this as I haven't had time to work on it, but hope to return to it at some point.

PR #4355, opened 2023-06-20
Fix setting region on immutable AmazonS3Client (s3-geotiff)

c-mckenna:
Fix for a `UnsupportedOperationException` when setting the region on `AmazonS3Client`. When created with the `AmazonS3ClientBuilder` the `AmazonS3Client` is immutable.

See comment on #4341 for additional details.

# Checklist

- [x] I have read the [contribution guidelines](https://github.com/geotools/geotools/blob/main/CONTRIBUTING.md).
- [ ] I have sent a [Contribution Licence Agreement](https://docs.geotools.org/latest/developer/procedures/contribution_license.html) (not required for small changes, e.g., fixing typos in documentation).
- [x] First PR targets the `main` branch (backports managed later; ignore for branch specific issues).
- [x] Avoid [Java 9+ split packages](http://tutorials.jenkov.com/java/modules.html#split-packages-not-allowed).
- [x] All the build checks are green ([see automated QA checks](https://docs.geotools.org/latest/developer/conventions/code/qa.html)).

For core and extension modules:

- [x] New unit tests have been added covering the changes.
- [ ] [Documentation](https://github.com/geotools/geotools/tree/main/docs) has been updated (if change is visible to end users).
- [ ] There is an issue in [GeoTools Jira](https://osgeo-org.atlassian.net/projects/GEOT) (except for changes not visible to end users). 
- [ ] Commit message(s) must be in the form ``[GEOT-XYZW] Title of the Jira ticket``.
- [x] Bug fixes and small new features are presented as a single commit.
- [x] The commit targets a single objective (if multiple focuses cannot be avoided, each one is in its own commit, and has a separate ticket describing it).

PR #4604, opened 2023-12-20
Fix updating immutable AmazonS3 client

mloeks:
When using a custom endpoint (for instance using a mock in integration tests), the S3Connector still tried to update the immutable client and throw.

A similar issue was already addressed in 27c6861858fe81979085bf41f2ed7e7a986e8e1d ([PR](https://github.com/geotools/geotools/pull/4355)), however, the chain which is used for custom endpoints was not covered.

I suppose this should be considered to backport to v29 as well?

Thanks for your review.

# Checklist

### DIFF
--- a/modules/unsupported/s3-geotiff/src/main/java/org/geotools/s3/S3Connector.java
+++ b/modules/unsupported/s3-geotiff/src/main/java/org/geotools/s3/S3Connector.java
@@ -19,10 +19,10 @@ package org.geotools.s3;
 import com.amazonaws.auth.AWSStaticCredentialsProvider;
 import com.amazonaws.auth.AnonymousAWSCredentials;
 import com.amazonaws.auth.BasicAWSCredentials;
+import com.amazonaws.client.builder.AwsClientBuilder;
 import com.amazonaws.regions.Regions;
 import com.amazonaws.services.s3.AmazonS3;
 import com.amazonaws.services.s3.AmazonS3ClientBuilder;
-import com.amazonaws.services.s3.S3ClientOptions;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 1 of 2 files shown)


modules/unsupported/s3-geotiff/src/main/java/org/geotools/s3/S3Connector.java
```diff
@@ -116,24 +116,21 @@ public AmazonS3 getS3Client() {
...
+                            .withEndpointConfiguration(endpointConfiguration)
+                            .withPathStyleAccessEnabled(true)
                             .build();
 
-            final S3ClientOptions clientOptions =
-                    S3ClientOptions.builder().setPathStyleAccess(true).build();
-            s3.setS3ClientOptions(clientOptions);
-            String endpoint = prop.getProperty(s3Alias + ".s3.endpoint");
-            if (!endpoint.endsWith("/")) {
-                endpoint = endpoint + "/";
```
mloeks, line 133: Not needed, as a leading slash is added to the resource path in `com.amazonaws.services.s3.internal.ServiceUtils.convertRequestToUrl(Request, boolean, boolean)`

Charmatzis:
Many thanks @mloeks

2mars:
Thanks @mloeks ! This bug is quite annoying.

modules/unsupported/s3-geotiff/src/main/java/org/geotools/s3/S3Connector.java
NielsCharlier: Indeed, and it looks like the code has also been simplified by proper use of the API.

NielsCharlier:
It seems to be a clear and simple improvement of the code, and three unit tests added, so I am confident to approve this.

NielsCharlier (approved):

aaime:
And thanks @NielsCharlier for th review!

mloeks:
Many thanks @NielsCharlier @aaime !

I cannot step in for maintenance of the module though I'm afraid. I don't know all too much about its functionality, and we also planned to drop it in our codebase since quite a while.

PR #4606, opened 2023-12-21
[GEOT-7511] Fix polygon marker fill

vuilleumierc:
[GEOT-7511](https://osgeo-org.atlassian.net/browse/GEOT-7511) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

Fix the error thrown when the margin is greater than the symbol size

  

# Checklist


modules/library/render/src/test/resources/org/geotools/renderer/lite/test-data/margin/mark-margin.sld
```diff
@@ -20,7 +20,7 @@
                 </Graphic>
               </GraphicFill>
             </Fill>
-            <VendorOption name="graphic-margin">5</VendorOption>
+            <VendorOption name="graphic-margin">20</VendorOption>
```
aaime, line 23: Can we have a new test, rather than reusing the existing one? 
So that the test covers the case of margin smaller than symbol, and the case of marging bigger than symbol, along with ensure that existing styles keep on working in the same way?

aaime (changes requested):

vuilleumierc:
Any idea why the Mac OS build fails here? Could an admin restart this workflow if that makes sense?

aaime:
@vuilleumierc there have been known problems with OSX setting up GDAL. Can you rebase your PR to the latest main? We have had fixes for that build, hopefully that will help.

vuilleumierc:
The regression fixed in this PR  was introduced by this commit: https://github.com/geotools/geotools/commit/c23bb389388223b578c6ec8d2bd5e53c3936b4ad which doesn't seem to be in the `29.x` branch. So it's probably OK to backport only to `30.x`?

PR #4633, opened 2024-01-27
[GEOT-7522] NetCDFMosaicReaderTest failure 

pcardinal:
[GEOT-7522](https://osgeo-org.atlassian.net/browse/GEOT-7522) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

Modification of the test NetCDFMosaicReaderTest.testGranuleSourceFileView() by the addition of a line of code to set the Timezone of the JVM before running two assertEquals using a timestamp. Without this change, the test uses the computer's Timezone  when running "mvn install". The Timezone is set to "UTC" for UTC+0:00.

  

# Checklist

- [X] All the build checks are green ([see automated QA checks](https://docs.geotools.org/latest/developer/conventions/code/qa.html)).

### DIFF
--- a/modules/plugin/coverage-multidim/netcdf/src/test/java/org/geotools/coverage/io/netcdf/NetCDFMosaicReaderTest.java
+++ b/modules/plugin/coverage-multidim/netcdf/src/test/java/org/geotools/coverage/io/netcdf/NetCDFMosaicReaderTest.java
@@ -1928,6 +1928,7 @@ public class NetCDFMosaicReaderTest {
                 nc1 = it.next();
                 nc2 = it.next();
             }
+						
             assertEquals("2017-02-06 00:00:00.0", nc1.getAttribute("time").toString());
             assertEquals("2017-02-06 12:00:00.0", nc2.getAttribute("time").toString());
         } finally {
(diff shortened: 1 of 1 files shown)


pcardinal:
Finsl

modules/plugin/coverage-multidim/netcdf/src/test/java/org/geotools/coverage/io/netcdf/NetCDFMosaicReaderTest.java
```diff
@@ -1916,6 +1916,7 @@ public void testGranuleSourceFileView() throws Exception {
                 nc2 = it.next();
             }
 
+            TimeZone.setDefault(TimeZone.getTimeZone("UTC"));
```
aaime, line 1919: Good call, but done like this could be source of trouble: the test JVM is re-used across modules, meaning other tests might get this same timezone while before they did not. Also, we use parallel builds, which other modules tests might be affected by this would be semi-random (depending on the performance of the hardware, how many parallel processes are used for the build, and so on).

Best to isolate this properly, like done in this test instead:
https://github.com/geotools/geotools/blob/main/modules/plugin/geopkg/src/test/java/org/geotools/geopkg/JSONArrayIOTest.java#L35

aaime (changes requested):

PR #4650, opened 2024-02-02
Vector mosaic: support granule filtering and picking granules from a repository

aaime:
Extend the vector mosaic store to support filtering granule stores, and picking them from a repository

# Checklist

### DIFF
--- a/modules/unsupported/vector-mosaic/src/main/java/org/geotools/vectormosaic/GranuleStoreFinderImpl.java
+++ b/modules/unsupported/vector-mosaic/src/main/java/org/geotools/vectormosaic/GranuleStoreFinderImpl.java
@@ -24,20 +24,34 @@ import java.util.logging.Logger;
 import org.geotools.api.data.DataStore;
 import org.geotools.api.data.DataStoreFactorySpi;
 import org.geotools.api.data.DataStoreFinder;
+import org.geotools.api.data.Repository;
+import org.geotools.data.store.DecoratingDataStore;
+import org.geotools.feature.NameImpl;
 import org.geotools.util.logging.Logging;
 
-/** Class used to find the DataStore for a VectorMosaicGranule. */
+/**
+ * Class used to find the DataStore for a VectorMosaicGranule. The lookup tries three possible
+ * avenues:
+ *
+ * <ul>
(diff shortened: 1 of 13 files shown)


modules/unsupported/vector-mosaic/src/main/java/org/geotools/vectormosaic/GranuleStoreFinderImpl.java
```diff
@@ -24,20 +24,25 @@
...
 
 /** Class used to find the DataStore for a VectorMosaicGranule. */
 public class GranuleStoreFinderImpl extends GranuleStoreFinder {
     static final Logger LOGGER = Logging.getLogger(GranuleStoreFinderImpl.class);
     protected final String preferredSPI;
+    protected final Repository repository;
 
     /**
      * Constructor that accepts a nullable preferred SPI.
      *
```
turingtestfail, line 40: Add Repository to doc?

modules/unsupported/vector-mosaic/src/main/java/org/geotools/vectormosaic/VectorMosaicFeatureSource.java
```diff
@@ -66,6 +67,9 @@
 /** FeatureSource for a vector mosaic. */
 public class VectorMosaicFeatureSource extends ContentFeatureSource {
     static final Logger LOGGER = Logging.getLogger(VectorMosaicFeatureSource.class);
+
```
turingtestfail, line 70: Add a note about what pattern this regex looks for with store names?

turingtestfail (changes requested):
Adding the option to use the repo with a filter seems like a big enough change to warrant some user documentation changes?

modules/unsupported/vector-mosaic/src/main/java/org/geotools/vectormosaic/VectorMosaicFeatureSource.java
aaime: Yes, done, and added the same in the GeoServer user docs.

aaime:
Documentation updates here: https://github.com/geoserver/geoserver/pull/7393

turingtestfail (approved):

PR #4668, opened 2024-02-22
[GEOT-7534] fix geojsonreader stops reading features if it encounters one with geometry=null

oberhamsi:
[GEOT-7534](https://osgeo-org.atlassian.net/browse/GEOT-7534) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

`geometry:null` is valid geojson and since a missing `geojson` field is also accepted, this should be fine.

# Checklist

### DIFF
--- a/modules/unsupported/geojson-core/src/main/java/org/geotools/data/geojson/GeoJSONReader.java
+++ b/modules/unsupported/geojson-core/src/main/java/org/geotools/data/geojson/GeoJSONReader.java
@@ -358,9 +358,12 @@ public class GeoJSONReader implements AutoCloseable {
                 }
                 while (parser.nextToken() == JsonToken.START_OBJECT) {
                     ObjectNode node = mapper.readTree(parser);
-
-                    SimpleFeature feature = getNextFeature(node);
-                    features.add(feature);
+                    try {
+                        SimpleFeature feature = getNextFeature(node);
+                        features.add(feature);
+                    } catch (IOException e) {
+                        LOGGER.log(Level.WARNING, e.getMessage(), e);
+                    }
                 }
             }
             // support paged feature collections, OGC API style
@@ -446,7 +449,7 @@ public class GeoJSONReader implements AutoCloseable {
 
         // the geometry might have been selected away by a property selection
         Geometry g = null;
-        if (geom != null) g = GEOM_PARSER.geometryFromJson(geom);
+        if (geom != null && !geom.isNull()) g = GEOM_PARSER.geometryFromJson(geom);
 
         JsonNode props = node.get("properties");
         // accommodate for STAC servers that remove the properties object altogether, when
--- a/modules/unsupported/geojson-core/src/main/java/org/geotools/data/geojson/GeoJSONReader.java
+++ b/modules/unsupported/geojson-core/src/main/java/org/geotools/data/geojson/GeoJSONReader.java
@@ -358,12 +358,8 @@ public class GeoJSONReader implements AutoCloseable {
                 }
                 while (parser.nextToken() == JsonToken.START_OBJECT) {
                     ObjectNode node = mapper.readTree(parser);
-                    try {
-                        SimpleFeature feature = getNextFeature(node);
-                        features.add(feature);
-                    } catch (IOException e) {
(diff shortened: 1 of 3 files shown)


modules/unsupported/geojson-core/src/main/java/org/geotools/data/geojson/GeoJSONReader.java
```diff
@@ -446,7 +449,7 @@ private SimpleFeature getNextFeature(ObjectNode node) throws IOException {
 
         // the geometry might have been selected away by a property selection
         Geometry g = null;
-        if (geom != null) g = GEOM_PARSER.geometryFromJson(geom);
+        if (geom != null && !geom.isNull()) g = GEOM_PARSER.geometryFromJson(geom);
```
aaime, line 452: This one seems to be the actual fix

modules/unsupported/geojson-core/src/main/java/org/geotools/data/geojson/GeoJSONReader.java
```diff
@@ -358,9 +358,12 @@ public SimpleFeatureCollection getFeatures() throws IOException {
...
                     ObjectNode node = mapper.readTree(parser);
-
-                    SimpleFeature feature = getNextFeature(node);
-                    features.add(feature);
+                    try {
+                        SimpleFeature feature = getNextFeature(node);
+                        features.add(feature);
+                    } catch (IOException e) {
+                        LOGGER.log(Level.WARNING, e.getMessage(), e);
+                    }
```
aaime, line 366: And here... it's more tolerance for whatever other case. However, the typical data source in Geotools throws a straight exception when anything unexpected happens during a read. Wondering if there should be some flag to control tolerant reading vs fail fast, with fail fast being the default for consistency.

@ianturton opinions?

modules/unsupported/geojson-core/src/test/java/org/geotools/data/geojson/GeoJSONReaderTest.java
```diff
@@ -690,6 +690,15 @@ public void testGeometryless() throws Exception {
         }
     }
 
+    @Test
+    public void testGeometrynull() throws Exception {
```
aaime, line 694:
```suggestion
    public void testGeometryNull() throws Exception {
```

modules/unsupported/geojson-core/src/main/java/org/geotools/data/geojson/GeoJSONReader.java
oberhamsi: i agree, i can undo this unless you want a flag.

aaime: Normally data sources fail fast. Indeed the rendering subsystem is tolerant. 
Keeping in line with the other source behavior would be best in my mind.

Adding a Query parameter for tolerance would also be a good idea, but having just one implementation seems a bit odd. Do we have other cases where we can "resume" the read like here in JSON? Maybe in JDBC data store if the issue happens converting from resultset to feature, for example?

jodygarnett: I think that would be wise, some datastore may have non simple curves we cannot support like a spline.

oberhamsi: thanks for all the input and pointers!

i would like to remove this catch from the PR and only submit a fix for geometry=null. which is legal per spec so should be okay to change. it's also the original problem which i care about :) 

is that alright for you?

aaime: Works for me, and @jodygarnett can take on adding new query parameters, and make their usage generic enough to be useful in 2-3 different places, control their activation, document them, and so on :laughing:

oberhamsi: okay! well, messy PR for such a simple change :grimacing:

PR #4709, opened 2024-03-28
Geostationary satellite projection improvements, and try harder to reproject envelopes

aaime:
For Geostationary satellite, add an AUTO code and a projection handler.

The projection is however so finicky that the above are not enough, a new algorithm based on bisection has been introduced to transform envelopes in projections that do not like "rectangular" areas (e.g., geostationary area of validity looks roughly like a circle in lon/lat, any reprojection outside of it would fail).

Since that involves a number of trials and errors with exceptions, TransformException stack trace filling is now optional, to work with situations where many errors are foreseeable (helps in envelope reprojection, helps a lot in raster reprojection).

Downstream build is expected to fail due to differences in envelope reprojection, has been adjusted as a consequence in this PR:
https://github.com/geoserver/geoserver/pull/7524

# Checklist

### DIFF
--- a/modules/library/referencing/src/main/java/org/geotools/referencing/factory/wms/Auto97004.java
+++ b/modules/library/referencing/src/main/java/org/geotools/referencing/factory/wms/Auto97004.java
@@ -1,0 +1,87 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ *
+ *    (C) 2019, Open Source Geospatial Foundation (OSGeo)
+ *
+ *    This library is free software; you can redistribute it and/or
+ *    modify it under the terms of the GNU Lesser General Public
+ *    License as published by the Free Software Foundation;
+ *    version 2.1 of the License.
+ *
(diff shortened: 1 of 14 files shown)


modules/library/referencing/src/main/java/org/geotools/referencing/factory/wms/Auto97004.java
```diff
@@ -0,0 +1,87 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ *
+ *    (C) 2019, Open Source Geospatial Foundation (OSGeo)
```
dromagnoli, line 5:
```suggestion
 *    (C) 2024, Open Source Geospatial Foundation (OSGeo)
```

dromagnoli (approved):

modules/library/referencing/src/main/java/org/geotools/referencing/factory/wms/Auto97004.java
aaime: Yep

PR #4710, opened 2024-04-03
[GEOT-7552] Improve flatgeobuf ability to quickly perform count/bounds

aaime:
[GEOT-7552](https://osgeo-org.atlassian.net/browse/GEOT-7552) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

  

# Checklist

### DIFF
--- a/modules/unsupported/flatgeobuf/src/main/java/org/geotools/data/flatgeobuf/FlatGeobufDataStore.java
+++ b/modules/unsupported/flatgeobuf/src/main/java/org/geotools/data/flatgeobuf/FlatGeobufDataStore.java
@@ -163,4 +163,9 @@ public class FlatGeobufDataStore extends ContentDataStore {
         }
         file.delete();
     }
+
+    /** Called internally to clear the header meta cache */
+    void clearHeaderMeta() {
+        this.headerMeta = null;
+    }
 }
--- a/modules/unsupported/flatgeobuf/src/main/java/org/geotools/data/flatgeobuf/FlatGeobufFeatureSource.java
+++ b/modules/unsupported/flatgeobuf/src/main/java/org/geotools/data/flatgeobuf/FlatGeobufFeatureSource.java
@@ -17,6 +17,8 @@
 package org.geotools.data.flatgeobuf;
 
 import java.io.IOException;
+import java.io.InputStream;
+import java.net.URL;
 import org.geotools.api.data.DataStore;
 import org.geotools.api.data.FeatureReader;
 import org.geotools.api.data.Query;
@@ -24,6 +26,7 @@ import org.geotools.api.feature.FeatureVisitor;
 import org.geotools.api.feature.simple.SimpleFeature;
 import org.geotools.api.feature.simple.SimpleFeatureType;
 import org.geotools.api.filter.Filter;
+import org.geotools.api.filter.spatial.BBOX;
 import org.geotools.api.referencing.FactoryException;
 import org.geotools.api.referencing.crs.CoordinateReferenceSystem;
 import org.geotools.data.store.ContentEntry;
@@ -32,6 +35,7 @@ import org.geotools.geometry.jts.ReferencedEnvelope;
 import org.geotools.referencing.CRS;
 import org.locationtech.jts.geom.Envelope;
 import org.wololo.flatgeobuf.HeaderMeta;
+import org.wololo.flatgeobuf.PackedRTree;
 
 public class FlatGeobufFeatureSource extends ContentFeatureSource {
 
@@ -57,7 +61,7 @@ public class FlatGeobufFeatureSource extends ContentFeatureSource {
 
     @Override
     protected ReferencedEnvelope getBoundsInternal(Query query) throws IOException {
-        if (query.getFilter() != Filter.INCLUDE) {
+        if (query.getFilter() != Filter.INCLUDE && !(query.getFilter() instanceof BBOX)) {
             return null;
         }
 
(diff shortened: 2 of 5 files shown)


aaime:
@bjornharrtell thoughts on this PR? Particularly interested in your opinion about the dateline crossing case.

bjornharrtell:
@aaime looks generally sensible but I don't quite understand the "dateline crossing case". Can you explain in more detail about the expectations in that case? Perhaps even visualise it, if that is not too much work.

aaime:
@bjornharrtell pheraps it's an issue with the test file, ``countries.fgb``? Here is the bit of test I'm referring to:

```java
        // now limit to the Austrailia region
        bbox = ff.bbox(ff.property(""), new ReferencedEnvelope(112, 154, -44, -11, WGS84));
        q.setFilter(bbox);
        // Australia, but also Fiji bbox happens to cross the dateline
        assertEquals(2, featureSource.getCount(q));
        // however, the actual features are only Australia...
        assertEquals(1, DataUtilities.count(featureSource.getFeatures(q)));
```

What I find is that, if I query the spatial index, that bbox matches two countries, Australia, and Fiji. Australia is the intended target, Fiji is actually far away and should not be matched. So why is it in the index, and why is it not, if I force the reader to go through instead?

The spatial index contains bbox elements that are normalized between -180 and 180. Fiji happens to overlap the 180 line, with a bit of geometry beyond it:

```
MULTIPOLYGON (((178.3736 -17.33992, 178.71806 -17.62846, 178.55271 -18.15059, 177.93266 -18.28799, 177.38146 -18.16432, 177.28504 -17.72465, 177.67087 -17.38114, 178.12557 -17.50481, 178.3736 -17.33992)), ((179.364143 -16.801354, 178.725059 -17.012042, 178.596839 -16.63915, 179.096609 -16.433984, 179.413509 -16.379054, 180 -16.067133, 180 -16.555217, 179.364143 -16.801354)), ((-179.917369 -16.501783, -180 -16.555217, -180 -16.067133, -179.79332 -16.020882, -179.917369 -16.501783)))
```

 The bbox stored in the index ends up looking as follows instead:

```
nodeMinX = -180.0
nodeMinY = -18.28799
nodeMaxX = 180.0
nodeMaxY = -16.020882
```

Visually, it looks like this (no intent to be precise, actually drawn by hand). Blue is the search area, red is the Fiji bbox:

image

So wondering, is this a feature of the Flatgeobuf format, or something that happened by accident?

bjornharrtell:
Hmm that does seem strange and unexpected. I would like to try and find the root cause.

bjornharrtell:
Most probable there is a bug somewhere in logic that determines/serialises the bbox on index generation, perhaps due to the dateline wrapping.

aaime:
To be found in the code that wrote the "countries.fgb" file. It might have been written using something other than the Java version (or it could be that the issue is shared across the various implementations).
Git history says you contributed that test file, any chance you remember how it was generated?

bjornharrtell:
Most likely GDAL.

aaime:
So, generally speaking, the code is ok, and the issue might be in the test data?
If so, I guess I can merge the improvement?

bjornharrtell:
@aaime agreed

bjornharrtell:
@aaime just figured out that there is no bug, the bbox is correct, because Fiju is a single feature feature multipolygon and existing on both side of the world and the index structure is one feature = one bbox it can only contain both parts of the multipolygon by making such stretched out bbox. So, Fiji can only be eliminated by a secondary filter on actual geometry.

aaime:
Ok so... generally speaking, we should not use the optimized count that only scans the index when:
* The data contains multi-geometries
* The result of using the index has both sides touching the dateline

In that case we cannot tell in advance, if the geometries happen to just touch the datelines on both ends, or there is one that spans the dateline, with bits on both ends, which results in its index bbox spanning the entire world.

But in the other cases, it should be fine. What do you think?

bjornharrtell:
I'm not sure it's worth doing anything, the count is approximate anyway?

aaime:
While implementing this I've reviewed the practice in other stores. PostGIS has a "loose bbox" parameter that enables the "bbox vs bbox" intersection logic, which is enabled by default for performance reasons, but when disabled, turns the "BBOX" filter into a "ST_Intersects" instead, which is slower but exact, "actual geometry vs bbox".

All the stores that do spatial filtering in memory, using the BBOXImpl logic, are in fact using "loose bbox", internally BBOXImpl does the "bbox vs bbox" comparison. So when going over the spatial index, we apply the same logic, and the count matches the actual number of features returned. In the case we're discussing, we are losing consistency.

We can be "loose" in matching features, but it needs to be consistent, in this particular case it's not. This may surprise paging clients, that expect as given total count, and get a different amount of features.
Had that been ok, I could have optimized also the other spatial filters that check for "interference" between geometries, e.g., intersection, overlap, containment, returning a count that's bigger than the features actually returned.

bjornharrtell:
I don't understand why/how results for postgis loose bbox for this case would be different. Are they really?

aaime:
Ooh, fair point. Indeed, they are not:

```
ogr2ogr -f PostgreSQL PG:"dbname='gttest' host=localhost user='cite' password='cite'" countries.fgb -lco GEOMETRY_NAME=geom -lco FID=gid -lco SPATIAL_INDEX=GIST -nln countries -overwrite

# ... getting into the database

gttest=# \d countries
                                       Table "public.countries"
 Column |            Type             | Collation | Nullable |                Default                 
--------+-----------------------------+-----------+----------+----------------------------------------
 gid    | integer                     |           | not null | nextval('countries_gid_seq'::regclass)
 id     | character varying           |           |          | 
 name   | character varying           |           |          | 
 geom   | geometry(MultiPolygon,4326) |           |          | 

gttest=# select name from countries where ST_MakeEnvelope(112,-44, 154, -11) && geom;
   name    
-----------
 Australia
 Fiji
(2 rows)
```

Hum.. ok, let's leave it at that. We should probably have a flag in the WFS that forces the replacement of the BBOX filter with an intersects one, in case the admin wants precise results (which is what the WFS standard states, BBBOX is just a shorthand for an intersects, but GeoServer has been doing otherwise for 20 years because the performance difference is significant).

PR #4725, opened 2024-04-15
[GEOT-7556] Generate javadoc for Maven deploy

roarbra:
[GEOT-7556](https://osgeo-org.atlassian.net/browse/GEOT-7556) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

There is no Javadoc jars published alongside the jar and sources. I believe this is because of a missing javadoc:jar call before the install.

In addition I've removed a warning for wrong configuration (additionalparam), and the keywords option is just necessary for the javadoc that are published to a web site.

# Checklist

### DIFF
--- a/build/release/build_release.sh
+++ b/build/release/build_release.sh
@@ -20,6 +20,7 @@ function usage() {
   echo " SKIP_BUILD : Skips main release build"
   echo " SKIP_TAG : Skips tag on release branch"
   echo " SKIP_JAVADOCS : Skips javadoc build"
+  echo " MAVEN_FLAGS : Flags used in calls to mvn"
 }
 
 # parse options
@@ -149,7 +150,7 @@ fi
 
 target=`pwd`/target
 
-# build the javadocs
+# build the javadocs for website
 if [ "$SKIP_JAVADOCS" != true ]; then
   echo "building javadocs"
   pushd modules > /dev/null
--- a/build/release/publish_release.sh
+++ b/build/release/publish_release.sh
@@ -77,7 +77,7 @@ git checkout tags/$tag -b rel_$tag
 if [ "$SKIP_DEPLOY"  != true ]; then
   echo "deploying with $MAVEN_FLAGS"
 
-  mvn clean deploy -DskipTests -Dall $MAVEN_FLAGS
+  mvn clean generate-sources javadoc:jar deploy -DskipTests -Dall $MAVEN_FLAGS
 fi
 
 # get <major.minor> for sf release dir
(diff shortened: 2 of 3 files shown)


build/release/build_release.sh
```diff
@@ -142,14 +143,14 @@ popd > /dev/null
 if [ "$SKIP_BUILD" != true ]; then
   export MAVEN_OPTS="-Xmx2048m"
   echo "building release"
-  mvn $MAVEN_FLAGS -DskipTests -Dall clean install
+  mvn $MAVEN_FLAGS -DskipTests -Dall clean generate-sources javadoc:jar install
```
aaime, line 146: I'm wondering if this helps at all, given that release publish does this:

```
mvn clean deploy -DskipTests -Dall
```

See that "clean"? I'm guessing that if you want to deploy javadocs, the building of them should be done there?
Also, the sources are already packaged, don't think we need to bother the build_release script with them, e.g.:

image

@jodygarnett knows the release procedures much better than I do, and might provide some more suggestions.

mprins:
probably you want something like

```xml

            <plugin>
                <artifactId>maven-javadoc-plugin</artifactId>
                <executions>
                    <execution>
                        <id>attach-javadoc</id>
                        <goals>
                            <goal>jar</goal>
                        </goals>
                        <phase>package</phase>
   ...

```
to "attach" the javadoc jar files as artifacts (possibly using `verify` phase instead)

roarbra:
> probably you want something like
> 
> ```
>             <plugin>
>                 <artifactId>maven-javadoc-plugin</artifactId>
>                 <executions>
>                     <execution>
>                         <id>attach-javadoc</id>
>                         <goals>
>                             <goal>jar</goal>
>                         </goals>
>                         <phase>package</phase>
>    ...
> ```
> 
> to "attach" the javadoc jar files as artifacts (possibly using `verify` phase instead)

I'm afraid that would make them build each time users are building with Maven. Don't think everyone would be happy about that. Having an exclusive goal is fine by me, as long as it's called within the release cycle.

build/release/build_release.sh
roarbra: > I'm wondering if this helps at all, given that release publish does this:
> 
> ```
> mvn clean deploy -DskipTests -Dall
> ```

It was somewhat hard to figure out how the release process worked. You're absolutely correct that it was unnecessary to add that generate-sources. Not sure why I did that.

Where in the scripts is this mvn-call you are referring to?

roarbra: > 
> It was somewhat hard to figure out how the release process worked. You're absolutely correct that it was unnecessary to add that generate-sources. Not sure why I did that.
> 

I remember now. And it is actually necessary. It isn't the sources.jar that is generated, but the Java-sources we generate from xsd and such. Javadoc-generation will fail without that goal.

jodygarnett: Can we add comment so we do not have to remember this detail again in the future?

PR #4727, opened 2024-04-17
[GEOT-7557] Provide the ability for downstream applications (e.g. GeoServer) to append additional querystring parameters e.g. AuthKey to every request made by the HTTPClient

petersmythe:
[GEOT-7557](https://osgeo-org.atlassian.net/browse/GEOT-7557) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

https://osgeo-org.atlassian.net/browse/GEOT-7557

Supports github.com/geoserver/geoserver/pull/7552

_Background: There is a GeoServer option to configure a WMTS data store with HTTP Headers (as well as basic username/password), however the HTTP Headers option is not available for WMS data stores (only username/password)._

In addition, specifically for AuthKeys/API keys, some WMS/WMTS servers are not aware of authentication mechanisms that they sit behind.  They then return Capabilities documents that do not include the authentication mechanism appended to the published URLs.  Clients that properly parse the Capabilities documents (QGIS, GeoServer) find that they can securely request the Capabilities documents however break on the subsequent Map/Tile requests.

I propose to add an AuthKey option to both WMS and WMTS data stores.

# Checklist

### DIFF
--- a/modules/extension/wms/src/main/java/org/geotools/ows/wms/WebMapServer.java
+++ b/modules/extension/wms/src/main/java/org/geotools/ows/wms/WebMapServer.java
@@ -411,6 +411,27 @@ public class WebMapServer extends AbstractOpenWebService<WMSCapabilities, Layer>
      * specified by serverURL.
      *
      * @param serverURL a URL that points to the capabilities document of a server
+     * @param httpClient The client to be used when performing HTTP requests
+     * @param hints A map of hints. Can be used to control some aspects of the XML parsing, see
+     *     {@link XMLHandlerHints} for a reference
+     * @param headers A map of headers. These will be added when making the HTTP request
+     * @throws IOException if there is an error communicating with the server
+     * @throws ServiceException if the server responds with an error
+     */
+    public WebMapServer(
+            final URL serverURL,
+            final HTTPClient httpClient,
+            Map<String, Object> hints,
+            Map<String, String> headers)
+            throws IOException, ServiceException {
+        super(serverURL, httpClient, null, hints, headers);
+    }
+
+    /**
+     * Creates a new WebMapServer instance and attempts to retrieve the Capabilities document
+     * specified by serverURL.
+     *
+     * @param serverURL a URL that points to the capabilities document of a server
      * @param timeout a time to be wait a server response
      * @throws IOException if there is an error communicating with the server
      * @throws ServiceException if the server responds with an error
--- a/modules/extension/wms/src/main/java/org/geotools/ows/wms/WebMapServer.java
+++ b/modules/extension/wms/src/main/java/org/geotools/ows/wms/WebMapServer.java
@@ -407,8 +407,7 @@ public class WebMapServer extends AbstractOpenWebService<WMSCapabilities, Layer>
     }
 
     /**
-     * Creates a new WebMapServer instance and retrieve the Capabilities document.
-     * specified by serverURL.
+     * Creates a new WebMapServer instance and retrieve the Capabilities document specified by serverURL.
      *
      * @param serverURL a URL that points to the capabilities document of a server
      * @param httpClient The client to be used when performing HTTP requests
--- a/modules/extension/wms/src/main/java/org/geotools/ows/wms/WebMapServer.java
+++ b/modules/extension/wms/src/main/java/org/geotools/ows/wms/WebMapServer.java
@@ -407,7 +407,7 @@ public class WebMapServer extends AbstractOpenWebService<WMSCapabilities, Layer>
     }
 
     /**
-     * Creates a new WebMapServer instance and attempts to retrieve the Capabilities document
+     * Creates a new WebMapServer instance and retrieve the Capabilities document.
      * specified by serverURL.
      *
      * @param serverURL a URL that points to the capabilities document of a server
--- a/modules/extension/wms/src/main/java/org/geotools/ows/wms/WebMapServer.java
+++ b/modules/extension/wms/src/main/java/org/geotools/ows/wms/WebMapServer.java
@@ -407,7 +407,8 @@ public class WebMapServer extends AbstractOpenWebService<WMSCapabilities, Layer>
     }
 
     /**
-     * Creates a new WebMapServer instance and retrieve the Capabilities document specified by serverURL.
+     * Creates a new WebMapServer instance and retrieve the Capabilities document specified by
+     * serverURL.
      *
      * @param serverURL a URL that points to the capabilities document of a server
      * @param httpClient The client to be used when performing HTTP requests
--- a/modules/library/http/src/main/java/org/geotools/http/AbstractHttpClient.java
+++ b/modules/library/http/src/main/java/org/geotools/http/AbstractHttpClient.java
@@ -21,6 +21,8 @@ import java.io.InputStream;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.net.URLConnection;
+import java.util.Map;
+import java.util.StringJoiner;
 
 /**
  * A base class for HTTPClient, that implements everything except the get and post methods.
@@ -33,7 +35,7 @@ public abstract class AbstractHttpClient implements HTTPClient {
 
     protected String password;
 
-    protected String authKey;
+    protected Map<String, Object> extraParams;
 
     protected int connectTimeout;
 
@@ -57,13 +59,13 @@ public abstract class AbstractHttpClient implements HTTPClient {
     }
 
     @Override
-    public void setAuthKey(String authKey) {
-        this.authKey = authKey;
+    public void setExtraParams(Map<String, Object> extraParams) {
+        this.extraParams = extraParams;
     }
 
     @Override
-    public String getAuthKey() {
-        return this.authKey;
+    public Map<String, Object> getExtraParams() {
+        return this.extraParams;
     }
 
     @Override
@@ -103,9 +105,15 @@ public abstract class AbstractHttpClient implements HTTPClient {
         return tryGzip;
     }
 
-    protected static URL appendURL(URL oldUrl, String appendQuery) throws MalformedURLException {
+    protected static URL appendURL(URL oldUrl, Map<String, Object> appendQuery)
+            throws MalformedURLException {
         String oldQuery = oldUrl.getQuery();
-        String newQuery = oldQuery != null ? oldQuery + "&" + appendQuery : appendQuery;
+
+        StringJoiner stringJoiner = new StringJoiner("&");
+        appendQuery.forEach((key, value) -> stringJoiner.add(key + "=" + value));
+        String query = stringJoiner.toString();
+
+        String newQuery = oldQuery != null ? oldQuery + "&" + query : query;
 
         return new URL(
                 oldUrl.getProtocol(),
--- a/modules/library/http/src/main/java/org/geotools/http/AbstractHttpClient.java
+++ b/modules/library/http/src/main/java/org/geotools/http/AbstractHttpClient.java
@@ -107,16 +107,11 @@ public abstract class AbstractHttpClient implements HTTPClient {
         String oldQuery = oldUrl.getQuery();
         String newQuery = oldQuery != null ? oldQuery + "&" + appendQuery : appendQuery;
 
-        String newUrlString =
-                oldUrl.getProtocol() + "://" + oldUrl.getAuthority() + oldUrl.getPath();
-        if (newQuery != null && !newQuery.isEmpty()) {
-            newUrlString += "?" + newQuery;
-        }
-        if (oldUrl.getRef() != null) {
-            newUrlString += "#" + oldUrl.getRef();
-        }
-
-        return new URL(newUrlString);
+        return new URL(
+                oldUrl.getProtocol(),
+                oldUrl.getAuthority(),
+                oldUrl.getPort(),
+                oldUrl.getPath() + "?" + newQuery);
     }
 
     protected boolean isFile(URL url) {
--- a/modules/library/http/src/main/java/org/geotools/http/AbstractHttpClient.java
+++ b/modules/library/http/src/main/java/org/geotools/http/AbstractHttpClient.java
@@ -109,7 +109,7 @@ public abstract class AbstractHttpClient implements HTTPClient {
 
         return new URL(
                 oldUrl.getProtocol(),
-                oldUrl.getAuthority(),
+                oldUrl.getHost(),
                 oldUrl.getPort(),
                 oldUrl.getPath() + "?" + newQuery);
     }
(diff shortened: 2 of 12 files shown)


petersmythe:
Downstream integration build with GeoServer fails with:

`/home/runner/geoserver/src/main/src/test/java/org/geoserver/test/http/MockHttpClient.java:[26,8] org.geoserver.test.http.MockHttpClient is not abstract and does not override abstract method setAuthKey(java.lang.String) in org.geotools.http.HTTPClient`

jodygarnett:
This is a geotools change so it will need a GeoTools ticket please.

aaime:
@roarbra do you have time to look at this change? Seems up your alley.

modules/library/http/src/main/java/org/geotools/http/AbstractHttpClient.java
```diff
@@ -90,6 +103,22 @@ public boolean isTryGzip() {
...
+        String newUrlString =
+                oldUrl.getProtocol() + "://" + oldUrl.getAuthority() + oldUrl.getPath();
+        if (newQuery != null && !newQuery.isEmpty()) {
+            newUrlString += "?" + newQuery;
+        }
+        if (oldUrl.getRef() != null) {
+            newUrlString += "#" + oldUrl.getRef();
+        }
+
+        return new URL(newUrlString);
```
mprins, line 119: This seems to be missing the TCP port.
Also please use one of the dedicated URL constructors such as https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/net/URL.html#%3Cinit%3E(java.lang.String,java.lang.String,int,java.lang.String)
It prevents a lot of - possibly error prone - String juggling

mprins (changes requested):

roarbra:
This was strange. Having a AuthKey that are appended to the query string of the url is not a common way of handling authentication. Rather the opposite. It's not recommended to have secrets in the url.

Having said that. I like the idea of having a way to add query parameters to the url from within Geoserver, but it shouldn't be named authKey and it should be open for more than just a single value. Like for instance a Map<String, Object> parameters.

petersmythe:
> This was strange. Having a AuthKey that are appended to the query string of the url is not a common way of handling authentication. Rather the opposite. It's not recommended to have secrets in the url.
> 
> Having said that. I like the idea of having a way to add query parameters to the url from within Geoserver, but it shouldn't be named authKey and it should be open for more than just a single value. Like for instance a Map<String, Object> parameters.

I agree, an AuthKey/API key in the querystring is not ideal (certainly AWS API Gateway refuses to implement it for that reason), however, it is still an extremely popular way to provide REST authentication, including our own GeoServer extension called AuthKey.  I think the HTTP headers option is far better but thought while I was busy with fixing WMS stores https://github.com/geoserver/geoserver/pull/7552 to include an API key option.  Should I rather undo the work?

I currently can't see the benefit of making it more generic, but I understand what you're asking for.  Should I spend the extra time to support a Map<String, Object> `additionalQuerystringParams` rather, and then the GeoServer panel can still be called AuthKey/API Key?

mprins:
> Having a AuthKey that are appended to the query string of the url is not a common way of handling authentication

Actually there are quite a few providers that have an "API user key" in the url as a query param, eg. Google, Thunderforest, Bing to name a few, just like there are many that actually provide a unique url with a token eg. ttps://s.map5.nl/map/bbpp.6488ab/service?layer=map5topo_simple&style=default&tilematrixset=geonovum_grid&Service=WMTS&Request=GetTile&Version=1.0.0&Format=image%2Fjpeg&TileMatrix=11&TileCol=967&TileRow=1032

aaime:
@petersmythe the API here could be a Map, and the GUI in GeoServer could be a simple text area that accepts a property file syntax. With the right Wicket converter/model you can turn that into the expected map object.

petersmythe:
> @petersmythe the API here could be a Map, and the GUI in GeoServer could be a simple text area that accepts a property file syntax. With the right Wicket converter/model you can turn that into the expected map object.

Understood, I will make the GeoTools API more generic and keep the GeoServer implementation more specific - just a single AuthKey text field is required, I believe.

modules/library/http/src/main/java/org/geotools/http/AbstractHttpClient.java
petersmythe: Done

petersmythe:
Would you mind taking another look over the PR @roarbra ?

modules/library/http/src/main/java/org/geotools/http/AbstractHttpClient.java
petersmythe: getAuthority returns both the hostname and port, but I have separated them to be more clear.

roarbra:
> Would you mind taking another look over the PR @roarbra ?

That looks a lot better. Thanks!

modules/library/http/src/main/java/org/geotools/http/AbstractHttpClient.java
```diff
@@ -90,6 +105,23 @@ public boolean isTryGzip() {
         return tryGzip;
     }
 
+    protected static URL appendURL(URL oldUrl, Map<String, Object> appendQuery)
+            throws MalformedURLException {
+        String oldQuery = oldUrl.getQuery();
+
+        StringJoiner stringJoiner = new StringJoiner("&");
+        appendQuery.forEach((key, value) -> stringJoiner.add(key + "=" + value));
```
aaime, line 113: Don't we need URL encoding here, for both key and value?

petersmythe: Done

mprins (approved):

petersmythe:
Thank you Mark.

Reviewers, please do not merge this PR yet, it should happen in conjunction with https://github.com/geoserver/geoserver/pull/7552 otherwise the downstream integration test will fail

modules/extension/wms/src/main/java/org/geotools/ows/wms/WebMapServer.java
```diff
@@ -406,6 +406,27 @@ public WebMapServer(final URL serverURL, final HTTPClient httpClient, Map<String
         super(serverURL, httpClient, null, hints);
     }
 
+    /**
+     * Creates a new WebMapServer instance and attempts to retrieve the Capabilities document
```
jodygarnett, line 410:
```suggestion
     * Creates a new WebMapServer instance and retrieve the Capabilities document.
```

modules/library/http/src/main/java/org/geotools/http/AbstractHttpClient.java
```diff
@@ -90,6 +108,33 @@ public boolean isTryGzip() {
         return tryGzip;
     }
 
+    protected static URL appendURL(URL oldUrl, Map<String, Object> appendQuery)
```
jodygarnett, line 111: We need a javadoc here as this ends up being a semi-public method due to protected visibility.

modules/library/http/src/main/java/org/geotools/http/AbstractHttpClient.java
```diff
@@ -90,6 +108,33 @@ public boolean isTryGzip() {
...
+        String oldQuery = oldUrl.getQuery();
+
+        StringJoiner stringJoiner = new StringJoiner("&");
+        appendQuery.forEach(
+                (key, value) -> {
+                    try {
+                        stringJoiner.add(
+                                URLEncoder.encode(key, "UTF-8")
+                                        + "="
+                                        + URLEncoder.encode(value.toString(), "UTF-8"));
```
jodygarnett, line 122: What kind of content is expected here as values? We do have the GeoTools converters API that can handle information with more of a text focus then `toString()` method.

modules/library/http/src/main/java/org/geotools/http/HTTPClient.java
```diff
@@ -133,6 +133,12 @@ default HTTPResponse get(URL url, Map<String, String> headers) throws IOExceptio
     /** @param password the HTTP BASIC Authentication password */
     public void setPassword(String password);
 
+    /** @return the optional querystring ExtraParams, or {@code null} if not set */
```
jodygarnett, line 136:
```suggestion
    /** Optional query parameters.
      * @return querystring ExtraParams, or {@code null} if not set
      */
```

We tend to prefer empty objects, for example an empty map, rather than null as per GeoTools style guide.

Reference "do not return null" https://docs.geotools.org/latest/developer/conventions/code/null.html

modules/library/http/src/main/java/org/geotools/http/AbstractHttpClient.java
jodygarnett: What is expected if oldQuery has some of the same keys?

modules/library/http/src/main/java/org/geotools/http/SimpleHttpClient.java
```diff
@@ -120,13 +120,19 @@ public HTTPResponse post(
 
     private URLConnection openConnection(URL finalURL, Map<String, String> headers)
             throws IOException {
+        Map<String, Object> extraParams = getExtraParams();
+        if (extraParams != null) {
```
jodygarnett, line 124: See [do not return null](https://docs.geotools.org/latest/developer/conventions/code/null.html).

If you like we could update this convention to make use of Optional, as an explicit alternative to returning null.

jodygarnett (changes requested):
Some API clarifications, and feedback based on geotools style guideline.

The need for this API change is clear; for URL arithmetic there may be better served by adding a utility method to [org.geotools.URLs](https://docs.geotools.org/latest/javadocs/org/geotools/util/URLs.html). I would like to avoid having similar utility methods scattered around the codebase each needing test cases written.

modules/library/http/src/main/java/org/geotools/http/AbstractHttpClient.java
petersmythe: The authkey value could have been a number, but there was no reason for it, so I changed it from Map<String, Object> to Map<String, String>, thus avoiding .toString()

Duplicate keys are now tested in https://github.com/geotools/geotools/compare/main...petersmythe:geotools:auth-for-cascaded-stores#diff-6ac812206449a240da095b61e785458b9c9837991a0e4531afafe37c1f46dffdR142 and work as expected (i.e. both values are present)

modules/library/http/src/main/java/org/geotools/http/SimpleHttpClient.java
petersmythe: Done, changing default to Collections.emptyMap()

petersmythe:
GeoServer still successfully builds with the PR at this time:

image

petersmythe:
Merging with https://github.com/geoserver/geoserver/pull/7552

aaime:
The backport to `31.x` failed:
```
The process '/usr/bin/git' failed with exit code 1
```
<details>
<summary>stderr</summary>

```
error: could not apply 5282fc86a2... Add AuthKey by appending to URL in HTTPClient
hint: After resolving the conflicts, mark them with
hint: "git add/rm <pathspec>", then run
hint: "git cherry-pick --continue".
hint: You can instead skip this commit with "git cherry-pick --skip".
hint: To abort and get back to the state before "git cherry-pick",
hint: run "git cherry-pick --abort".

```
</details>

<details>
<summary>stdout</summary>

```
[backport-4727-to-31.x 70bb412cc6] To support HTTP Headers see github.com/geoserver/geoserver/pull/7552
 Author: Peter Smythe <peter@afrigis.co.za>
 Date: Wed Apr 17 12:57:08 2024 +0200
 1 file changed, 21 insertions(+)
Auto-merging modules/library/http/pom.xml
CONFLICT (content): Merge conflict in modules/library/http/pom.xml
Auto-merging modules/plugin/http-commons/pom.xml
CONFLICT (content): Merge conflict in modules/plugin/http-commons/pom.xml
Auto-merging modules/plugin/mongodb/pom.xml
CONFLICT (content): Merge conflict in modules/plugin/mongodb/pom.xml

```
</details>

To backport manually, run these commands in your terminal:
```bash
# Fetch latest updates from GitHub
git fetch
# Create a new working tree
git worktree add .worktrees/backport-31.x 31.x
# Navigate to the new working tree
cd .worktrees/backport-31.x
# Create a new branch
git switch --create backport-4727-to-31.x
# Cherry-pick the merged commit of this pull request and resolve the conflicts
git cherry-pick 2c64415c672901095533b5c75fe25a8e7bcbc196,5282fc86a283dac457aa0ee9ba9e8ed4c9c4d93b,e75b49195d58f635b1bb0a9788956934a00aaee8,8c9e05cc9dbb273a049b473dc3a7a3fec4bfccde,cbcd07ce43e7d2caa109579685edf3c273508890,c2e8460f88fd72d59af4ea56d1376ee97e34fef0,bfab9b24983eb6f6207581c548b5589dc5a2c5f4,9cac6a53965d2cb69780222849bbcafec3aee201,da90c1663aa0de0ed0622b5b041866c772bf214c,20993c164f6d46e6e495cd956207ce199a4696b6,f22ce7a77a9d47c49721cf5fa450e57e404862fe,711c099db0ed9fbdb67adba7113d63e1cace7910,2ad536e9990454870e7bde15386b7fe2424f20a9,8dd3aa97f3d77237eeb1b87f2f1bcdde2276bf53,0f8d61aed046585763591a66619e968b0de2826f,6fdafbd7e9b5eb3ff7510be19c07718bed4f4551,5b00f2ef7c4000ce86083f6877933963a86593cb,a754bea7af2a2856b3601db6ff4c67e4673d07f2,741510a74bf1660f8e21fee3a86ccc08e22069f0,d1807c6a4f9406fbb9bb414c18db14e2f93c00e7,16fd45686aa23eb7a3e9c70d348b1381d39ad683,a29aabc35cef1610ed095bf0a20ae9ccc9b749cd,5a04c3a3bd328e8c3d25cc6afde012449105ecac,d79060463c0649659439eafeefa6c83633cb7d3c,7ef354bf7f00c59297ba74b1672a1ab279a64603,97321be859c58b1e323153c89ee38bf99b39620c,7ad1c8177df106e826f843b4c3dc49ae9d1e8c07,40ea5b296e5904939a2c0a3a2f8d2259d40691b8,f7560f897e42757028223e2d951c8cc2f35156ec,0f22d69735f97201e6132650f50e1c2be31b5934
# Push it to GitHub
git push --set-upstream origin backport-4727-to-31.x
# Go back to the original working tree
cd ../..
# Delete the working tree
git worktree remove .worktrees/backport-31.x
```
Then, create a pull request where the `base` branch is `31.x` and the `compare`/`head` branch is `backport-4727-to-31.x`.

PR #4728, opened 2024-04-21
Test Eclipse suggested fix

petersmythe:
The current GeoTools main branch shows a build error in Eclipse 2023-06 IDE with Java 11, and this is the recommended fix for it.

According to ChatGPT:

> In this line:
> 
> (Comparable<? super T>) is a bounded wildcard. It means that range.getMinValue() should be cast to something that is a superclass of T and is also comparable to that superclass.
> This ensures that you're casting range.getMinValue() to a type that is comparable to T or its superclasses. This can be important when T is not itself comparable, but its superclass is.

# Checklist


aaime:
Looks good, merged. 

If we want to prevent these incompatibilities from happening in the first place, a dedicated GitHub action building with the Eclipse compiler plugin may be an option, e.g. dedicated profile in the main pom following these two suggestions:

* https://stackoverflow.com/questions/33164976/using-eclipse-java-compiler-ecj-in-maven-builds
* https://maven.apache.org/plugins/maven-compiler-plugin/non-javac-compilers.html

and a Github action that enables the profile.

aaime:
Merged on 31.x too

PR #4730, opened 2024-04-24
[GEOT-7559] App-Schema throws an error when mappings have duplicate names, even when they come from includes

turingtestfail:
[GEOT-7559](https://osgeo-org.atlassian.net/browse/GEOT-7559) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

deep equality trimmed down

fixed equality compare and started on test

Cleanup

more cleanup

moved equal to local

  

# Checklist

### DIFF
--- a/modules/extension/app-schema/app-schema/src/main/java/org/geotools/data/complex/AppSchemaDataAccess.java
+++ b/modules/extension/app-schema/app-schema/src/main/java/org/geotools/data/complex/AppSchemaDataAccess.java
@@ -144,11 +144,20 @@ public class AppSchemaDataAccess implements DataAccess<FeatureType, Feature> {
                     // check both mappings and the registry, because the data access is
                     // only registered at the bottom of this constructor, so it might not
                     // be in the registry yet
-                    throw new DataSourceException(
-                            "Duplicate mappingName or targetElement across FeatureTypeMapping instances detected.\n"
-                                    + "They have to be unique, or app-schema doesn't know which one to get.\n"
-                                    + "Please check your mapping file(s) with mappingName or targetElement of: "
-                                    + name);
+                    // if the mapping is from an include and is identical to the previous one, log
+                    // and continue
+                    if (isIncludeAndDeepSame(mapping)) {
+                        LOGGER.fine(
+                                "Duplicate mappingName detected, but no error thrown because it comes from an include: "
+                                        + name);
+                        continue;
+                    } else {
+                        throw new DataSourceException(
+                                "Duplicate mappingName or targetElement across FeatureTypeMapping instances detected.\n"
+                                        + "They have to be unique, or app-schema doesn't know which one to get.\n"
+                                        + "Please check your mapping file(s) with mappingName or targetElement of: "
+                                        + name);
+                    }
                 }
                 this.mappings.put(name, mapping);
                 // if the type is not a feature, it should be wrapped with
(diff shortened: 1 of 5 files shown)


modules/extension/app-schema/app-schema/src/main/java/org/geotools/data/complex/AppSchemaDataAccess.java
```diff
@@ -169,6 +179,57 @@ public AppSchemaDataAccess(Set<FeatureTypeMapping> mappings, boolean hidden)
...
+                        compareMapping = appSchemaDataAccess.getMappingByName(name);
+                    }
+                } catch (IOException | ClassCastException e) {
+                    LOGGER.fine(
+                            "Could not get mapping from registry for "
+                                    + name
+                                    + " while trying to test for duplicates");
+                    return false;
+                }
+            }
```
nmco, line 213: I'm uncertain about this aspect, but if my understanding is accurate, the objective is to prevent App-Schema from raising errors about duplicate mappings when the exact same feature type mapping is encountered multiple times due to being included by different mappings.

It appears that we're conducting a search in two different places:

1. Firstly, in the mappings hashmap of this `AppSchemaDataAccess` instance. I presume that an `AppSchemaDataAccess` instance is associated with a root mapping in App-Schema.
2. Secondly, in the `DataAccessRegistry` registry, which encompasses all the root mappings registered in App-Schema.

The search for duplicates encompasses both:

```java
if (this.mappings.containsKey(name) || DataAccessRegistry.hasName(name)) {
```

I would suggest adding comments in the code to elucidate the process. It seems that the initial check is performed within the current mappings, as if the mapping is found there, it implies it has already been validated against the common registry.

modules/extension/app-schema/app-schema/src/main/java/org/geotools/data/complex/AppSchemaDataAccess.java
```diff
@@ -169,6 +179,57 @@ public AppSchemaDataAccess(Set<FeatureTypeMapping> mappings, boolean hidden)
...
+     *
+     * @param mapping the mapping to compare
+     * @param compareMapping the mapping to compare with
+     * @return true if the mappings are deep equal
+     */
+    private boolean mappingsDeepCompare(
+            FeatureTypeMapping mapping, FeatureTypeMapping compareMapping) {
+        if (mapping == compareMapping) return true;
+        return Objects.equals(mapping.getAttributeMappings(), compareMapping.getAttributeMappings())
+                && Objects.equals(mapping.getMappingName(), compareMapping.getMappingName());
```
nmco, line 230: The same comment applies here as above. So, it's not only the mapping name, which is optional, but we also need to check the target element name.

That said, should we also check for other properties of the mappings? Such as data sources, normalization, etc.? Would it make more sense to implement this equality check in the FeatureTypeMapping class itself?

nmco (commented):
Thx @turingtestfail, I add some review comments.

modules/extension/app-schema/app-schema/src/main/java/org/geotools/data/complex/config/AppSchemaDataAccessConfigurator.java
```diff
@@ -121,6 +121,9 @@ public class AppSchemaDataAccessConfigurator {
 
     public static final String PROPERTY_REPLACE_OR_UNION = "app-schema.orUnionReplace";
 
+    /** Whether the mapping is for an include. */
+    private final Boolean isInclude;
```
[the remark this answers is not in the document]
turingtestfail: There were a lot of tests that have calls to the method so to avoid a lot of rewrites I overloaded the method with this as an argument.  The latest push switched it to have an Optional wrapper to make it more explicit.  Does that work?

modules/extension/app-schema/app-schema/src/main/java/org/geotools/data/complex/AppSchemaDataAccess.java
```diff
@@ -169,6 +179,57 @@ public AppSchemaDataAccess(Set<FeatureTypeMapping> mappings, boolean hidden)
         register();
     }
 
+    /**
+     * Checks if the mapping is an include and is identical to the previous one.
```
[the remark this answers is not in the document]
turingtestfail: Comments added

modules/extension/app-schema/app-schema/src/main/java/org/geotools/data/complex/AppSchemaDataAccess.java
```diff
@@ -169,6 +179,57 @@ public AppSchemaDataAccess(Set<FeatureTypeMapping> mappings, boolean hidden)
...
+     * @return true if the mapping is an include and is identical to the previous one
+     */
+    private boolean isIncludeAndDeepSame(FeatureTypeMapping mapping) {
+        // Not an include, so no need to check
+        if (!Boolean.TRUE.equals(mapping.isInclude())) {
+            return false;
+        }
+        FeatureTypeMapping compareMapping = null;
+        // Get the mapping to compare with
+        if (mapping.getMappingName() != null) {
```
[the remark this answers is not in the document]
turingtestfail: Added logic to check of target element.

modules/extension/app-schema/app-schema/src/main/java/org/geotools/data/complex/AppSchemaDataAccess.java
turingtestfail: Comments added.

modules/extension/app-schema/app-schema/src/main/java/org/geotools/data/complex/AppSchemaDataAccess.java
turingtestfail: Target check added.  Data source id and normalization comparisons added.  Previously this was the equals method in FeatureTypeMapping but the accompanying hash method was causing test failures so I moved it here.

turingtestfail: Actually I just pushed a version with the logic moved to the equals method in FeatureTypeMapping and a corrected hash method and the tests are passing.

modules/extension/app-schema/app-schema/src/main/java/org/geotools/data/complex/config/AppSchemaDataAccessConfigurator.java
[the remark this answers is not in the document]
turingtestfail: Switched to plain boolean primitive type with default of false.

sikeoka:
This PR may be failing GeoServer builds due to an API change in AppSchemaDataAccessConfigurator:
`Error: 8,713 [ERROR] /home/runner/work/geoserver/geoserver/src/community/smart-data-loader/src/main/java/org/geoserver/smartdataloader/data/SmartDataLoaderDataAccessFactory.java:[205,74] incompatible types: org.geotools.data.complex.config.DataAccessMap cannot be converted to boolean`

turingtestfail:
Yes, was made aware of that today.  This PR provides a fix on the GeoServer
side:
https://github.com/geoserver/geoserver/pull/7633

PR #4752, opened 2024-05-14
Updated FeatureJSON to (optionally) use Geometry as the geometry type…

jwomeara:
… if mixed geometries are present in a feature collection.

I ran into an issue recently where I was trying to read a feature collection containing features with a mix of geometry types (point, linestring, and polygon).  Using the `readFeatureCollectionSchema` method in FeatureJSON to extract the schema as a simple feature type, the simple feature type returned specifies `Point` as the geometry type.  This causes the GeoJSON geometries to be incorrectly read as `Point` instead of their original Geometry type.  

For my application I was able to create a new SimpleFeatureType from the returned SimpleFeatureType and update the geometry type to be `Geometry`.  Doing that, all of the original geometry types are preserved.  It would be nice if I didn't have to do that though, which is why I created this PR.  

The unit tests I added showcase the issue with the original logic, and the fixed output from the proposed PR.

Sent my license agreement in, and tried creating an issue in jira, but I don't have sufficient permissions at the moment.  Happy to do that if that's resolved.

Thanks!

### DIFF
--- a/modules/unsupported/geojson/src/main/java/org/geotools/geojson/feature/FeatureJSON.java
+++ b/modules/unsupported/geojson/src/main/java/org/geotools/geojson/feature/FeatureJSON.java
@@ -492,12 +492,56 @@ public class FeatureJSON {
      * @param nullValuesEncoded if the input has null values encoded. If this flag is set to true
      *     and the GeoJSON doesn't actually encode null values, only the first feature attributes
      *     will be discovered.
+     * @param checkAllGeometryTypes if we should check the geometry type of each feature, and not
+     *     just the first. If this flag is set to true and the GeoJSON contains a mix of geometry
+     *     types, the type will be set to {@link Geometry}. If this flag is set to false the
+     *     geometry type of the first feature will be used.
+     * @return The feature collection schema
+     * @throws IOException In the event of a parsing error or if the input json is invalid.
+     */
+    public SimpleFeatureType readFeatureCollectionSchema(
+            Object input, boolean nullValuesEncoded, boolean checkAllGeometryTypes)
+            throws IOException {
+        return GeoJSONUtil.parse(
+                new FeatureTypeHandler(nullValuesEncoded, checkAllGeometryTypes), input, false);
+    }
+
+    /**
+     * Reads the {@link SimpleFeatureType} of a GeoJSON feature collection. In the worst case, it
+     * will parse all features searching for attributes not present in previous features.
+     *
+     * @param input The input. See {@link GeoJSONUtil#toReader(Object)} for details.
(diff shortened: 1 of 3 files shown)


mprins:
This is not according to spec: https://datatracker.ietf.org/doc/html/rfc7946#section-1.4

Also here doesn't seem to be a test for reading such a GeoJSON? Ad it breaks backward compatibility; the original signatures should at least be preserved

jwomeara:
Hey there, thank you so much for the feedback.  Let me make sure I understand your concerns.

> This is not according to spec: https://datatracker.ietf.org/doc/html/rfc7946#section-1.4
Thanks for sharing this as I hadn't seen it before.  I don't think this changes anything w.r.t. this PR though as I am not changing how the JSON itself is formatted, only how GeoJSON schema is represented in memory/Java.  The current code actually *misrepresents* the GeoJSON in the case outlined in my tests.  

In the tests I added you can see that when we read and write the example GeoJSON using the existing code, the geometry types of all of the features in the collection change from their original type to just `Point`.  Whereas with this change, when you read/write the feature collection the input matches the output exactly.  

> Also here doesn't seem to be a test for reading such a GeoJSON?
My PR does contain tests which showcase the issue.  I pointed that out in the original PR description.  Here is a direct link to the tests that I added.
https://github.com/geotools/geotools/blob/f7e9eec9eb36c70eb5fecc0c5d14ae0a8d28e12a/modules/unsupported/geojson/src/test/java/org/geotools/geojson/FeatureJSONTest.java#L1384
https://github.com/geotools/geotools/blob/f7e9eec9eb36c70eb5fecc0c5d14ae0a8d28e12a/modules/unsupported/geojson/src/test/java/org/geotools/geojson/FeatureJSONTest.java#L1432

> Ad it breaks backward compatibility; the original signatures should at least be preserved
Thank you for pointing this out.  I added the methods with the original signatures back into my branch so as not to break backwards compatibility.  

At a minimum, I want to draw attention to the presence of a simple bug in the `FeatureJSON.readFeatureCollectionSchema` method which accepts an `InputStream` as it's first argument.  In that method, you allow the user to specify `nullValuesEncoded`, but that value is ignored and `false` is always passed.

jwomeara:
I appreciate you enabling the github actions for this.  Looking at the logs for the failing task, I see that the geoserver TabsTest is failing.  Looking at some of the other PRs against geoserver, it looks like this might be an existing or intermittent issue, unrelated to this change specifically.  If that's not the case, please let me know and I will start to dig into that test to figure out what's going on.  Thanks again!

aaime:
Restarted build. Also noticing this is the old GeoJSON module... it will eventually be removed in favor of gt-geojson-core, I'd recommend putting effort in migrating to it, rather than investing in this one.

PR #4785, opened 2024-05-27
[Backport 30.x] [GEOT-7579] FlatGeobuf handle java.util.Date and iso correctness

aaime:
[GEOT-7579](https://osgeo-org.atlassian.net/browse/GEOT-7579) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

Backport #4768
 **Authored by:** @bjornharrtell


bjornharrtell:
Hmm it's something flaky about the timestamps serialization... I saw it earlier and cut off fraction on seconds to fix but pops up here. Strange... I should have a look.

bjornharrtell:
To clarify I think it's in the tests rather than in actual logic the flakyness exists.

bjornharrtell:
@aaime I'm worried this flakyness can happen randomly regardless of backport/java versions etc.

aaime:
Yes, looks like it's random. Ouch, we don't need another random failure. Could you rework the test so that this does not happen=

PR #4787, opened 2024-05-27
FlatGeobuf fix AttributeRoundtripTest exotic1 flakyness

bjornharrtell:


bjornharrtell:
@aaime this is hopefully fixing the flakyness as seen after https://github.com/geotools/geotools/pull/4768, as well improving the robustness of the test in general.

PR #4788, opened 2024-05-27
[GEOT-7586] GeoPackage output contains invalid field types

davidblasby:
[GEOT-7586](https://osgeo-org.atlassian.net/browse/GEOT-7586) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

GeoServer issue - https://osgeo-org.atlassian.net/browse/GEOS-11415

Geotools can create GeoPackages with incorrect column types.  Typically this will happen if you have a JDBC (i.e. postgis) layer and then request a WFS with GeoPackage output.

The issue is caused because some Datastores (i.e. postgis) add JDBC_NATIVE_TYPENAME and JDBC_NATIVE_TYPE column metadata (descriptor userdata) to the schema.  The JDBCDatastore then treats this as "gospel" and will create the DB schema with these specific types.  For most JDBC systems, this is correct.  However, the [GeoPackage spec](https://www.geopackage.org/spec/) (Table 1) defines specific sqllite types for columns instead of "normal" JDBC Column types.

You can see that JDBCDatastore#getSQLTypeNames() absolutely uses JDBC_NATIVE_TYPENAME and JDBC_NATIVE_TYPE over any conversion.

https://github.com/geotools/geotools/blob/main/modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java#L3392-L3396
and
https://github.com/geotools/geotools/blob/main/modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStore.java#L3400-L3402

This PR fixes this by updating the schema creation component (and nothing else) of GeoPackage.java so it:

1. copies the incoming schema in the #create() 
2. removes the JDBC_NATIVE_TYPENAME and JDBC_NATIVE_TYPE descriptor metadata (userdata)

This is only done when you #create()  a new schema.

# Checklist

- [n/a] [Documentation](https://github.com/geotools/geotools/tree/main/docs) has been updated (if change is visible to end users).

### DIFF
--- a/modules/plugin/geopkg/src/main/java/org/geotools/geopkg/GeoPackage.java
+++ b/modules/plugin/geopkg/src/main/java/org/geotools/geopkg/GeoPackage.java
@@ -17,6 +17,8 @@
 package org.geotools.geopkg;
 
 import static java.lang.String.format;
+import static org.geotools.jdbc.JDBCDataStore.JDBC_NATIVE_TYPE;
+import static org.geotools.jdbc.JDBCDataStore.JDBC_NATIVE_TYPENAME;
 import static org.geotools.jdbc.util.SqlUtil.prepare;
 
 import java.io.Closeable;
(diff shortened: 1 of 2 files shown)


aaime:
Good hunting. Can you have a GeoTools Jira for a GeoTools PR? This allows to publish this change amongst the GeoTools release notes too.

jodygarnett:
As Andrea says good debugging; I have added backport labels as this is a big fix.

jodygarnett (approved):

aaime:
@davidblasby for next time, best to have it on both ends, we have two separate groups of users here, they both deserve being informed on release.

jodygarnett:
Aside: if you wish the fix listed for geoserver release notes you would want a bug report there as well :)

There is work like this extra bug reports so that release manager does not get overwhelmed so much.  It is up to you if you wish the fix highlighted for GeoServer also.

davidblasby:
I created an issue in GEOS Jira - https://osgeo-org.atlassian.net/browse/GEOS-11415

PR #4793, opened 2024-05-28
[GEOT-7537] GeoPackage: TIMESTAMP must be DATETIME according to the GeoPackage spec

davidblasby:
[GEOT-7537](https://osgeo-org.atlassian.net/browse/GEOT-7537) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

Please give this a good review!

This fixes some problems with Timestamp support in GeoPackages.  According to the spec, these should be in a "DATETIME" column (TEXT) and in a specific format (GMT - ISO-8601 - YYYY-MM-DDTHH:MM[:SS.SSS]Z).

I had some issues with - GeoPkgDatetimeTest#testBetween_timestamp - I added a few comments.  Its quite difficult to read the test case - the month is changing (unexpected) and its hard to see the difference.  Also, hard to interpret 00:00:00 24:00:00.  I think its correct. 

NOTE: This includes the test case from the bug report.  Is this ok in terms of IP?

NOTE: Timestamp are timezone sensitive.  In the test case I explicitly use GMT.  If a timestamp is create, its likely to be in the machine's timezone.  This isn't necessarily an issue, but worth mentioning. 

NOTE: I noticed that there is some type of support for "TIME" - however I don't think it working correctly and its unclear (from the geopkg spec) if its even allowed.   

# Checklist

- [n/a] [Documentation](https://github.com/geotools/geotools/tree/main/docs) has been updated (if change is visible to end users).


davidblasby:
The problematic test case is failing.  I expect this is a timezone issue.  I will look into it.

jodygarnett:
A: We can always ask in the bug report, but I am much more relaxed about non production code (test does not from part of the running library)

modules/plugin/geopkg/src/test/java/org/geotools/geopkg/GeoPackageTestWriteDatetime.java
```diff
@@ -0,0 +1,168 @@
...
+ *    modify it under the terms of the GNU Lesser General Public
+ *    License as published by the Free Software Foundation;
+ *    version 2.1 of the License.
+ *
+ *    This library is distributed in the hope that it will be useful,
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ *    Lesser General Public License for more details.
+ *
+ *    @author Stijn Goedertier
```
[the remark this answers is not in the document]
jodygarnett: We do day in contributing for small changes like docs (and tests) a committer can contribute their behalf. But yes we can ask.

modules/plugin/geopkg/src/main/java/org/geotools/geopkg/GeoPkgDialect.java
```diff
@@ -97,14 +102,19 @@ public class GeoPkgDialect extends PreparedStatementSQLDialect {
 
     private JSONArrayIO jsonArrayIO = new JSONArrayIO();
 
+    // geopkg - format will be ISO-8601 - YYYY-MM-DDTHH:MM[:SS.SSS]Z
+    SimpleDateFormat geopkgDateFormat = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'");
```
jodygarnett, line 106: Nice, I never remember if format is thread safe or can be a constant.

aaime: Not thread safe

aaime:
Many failures in GeoPackage online tests:

```
Error:    GeoPkgGroupByVisitorOnlineTest>JDBCGroupByVisitorOnlineTest.testGroupByWithAggregateAllValuesNull:368->JDBCGroupByVisitorOnlineTest.testGroupByWithAggregateAllValuesNull:354->JDBCGroupByVisitorOnlineTest.genericGroupByTestTest:476->JDBCGroupByVisitorOnlineTest.genericGroupByTestTest:504 » DateTimeParse
Error:    GeoPkgGroupByVisitorOnlineTest>JDBCGroupByVisitorOnlineTest.testNumericHistogram:159->JDBCGroupByVisitorOnlineTest.genericGroupByTestTest:463->JDBCGroupByVisitorOnlineTest.genericGroupByTestTest:476->JDBCGroupByVisitorOnlineTest.genericGroupByTestTest:504 » DateTimeParse
Error:    GeoPkgGroupByVisitorOnlineTest>JDBCGroupByVisitorOnlineTest.testTimestampHistogram:170->JDBCGroupByVisitorOnlineTest.testTimestampHistogram:189->JDBCGroupByVisitorOnlineTest.genericGroupByTestTest:463->JDBCGroupByVisitorOnlineTest.genericGroupByTestTest:476->JDBCGroupByVisitorOnlineTest.genericGroupByTestTest:504 » DateTimeParse
Error:    GeoPkgGroupByVisitorOnlineTest>JDBCGroupByVisitorOnlineTest.testTimestampHistogramDateWithDifferenceInDays:544->JDBCGroupByVisitorOnlineTest.testTimestampHistogramDateWithDifferenceInSpecificTimeUnits:578->JDBCGroupByVisitorOnlineTest.genericGroupByTestTest:463->JDBCGroupByVisitorOnlineTest.genericGroupByTestTest:476->JDBCGroupByVisitorOnlineTest.genericGroupByTestTest:504 » DateTimeParse
Error:    GeoPkgGroupByVisitorOnlineTest>JDBCGroupByVisitorOnlineTest.testTimestampHistogramDateWithDifferenceInHours:550->JDBCGroupByVisitorOnlineTest.testTimestampHistogramDateWithDifferenceInSpecificTimeUnits:578->JDBCGroupByVisitorOnlineTest.genericGroupByTestTest:463->JDBCGroupByVisitorOnlineTest.genericGroupByTestTest:476->JDBCGroupByVisitorOnlineTest.genericGroupByTestTest:504 » DateTimeParse
Error:    GeoPkgGroupByVisitorOnlineTest>JDBCGroupByVisitorOnlineTest.testTimestampHistogramDateWithDifferenceInMinutes:556->JDBCGroupByVisitorOnlineTest.testTimestampHistogramDateWithDifferenceInSpecificTimeUnits:578->JDBCGroupByVisitorOnlineTest.genericGroupByTestTest:463->JDBCGroupByVisitorOnlineTest.genericGroupByTestTest:476->JDBCGroupByVisitorOnlineTest.genericGroupByTestTest:504 » DateTimeParse
Error:    GeoPkgGroupByVisitorOnlineTest>JDBCGroupByVisitorOnlineTest.testTimestampHistogramDateWithDifferenceInSeconds:562->JDBCGroupByVisitorOnlineTest.testTimestampHistogramDateWithDifferenceInSpecificTimeUnits:578->JDBCGroupByVisitorOnlineTest.genericGroupByTestTest:463->JDBCGroupByVisitorOnlineTest.genericGroupByTestTest:476->JDBCGroupByVisitorOnlineTest.genericGroupByTestTest:504 » DateTimeParse
Error:    GeoPkgGroupByVisitorOnlineTest>JDBCGroupByVisitorOnlineTest.testUnkonwnFunction:81->JDBCGroupByVisitorOnlineTest.genericGroupByTestTest:463->JDBCGroupByVisitorOnlineTest.genericGroupByTestTest:476->JDBCGroupByVisitorOnlineTest.genericGroupByTestTest:504 » DateTimeParse
```

You may be asking why they are even online...  I believe it's because the native sqlite libs are not available on all platform, but not sure.

davidblasby:
There’s a lot of problems! Dates in SQLite dont behave as you would expect. I am finding that the same query in GT and in a SQLite db viewer app give different results.  


If there is a literal date in a filter, what time zone is it in?

aaime:
I believe in the JVM current time zone. @ianturton years ago worked on a flag to handle timezone-less dates and timestamps I believe, he might have some useful feedback to this one.

davidblasby:
Ok - making some progress.

I was trying to make use of "subsec" (sub-second accuracy for datetimes) - however, they are only supported after sqlite 3.42.0 (we are using 3.41.2.2).  However, upgrading to a later version caused a bunch of problems - likely some type of compatibility issues.

I've left in code that will use "subsec" if you are using 3.42.0 or higher.  However, upgrading to 3.42 doesn't work...  I'll leave that for someone in the future.  NOTE: 3.43.0 fixed a typo in one of the method names (readUnCommitted - with 2 T's).

I managed to get all the test cases working with geopkg - but I've broken some of the other DBs.  I expect this has to do with how geopackage is required to handle and store dates.  I will move my specific tests to a geopackage-only test and go back to the old JDBC test.

jodygarnett:
David dates and timezones are really hard in Java. 

Here are two classes from GeoNetwork codebase that may serve as reference point:
- https://github.com/geonetwork/core-geonetwork/blob/main/domain/src/main/java/org/fao/geonet/domain/ISODate.java
- https://github.com/geonetwork/core-geonetwork/blob/main/domain/src/test/java/org/fao/geonet/domain/ISODateTest.java

This shows the use of java ZonedDateTime as something stable for parsing.

modules/plugin/geopkg/src/main/java/org/geotools/geopkg/GeoPackage.java
```diff
@@ -345,6 +345,16 @@ static void createFunctions(Connection cx) throws SQLException {
...
+        try (var stmt = cx.createStatement()) {
+            try (var rs =
+                    stmt.executeQuery(
+                            "select exists(select 1 from pragma_function_list where name='ST_MinX')")) {
+                if (rs.getInt(1) == 1) {
+                    return; // already exist
+                }
+            }
+        }
+
```
aaime, line 357: This seems to be unrelated to the DATETIME management, but came from an attempt at upgrading to newer versions of SQLite, that don't accept a second registration of the same function, right? Nevertheless, does not hurt.

modules/plugin/geopkg/src/main/java/org/geotools/geopkg/GeoPkgDialect.java
```diff
@@ -107,9 +109,25 @@ public GeoPkgDialect(JDBCDataStore dataStore) {
         geomWriterConfig = new GeoPkgGeomWriter.Configuration();
     }
 
+    public String sqlLiteVersion;
+    public boolean supportsSubSeconds;
+
+    public void setupVersion(Connection cx) throws SQLException {
```
aaime, line 115: Does not seem to need being public, could use a javadoc, and it's not a "setup of the version" but a way to recognize it and then decide if certain abilities are found.

I'd recommend checking the PostGIS dialect it does this a lot, uses GeoTools own Version class, has version comparison, and caches the result to avoid inspecting the version over and over.

modules/plugin/geopkg/src/main/java/org/geotools/geopkg/GeoPkgDialect.java
```diff
@@ -688,9 +707,15 @@ public void setValue(
...
+                // 2020-02-19 23:00:00.0  --> 2020-02-19T23:00:00.0Z
                 // We need the "Z" or sql lite will interpret the value as local time
-                ps.setString(column, value.toString() + "Z");
+                // geopkg - format will be ISO-8601 - YYYY-MM-DDTHH:MM[:SS.SSS]Z
+                if (value != null) {
+                    var v = ((Timestamp) value).toInstant().toString();
+                    ps.setString(column, v);
+                } else {
+                    ps.setString(column, null);
+                }
```
aaime, line 718: Unnecessary, the code at the beginning of the method is already checking for null, using ``setNull``, and exiting.

modules/plugin/geopkg/src/main/java/org/geotools/geopkg/GeoPkgDialect.java
```diff
@@ -995,6 +1020,21 @@ public Object convertValue(Object value, AttributeDescriptor ad) {
...
+            // geopkg - format will be ISO-8601 - YYYY-MM-DDTHH:MM[:SS.SSS]Z
+            var strValue = (String) value;
+            // this is for support with "bad" geopkgs
+            if (!strValue.endsWith("Z")) {
+                strValue += "Z";
+            }
+            Instant instant = Instant.parse(strValue);
+            Timestamp timestamp = Timestamp.from(instant);
+            return timestamp; // this will be in local time
+        }
```
aaime, line 1036: Makes sense, but isn't this overly rigid? I would not read anymore geopackages written by a previous version of GeoTools, or files written without strict adherence to the the GeoPackage standard. I'd keep the code within try/catch, and fall back on Converters (what ``super.convertValue`` is doing already) to try and interpret values in a less strict version.

Generally speaking, it's what we try to do in GeoTools data reading in general, be lenient and try to read the input, rather than playing "specification cop".

modules/plugin/geopkg/src/main/java/org/geotools/geopkg/GeoPkgFilterToSQL.java
```diff
@@ -51,9 +51,9 @@ public GeoPkgFilterToSQL(GeoPkgDialect dialect) {
      *
      * <p>There is different handling for Date (DATE) and Timestamp (DATETIME).
      *
-     * <p>For Timestamp (DATETIME), we use the datetime(XYZ, 'utc'):
+     * <p>For Timestamp (DATETIME), we use the datetime(XYZ, 'utc','subsec'):
```
aaime, line 54: Only if subsec is supported, though.

modules/plugin/geopkg/src/main/java/org/geotools/geopkg/GeoPkgFilterToSQL.java
```diff
@@ -74,9 +74,12 @@ public String escapeName(String name) {
             if (Time.class.isAssignableFrom(binding)) {
                 return "time(" + super_result + ",'utc')";
             } else if (Timestamp.class.isAssignableFrom(binding)) {
+                if (dialect.supportsSubSeconds) {
```
aaime, line 77: Please encapsulate and use a getter?

modules/plugin/geopkg/src/test/java/org/geotools/geopkg/GeoPkgDatetimeTest.java
```diff
@@ -278,13 +278,24 @@ public void testBetween() throws IOException, CQLException {
...
+    //  '2020-03-19 24:00:00' is the start of the next day
+    //
     @Test
     public void testBetween_timestamp() throws IOException, CQLException {
         Filter between =
                 ECQL.toFilter("time BETWEEN '2020-02-19 23:00:00' AND '2020-03-19 00:00:00'");
 
         SimpleFeatureSource fs = gpkg.getFeatureSource(gpkg.getTypeNames()[0]);
         SimpleFeatureCollection features = fs.getFeatures(between);
+        // var list = new ListFeatureCollection(features).stream().collect(Collectors.toList());
```
aaime, line 298: Don't leave around commented out code please.

modules/plugin/geopkg/src/test/java/org/geotools/geopkg/GeoPkgDatetimeTest.java
```diff
@@ -278,13 +278,24 @@ public void testBetween() throws IOException, CQLException {
...
+    //     time                lower      upper   result
+    // 2020-02-19T22:00:00Z     FAIL       PASS    FAIL
+    // 2020-02-19T23:00:00Z     PASS       PASS    PASS
+    // 2020-03-19T00:00:00Z     PASS       PASS    PASS
+    // 2020-03-19T01:00:00Z     PASS       FAIL    FAIL
+    // 2020-02-20T02:00:00Z     PASS       PASS    PASS
+    //
+    //  '2020-03-19 00:00:00' is the start of the day
+    //  '2020-03-19 24:00:00' is the start of the next day
+    //
```
aaime, line 290: I don't think this is needed any more?

aaime (changes requested):

davidblasby:
@aaime  - thanks for review.  I will work on this today.
1. I will remove all the version stuff.  Its, currently, not in use.  I think upgrading to a new version will be quite a bit of work and this is just going to confuse whoever does this (I didnt get the new sqlite version working).
2. I will look at a the old GT and make sure they are readable - and also test with QGIS.

Thanks for feedback!

davidblasby:
Ok, tested the output (from the `GeoPackageTestWriteDateTime.java` test case).

NOTE: geopackage file must end in `.gpkg` for qgis/ogr to recognize the file.

1. Tested with QGIS
2. Tested with `ogrinfo` (i expect QGIS is using ogr to read the geopackage).
3. Tested with DB Browser (sqlite GUI)

image

modules/plugin/geopkg/src/main/java/org/geotools/geopkg/GeoPkgDialect.java
```diff
@@ -685,12 +691,25 @@ public void setValue(
...
+                //                if (!time.endsWith("Z")) {
+                //                     //convert from local to Zulu
+                //                    var time1 = convert(value, Time.class);
+                //                    var gmt = new Date(time1.getTime() - Calendar.getInstance()
+                // .getTimeZone().getOffset(time1.getTime()));
+                //                    SimpleDateFormat sdf = new SimpleDateFormat("HH:mm:ss'Z'");
+                //                    var gmt2 = sdf.format(gmt);
+                //                    ps.setString(column, gmt2);
+                //                    break;
+                //                }
```
aaime, line 704: Please don't leave commented out code hanging around

aaime (changes requested):

davidblasby:
@aaime - I added a test case for handling of the "old" way GT was putting datetime columns in.  Also, removed the commented out code.

PR #4802, opened 2024-06-05
Add Nullable annotation to GeometryClipper

ricksurya:
Hi all, I'm proposing a change to add a `@Nullable` annotation to the `GeometryClipper#clip` and `Geometry#clipSafe` functions. This would help prevent NPEs by callers of this library method. Would appreciate any input, thank you!

# Checklist

### DIFF
--- a/modules/library/main/pom.xml
+++ b/modules/library/main/pom.xml
@@ -267,6 +267,7 @@
     <dependency>
       <groupId>com.google.code.findbugs</groupId>
       <artifactId>jsr305</artifactId>
+      <version>3.0.0</version>
     </dependency>
   </dependencies>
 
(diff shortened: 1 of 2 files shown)


modules/library/main/pom.xml
```diff
@@ -264,6 +264,11 @@
       <artifactId>hamcrest-library</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.google.code.findbugs</groupId>
+      <artifactId>jsr305</artifactId>
+      <version>3.0.0</version>
```
mprins, line 270: this is an obsoleted version from 2014

mprins (changes requested):
the API doc of `#clipSafe(...)` already tells you that the result may be `null`... just introducing the annotation in this one class does not do justice to all those other locations where a `null` may be returned

ricksurya:
@mprins yup I see the API doc already specifies that. I'm still biased to add this annotation as this allows errorprone lint checkers to warn if anyone missed the doc. I feel this adds an extra layer of safety with minimal downside.

PR #4803, opened 2024-06-06
[GEOT-7576] Use HttpClient in SchemaCache

vuilleumierc:
[GEOT-7576](https://osgeo-org.atlassian.net/browse/GEOT-7576) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

Currently the XML SchemaCache makes direct HTTP calls for downloading XML schemas.
The goal of this PR is for the SchemaCache to use the geotools HttpClient instead. This ensures that if a specific HttpClient is set (see https://github.com/geotools/geotools/blob/main/docs/user/library/http/index.rst) it will be used for all outgoing HTTP calls.

Since there was a bit of logic for following redirections in SchemaCache, I have moved it to the SimpleHttpClient so that the default behavior is preserved. This won't be the case for applications already using another specific HttpClient. Note that after testing it appeared that `java.net.HttpURLConnection` already follows redirections if the protocol is the same (i.e. HTTP or HTTPS) so this bit of code seems to be needed only for HTTP to HTTPS redirection.

Also changed a URL in the online tests which was causing a SSL error.

# Checklist

### DIFF
--- a/modules/library/http/src/main/java/org/geotools/http/SimpleHttpClient.java
+++ b/modules/library/http/src/main/java/org/geotools/http/SimpleHttpClient.java
@@ -20,6 +20,9 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.net.HttpURLConnection;
+import java.net.MalformedURLException;
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.net.URL;
 import java.net.URLConnection;
 import java.nio.charset.StandardCharsets;
@@ -42,6 +45,10 @@ public class SimpleHttpClient extends AbstractHttpClient implements HTTPProxy {
     private static final Logger LOGGER = Logging.getLogger(SimpleHttpClient.class);
 
     private static final int DEFAULT_TIMEOUT = 30; // 30 seconds
+
+    /** Max number of attempts to follow a redirect to avoid redirect loop * */
+    private static final int MAX_FOLLOW_REDIRECT =
+            Integer.getInteger("org.geotools.http.simpleHttpClient.maxFollowRedirect", 16);
 
     /**
      * A SimpleHttpClient should be initiated by a call to
@@ -66,12 +73,11 @@ public class SimpleHttpClient extends AbstractHttpClient implements HTTPProxy {
         if (isFile(url)) {
             return this.createFileResponse(url);
         }
-        URLConnection connection = openConnection(url, headers);
-        if (connection instanceof HttpURLConnection) {
-            ((HttpURLConnection) connection).setRequestMethod("GET");
-        }
-        connection.connect();
 
+        URLConnection connection = this.get(url, headers, 0);
+        if (connection == null) {
+            return null;
+        }
         return new DefaultHttpResponse(connection);
     }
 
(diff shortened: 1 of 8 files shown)


modules/library/http/src/main/java/org/geotools/http/SimpleHttpClient.java
```diff
@@ -171,4 +196,36 @@ private URLConnection openConnection(URL finalURL, Map<String, String> headers)
...
+            HttpURLConnection connection, Map<String, String> headers, int redirectionCount)
+            throws IOException {
+        String redirect = connection.getHeaderField("Location");
+        if (redirect == null) {
+            LOGGER.warning("Tried to follow redirect but no url was provided in Location header");
+        } else if (redirectionCount < MAX_FOLLOW_REDIRECT) {
+            redirectionCount++;
+            try {
+                URL redirectURL = new URI(redirect).toURL();
+                LOGGER.info("Following redirect to " + redirect);
```
aaime, line 216: This log level seems to high, it's a detail, but it would show up, for example, in the GeoServer default log level (which logs INFO and above).

modules/library/xml/src/main/java/org/geotools/xml/resolver/SchemaCache.java
```diff
@@ -219,48 +217,38 @@ static byte[] download(URI location) {
...
-            if (responseCode != HttpURLConnection.HTTP_OK) {
-                if (hasRedirect(responseCode))
-                    return followRedirect(connection, blockSize, redirectionCount);
-                LOGGER.warning(
-                        String.format(
-                                "Unexpected response \"%d %s\" while downloading %s",
-                                connection.getResponseCode(),
-                                connection.getResponseMessage(),
-                                location.toString()));
+            HTTPClient httpClient = HTTPClientFinder.createClient();
```
aaime, line 235: Interesting bit here... it switches from HttpURLConnection to something that may, or may not, be a SimpleHttpClient (depends on the default system hints AFAIK).

I'm not sure what would happen if the multithreaded client (commons-http base) is used instead. Does it have a redirection limit? In GeoServer, when we want to ensure a simple HTTP client is used, we do the following:

```java
client =
                    HTTPClientFinder.createClient(new Hints(Hints.HTTP_CLIENT, SimpleHttpClient.class));
```

However, at the same time, doing so would break the tests below, which rely on default system hints to inject a mock client. Hum... maybe just verify that if the multithreaded client is used, we do not risk a loop of redirects?

modules/library/xml/src/main/java/org/geotools/xml/resolver/SchemaCache.java
```diff
@@ -219,48 +217,38 @@ static byte[] download(URI location) {
...
+            httpClient.setConnectTimeout(downloadTimeout);
+            httpClient.setReadTimeout(downloadTimeout);
+            HTTPResponse httpResponse;
+            try {
+                httpResponse = httpClient.get(url);
+            } catch (IOException e) {
+                // don't throw an exception if status code >= 400
```
aaime, line 242: The previous code was logging something when an unexpected response was received:

```
LOGGER.warning(
                        String.format(
                                "Unexpected response \"%d %s\" while downloading %s",
                                connection.getResponseCode(),
                                connection.getResponseMessage(),
                                location.toString()));
```

aaime (changes requested):
Overall looks good, thank you! 
See some comments below about small changes in behavior.

vuilleumierc:
Thank you for reviewing @aaime  - I did the two changes about logging and added a test for the multithreaded client (see inline comment)

modules/library/xml/src/main/java/org/geotools/xml/resolver/SchemaCache.java
vuilleumierc: Good point, I did some testing and found out the Apache client does check for circular redirections and raises an IOException in this case (in this class: https://www.javadoc.io/doc/org.apache.httpcomponents/httpclient/4.3.5/org/apache/http/impl/execchain/RedirectExec.html, default is 50 redirects). With the current code the exception is caught and logged by the schema cache (similarly to >= 400 status codes).
I added a test for this, let me know what you think. Thanks in advance

aaime:
Looks good to me, merging.

PR #4805, opened 2024-06-06
[GEOT-7589] Fix the JsonArrayDelegation for Postgres to support searching in root level array

Mike7311:
[GEOT-7589](https://osgeo-org.atlassian.net/browse/GEOT-7589) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

This PR is aimed to add support for searching root level arrays using postgres implementation of JsonArrayContains.

Current implementation is using the @> operator, however to use it with array at root level the searched object should be wrapped in square brackets. Below is an examples of working queries for array wrapped in object and non wrapped:
```
("jsonColumn"::jsonb @> '{ "arrayStrValues": ["EL2"] }'::jsonb
```
for {"arrayStrValues":["EL1","EL2","EL3"]} and
```
("jsonColumn"::jsonb @> '[{ "arrayStrValues": "EL2" }]'::jsonb
```
for ["EL1","EL2","EL3"]

That is, to properly resolve this we should know if we are targeting root array during building query, which is currently not possible.

The new approach relies on jsonb_path_exists, which does not have this drawback.
```
jsonb_path_exists(jsonColumn::jsonb, '$ ? (@.arrayStrValues == "EL2")')
```

The check "PostGIS online tests / build (pull_request)" is failing because this functionality was added in Postgres 12, and VM that runs the check apparently uses older version, resulting in "ERROR: function jsonb_path_exists(jsonb, unknown) does not exist". ~~Thus this functionality is disabled by default, and can be enabled by using the "org.geotools.data.postgis.useJsonPathExists" property.~~
Code will use new implementation for versions >= 12, and old implementation for previous ones.

# Checklist

### DIFF
--- a/docs/user/library/jdbc/postgis.rst
+++ b/docs/user/library/jdbc/postgis.rst
@@ -64,6 +64,9 @@ Advanced
 |                        | be encoded into their SQL equivalent           |
 +------------------------+------------------------------------------------+
 
+By default JsonArrayContains function will be delegated to @> operator. However for postgres versions >= 12 it will be delegated to jsonPathExists function.
+jsonPathExists is able to search values in the root level array.
+
 Example use::
   
   params.put(PostgisNGDataStoreFactory.LOOSEBBOX, true );
--- a/modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/FilterToSqlHelper.java
+++ b/modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/FilterToSqlHelper.java
@@ -16,6 +16,8 @@
  */
 package org.geotools.data.postgis;
 
+import static org.geotools.data.postgis.PostGISDialect.PGSQL_V_12_0;
+
 import java.io.IOException;
 import java.io.StringWriter;
 import java.io.Writer;
@@ -26,6 +28,7 @@ import java.sql.Time;
 import java.sql.Timestamp;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.StringJoiner;
 import java.util.concurrent.TimeUnit;
 import java.util.function.BiConsumer;
 import java.util.stream.Collectors;
@@ -108,6 +111,7 @@ import org.geotools.jdbc.JDBCDataStore;
 import org.geotools.jdbc.PreparedFilterToSQL;
 import org.geotools.jdbc.PrimaryKeyColumn;
 import org.geotools.jdbc.SQLDialect;
+import org.geotools.util.Version;
 import org.geotools.util.factory.Hints;
 import org.locationtech.jts.geom.Envelope;
 import org.locationtech.jts.geom.Geometry;
@@ -126,9 +130,17 @@ class FilterToSqlHelper {
     Writer out;
     boolean looseBBOXEnabled;
     boolean encodeBBOXFilterAsEnvelope;
+    boolean jsonPathExistsSupported;
 
     public FilterToSqlHelper(FilterToSQL delegate) {
         this.delegate = delegate;
+        this.jsonPathExistsSupported = false;
+    }
+
+    public FilterToSqlHelper(FilterToSQL delegate, Version pgVersion) {
+        this.delegate = delegate;
+        this.jsonPathExistsSupported =
+                postgresMajorVersionIsEqualOrGreaterThan(pgVersion, PGSQL_V_12_0);
     }
 
     public static FilterCapabilities createFilterCapabilities(boolean encodeFunctions) {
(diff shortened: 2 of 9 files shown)


mprins:
> The check "PostGIS online tests / build (pull_request)" is failing because this functionality was added in Postgres 12, and VM that runs the check apperently uses older version, resulting in "ERROR: function jsonb_path_exists(jsonb, unknown) does not exist". Thus this functionality is disabled by default, and can be enabled by using the "org.geotools.data.postgis.useJsonPathExists" property.

I'd be OK with updating the version in the workflow (https://github.com/geotools/geotools/blob/f0c9961dc4d40c5acfce2169fab92805738de5ea/.github/workflows/postgis_online.yml#L17) to a current release as PG 11 will be EOL in 5 months.

ianturton:
It would be better to make the change version dependent. Just because a pg
version is eol doesn't mean people will stop using it. But updating the
work flow version is also useful.

modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/FilterToSqlHelper.java
```diff
@@ -703,23 +704,66 @@ public String buildJsonFromStrPointer(String[] pointers, int index, Expression e
...
 
+    /**
+     * Encode the jsonArrayContains call using the jsonb_path_exists function from Postgres
+     *
+     * @param jsonArrayContains
+     * @throws IOException
+     */
     private void encodeJsonArrayContains(Function jsonArrayContains) throws IOException {
+        String useJsonPathExists =
+                System.getProperty("org.geotools.data.postgis.useJsonPathExists");
```
aaime, line 715: If you keep the variable, please use a constant for "org.geotools.data.postgis.useJsonPathExists" and use it consistently here and in the test code.

docs/user/library/jdbc/postgis.rst
```diff
@@ -77,3 +77,7 @@ An optimization on spatial queries can be applied, to avoid a bug with PostGIS q
 This optimization can be enabled using a system-wide default from the command line::
   
   java -Dorg.geotools.data.postgis.largeGeometriesOptimize=true
+
+By default JsonArrayContains will be delegated to @> operator. It is possible to instead delegate it to jsonb_path_exists, which was added in postgres version 12, by setting following property::
+
+  java -Dorg.geotools.data.postgis.useJsonPathExists=true
```
aaime, line 83: The decision could be automatic. The PostgisDialect is responsible for creating the FilterToSQL, and it has a field that contains the PostgreSQL version. The field could theoretically be null, but in practice, it won't be by the time you try to use it for a filter to SQL in real usage.

Suggestion: in both the prepared and non prepared SQL dialects, pass down a flag to indicate whether to use the new syntax, and default to the old one if the field happens to be null.
In theory if missing you could grab a connection from the store, but don't do it: the filter to SQL creation might be called from a method that already has one, and grabbing 2+ connections to complete any operations is deadlock prone (every thread manages to get the first connection, the pool gets depleted, no thread can grab the second, deadlock happens).

aaime (changes requested):
See comment below. The flag is not bad, but there is a possibility to just recognize the PostgreSQL version instead.

modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/FilterToSqlHelper.java
Mike7311: Removed it

docs/user/library/jdbc/postgis.rst
Mike7311: Updated the both PG dialects to check the version and pass it down to FilterToSQL.

modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/PostGISDialect.java
```diff
@@ -1407,14 +1409,26 @@ public void encodeGeometryValue(Geometry value, int dimension, int srid, StringB
...
+        PostgisFilterToSQL sql =
+                new PostgisFilterToSQL(this, postgresMajorVersionIsHigherThen(PGSQL_V_12_0));
         sql.setLooseBBOXEnabled(looseBBOXEnabled);
         sql.setEncodeBBOXFilterAsEnvelope(encodeBBOXFilterAsEnvelope);
         sql.setFunctionEncodingEnabled(functionEncodingEnabled);
         sql.setEscapeBackslash(escapeBackslash);
         return sql;
     }
 
+    public boolean postgresMajorVersionIsHigherThen(Version version) {
```
mprins, line 1421: it actually returns the value from a equal or greater than comparison... 

```suggestion
    public boolean postgresMajorVersionIsEqualOrGreaterThan(Version version) {
```
Aside from that not sure if this needs to be `public`; I would think it is only used in the PG module

modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/FilterToSqlHelper.java
```diff
@@ -126,9 +127,11 @@ class FilterToSqlHelper {
     Writer out;
     boolean looseBBOXEnabled;
     boolean encodeBBOXFilterAsEnvelope;
+    boolean pg12OrNewer;
 
-    public FilterToSqlHelper(FilterToSQL delegate) {
+    public FilterToSqlHelper(FilterToSQL delegate, boolean pg12OrNewer) {
```
mprins, line 132: to keep backwards API compatibility you might want to keep the old constructor around with a default value of `false` for `pg12OrNewer`. eg.

```java
public FilterToSqlHelper(FilterToSQL delegate) {
        this.delegate = delegate;
        this.pg12OrNewer = false;
}
```
Perhaps with `@Deprecated` annotation

aaime: I would also just get a Version field, otherwise these flags will profilerate... 10 years down the line we might have a pg120Newer, pg150Newer, pg180Newer and so on. A single Version that can be compared to a reference value is probably going to be better.

Mike7311: Returned old constructor and changed parameter to Version

modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/PostGISDialect.java
Mike7311: Renamed and moved to FilterToSqlHelper (since it is used only there).

modules/plugin/jdbc/jdbc-postgis/src/test/java/org/geotools/data/postgis/PostGISJsonOnlineTest.java
```diff
@@ -442,6 +444,87 @@ public void testJSONArrayContainsFunction() throws Exception {
...
+                                        feature.getAttribute("jsonbColumn")
+                                                .toString()
+                                                .contains("\"version\": 3")));
+            }
+        }
+    }
+
+    @Test
+    public void testJSONArrayContainsWithJsonPathExists() throws Exception {
+        if (pgJsonTestSetup.supportJsonPathExists) {
```
aaime, line 521: This if has no reason to exist, if methods invoked are also marked with @Test, they are going to be invoked directly anyways. If those tests rely on having json path exist support to work, then you should add an Assume at their beginning, so that they can be skipped if the assumption fails.

aaime (changes requested):
Can you clean up the history of your PR?

image

Those first two commits are several months old and are already in the history of the main branch.

modules/plugin/jdbc/jdbc-postgis/src/test/java/org/geotools/data/postgis/PostGISJsonOnlineTest.java
Mike7311: removed

aaime (approved):

PR #4818, opened 2024-06-20
[GEOT-7601] Transform FeatureCollection Query handling of reprojection

jodygarnett:
[GEOT-7601](https://osgeo-org.atlassian.net/browse/GEOT-7601) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

Transform FeatureCollection handling of a query that both limits attributes and changes CRS.

During this PR several other issues came up as I ran into missing or inconsistent functionality. 

* [GEOT-7612](https://osgeo-org.atlassian.net/browse/GEOT-7612) Types.validate(Attribute) does not check ComplexAttribute values individually
* [GEOT-7601](https://osgeo-org.atlassian.net/browse/GEOT-7601) Transform FeatureCollection Query handling of reproduction
* [GEOT-7602](https://osgeo-org.atlassian.net/browse/GEOT-7602) Implement SimpleFeatureTypeBuilder.retype(schema,query) for properyNames and crs together
* [GEOT-7603](https://osgeo-org.atlassian.net/browse/GEOT-7603) Use of JTS setCRS / getCRS methods for use of geometry user data
* [GEOT-7604](https://osgeo-org.atlassian.net/browse/GEOT-7604) GeoJSON test case was timezone senstive

The approach used is to pass the query crs and retrojectCRS onto to the source and trust it to handle reproduction and the schema change.

aside: An improvement would be handle reproduction and crs change as part of the transform, so that any expressions such as `geom=buffer(geom,0.1)` could be respected (as they would always be executed in the source crs).

# Checklist

### DIFF
--- a/modules/extension/transform/src/main/java/org/geotools/data/transform/TransformFeatureCollection.java
+++ b/modules/extension/transform/src/main/java/org/geotools/data/transform/TransformFeatureCollection.java
@@ -44,6 +44,7 @@ import org.geotools.feature.visitor.MaxVisitor;
 import org.geotools.feature.visitor.MinVisitor;
 import org.geotools.feature.visitor.UniqueVisitor;
 import org.geotools.geometry.jts.ReferencedEnvelope;
+import org.geotools.referencing.CRS;
 import org.geotools.util.logging.Logging;
 
 /**
@@ -69,12 +70,19 @@ class TransformFeatureCollection extends AbstractFeatureCollection {
         this.query = query;
     }
 
-    /** Creates a sub-schema with only the selected attributes */
+    /**
+     * Creates a sub-schema with only the selected attributes.
+     *
+     * <p>Transformer is going to trust that the source can handle any reprojection.
+     */
     static SimpleFeatureType retypeSchema(SimpleFeatureType schema, Query query) {
-        if (query.getPropertyNames() == Query.ALL_NAMES) {
+        if (query.getPropertyNames() == Query.ALL_NAMES
+                && CRS.equalsIgnoreMetadata(
+                        schema.getCoordinateReferenceSystem(),
+                        query.getCoordinateSystemReproject())) {
             return schema;
         } else {
-            return SimpleFeatureTypeBuilder.retype(schema, query.getPropertyNames());
+            return SimpleFeatureTypeBuilder.retype(schema, query);
         }
     }
 
@@ -123,7 +131,7 @@ class TransformFeatureCollection extends AbstractFeatureCollection {
 
             result = new SimpleFeatureIteratorIterator(transformed);
         } catch (IOException e) {
-            throw new RuntimeException(e);
+            throw new RuntimeException("Transform failure: " + e.getMessage(), e);
         } finally {
             // if result is null, an exception has occurred, close the wrapped iterator
             if (result == null) {
--- a/modules/library/main/src/main/java/org/geotools/feature/simple/SimpleFeatureBuilder.java
+++ b/modules/library/main/src/main/java/org/geotools/feature/simple/SimpleFeatureBuilder.java
@@ -361,6 +361,12 @@ public class SimpleFeatureBuilder extends FeatureBuilder<FeatureType, Feature> {
      * @param id FeatureID for the generated feature, use null to allow one to be supplied for you
      */
     public static SimpleFeature build(SimpleFeatureType type, List<Object> values, String id) {
+        if (type.getAttributeCount() != values.size()) {
+            throw new IllegalArgumentException(
+                    String.format(
+                            "%s attributes %d count does not match %d values provided",
+                            type.getTypeName(), type.getAttributeCount(), values.size()));
+        }
         return build(type, values.toArray(), id);
     }
 
--- a/modules/library/main/src/main/java/org/geotools/feature/simple/SimpleFeatureBuilder.java
+++ b/modules/library/main/src/main/java/org/geotools/feature/simple/SimpleFeatureBuilder.java
@@ -16,7 +16,8 @@
  */
 package org.geotools.feature.simple;
 
-import java.util.Arrays;
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -123,12 +124,12 @@ public class SimpleFeatureBuilder extends FeatureBuilder<FeatureType, Feature> {
     Map<String, Integer> index;
 
     /** the values */
-    // List<Object> values;
     Object[] values;
 
     /** pointer for next attribute */
     int next;
 
+    /** Attribute userData by index. */
     Map<Object, Object>[] userData;
 
     Map<Object, Object> featureUserData;
@@ -152,6 +153,7 @@ public class SimpleFeatureBuilder extends FeatureBuilder<FeatureType, Feature> {
         reset();
     }
 
+    /** Reset the builder, for generating a new feature. */
     public void reset() {
         values = new Object[featureType.getAttributeCount()];
         next = 0;
@@ -159,7 +161,7 @@ public class SimpleFeatureBuilder extends FeatureBuilder<FeatureType, Feature> {
         featureUserData = null;
     }
 
-    /** Returns the simple feature type used by this builder as a feature template */
+    /** Returns the simple feature type used by this builder as a feature template. */
     @Override
     public SimpleFeatureType getFeatureType() {
         return featureType;
@@ -168,8 +170,10 @@ public class SimpleFeatureBuilder extends FeatureBuilder<FeatureType, Feature> {
     /**
      * Initialize the builder with the provided feature.
      *
-     * <p>This method adds all the attributes from the provided feature. It is useful when copying a
-     * feature.
+     * <p>This method adds all the attributes from the provided feature (along with user data if
+     * provided). It is useful when copying a feature.
+     *
+     * @param feature Feature to copy
      */
     public void init(SimpleFeature feature) {
         reset();
@@ -204,16 +208,18 @@ public class SimpleFeatureBuilder extends FeatureBuilder<FeatureType, Feature> {
         next++;
     }
 
-    /** Adds a list of attributes. */
+    /** Adds a list of attributes, in order provided. */
     public void addAll(List<Object> values) {
         for (Object value : values) {
             add(value);
         }
     }
 
-    /** Adds an array of attributes. */
+    /** Adds an array of attributes, in order provided. */
     public void addAll(Object... values) {
-        addAll(Arrays.asList(values));
+        for (Object value : values) {
+            add(value);
+        }
     }
 
     /**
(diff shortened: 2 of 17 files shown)


jodygarnett:
@NielsCharlier you were the last person to adjust JTS.setCRS / getCRS methods. You may wish to review? I was unaware of the use of a HashMap for user data so it tripped me up in debugging / testing.

modules/unsupported/geojson-core/src/test/java/org/geotools/data/geojson/GeoJSONWriteTest.java
```diff
@@ -297,7 +299,12 @@ public void testDateFormat() throws Exception {
         String json = out.toString(StandardCharsets.UTF_8);
         JsonNode root = new ObjectMapper().readTree(json);
         String date = root.get("properties").get("date").textValue();
-        assertEquals(DateFormatUtils.ISO_8601_EXTENDED_DATE_FORMAT.format(testDate), date);
+
+        FastDateFormat utm =
+                FastDateFormat.getInstance(
+                        DateFormatUtils.ISO_8601_EXTENDED_DATE_FORMAT.getPattern(),
+                        GeoJSONWriter.DEFAULT_TIME_ZONE);
```
jodygarnett, line 306: writer.setDatePattern method above reuses the timezone of the existing writer formatter, which by default is DEFAULT_TIME_ZONE. 

Using GMT is correct as I understand it for JSON, so I am correcting the test case with this in mind.

modules/library/main/src/main/java/org/geotools/geometry/jts/JTS.java
```diff
@@ -1685,23 +1688,61 @@ public static CoordinateReferenceSystem getCRS(Geometry geometry) {
...
-            userData.putAll((Map<?, ?>) obj);
+        if (crs != null) {
+            try {
+                Integer srid = CRS.lookupEpsgCode(crs, false);
+                if (srid != null) {
+                    geometry.setSRID(srid);
+                }
+            } catch (FactoryException noSRID) {
+                geometry.setSRID(0);
+            }
```
aaime, line 1721: Oh gosh no no no... the lookup is not free, doing it for every geometry is criminal performance-wise, but more importantly, usage of SRID should be banned from all the codebase, with the exception of places where it's the only option (spatial databases), because it does not carry an authority (since the IAU work we live in a post-EPSG world).

jodygarnett: I was gathering up this code from  CRSEvaluator that did this, in part to catch assumptions like this :)

modules/library/main/src/main/java/org/geotools/geometry/jts/JTS.java
```diff
@@ -1685,23 +1688,61 @@ public static CoordinateReferenceSystem getCRS(Geometry geometry) {
...
+
+            Object obj = geometry.getUserData();
+            if (obj != null && obj instanceof CoordinateReferenceSystem) {
+                geometry.setUserData(crs);
+            }
+            Map<Object, Object> userData = new HashMap<>();
+            if (obj != null && obj instanceof Map) {
+                // preserve existing user data map contents
+                userData.putAll((Map<?, ?>) obj);
```
jodygarnett, line 1730: @aaime in the interests of doing less working in possibly time sensitive method, is there any reason I should not update the existing user data map in place?

aaime: I was wondering the same, but I cannot offer an answer. I'd suggest checking the history ... @NielsCharlier introduced that method, maybe he can explain the reasoning (maybe it's just basic defensive programming).

jodygarnett: Yeah I already asked @NielsCharlier for input in an earlier message. I am going to try the change locally and let it build over night.

jodygarnett:
Looks like we have one regression in importer rest:
```
Error:  Tests run: 4, Failures: 0, Errors: 2, Skipped: 0, Time elapsed: 2.113 s <<< FAILURE! - in org.geoserver.importer.rest.AttributeComputeTransformTest
Error:  testTransformLiteral(org.geoserver.importer.rest.AttributeComputeTransformTest)  Time elapsed: 0.009 s  <<< ERROR!
java.lang.IllegalArgumentException: river attributes 5 count does not match 4 values provided
	at org.geoserver.importer.rest.AttributeComputeTransformTest.testTransformLiteral(AttributeComputeTransformTest.java:41)

Error:  testTransformExpression(org.geoserver.importer.rest.AttributeComputeTransformTest)  Time elapsed: 0.002 s  <<< ERROR!
java.lang.IllegalArgumentException: river attributes 5 count does not match 4 values provided
	at org.geoserver.importer.rest.AttributeComputeTransformTest.testTransformExpression(AttributeComputeTransformTest.java:66)
```

Chances are the regression is actually making things correct...

modules/library/main/src/main/java/org/geotools/feature/simple/SimpleFeatureBuilder.java
```diff
@@ -361,6 +361,12 @@ public static SimpleFeature build(SimpleFeatureType type, Object[] values, Strin
      * @param id FeatureID for the generated feature, use null to allow one to be supplied for you
      */
     public static SimpleFeature build(SimpleFeatureType type, List<Object> values, String id) {
+        if (type.getAttributeCount() != values.size()) {
```
aaime, line 364: This is a backward incompatible behavioral change, it will cause issues outside of GeoTools, and it's the reason for the importer test breaking.

The computed feature types are the expected ones (original, transformed):

```
SimpleFeatureTypeImpl AttributeComputeTransformTest:river identified extends Feature(id:id,geom:geom,river:river,flow:flow)
SimpleFeatureTypeImpl AttributeComputeTransformTest:river identified extends Feature(id:id,geom:geom,river:river,flow:flow,theDate:theDate)
```

But the code is building a new feature with the new feature type, providing a list of values from the old type (thus, it's missing one value).

The current builder can build a feature with values missing. The Javadoc did not say anything about matching the number of attributes, so there are likely implementations that provide more or fewer attributes than planned.

The change is not wrong per se mind, but if you stick to it, you'll have to change downstream code and add some instructions in the upgrade guide.

jodygarnett:
Oh hey thanks Andrea, I was indeed already aware of the problem. The javadocs indicate that the array has to be the correct number of attributes - and it is supposed to be an optimization (get the parameters right and go fast). Having more (or less) values defined seems better when you use the builder and provide values one at a time. that way the missing values get a default of some sort.

I am going to think a bit more and review the javadocs. I believe I prefer the exception message if it provides an oppertunity for a fast path.

modules/library/main/src/main/java/org/geotools/feature/simple/SimpleFeatureTypeBuilder.java
```diff
@@ -1050,10 +1065,80 @@ public static SimpleFeatureType retype(SimpleFeatureType original, Query query)
...
-            SimpleFeatureType original, String[] propertyNames, CoordinateReferenceSystem crs) {
-        // TODO Auto-generated method stub
-        return null;
+            SimpleFeatureType original, String[] attributes, CoordinateReferenceSystem crs) {
+        SimpleFeatureTypeBuilder b = new SimpleFeatureTypeBuilder();
+
+        // initialize the builder
+        b.init(original);
+
+        List<String> properties;
```
aaime, line 1083: In simple features we call these attributes. Actually, these are attribute names. But they are also provided as arguments, which makes things even more confusing. Maybe the argument should be called attributNames, and you can have the method just use arrays rather than a list? I promise, it's a bit old fashioned, but does not hurt :rofl:, and you can do for loops on them too.

aaime:
@jodygarnett the javadoc says:

```
    /**
     * * Static method to build a new feature.
     *
     * <p>If multiple features need to be created, this method should not be used and instead an
     * instance should be instantiated directly.
     *
     * @param type SimpleFeatureType defining the structure for the created feature
     * @param values Attribute values, must be in the order defined by SimpleFeatureType
     * @param id FeatureID for the generated feature, use null to allow one to be supplied for you
```

I see mention of the order, not of the count.

jodygarnett:
Yes it never occurred to me that people works do more or less 😆

NielsCharlier:
@jodygarnett Sorry for my late response. Looking back at my commit (https://github.com/geotools/geotools/commit/f5ece02748173969cc359727d73de5a6d35e153d) there was already code using UserData inside of a geometry as a map, but sometimes just a CRS. Hence it was always necessary to do a few instanceof to get the CRS. What I did is move the code that does that to one central place (JTS.getCRS) so that it can be called from anywhere and behaviour is consistent.

I'm not sure if there are any other questions for me?

modules/library/main/src/main/java/org/geotools/feature/simple/SimpleFeatureBuilder.java
jodygarnett: Andrea re thought about this and decided that builder needs to be forgiving as much as possible. It already is doing type conversion and anything possible to get a valid feature out - so supplying a few default values is on brand!

Also the way addAll was written it was adding from the present "index" ... so it was just undocumented; it was always intended to allow less attributes. 

I did notice when writing test cases that builder setValidation(true) was not managing to validate individuals values resulting in another new jira ticket.

modules/extension/transform/src/main/java/org/geotools/data/transform/TransformFeatureCollection.java
```diff
@@ -123,7 +133,7 @@ protected Iterator<SimpleFeature> openIterator() {
 
             result = new SimpleFeatureIteratorIterator(transformed);
         } catch (IOException e) {
-            throw new RuntimeException(e);
+            throw new RuntimeException("Transform failure: " + e.getLocalizedMessage(), e);
```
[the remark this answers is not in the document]
jodygarnett: oh, okay!

modules/extension/transform/src/main/java/org/geotools/data/transform/TransformFeatureCollection.java
```diff
@@ -103,6 +111,8 @@ protected Iterator<SimpleFeature> openIterator() {
             SimpleFeatureIterator transformed =
                     new TransformFeatureIteratorWrapper(fi, transformer);
 
+            // See if we need to reproject
+
```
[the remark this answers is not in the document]
jodygarnett: Removing, I was unable to obtain the returned FeatureType in order to determine if reproject needed to be applied. I was hoping to make TransformerFeatureCollection very tolerant of incomplete implementations...

modules/library/main/src/main/java/org/geotools/feature/simple/SimpleFeatureTypeBuilder.java
jodygarnett: yeah you are right, fixed

jodygarnett:
Thanks Niels, it helps to have the history. If you can double check that part of the PR it would be good (or you know trust the test cases)

jodygarnett:
Okay if the downstream integration build works I will be darn impressed.

NielsCharlier:
@jodygarnett yes it looks good, you might want to run the app-schema online tests as well

jodygarnett:
> @jodygarnett yes it looks good, you might want to run the app-schema online tests as well

Ok how do I do that?

PR #4819, opened 2024-06-21
[GEOT-7607] Error using function Interpolate in COLOR mode from SLD

groldan:
[GEOT-7607](https://osgeo-org.atlassian.net/browse/GEOT-7607) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

Fix `org.geotools.filter.function.InterpolateFunction` argument validateion and context fallback.

The defect was exposed by [GEOT-7414] at commit d7991fdc, as `InterpolateFunctionImpl` wouldn't adhere to the interface contract when receiving `null` or `Object.class` as `context` parameter.

  

# Checklist

### DIFF
--- a/modules/library/main/src/main/java/org/geotools/filter/function/InterpolateFunction.java
+++ b/modules/library/main/src/main/java/org/geotools/filter/function/InterpolateFunction.java
@@ -301,28 +301,22 @@ public class InterpolateFunction implements Function {
         return evaluate(object, Object.class);
     }
 
+    /**
+     * {@inheritDoc}
+     *
+     * <p>When {@code context} is unspecified (i.e. {@code null} or {@code Object.class}), it'll be
+     * derived from the {@code method} parameter, as {@code java.awt.Color.class} when {@code method
+     * == COLOR}, and as {@code java.lang.Double} when {@code method == NUMERIC}.
+     *
+     * @throws IllegalArgumentException if the {@code method} parameter is {@code NUMERIC} and
+     *     {@code context == java.awt.Color.class}, or {@code method} is {@code COLOR} and {@code
+     *     context} is specified and is not {@code context == java.awt.Color.class}
+     */
     @Override
     public <T> T evaluate(Object object, Class<T> context) {
-        // initialize the lookup data structures only once and in a thread safe way please
-        if (interpPoints == null) {
-            synchronized (this) {
-                if (interpPoints == null) {
-                    initialize();
-                }
-            }
-        }
-
-        if (method == Method.NUMERIC && Color.class.isAssignableFrom(context)) {
-            throw new IllegalArgumentException(
-                    "Trying to evaluate the function as Color but the method parameter is set as NUMERIC");
-        }
-
-        if (method == Method.COLOR && !Color.class.isAssignableFrom(context)) {
-            throw new IllegalArgumentException(
-                    "Trying to evaluate the function as "
-                            + context.getSimpleName()
-                            + " but the method parameter is set as COLOR");
-        }
+        threadSafeInitialize();
+        context = sanitizeContext(context);
+        validateArguments(context);
 
         /**
          * Lookup value should be either the name of a feature property which can be evaluated to a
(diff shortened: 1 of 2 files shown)


modules/library/main/src/main/java/org/geotools/filter/function/InterpolateFunction.java
```diff
@@ -378,6 +372,32 @@ public <T> T evaluate(Object object, Class<T> context) {
...
 
+    @SuppressWarnings("unchecked")
+    private <T> Class<T> sanitizeContext(Class<T> context) {
+        if (null == context || Object.class.equals(context)) {
+            if (method == Method.NUMERIC) {
+                context = (Class<T>) Double.class;
+            } else if (method == Method.COLOR) {
+                context = (Class<T>) Color.class;
+            } else {
+                throw new IllegalStateException("Unknown method, expected NUMERIC or COLOR");
```
aaime, line 383: Can you double-check the Symbology Encoding 1.1 specification and see if there is a default? 
I have vague memories it should be numeric, and the "sample" value for the parameter is indeed that, but I'm not sure.

aaime (commented):
Overall looks good, see note about the possibility of a default method below.

modules/library/main/src/main/java/org/geotools/filter/function/InterpolateFunction.java
groldan: The default of mode LINEAR and method NUMERIC are set in `initialize()`, hence the IAE here, it should never be reached for as long as `initialize()` is called before

groldan: note both cases are fully covered by tests:
image

jodygarnett:
Do we need to backport further?

modules/library/main/src/main/java/org/geotools/filter/function/InterpolateFunction.java
```diff
@@ -301,28 +301,22 @@ public Object evaluate(Object object) {
...
+    /**
+     * {@inheritDoc}
+     *
+     * <p>When {@code context} is unspecified (i.e. {@code null} or {@code Object.class}), it'll be
+     * derived from the {@code method} parameter, as {@code java.awt.Color.class} when {@code method
+     * == COLOR}, and as {@code java.lang.Double} when {@code method == NUMERIC}.
+     *
+     * @throws IllegalArgumentException if the {@code method} parameter is {@code NUMERIC} and
+     *     {@code context == java.awt.Color.class}, or {@code method} is {@code COLOR} and {@code
+     *     context} is specified and is not {@code context == java.awt.Color.class}
```
jodygarnett, line 313: Hold up, we should be able to get a String out of a colour. We went to a lot of trouble with converters to do so ...
We went to a lot of trouble to with converters to avoid being strongly typed ...

aaime: Yeah you're right... the original implementation was rigid, but functions should try to use converters to go from the internal representation they come up with, to the desired output, and return null if they fail (see Expression javadoc). 
Converters are also pluggable, so it's not possible to know in advance what would work, and what wouldn't, one just has to try.

groldan: Not sure I understand. What changed is it'll output Color or Double if no context was provided. It changes nothing otherwise and the interpolate functions still return `Converters.convert(interpolated, context)`, where `context` hasn't been changed if it was `String.class` or anything else but `null` or `Object.java`. both cases were failing.

aaime: What Jody is saying is that the original implementation is broken in more ways than you've fixed.

groldan:
> Do we need to backport further?

[GEOT-7414] is present in 30.x too so I guess it wouldn't hurt?

PR #4833, opened 2024-07-10
[WiP] Update jaxb to jakarta

Charmatzis:
Update jaxb to jakarta
Following this
https://docs.openrewrite.org/recipes/java/migrate/jakarta/javaxxmlbindmigrationtojakartaxmlbind


mprins (commented):
please respect project guidelines, such as using the mandatory pull request template, as described in https://docs.geotools.org/latest/developer/procedures/pull_requests.html

aaime:
If you want to get rid of jaxb fully, I believe you'll also have to upgrade upstream projects, for example, 
imageio-ext depends on it here:
https://github.com/geosolutions-it/imageio-ext/blob/master/library/geocore/pom.xml

(don't remember if it's used anywhere else in imageio-ext, it might).

Charmatzis:
@aaime true, thx. I will and come back

Charmatzis:
Raised a PR for `imageio-ext`
https://github.com/geosolutions-it/imageio-ext/pull/307

aaime:
Another bit that is troublesome, is that GeoServer is also using jaxb in a number of modules, both in supported land, and amongst the community modules. All these projects move as one (as you can see from the "downstream integration build") so the upgrade needs to be performed end-to-end.

Also a note, I've found down in the imageio-ext build that the jars require at least java 11. Can you confirm, and in particular, make sure that Java 17 is not required? The "geo-trio" (GeoTools, GeoWebCache and GeoServer) will eventually start requiring Java 17 to build, but for now, the build must run with Java 11.

PR #4835, opened 2024-07-11
[GEOT-7611] Fix logoUrl parsing when missing format element

airnez:
[GEOT-7611](https://osgeo-org.atlassian.net/browse/GEOT-7611) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

Prevents ArrayIndexOutOfBoundsException while parsing attribution logoUrl element in capabilities XML.
Happens when the format element is missing. Prevents from parsing the whole capabilities response.
See issue https://osgeo-org.atlassian.net/browse/GEOT-7611
  

# Checklist

-  There is no change visible to the end user, i don't think  that any documentation update is needed.


airnez:
Hi,
I don't think that the Mac OS CI / build failure is related to my commits. What do you think ?

aaime:
Unlikely. Restarted.

aaime:
Looked at the changes, they seem fine. I've restarted the OSX build once more.

airnez:
Hi. Is there anything I missed blocking the PR ? I just sent the contributor license agreement.

jodygarnett:
Thanks for sending the contributor thing. I am just kicking off the build and this should be fine... let me review.

modules/extension/wms/src/test/resources/org/geotools/ows/wms/xml/test/test-data/1.1.0CapabilitiesMissingLogoUrlFormat.xml
```diff
@@ -0,0 +1,110 @@
...
+    <SRS>EPSG:42304 EPSG:42101 EPSG:4269 EPSG:4326</SRS>
+    <LatLonBoundingBox minx="-172.367" miny="35.6673" maxx="-11.5624" maxy="83.8293" />
+    <BoundingBox SRS="EPSG:42304"
+                minx="-2.2e+06" miny="-712631" maxx="3.0728e+06" maxy="3.84e+06" />
+    <Attribution>
+      <Title>test</Title>
+      <OnlineResource xmlns:xlink="http://www.w3.org/1999/xlink" xlink:type="simple" xlink:href="http://www.example.com"/>
+      <LogoURL height="100" width="100">
+            <!-- <Format>image/png</Format> --> <!-- Removed for GEOT-7611 test (ArrayIndexOutOfBoundsException when missing format) -->
+          <OnlineResource xmlns:xlink="http://www.w3.org/1999/xlink" xlink:type="simple" xlink:href="http://www.osgeo.org/sites/all/themes/osgeo/logo.png"/>
```
jodygarnett, line 105: So it is actually invalid, you are just trying to be forgiving?

jodygarnett (approved):

jodygarnett (commented):
Going to add a comment here. I had a look at the bug report which was based on:

```
<LogoURL width="274" height="100">
<OnlineResource xlink:href="https://www.craig.fr/sites/www.craig.fr/files/images/logo_craig_2018.svg"/>
</LogoURL>
```

I had a look and the WMS Client code does not appear to use this information (if it did we could try guessing based on extension).

Q: Is a similar change needed for LegendURL?

aaime:
@airnez could you answer @jodygarnett feedback?

modules/extension/wms/src/test/resources/org/geotools/ows/wms/xml/test/test-data/1.1.0CapabilitiesMissingLogoUrlFormat.xml
airnez: You are making a good point. I was concerned about the whole getcapabilities response parsing failing because of one malformed element for one layer. Even if we don't consider the fact that we don't use the format information later on, we could still consume the reste of the layers while having just that one failing.

The previous test written are using validation hints while it is disabled elsewhere : 
https://github.com/geotools/geotools/blob/7033c917fe2c3aedb02eb19a19f4034d4da868ff/modules/extension/wms/src/main/java/org/geotools/ows/wms/response/WMSGetCapabilitiesResponse.java#L51C1-L53C14

I thought that the fact that document validation is disabled meant that we are indeed trying to be forgiving.
The previous code works properly with validation enabled, because this ensures Format and OnlineResource elements presence and order. See prevous implementation : 
```
if (sameName(elems[1], value[i])) {
        url = (URL) value[1].getValue();
}
```
Because this validation is disabled outside tests, the code seems misleading to me. It is testing value[i] and then accessing value[1].

airnez: I agree concerning LegendUrl. I just pushed the corresponding fix.

airnez:
I fixed the formatting issue blocking test worklflows, could you restart them ?

jodygarnett:
thanks, going to run the workflows and see if we can get this in.

airnez:
Hi :) Would you mind restarting workflows ? I expect everything to pass now

jodygarnett:
Thanks for you contribution :) restarting! We may or may not make the release cycle...

aaime:
Merged after 32.x was created, added a new backport label for 32.x, testing it.

aaime:
Oh, looks like there was a merge commit in the history of this one... can only be backported manually then.

PR #4846, opened 2024-07-19
[GEOT-7614] support for postgres reWriteBatchedInserts

j-heaton:
[GEOT-7614](https://osgeo-org.atlassian.net/browse/GEOT-7614) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

adds support for setting `reWriteBatchedInserts` as a jdbc property for postgis

# Checklist

### DIFF
--- a/modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/PostgisNGDataStoreFactory.java
+++ b/modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/PostgisNGDataStoreFactory.java
@@ -91,13 +91,13 @@ public class PostgisNGDataStoreFactory extends JDBCDataStoreFactory {
             new Param(
                     "create database params",
                     String.class,
-                    "Extra specifications appeneded to the CREATE DATABASE command",
+                    "Extra specifications appended to the CREATE DATABASE command",
                     false,
                     "",
                     Param.LEVEL,
                     "advanced");
 
-    /** Wheter a prepared statements based dialect should be used, or not */
+    /** Whether a prepared statements based dialect should be used, or not */
     public static final Param PREPARED_STATEMENTS =
             new Param(
                     "preparedStatements",
@@ -146,10 +146,21 @@ public class PostgisNGDataStoreFactory extends JDBCDataStoreFactory {
             new Param(
                     "SSL mode",
                     SslMode.class,
-                    "The connectin SSL mode",
+                    "The connection SSL mode",
                     false,
                     SslMode.DISABLE,
                     new KVP(Param.OPTIONS, Arrays.asList(SslMode.values())));
+
+    public static final Param REWRITE_BATCHED_INSERTS =
+            new Param(
+                    "reWriteBatchedInserts",
+                    Boolean.class,
+                    "This will change batch inserts from INSERT INTO foo (col1, col2, col3) VALUES (1, 2, 3) "
+                            + "to INSERT INTO foo (col1, col2, col3) VALUES (1, 2, 3), (4, 5, 6) . "
+                            + "When setting this parameter to true, you should also set the parameter BATCH_INSERT_SIZE = 1000. "
+                            + "Enabling this parameter and setting BATCH_INSERT_SIZE too low may result in decreased insert performance.",
+                    false,
+                    Boolean.FALSE);
 
     @Override
     protected SQLDialect createSQLDialect(JDBCDataStore dataStore, Map<String, ?> params) {
@@ -162,7 +173,7 @@ public class PostgisNGDataStoreFactory extends JDBCDataStoreFactory {
             if (LOGGER.isLoggable(Level.FINE))
                 LOGGER.log(
                         Level.FINE,
-                        "Failed to lookup prepared statement parameter, continuining with non prepared dialect",
+                        "Failed to lookup prepared statement parameter, continuing with non prepared dialect",
                         e);
         }
 
@@ -290,6 +301,7 @@ public class PostgisNGDataStoreFactory extends JDBCDataStoreFactory {
         parameters.put(SIMPLIFICATION_METHOD.key, SIMPLIFICATION_METHOD);
         parameters.put(CREATE_DB_IF_MISSING.key, CREATE_DB_IF_MISSING);
         parameters.put(CREATE_PARAMS.key, CREATE_PARAMS);
+        parameters.put(REWRITE_BATCHED_INSERTS.key, REWRITE_BATCHED_INSERTS);
     }
 
     @Override
@@ -302,10 +314,20 @@ public class PostgisNGDataStoreFactory extends JDBCDataStoreFactory {
         String host = (String) HOST.lookUp(params);
         String db = (String) DATABASE.lookUp(params);
         int port = (Integer) PORT.lookUp(params);
-        String url = "jdbc:postgresql" + "://" + host + ":" + port + "/" + db;
+        boolean reWriteBatchedInserts = Boolean.TRUE.equals(REWRITE_BATCHED_INSERTS.lookUp(params));
+        String url =
+                "jdbc:postgresql"
+                        + "://"
+                        + host
+                        + ":"
+                        + port
+                        + "/"
+                        + db
+                        + "?reWriteBatchedInserts="
+                        + reWriteBatchedInserts;
         SslMode mode = (SslMode) SSL_MODE.lookUp(params);
         if (mode != null) {
-            url = url + "?sslmode=" + mode + "&binaryTransferEnable=bytea";
+            url = url + "&sslmode=" + mode + "&binaryTransferEnable=bytea";
         }
 
         return url;
--- a/modules/plugin/jdbc/jdbc-postgis/src/test/java/org/geotools/data/postgis/PostgisJNDIDataSourceOnlineTest.java
+++ b/modules/plugin/jdbc/jdbc-postgis/src/test/java/org/geotools/data/postgis/PostgisJNDIDataSourceOnlineTest.java
@@ -52,6 +52,7 @@ public class PostgisJNDIDataSourceOnlineTest extends JDBCJNDIDataSourceOnlineTes
         standardParams.remove(PostgisNGDataStoreFactory.CREATE_DB_IF_MISSING.key);
         standardParams.remove(PostgisNGDataStoreFactory.CREATE_PARAMS.key);
         standardParams.remove(PostgisNGDataStoreFactory.SSL_MODE.key);
+        standardParams.remove(PostgisNGDataStoreFactory.REWRITE_BATCHED_INSERTS.key);
         standardParams.removeAll(baseParams);
         List<String> baseJndiParams = getBaseJNDIParams();
         List<String> jndiParams = getParamKeys(getJNDIStoreFactory());
--- a/modules/plugin/jdbc/jdbc-postgis/src/test/java/org/geotools/data/postgis/PostgisNGDataStoreFactoryTest.java
+++ b/modules/plugin/jdbc/jdbc-postgis/src/test/java/org/geotools/data/postgis/PostgisNGDataStoreFactoryTest.java
@@ -160,4 +160,16 @@ public class PostgisNGDataStoreFactoryTest {
         verify(this.st2, withQuery ? times(1) : never()).close();
         verify(this.rs2, withQuery ? times(1) : never()).close();
     }
+
+    @Test
+    public void testGetJDBCUrl() throws Exception {
+        Map<String, Object> params = new HashMap<>();
+        params.put(PostgisNGDataStoreFactory.HOST.key, "localhost");
+        params.put(PostgisNGDataStoreFactory.PORT.key, 5432);
+        params.put(PostgisNGDataStoreFactory.DATABASE.key, "template1");
+        params.put(PostgisNGDataStoreFactory.REWRITE_BATCHED_INSERTS.key, Boolean.TRUE);
+        assertEquals(
+                "jdbc:postgresql://localhost:5432/template1?reWriteBatchedInserts=true",
+                new PostgisNGDataStoreFactory().getJDBCUrl(params));
+    }
 }
--- a/modules/plugin/jdbc/jdbc-postgis/src/test/java/org/geotools/data/postgis/ps/PostgisReWriteBatchedInsertsFeatureStoreOnlineTest.java
+++ b/modules/plugin/jdbc/jdbc-postgis/src/test/java/org/geotools/data/postgis/ps/PostgisReWriteBatchedInsertsFeatureStoreOnlineTest.java
@@ -1,0 +1,31 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ *
+ *    (C) 2002-2009, Open Source Geospatial Foundation (OSGeo)
+ *
+ *    This library is free software; you can redistribute it and/or
+ *    modify it under the terms of the GNU Lesser General Public
+ *    License as published by the Free Software Foundation;
+ *    version 2.1 of the License.
+ *
+ *    This library is distributed in the hope that it will be useful,
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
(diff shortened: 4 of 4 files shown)


modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/PostgisNGDataStoreFactory.java
```diff
@@ -302,10 +312,11 @@ protected String getJDBCUrl(Map<String, ?> params) throws IOException {
         String host = (String) HOST.lookUp(params);
         String db = (String) DATABASE.lookUp(params);
         int port = (Integer) PORT.lookUp(params);
-        String url = "jdbc:postgresql" + "://" + host + ":" + port + "/" + db;
+        Boolean reWriteBatchedInserts = (Boolean) REWRITE_BATCHED_INSERTS.lookUp(params);
+        String url = "jdbc:postgresql" + "://" + host + ":" + port + "/" + db + "?reWriteBatchedInserts=" + reWriteBatchedInserts;
         SslMode mode = (SslMode) SSL_MODE.lookUp(params);
         if (mode != null) {
-            url = url + "?sslmode=" + mode + "&binaryTransferEnable=bytea";
+            url = url + "&sslmode=" + mode.value + "&binaryTransferEnable=bytea";
```
sikeoka, line 319: This should be a separate JIRA ticket and PR with its own unit test.

modules/plugin/jdbc/jdbc-postgis/src/test/java/org/geotools/data/postgis/PostgisNGDataStoreFactoryTest.java
```diff
@@ -38,11 +38,13 @@
 import org.geotools.jdbc.SQLDialect;
 import org.geotools.util.Version;
 import org.junit.After;
+import org.junit.Assert;
```
sikeoka, line 41: This file already has a static import for Assert.assertEquals. Why not just use that?

j-heaton: whoops. removed this import.

modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/PostgisNGDataStoreFactory.java
j-heaton: updated the testGetJDBCUrl to exclude the sslmode param for now

modules/plugin/jdbc/jdbc-postgis/src/main/java/org/geotools/data/postgis/PostgisNGDataStoreFactory.java
```diff
@@ -302,10 +312,11 @@ protected String getJDBCUrl(Map<String, ?> params) throws IOException {
         String host = (String) HOST.lookUp(params);
         String db = (String) DATABASE.lookUp(params);
         int port = (Integer) PORT.lookUp(params);
-        String url = "jdbc:postgresql" + "://" + host + ":" + port + "/" + db;
+        Boolean reWriteBatchedInserts = (Boolean) REWRITE_BATCHED_INSERTS.lookUp(params);
+        String url = "jdbc:postgresql" + "://" + host + ":" + port + "/" + db + "?reWriteBatchedInserts=" + reWriteBatchedInserts;
```
sikeoka, line 316: Should there be a null check here like with the sslmode below? It would probably be better to make sure that the URL does not have `reWriteBatchedInserts=null` even if that doesn't cause an exception.

j-heaton: i have updated the code to `boolean reWriteBatchedInserts = Boolean.TRUE.equals(REWRITE_BATCHED_INSERTS.lookUp(params));`

`Boolean.TRUE.equals(null)` will evaluate to `false`

modules/plugin/jdbc/jdbc-postgis/src/test/java/org/geotools/data/postgis/PostgisJNDIDataSourceOnlineTest.java
```diff
@@ -52,6 +52,7 @@ public void testExtraParams() {
         standardParams.remove(PostgisNGDataStoreFactory.CREATE_DB_IF_MISSING.key);
         standardParams.remove(PostgisNGDataStoreFactory.CREATE_PARAMS.key);
         standardParams.remove(PostgisNGDataStoreFactory.SSL_MODE.key);
+        standardParams.remove(PostgisNGDataStoreFactory.REWRITE_BATCHED_INSERTS.key);
```
jodygarnett, line 55: So I am not seeing any test coverage of this new feature? Can you add a method that uses your code (we want to be sure that something is actually happening and that it produces a SQL Insert statement that works).

jodygarnett (changes requested):
Looks like a nice clean new feature; can you add something to the OnlineTest so we know it works...

modules/plugin/jdbc/jdbc-postgis/src/test/java/org/geotools/data/postgis/PostgisJNDIDataSourceOnlineTest.java
aaime: I think this is all happening inside the driver, so you won't see any difference in the SQL generated by JDBCDataStore. Nevertheless, it would be useful to have one write test run with these settings, to prove they work. Given that prepared statements are already batching write requests, I'm guessing this may be  useful for the case when prepared statements are not enabled. 

@j-heaton can you make sure there is a subclass of PostgisFeatureStoreOnlineTest that forces the new flag on? Also, did you measure performance improvements with the new settings?
How does it compare with the case where prepared statements are enabled?

j-heaton: you are correct that it happens inside the driver. i was able to verify the behavior by setting `log_statements = 'all'` on the postgres DB and checking logs.

i have added a subclass of PostgisFeatureStoreOnlineTest that sets the new flag.

i was able to measure the performance and saw significant gains in ingest speed. i wrote a simple test to prove this although i'm not sure if it belongs in a commit.
couple caveats for this:
- statements will NOT be rewritten if geotools is generating fids for features. therefore features need to already have a fid and the hints `PROVIDED_FID` and `USE_PROVIDED_FID` must be used.
- `BATCH_INSERT_SIZE` needs to be increased or else performance will be worse than the case where `REWRITE_BATCHED_INSERTS` is disabled

here is the test i used along with the results:
```
REWRITE_BATCHED_INSERTS=false BATCH_INSERT_SIZE=100
  avg: 1989.375 min: 1875 max: 2125
REWRITE_BATCHED_INSERTS=false BATCH_INSERT_SIZE=1000
  avg: 1367.375 min: 1343 max: 1398
REWRITE_BATCHED_INSERTS=true BATCH_INSERT_SIZE=100
  avg: 2449.0 min: 2375 max: 2620
REWRITE_BATCHED_INSERTS=true BATCH_INSERT_SIZE=1000
  avg: 895.25 min: 872 max: 916
```
  
```
package org.geotools.data.postgis.ps;

import java.io.IOException;
import java.sql.SQLException;
import java.util.*;

import org.geotools.api.data.DataStore;
import org.geotools.api.data.Transaction;
import org.geotools.api.feature.simple.SimpleFeature;
import org.geotools.api.feature.simple.SimpleFeatureType;
import org.geotools.data.DefaultTransaction;
import org.geotools.data.postgis.PostgisNGDataStoreFactory;
import org.geotools.data.simple.SimpleFeatureCollection;
import org.geotools.feature.DefaultFeatureCollection;
import org.geotools.feature.SchemaException;
import org.geotools.feature.simple.SimpleFeatureBuilder;
import org.geotools.feature.simple.SimpleFeatureTypeBuilder;
import org.geotools.jdbc.JDBCFeatureStore;
import org.geotools.referencing.crs.DefaultGeographicCRS;
import org.geotools.test.OnlineTestSupport;
import org.geotools.util.factory.Hints;
import org.junit.AfterClass;
import org.junit.Test;
import org.junit.runner.RunWith;
import org.junit.runners.Parameterized;
import org.locationtech.jts.geom.Coordinate;
import org.locationtech.jts.geom.GeometryFactory;
import org.locationtech.jts.geom.Point;

@RunWith(Parameterized.class)
public class PostGISReWriteBatchedInsertsPerfOnlineTest extends OnlineTestSupport {
    private static final String TYPE_NAME = "PostgisBatchInsertTest";
    private static final int TOTAL_FEATURES = 100000;
    private static final int NUM_FEATURE_ATTRIBUTES = 0;
    private static final int ITERATIONS = 8;

    private static final List<SimpleFeature> preGeneratedFeatures = generateFeatures();
    private static final List<Stats> results = new ArrayList<>();

    private final boolean reWriteBatchedInserts;
    private final int batchSize;
    @Parameterized.Parameters(name = "rewrite batches={0}, batch size={1}")
    public static Collection<Object[]> data() {
        return Arrays.asList(new Object[][] {
                { false, 100 },
                { false, 1000 },
                { true, 100 },
                { true, 1000 }
        });
    }

    public PostGISReWriteBatchedInsertsPerfOnlineTest(boolean reWriteBatchedInserts, int batchSize) {
        this.reWriteBatchedInserts = reWriteBatchedInserts;
        this.batchSize = batchSize;
    }

    @Override
    protected String getFixtureId() {
        return "postgis";
    }

    @Test
    public void testReWriteBatchedInserts() throws IOException, SQLException, SchemaException {
        System.out.printf("running test with REWRITE_BATCHED_INSERTS=%s, BATCH_INSERT_SIZE=%s\n", reWriteBatchedInserts, batchSize);
        System.out.println("warm up");
        Stats stats = new Stats(reWriteBatchedInserts, batchSize);
        runBatchInsertTest(reWriteBatchedInserts, batchSize, stats);
        stats.reset();
        System.out.println("begin recording iterations");
        for (int i=0; i<ITERATIONS; i++) {
            runBatchInsertTest(reWriteBatchedInserts, batchSize, stats);
        }
        System.out.printf("iteration stats:\n  avg: %sms\n  min: %sms\n  max: %sms\n\n", stats.getAverage(), stats.getMin(), stats.getMax());
        results.add(stats);
    }

    @AfterClass
    public static void summarizeResults() {
        System.out.println("all tests complete. results:");
        for (Stats result : results) {
            System.out.printf("REWRITE_BATCHED_INSERTS=%s BATCH_INSERT_SIZE=%s\n  avg: %s min: %s max: %s\n", result.isReWriteBatchedInserts(), result.getBatchSize(), result.getAverage(), result.getMin(), result.getMax());
        }
    }

    private static List<SimpleFeature> generateFeatures() {
        List<SimpleFeature> features = new ArrayList<>();
        SimpleFeatureType featureType = buildFeatureType();
        SimpleFeatureBuilder featureBuilder = new SimpleFeatureBuilder(featureType);
        GeometryFactory geometryFactory = new GeometryFactory();
        for (int i = 0; i < PostGISReWriteBatchedInsertsPerfOnlineTest.TOTAL_FEATURES; i++) {
            featureBuilder.set("geom", geometryFactory.createPoint(new Coordinate(i, i)));
            for (int j = 0; j < NUM_FEATURE_ATTRIBUTES; j++) {
                featureBuilder.set("field_" + j, String.valueOf(j));
            }
            featureBuilder.featureUserData(Hints.PROVIDED_FID, "fid");
            featureBuilder.featureUserData(Hints.USE_PROVIDED_FID, Boolean.TRUE);
            features.add(featureBuilder.buildFeature(String.valueOf(i)));
        }

        return features;
    }

    private void runBatchInsertTest(boolean reWriteBatchedInserts, int batchSize, Stats stats) throws IOException {
        DataStore store = createDataStore(reWriteBatchedInserts, batchSize);
        SimpleFeatureType featureType = buildFeatureType();
        try {
            store.removeSchema(TYPE_NAME);
        } catch (Exception ignored) { }
        store.createSchema(featureType);
        JDBCFeatureStore featureStore = (JDBCFeatureStore) store.getFeatureSource(TYPE_NAME);
        Transaction tx = new DefaultTransaction();
        DefaultFeatureCollection collection = new DefaultFeatureCollection();
        collection.addAll(preGeneratedFeatures);

        long startTime = System.currentTimeMillis();
        try {
            featureStore.setTransaction(tx);
            featureStore.addFeatures((SimpleFeatureCollection) collection);
            tx.commit();
        } catch (Exception e) {
            e.printStackTrace();
            tx.rollback();
            throw e;
        } finally {
            tx.close();
        }
        long endTime = System.currentTimeMillis();
        long totalTime = endTime - startTime;

        System.out.printf("iteration completed in %sms\n", totalTime);
        stats.update(totalTime);
        store.removeSchema(TYPE_NAME);
        store.dispose();
    }

    private DataStore createDataStore(boolean reWriteBatchedInserts, int batchSize) throws IOException {
        PostgisNGDataStoreFactory factory = new PostgisNGDataStoreFactory();
        Properties db = getFixture();
        Map<String, Object> params = new HashMap<>();

        params.put(PostgisNGDataStoreFactory.HOST.key, db.getProperty(PostgisNGDataStoreFactory.HOST.key));
        params.put(PostgisNGDataStoreFactory.DATABASE.key, db.getProperty(PostgisNGDataStoreFactory.DATABASE.key));
        params.put(PostgisNGDataStoreFactory.PORT.key, db.getProperty(PostgisNGDataStoreFactory.PORT.key));
        params.put(PostgisNGDataStoreFactory.USER.key, db.getProperty(PostgisNGDataStoreFactory.USER.key));
        params.put(PostgisNGDataStoreFactory.PASSWD.key, db.getProperty(PostgisNGDataStoreFactory.PASSWD.key));
        params.put(PostgisNGDataStoreFactory.DBTYPE.key, "postgis");

        params.put(PostgisNGDataStoreFactory.PREPARED_STATEMENTS.key, Boolean.TRUE);

        params.put(PostgisNGDataStoreFactory.REWRITE_BATCHED_INSERTS.key, reWriteBatchedInserts);
        params.put(PostgisNGDataStoreFactory.BATCH_INSERT_SIZE.key, batchSize);

        return factory.createDataStore(params);
    }

    private static SimpleFeatureType buildFeatureType() {
        SimpleFeatureTypeBuilder typeBuilder = new SimpleFeatureTypeBuilder();
        typeBuilder.setName(TYPE_NAME);
        DefaultGeographicCRS crs = DefaultGeographicCRS.WGS84;
        typeBuilder.setCRS(crs);
        typeBuilder.add("geom", Point.class);
        for (int j = 0; j < NUM_FEATURE_ATTRIBUTES; j++) {
            typeBuilder.add("field_" + j, String.class);
        }
        return typeBuilder.buildFeatureType();
    }

    private static class Stats {
        private final boolean reWriteBatchedInserts;
        private final int batchSize;

        private float totalTime = 0;
        private long minTime = Long.MAX_VALUE;
        private long maxTime = Long.MIN_VALUE;
        private int count = 0;

        public Stats(boolean reWriteBatchedInserts, int batchSize) {
            this.reWriteBatchedInserts = reWriteBatchedInserts;
            this.batchSize = batchSize;
        }

        public void update(long time) {
            totalTime += time;
            minTime = Math.min(minTime, time);
            maxTime = Math.max(maxTime, time);
            count++;
        }

        public boolean isReWriteBatchedInserts() {
            return reWriteBatchedInserts;
        }

        public int getBatchSize() {
            return batchSize;
        }

        public float getAverage() {
            return count > 0 ? totalTime / count : 0;
        }

        public long getMin() {
            return minTime;
        }

        public long getMax() {
            return maxTime;
        }

        public void reset() {
            totalTime = 0;
            minTime = Long.MAX_VALUE;
            maxTime = Long.MIN_VALUE;
            count = 0;
        }
    }

}
```

aaime:
The QA Build has a legitimate failure, code is not formatted properly. If you just run "mvn clean install" from the command line, the build will reformat the code to follow guidelines.

Also, the information about having to set up a larger batch size is important, it should be documented somewhere.
The easiest place I can think of is the REWRITE_BATCHED_INSERTS description itself.

j-heaton:
i've added documentation to the parameter description. all build checks look green.

aaime:
Yep, merged, thanks.

PR #4848, opened 2024-07-30
Pinning MSSQL 2022 tag to avoid changes in tools directories. 

aaime:
See changes in directory names, making the latest 2022 docker image incompatible with the 2019 one: https://hub.docker.com/r/microsoft/mssql-server

This PR tries to pin the previous tag to avoid the issue.

# Checklist

### DIFF
--- a/.github/workflows/mssql_online.yml
+++ b/.github/workflows/mssql_online.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: [ubuntu-latest]
     strategy:
       matrix:
-        mssql: [2019-latest, 2022-latest]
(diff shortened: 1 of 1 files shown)


aaime:
Looks like it's working. An alternative would be to use latest, but then recognize the 2022 version and change the directories being used.

mprins (commented):
it's a pain [this happened for both 2022 and 2019](https://github.com/microsoft/mssql-docker/issues/892), the other workaround option is to add `-No` and update the path to `/opt/mssql-tools18/bin` as described in https://github.com/microsoft/mssql-docker/issues/892#issuecomment-2249029917

PR #4865, opened 2024-08-06
Make this job remind devs to update the GWC/GS downstream branches when not running on the main branch

aaime:
The idea here is to check somehow that the current GT branch is no longer main, and tell the devs to update the corresponding branches of GWC/GS for the downstream builds. Not sure this is the right way, maybe not, but we should work out something, to avoid surprises in backports.

# Checklist

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -1,10 +1,6 @@
 name: Downstream integration build (GeoWebCache and GeoServer)
 
-on:
-  # trigger on PR, but only on main branch, the checkouts of the downstream projects are also targeting main (default branch)
-  pull_request:
-    branches:
-      - main
+on: [pull_request]
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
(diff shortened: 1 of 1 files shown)


.github/workflows/integration.yml
```diff
@@ -16,6 +12,11 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+      - name: Verify branch
+        if: github.ref != 'refs/heads/main'
```
mprins, line 16: Not sure but I think using `github.base_ref` might work better

aaime: Yep, I was thinking the same, will try it, thanks for the suggestion.

aaime:
Hum... the current structure works, test wise: it's skipped here on main, but produces the error message [on a PR targeting 31.x](https://github.com/geotools/geotools/actions/runs/10286677571/job/28467929519?pr=4866).
However, "exit 1" does not seem to fail the workflow, which is a problem... Haven't figured out a good replacement so far.

PR #4866, opened 2024-08-07
Make this job remind devs to update the GWC/GS downstream branches when not running on the main branch  (31.x test, should fail)

aaime:
# Checklist

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -15,8 +15,11 @@ jobs:
       - name: Verify branch
         if: ${{ github.base_ref != 'main' }}
         uses: actions/github-script@v7
-        script: |
-          core.setFailed("This workflow can only run as-is on the main branch. Please remove this conditional check and make sure GWC and GS are checked out on the right branches for this GeoTools series (see the --branch hints below). Current branch is ${GITHUB_BASE_REF}")
+        with:
+          script: |
+            core.setFailed(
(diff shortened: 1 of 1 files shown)


.github/workflows/integration.yml
```diff
@@ -12,6 +12,10 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+      - name: Verify branch
+        if: ${{ github.base_ref != 'main' }}
+        run: |
+          echo "This workflow can only run as-is on the main branch. Please remove this conditional check and make sure GWC and GS are checked out on the right branches for this GeoTools series (see the --branch hints below). Current branch is ${GITHUB_BASE_REF}"
```
mprins, line 18:
```suggestion
        uses: actions/github-script@v7
        script: |
          core.setFailed("This workflow can only run as-is on the main branch. Please remove this conditional check and make sure GWC and GS are checked out on the right branches for this GeoTools series (see the --branch hints below). Current branch is ${GITHUB_BASE_REF}")
```

this might work, according to https://stackoverflow.com/a/68777117/5454667

aaime: Let's give it a try

aaime: Apparently not:

```
[Invalid workflow file: .github/workflows/integration.yml#L18](https://github.com/geotools/geotools/actions/runs/10301727692/workflow)
The workflow is not valid. .github/workflows/integration.yml (Line: 18, Col: 9): Unexpected value 'script'
```

mprins:
@aaime It seems to be working now

aaime:
Awesome, thanks!

PR #4868, opened 2024-08-07
[GEOT-7629] Add validation when deserializing objects

sikeoka:
[GEOT-7629](https://osgeo-org.atlassian.net/browse/GEOT-7629) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)
This PR uses ValidatingObjectInputStream from the Commons IO library to to restrict the classes that are allowed when deserializing imagemosaic sample image and histogram files and graph files. There are multiple existing unit tests that test deserializing valid graph files. The sample image files are the primary concern for potential backwards-compatibility issues.

SimpleFeatureIO was restricted to only allow deserialization when the constructor is called with an empty file with a system property to completely enable or disable deserialization. This allows the existing use cases where a new temp file is created and a single SimpleFeatureIO instance will serialize data to the file and then deserialize the data back from the file.

An unused method in the Base64 class was deprecated with a warning to discourage people from using it.

  

# Checklist

### DIFF
--- a/modules/library/main/src/main/java/org/geotools/data/sort/SimpleFeatureIO.java
+++ b/modules/library/main/src/main/java/org/geotools/data/sort/SimpleFeatureIO.java
@@ -51,6 +51,8 @@ public class SimpleFeatureIO {
     public static final int MAX_BYTES_LENGTH = 65535;
     public static final String BIG_STRING = "bigString";
 
+    public static final String ENABLE_DESERIALIZATION = "geotools.simplefeatureio.deserialization";
+
     RandomAccessFile raf;
 
     SimpleFeatureType schema;
@@ -65,11 +67,16 @@ public class SimpleFeatureIO {
     ByteArrayOutStream bos;
     byte[] buffer = new byte[8192]; // reusable buffer for small geometries
 
+    private final boolean initialFileEmpty;
+    private final String enableDeserialization;
+
     public SimpleFeatureIO(File file, SimpleFeatureType schema) throws FileNotFoundException {
         this.file = file;
         this.raf = new RandomAccessFile(file, "rw");
         this.schema = schema;
         this.builder = new SimpleFeatureBuilder(schema);
+        this.initialFileEmpty = this.file.length() <= 0;
+        this.enableDeserialization = System.getProperty(ENABLE_DESERIALIZATION);
     }
 
     /** Writes the feature to the file */
@@ -269,6 +276,12 @@ public class SimpleFeatureIO {
 
     @SuppressWarnings("BanSerializableRead")
     private Object readObject() throws IOException {
+        if ("false".equalsIgnoreCase(this.enableDeserialization)) {
+            throw new IllegalStateException("Object deserialization is not allowed");
+        } else if (!this.initialFileEmpty && !"true".equalsIgnoreCase(this.enableDeserialization)) {
+            throw new IllegalStateException(
+                    "Object deserialization is only allowed when created with an empty file");
+        }
         int length = raf.readInt();
         byte[] buffer = new byte[length];
         raf.read(buffer);
--- a/modules/library/metadata/src/main/java/org/geotools/util/Base64.java
+++ b/modules/library/metadata/src/main/java/org/geotools/util/Base64.java
@@ -898,7 +898,9 @@ public class Base64 {
      * @param encodedObject The Base64 data to decode
      * @return The decoded and deserialized object
      * @since 1.5
+     * @deprecated This method is not secure and will be removed in a future release.
      */
+    @Deprecated
     @SuppressWarnings("PMD.UseTryWithResources") // works fully in memory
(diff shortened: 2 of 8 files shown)


modules/library/main/src/main/java/org/geotools/data/sort/SimpleFeatureIO.java
```diff
@@ -269,6 +276,12 @@ private byte[] getByteBuffer(int length) {
 
     @SuppressWarnings("BanSerializableRead")
     private Object readObject() throws IOException {
+        if ("false".equalsIgnoreCase(this.enableDeserialization)) {
+            throw new IllegalStateException("Object deserialization is not allowed");
+        } else if (!this.initialFileEmpty && !"true".equalsIgnoreCase(this.enableDeserialization)) {
+            throw new IllegalStateException(
+                    "Object deserialization is only allowed when created with an empty file");
+        }
```
aaime, line 284: SimpleFeatureIO is used to perform sort/merge over secondary memory (when past a certain limit of features, below it works in memory). This is the default behavior of ContentDataStore, so all stores inherit it unless they can provide native sorting. 

Now, most attributes will be simple types and won't be affected, but it's not uncommon for stores nowadays to add Map objects as attributes to support more complex types. For example, PostgreSQL support for hstore and JSON uses maps (likely saved by native order by encoding), and I'm guessing ElasticSearch and Mongo too (but they also support native sorting).

Long story short, while I cannot find an example right away, the chance of breaking some existing implementation out there is not insignificant. This should probably be marked in the upgrade notes for both GeoTools and GeoServer.

modules/library/metadata/src/main/java/org/geotools/util/Base64.java
```diff
@@ -898,7 +898,9 @@ public static byte[] decode(String s) {
      * @param encodedObject The Base64 data to decode
      * @return The decoded and deserialized object
      * @since 1.5
+     * @deprecated This method is not secure and will be removed in a future release.
```
aaime, line 901: Doesn't seem to be used anyways, so looks like a safe removal

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/Utils.java
```diff
@@ -236,6 +235,25 @@ public class Utils {
...
+        float[].class,
+        int[].class,
+        short[].class,
+        java.math.BigInteger.class,
+        java.net.InetAddress.class,
+        java.util.Hashtable.class
+    };
+
+    /** Patterns for qualified class names to allow when deserializing SampleImage objects */
+    private static final String[] SAMPLE_IMAGE_PATTERNS = {
```
aaime, line 253: Yep, you stumbled into the reason why serialization was used in the first place. There are so many possible combinations of sample and color models that it's not practical to try and enumerate them all.

This has some potential to break existing implementations that use custom-written image readers. The patterns here should be made configurable with a system variable.

aaime (changes requested):
Looks good, seem comment below

modules/library/main/src/main/java/org/geotools/data/sort/SimpleFeatureIO.java
sikeoka: There are two existing cases where SimpleFeatureIO is used and both create a single SimpleFeatureIO object on an empty temporary file and then write to and read from the file from the same object. The default behavior was specifically chosen to allow this behavior.
https://github.com/geotools/geotools/blob/main/modules/library/main/src/main/java/org/geotools/data/sort/MergeSortDumper.java#L137
https://github.com/geotools/geotools/blob/main/modules/library/render/src/main/java/org/geotools/renderer/lite/MarkFeatureIterator.java#L161

aaime: Isn't "enableDeserialization" is false by default? Then the second if won't be reached at all... Oh... by default it's the empty string... that's why I found it confusing, normally for boolean system variables we use Boolean.getBoolean instead.

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/Utils.java
sikeoka: I added a property to allow administrators to allow additional classes for sample image files.

aaime: Thanks, that helps. It should be also noted in the upgrade notes (both the stricter checks and the system variable to allow other classes be deserialized).

sikeoka:
I created https://github.com/geoserver/geoserver/pull/7855 to add the new property to the GeoServer upgrade notes.

jodygarnett:
Are we comfortable back porting this one? I guess we are ..

PR #4880, opened 2024-08-19
[GEOT-7634] Update DWITHIN delegation in Mongo to support linestring

Mike7311:
https://osgeo-org.atlassian.net/browse/GEOT-7634

Current implementation of DWITHIN for Mongo is being delegated to Mongo's $near operator, however, this operator accepts only points. This commit refactors delegation to instead make use of $geoIntersects, which accepts other geometry types.
The buffer is being applied to input geometry, and resulting geometry is being used in $geoIntersects.

# Checklist

### DIFF
--- a/modules/plugin/mongodb/src/main/java/org/geotools/data/mongodb/AbstractFilterToMongo.java
+++ b/modules/plugin/mongodb/src/main/java/org/geotools/data/mongodb/AbstractFilterToMongo.java
@@ -28,8 +28,10 @@ import java.util.ArrayList;
 import java.util.Date;
 import java.util.List;
 import java.util.Set;
+import java.util.logging.Logger;
 import java.util.regex.Pattern;
 import org.bson.types.ObjectId;
+import org.geotools.api.feature.type.FeatureType;
 import org.geotools.api.filter.And;
 import org.geotools.api.filter.BinaryComparisonOperator;
 import org.geotools.api.filter.ExcludeFilter;
@@ -85,13 +87,23 @@ import org.geotools.api.filter.temporal.OverlappedBy;
 import org.geotools.api.filter.temporal.TContains;
 import org.geotools.api.filter.temporal.TEquals;
 import org.geotools.api.filter.temporal.TOverlaps;
+import org.geotools.api.referencing.FactoryException;
+import org.geotools.api.referencing.crs.CoordinateReferenceSystem;
+import org.geotools.api.referencing.crs.GeographicCRS;
+import org.geotools.api.referencing.operation.MathTransform;
+import org.geotools.api.referencing.operation.TransformException;
 import org.geotools.data.mongodb.complex.JsonSelectAllFunction;
 import org.geotools.data.mongodb.complex.JsonSelectFunction;
 import org.geotools.data.util.DistanceBufferUtil;
 import org.geotools.filter.FilterAttributeExtractor;
+import org.geotools.geometry.jts.JTS;
+import org.geotools.referencing.CRS;
+import org.geotools.referencing.crs.DefaultGeographicCRS;
 import org.geotools.util.Converters;
+import org.locationtech.jts.geom.Coordinate;
 import org.locationtech.jts.geom.Envelope;
 import org.locationtech.jts.geom.Geometry;
+import org.locationtech.jts.geom.Point;
 
 /**
  * Abstract visitor responsible for generating a BasicDBObject to use as a MongoDB query.
@@ -103,7 +115,15 @@ import org.locationtech.jts.geom.Geometry;
  */
 public abstract class AbstractFilterToMongo implements FilterVisitor, ExpressionVisitor {
 
+    public static final int HUNDRED_KM_IN_METERS = 100000;
+    protected static Logger LOGGER =
+            org.geotools.util.logging.Logging.getLogger(AbstractFilterToMongo.class);
+
+    private static final int DEFAULT_SEGMENTS = 10;
     protected final MongoGeometryBuilder geometryBuilder;
+
+    /** The schmema the encoder will use as reference to drive filter encoding */
+    protected FeatureType featureType;
 
     /** The whole world in WGS84 */
     private static final Envelope WORLD = new Envelope(-179.99, 179.99, -89.99, 89.99);
@@ -123,6 +143,15 @@ public abstract class AbstractFilterToMongo implements FilterVisitor, Expression
         return new BasicDBObject();
     }
 
+    /**
+     * Sets the feature type the encoder is encoding a filter for.
+     *
+     * <p>The type of the attributes may drive how the filter is translated to a mongodb query
+     * document.
+     */
+    public void setFeatureType(FeatureType featureType) {
+        this.featureType = featureType;
+    }
     //
     // primitives
     //
@@ -146,14 +175,14 @@ public abstract class AbstractFilterToMongo implements FilterVisitor, Expression
     }
 
     /**
-     * Method responsible of retrieving the MongoDB geometry json path
+     * Method responsible for retrieving the MongoDB geometry json path
      *
      * @return the MongoDB json paath for the default geometry
      */
     protected abstract String getGeometryPath();
 
     /**
-     * Method responsible of mapping a PropertyName to the corresponding MongoDB json path
+     * Method responsible for mapping a PropertyName to the corresponding MongoDB json path
      *
      * @param property the string property name to map
      * @return the MongoDB json path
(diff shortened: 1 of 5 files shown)


modules/plugin/mongodb/src/main/java/org/geotools/data/mongodb/AbstractFilterToMongo.java
```diff
@@ -527,14 +559,35 @@ public Object visit(DWithin filter, Object extraData) {
...
 
+        CoordinateReferenceSystem coordinateReferenceSystem =
+                featureType.getCoordinateReferenceSystem();
+        Unit<?> unit = coordinateReferenceSystem.getCoordinateSystem().getAxis(0).getUnit();
+        double distanceInMeters = DistanceBufferUtil.getDistanceInMeters(filter);
+        if (!METRIC_UNITS.contains(unit.toString())) {
+            MathTransform transform = null;
+            try {
+                // Transform geometry to CRS that supports meters
+                transform = CRS.findMathTransform(coordinateReferenceSystem, EPSG_3857);
```
aaime, line 571: If the reference geometry is a point, I would leave the $near operator, it's likely to use proper ellipsoidic distances.

For all other geometries, EPSG:3857 is a poor choice for the task at hand, the deformation factor for lengths diverges towards the poles, and it's already around 2 times at 60° north/south.

A good projection for local geometries with equidistancy properties (in a small area at least) is the Azimuthal Equidistant, and there is an auto-code to re-center it. Suggestion:

* Determine the centroid of the geometry
* Create an azimuthal equidistant projection centered on it using the "AUTO:97003,9001,{lon},{lat}" projection, e.g., ``CRS.decode("AUTO:97003,9001,-110,20")``
* Then apply the same algorithm as you are now.

Mind, this will still just work for "small" areas , e.g., less than 100km from the center. For bigger geometries you'll probably have to split the reference geometry into multiple smaller bits, buffer each one within its own local projection, reproject back to native, then union them into a reference geometry.

aaime (changes requested):

modules/plugin/mongodb/src/main/java/org/geotools/data/mongodb/AbstractFilterToMongo.java
Mike7311: Refactored the code as proposed to make use of azimuthal equidistant projection. However one doubt I have is splitting the "big" geometry into smaller parts and processing them separately before merging them again. The implementation sounds not trivial and as I understand can affect the performance negatively. Checking the buffer applied to 2 linestrings, one along the shore of Norway and another from south pole to north I did not encountered any noticeable inaccuracies (visible to my eye), so I wonder if such if precision offered by splitting the geometry is required.

modules/plugin/mongodb/src/main/java/org/geotools/data/mongodb/AbstractFilterToMongo.java
```diff
@@ -520,27 +543,88 @@ public Object visit(Disjoint filter, Object extraData) {
...
+            DBObject geometryDBObject = geometryBuilder.toObject(geometry);
+            dbo =
+                    BasicDBObjectBuilder.start()
+                            .push("$near")
+                            .add("$geometry", geometryDBObject)
+                            .add("$maxDistance", distanceInMeters)
+                            .get();
+            // Other type of geometry, use $geoIntersects
+        } else {
+            MathTransform transform = null;
```
aaime, line 578: Transform is not used outside of the try/catch... so why declare it before it?

modules/plugin/mongodb/src/main/java/org/geotools/data/mongodb/AbstractFilterToMongo.java
```diff
@@ -520,27 +543,88 @@ public Object visit(Disjoint filter, Object extraData) {
...
+
+                // Apply buffer in meters to geometry
+                Geometry bufferedGeometry = geometry.buffer(distanceInMeters, DEFAULT_SEGMENTS);
+
+                // Transform back to original with buffer applied
+                transform =
+                        CRS.findMathTransform(azimuthalEquidistantCRS, coordinateReferenceSystem);
+                geometry = JTS.transform(bufferedGeometry, transform);
+            } catch (FactoryException | TransformException e) {
+                throw new RuntimeException(e);
```
aaime, line 595: Provide some context, e.g.:

```suggestion
                throw new RuntimeException("Failed to compute polygon within distance from reference geometry", e);
```

modules/plugin/mongodb/src/main/java/org/geotools/data/mongodb/AbstractFilterToMongo.java
```diff
@@ -520,27 +543,88 @@ public Object visit(Disjoint filter, Object extraData) {
...
+            throws FactoryException, TransformException {
+        Point centroid = geometry.getCentroid();
+        if (!(coordinateReferenceSystem instanceof GeographicCRS)) {
+            MathTransform mathTransform =
+                    CRS.findMathTransform(coordinateReferenceSystem, DefaultGeographicCRS.WGS84);
+            centroid = (Point) JTS.transform(centroid, mathTransform);
+        }
+
+        double lon = centroid.getX();
+        double lat = centroid.getY();
```
aaime, line 620: Mind that a GeographicCRS can still be in NORTH/EAST axis order (where x is latitude). One easy way out is to add a condition to the if above, testing CRS.getAxisOrder(coordinateReferenceSsytem)... the reprojection to WGS84 will also give you the axis order you want.

aaime (changes requested):

modules/plugin/mongodb/src/main/java/org/geotools/data/mongodb/AbstractFilterToMongo.java
aaime: You can start with the current implementation, it's better than nothing. If the distance happens to be greater than 100km maybe just log (at fine level) that the accuracy is likely not good enough.

modules/plugin/mongodb/src/main/java/org/geotools/data/mongodb/AbstractFilterToMongo.java
Mike7311: Moved to place where it used

modules/plugin/mongodb/src/main/java/org/geotools/data/mongodb/AbstractFilterToMongo.java
Mike7311: Added message

modules/plugin/mongodb/src/main/java/org/geotools/data/mongodb/AbstractFilterToMongo.java
Mike7311: Added check to if condition mentioned above

modules/plugin/mongodb/src/main/java/org/geotools/data/mongodb/AbstractFilterToMongo.java
Mike7311: Added suggested logging in logAdditionalInformation method.

modules/plugin/mongodb/src/main/java/org/geotools/data/mongodb/AbstractFilterToMongo.java
```diff
@@ -520,27 +548,97 @@ public Object visit(Disjoint filter, Object extraData) {
...
+                            .add("$geometry", geometryDBObject)
+                            .get();
+        }
 
         output.put((String) e1, dbo);
         return output;
     }
 
+    private static void logAdditionalInformation(Geometry geometry, Geometry transformedCentroid) {
+        if (geometry.intersects(transformedCentroid.buffer(HUNDRED_KM_IN_METERS).getBoundary())) {
```
aaime, line 618: Sigh no... if the dwithin distance is > 100KM then log the warning.

aaime: Wait .. hum.. actually... both! Your observation is right, the distances are preserved properly within 100km from the center of the projection. So it might be that the geometry is too big, or the distance is too big. But I'm not sure about the test... it would probably be easier if you take the buffered geometry, get its bbox, and compute the distance between one of its corners and the centroid. If > 100km, then we went too far.

Mike7311: updated accordingly

Mike7311: Changed to checking both distance from centroid to corners of envelope and input buffer size.

aaime (approved):

PR #4904, opened 2024-09-16
Release 32.0

petersmythe:
# Checklist

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -20,10 +20,3 @@ updates:
       interval: 'monthly'
     target-branch: '31.x'
     open-pull-requests-limit: 10
-
-  - package-ecosystem: 'github-actions'
-    directory: '/'
-    schedule:
-      interval: 'monthly'
-    target-branch: '30.x'
-    open-pull-requests-limit: 10
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,6 +11,13 @@ updates:
     directory: '/'
     schedule:
       interval: 'monthly'
+    target-branch: '32.x'
+    open-pull-requests-limit: 10
+
+  - package-ecosystem: 'github-actions'
+    directory: '/'
+    schedule:
+      interval: 'monthly'
     target-branch: '31.x'
     open-pull-requests-limit: 10
 
--- a/docs/developer/procedures/release.rst
+++ b/docs/developer/procedures/release.rst
@@ -121,13 +121,13 @@ When creating the first release candidate of a series, there are some extra step
     
     Even if you wish to continue build prior branches please disable the documentation builds.
 
-  * For the new stable branch create new jobs, duplicate from the two existing ``stable`` jobs (geotools-27.x and geotools-27.x-docs), editing branch specifier to the new branch (e.g. `27.x`)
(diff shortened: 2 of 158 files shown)


docs/developer/procedures/release.rst
```diff
@@ -136,14 +136,14 @@ When creating the first release candidate of a series, there are some extra step
     git commit -m "Change 28.x to stable branch"
     git push upstream 28.x
 
-  For the new `maintenance` branch, and the remote for the official GeoTools is called ``upstream``::
+  For the new `maintenance` branch (which was the previous `stable` branch)::
   
-    git checkout 26.x
```
petersmythe, line 153: @jodygarnett please check that these changes below are correct.

If 27.x was the stable branch, the new stable branch is 28.x (set above) and then 27.x should become the maintenance branch.  Does the 26.x branch ever get set to archive or similar?

.github/dependabot.yml
```diff
@@ -7,6 +7,13 @@ updates:
     target-branch: 'main'
     open-pull-requests-limit: 10
 
+  - package-ecosystem: 'github-actions'
+    directory: '/'
+    schedule:
+      interval: 'monthly'
+    target-branch: '32.x'
+    open-pull-requests-limit: 10
+
```
petersmythe, line 16: Should 30.x below not be deleted i.e. every branch shifts along one version?

docs/developer/procedures/release.rst
jodygarnett: yes, I think that happens further down the page.

.github/dependabot.yml
jodygarnett: Yes; although I also wonder why this is not just *main" branch, and then backport dependency change if appropriate? Like our policy is to always commit to "main" 1st and then backport...

petersmythe: That makes sense to me, I will adjust

mprins: No, read the doc plz

mprins: As discussed in the PR and ML each branch gets it's own PRs for these updates

petersmythe: Thanks Mark, are you perhaps talking about this documentation: https://docs.github.com/en/code-security/getting-started/dependabot-quickstart-guide?  It does make sense to have dependabot running on older branches (e.g. maintenance 31.x) as they might have different dependencies being checked, and the dependabot will prepare the appropriate PRs (I am not sure what a ML is).

So then my original question stands: the archive 30.x should be disabled/removed?  I think so.

petersmythe:
Does `backport-32.x` label get created automatically, or does that need to be added to the instructions @jodygarnett ?

docs/developer/procedures/release.rst
petersmythe: Suggested fix: https://github.com/geotools/geotools/pull/4907

.github/dependabot.yml
mprins: I apologise for being short previously; dependabot is not configured to check dependencies (apart from security updates by default on the main branch); the idea is to keep the workflows/actions for each branch up-2-date, at some point some of the older branches can be removed from the dependabot.yml config, but we still see commits to branches that have reached EOL when there is a crucial security update

PR #4932, opened 2024-10-01
Improve rendering of one ways from MBStyle styles

aaime:
A couple of fixes ensure one way is rendered properly (pointing in the right direction and centered).

# Checklist


modules/extension/mbstyle/src/main/java/org/geotools/mbstyle/layer/SymbolMBLayer.java
```diff
@@ -1846,38 +1851,42 @@ public List<FeatureTypeStyle> transformInternal(MBStyle styleContext) {
                         halo,
                         fill);
         symbolizer.fonts().clear();
-        symbolizer.fonts().addAll(fonts);
+        // no need to add fonts and alternatives if we are displaying a simple symbol
+        if (!pureSymbol) {
+            symbolizer.fonts().addAll(fonts);
+        } else {
+            // tiny font
+            Font font =
```
turingtestfail, line 1859: This should be created statically with a comment at the top of the class since it is basically a constant representing the tiny font

turingtestfail (changes requested):

modules/extension/mbstyle/src/main/java/org/geotools/mbstyle/layer/SymbolMBLayer.java
aaime: Yep, done

aaime:
One funny side effect, the "symbol-spacing" is now working as intended, the goldfish were supposed to have a symbol-spacing of 50, which according to spec is "Distance between two symbol anchors". And indeed now it's 50 between goldfishes. I had to update the expected image result .

turingtestfail (approved):

PR #4984, opened 2024-11-13
[GEOT-7680] fix set UseExisting feature id

kfgisbis:
My bug fix: https://osgeo-org.atlassian.net/browse/GEOT-7680

## Checklist
 
#### For core and extension modules:
 [Documentation](https://github.com/geotools/geotools/tree/main/docs) has been updated (if change is visible to end users).


modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/internal/v1_x/StrictWFS_1_x_Strategy.java
```diff
@@ -47,20 +47,7 @@
...
-import net.opengis.wfs.InsertElementType;
-import net.opengis.wfs.OperationsType;
-import net.opengis.wfs.PropertyType;
-import net.opengis.wfs.QueryType;
-import net.opengis.wfs.ResultTypeType;
-import net.opengis.wfs.TransactionType;
-import net.opengis.wfs.UpdateElementType;
-import net.opengis.wfs.WFSCapabilitiesType;
-import net.opengis.wfs.WfsFactory;
+import net.opengis.wfs.*;
```
mprins, line 50: wildcard imports are not allowed

modules/unsupported/wfs-ng/src/test/java/org/geotools/data/wfs/WFSFeatureStoreTest.java
```diff
@@ -41,20 +46,35 @@
...
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.BeforeClass;
 import org.junit.Test;
-import org.locationtech.jts.geom.Coordinate;
-import org.locationtech.jts.geom.GeometryFactory;
-import org.locationtech.jts.geom.Point;
-import org.locationtech.jts.geom.PrecisionModel;
+import org.locationtech.jts.geom.*;
+import org.w3c.dom.*;
```
mprins, line 75: wildcard imports are not allowed

modules/library/main/src/main/java/org/geotools/api/data/QueryCapabilities.java
```diff
@@ -95,7 +95,7 @@ public boolean isReliableFIDSupported() {
      * identifier.
      */
     public boolean isUseProvidedFIDSupported() {
-        return false;
+        return true;
```
mprins, line 98: this is likely to break existing applications as it inverts the logic

modules/unsupported/wfs-ng/src/test/java/org/geotools/data/wfs/WFSFeatureStoreTest.java
```diff
@@ -21,12 +21,17 @@
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
 
-import java.io.IOException;
-import java.io.PrintWriter;
-import java.io.StringWriter;
+import java.io.*;
```
mprins, line 24: wildcard imports are not allowed

modules/unsupported/wfs-ng/src/test/java/org/geotools/data/wfs/integration/v1_1/GeoServerIntegrationTest.java
```diff
@@ -16,28 +16,50 @@
...
+import org.geotools.feature.simple.SimpleFeatureTypeBuilder;
+import org.geotools.filter.identity.FeatureIdImpl;
+import org.geotools.ows.ServiceException;
 import org.geotools.referencing.CRS;
+import org.geotools.util.factory.Hints;
 import org.junit.Ignore;
 import org.junit.Test;
-import org.locationtech.jts.geom.Coordinate;
-import org.locationtech.jts.geom.GeometryFactory;
+import org.locationtech.jts.geom.*;
```
mprins, line 50: wildcard imports are not allowed

mprins (changes requested):
please make sure to follow the steps in the developer manual, spcifically regarding running maven commands for formatting and Q&A

modules/library/main/src/main/java/org/geotools/api/data/QueryCapabilities.java
kfgisbis: It is not clear, because the method does not carry any logic, it only returns false. I have only one option, not to take this parameter into account when adding feature. I hope this goes well for you.

aaime: This change is not acceptable, not here at least. 

The QueryCapabilities default state indicates what a generic store without any special ability and support can do. Take for example the shapefile data store, it cannot possibly accept user provided identifies, because identifiers are generated from the row number in the shapefile itself... it just cannot be modified.

There are currently a few stores that override this method:
* PropertyDataStore, MemoryDataStore (always supported, they use a string)
* JDBCDataStore, checking if the table has a primary key column 

In your case, you have to build a custom subclass of QueryCapabilities in the WFS store, and return true or false depending on the abilities of the WFS server you're connecting to (it must be a WFS 1.1 server, idgen is only supported in that version).

aaime (changes requested):
As indicated in comments, should be reworked with a query capabilities that's specific to the WFS store, checking the remote server is a version 1.1

kfgisbis:
first failed test 
```
package org.geotools.referencing.proj does not exist
```
I did not do something that can break it

second test - project gt-wmts
I assume that it does not be linked with my changes

And third test failed not from my changes too
```
The `brew link` step did not complete successfully
```

modules/library/main/src/main/java/org/geotools/data/store/ContentFeatureStore.java
```diff
@@ -283,7 +283,7 @@ FeatureId addFeature(
         // pass through the fid if the user asked so
         boolean useExisting =
                 Boolean.TRUE.equals(feature.getUserData().get(Hints.USE_PROVIDED_FID));
-        if (getQueryCapabilities().isUseProvidedFIDSupported() && useExisting) {
```
aaime, line 286: This change needs to be removed as well. You need to implement a custom QueryCapabilties in the WFSFeatureSource, but instead of having it return "true", if you need to check if the remove WFS server can support the idgen attribute:

https://github.com/geotools/geotools/blob/dad50dbda0d4204ce5e5dae099c3c1da3c096e4a/modules/library/main/src/main/java/org/geotools/data/memory/MemoryFeatureStore.java#L103

aaime (changes requested):

modules/library/main/src/main/java/org/geotools/data/store/ContentFeatureStore.java
kfgisbis: we must look on section Transaction in getCapabilities and if we have value UseExisting in parameter idgen and after this  set in our realisation QueryCapabilties useProvidedFIDSupported = true, right?

aaime (approved):
Looks good to me

PR #5000, opened 2024-11-27
[GEOT-7688] Graduate cql2-text and cql2-json to supported status

aaime:
[GEOT-7688](https://osgeo-org.atlassian.net/browse/GEOT-7688) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

See geotools-devel discussion

# Checklist


modules/library/cql2-json/src/main/java/org/geotools/filter/text/cqljson/ExpressionToCQL2Json.java
```diff
@@ -182,31 +183,46 @@ private void toPeriod(ArrayNode output, Period period) {
         }
     }
 
-    private void toDate(ArrayNode output, Object value) {
+    private void toDate(ArrayNode output, java.sql.Date value) {
```
pmauduit, line 186: CQL requires `java.sql.Date` instead of the other Date classes from the JVM, which are less specific to SQL ?

pmauduit (approved):
I was just surprised by the use of `java.sql.Date` in the CQL module (which is not necessarily about SQL databases ?) instead of the other types from the JVM, but maybe I am lacking of deep knowledge of the codebase.

modules/library/cql2-json/src/main/java/org/geotools/filter/text/cqljson/ExpressionToCQL2Json.java
aaime: java.util.Date is actually a full timestamp, so the convention is GeoTools is to use java.sql.Date to represent simple dates, as it's unmistakably just a date without a time.
Eventually GT should switch to use the new Java date/time framework, but nobody is hard pressed to do it.

jodygarnett:
congrats!

PR #5014, opened 2024-12-04
[GEOT-7681] encoding fix when updating geometry in WFS 1.1

kfgisbis:
[GEOT-7681](https://osgeo-org.atlassian.net/browse/GEOT-7681) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

In WFS1.0-1.1, GML2 was used when updating geometry. As I understand it, WFS1.1 must have GML3, then the geometry is saved correctly.

My fix bug for [GEOT-7681](https://osgeo-org.atlassian.net/browse/GEOT-7681)

# Checklist


modules/extension/xsd/xsd-wfs/src/main/java/org/geotools/wfs/v1_1/PropertyTypeBinding_1_1.java
```diff
@@ -0,0 +1,87 @@
+package org.geotools.wfs.v1_1;
```
mprins, line 1: This file is missing the header

```suggestion
/*
 *    GeoTools - The Open Source Java GIS Toolkit
 *    http://geotools.org
 *
 *    (C) 2024, Open Source Geospatial Foundation (OSGeo)
 *
 *    This library is free software; you can redistribute it and/or
 *    modify it under the terms of the GNU Lesser General Public
 *    License as published by the Free Software Foundation;
 *    version 2.1 of the License.
 *
 *    This library is distributed in the hope that it will be useful,
 *    but WITHOUT ANY WARRANTY; without even the implied warranty of
 *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
 *    Lesser General Public License for more details.
 */

package org.geotools.wfs.v1_1;
```

modules/extension/xsd/xsd-wfs/src/main/java/org/geotools/wfs/v1_1/PropertyTypeBinding_1_1.java
```diff
@@ -0,0 +1,103 @@
...
+                                            gml.getNumDecimals(),
+                                            gml.getForceDecimalEncoding(),
+                                            gml.getPadWithZeros(),
+                                            "gml",
+                                            gml.getEncodeMeasures());
+                            geometryEncoder.encode(geometry, new AttributesImpl(), handler);
+
+                        } else {
+                            String s = value.toString();
+                            output.characters(s.toCharArray(), 0, s.length());
```
aaime, line 95: I see you basically adapted the WFS 1.0 PropertyTypeBinding, so it should work, but this part is leaving me wondering.... are dates encoded as ISO timestamps when going through this path?
All other types I can think of should "toString" into a valid GML representation.

Not that it's necessarily a blocker either... what happens now? If java.util.Data is toStringed anyways, this change would not introduce a regression and would be good to go as-is

kfgisbis: I copied this code from the PropertyTypeBinding class and changed only the geometry transformation in it. 
I had to do this because WFS1.0 should be handled as before, and PropertyTypeBinding_1_1 should be used in WFS1.1. 
My goal was to fix the geometry handling and not change the rest of the code, I think this should be a new fix task.

Also, I think the problem will be not only with the date; if the user wants to set the field as null, an error will occur. 
But this is now beyond the scope of my correction. In the future, I will try to offer my own edits on this matter.

aaime: You basically repeated what I said above... with one exception. Do you know what the current encoding output for date is, or not?

I've modified the WFSEncodingTest.encodeUpdate method to encode a Date, and it comes out as "Mon Dec 16 14:20:35 CET 2024"... it's not conformant. However, if I modify encodeInsert the same way (giving the encoder knowledge that I want a timestamp thought the attribute descriptor), then I get it encoded properly "2024-12-16T13:23:35.355Z"... it's however a different code path and has more context to judge the correct course of action.

I have no proof that the new code might introduce regressions in dates encoding, just a suspicion, so I'll let it go. Finger crossed.

kfgisbis: We did not pay attention to the date encoding; if I encounter problems regarding this, I will definitely offer my solution.

PR #5015, opened 2024-12-04
[GEOT-7556] Generate javadoc for Maven deploy

roarbra:
I would like to make a new try on generating javadoc jars when releasing a new version of GeoTools. This will ensure that we have a javadocs jar in the maven repository for each release.

I have ran the command within publish_release.sh with JDK 11, JDK 17 and JDK 21, without errors. Just a bunch of warnings. I started to fix them as well, but that task is a far fetch.

I have splitted in two commits. The last one is for publish_release.sh only, and the reason is that this easily can be reverted if a release fails.

# Checklist


mprins (approved):
lgtm, would be nice to have in eg. Intellij (though that will use the source jars as an alternative)

PR #5060, opened 2025-01-23
[GEOT-7709] WMTS GetCapabilities will not recognize ServiceException

roarbra:
[GEOT-7709](https://osgeo-org.atlassian.net/browse/GEOT-7709) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

Rearranged the constructor to avoid ClassCastException and added tests for two situations.

# Checklist

### DIFF
--- a/modules/extension/wmts/src/main/java/org/geotools/ows/wmts/response/WMTSGetCapabilitiesResponse.java
+++ b/modules/extension/wmts/src/main/java/org/geotools/ows/wmts/response/WMTSGetCapabilitiesResponse.java
@@ -17,29 +17,25 @@
 
 package org.geotools.ows.wmts.response;
 
-import java.io.BufferedReader;
-import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
-import java.io.InputStreamReader;
-import java.nio.charset.StandardCharsets;
 import java.util.Map;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.xml.parsers.ParserConfigurationException;
 import net.opengis.wmts.v_1.CapabilitiesType;
 import org.geotools.data.ows.GetCapabilitiesResponse;
+import org.geotools.data.ows.ServiceExceptionParser;
 import org.geotools.http.HTTPResponse;
 import org.geotools.ows.ServiceException;
 import org.geotools.ows.wmts.model.WMTSCapabilities;
 import org.geotools.util.logging.Logging;
 import org.geotools.wmts.WMTSConfiguration;
 import org.geotools.xsd.Parser;
-import org.xml.sax.InputSource;
 import org.xml.sax.SAXException;
 
 /**
- * Provides a hook up to parse the capabilties document from inputstream.
+ * Provides a hook up to parse the capabilities document from input stream.
  *
  * <p>(Based on existing work by rgould for WMS service)
  *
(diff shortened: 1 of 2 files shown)


modules/extension/wmts/src/main/java/org/geotools/ows/wmts/response/WMTSGetCapabilitiesResponse.java
```diff
@@ -54,43 +50,38 @@ public WMTSGetCapabilitiesResponse(HTTPResponse response) throws ServiceExceptio
...
-                throw (ServiceException) new ServiceException("Error while parsing XML.").initCause(e);
+                throw new IOException("Error while parsing XML.", e);
             }
-
-            if (object instanceof ServiceException) {
-                throw (ServiceException) object;
+            if (object instanceof CapabilitiesType) {
+                this.capabilities = new WMTSCapabilities((CapabilitiesType) object);
+            } else {
+                inputStream.reset();
```
aaime, line 75: This part is giving me pause. HTTPResponse returns a generic InputStream, which means reset might not be supported and could throw an exception.

Also, I don't see mark being called at the beginning, which has other implications, according to the javadoc:

```java
     * <li> If the method <code>markSupported</code> returns
     * <code>true</code>, then:
     *
     *     <ul><li> If the method <code>mark</code> has not been called since
     *     the stream was created, or the number of bytes read from the stream
     *     since <code>mark</code> was last called is larger than the argument
     *     to <code>mark</code> at that last call, then an
     *     <code>IOException</code> might be thrown.
     * ...
     * <li> If the method <code>markSupported</code> returns
     * <code>false</code>, then:
     *
     *     <ul><li> The call to <code>reset</code> may throw an
     *     <code>IOException</code>.
     *
     *     <li> If an <code>IOException</code> is not thrown, then the stream
     *     is reset to a fixed state that depends on the particular type of the
     *     input stream and how it was created. The bytes that will be supplied
     *     to subsequent callers of the <code>read</code> method depend on the
     *     particular type of the input stream. </ul></ul>
```

In other words, there is no guarantee of where the reset will end up, if ``mark`` is not getting called. And then, the code should somehow manage the possibility of mark not being supported to start with.

modules/extension/wmts/src/main/java/org/geotools/ows/wmts/response/WMTSGetCapabilitiesResponse.java
```diff
@@ -54,43 +50,38 @@ public WMTSGetCapabilitiesResponse(HTTPResponse response) throws ServiceExceptio
...
     public WMTSGetCapabilitiesResponse(HTTPResponse response, Map<String, Object> hints)
             throws ServiceException, IOException {
         super(response);
-
-        try {
-
-            Object object;
-            try (InputStream inputStream = response.getResponseStream()) {
+        Object object;
+        try (InputStream inputStream = response.getResponseStream()) {
```
aaime, line 65: The two responses... don't they have a different content type, to start with?
You could use that to decide what to parse.

However, that assumes the server is setting the right content type, making the code less tolerant to non compliant servers... So not sure it's actually a good idea...

aaime (changes requested):

modules/extension/wmts/src/main/java/org/geotools/ows/wmts/response/WMTSGetCapabilitiesResponse.java
roarbra: The fix is for non compliant servers. That don't set content-type:application/vnd.ogc.se_xml.
If the content-type is set according to the spec, a superclass will throw the ServiceException.

modules/extension/wmts/src/main/java/org/geotools/ows/wmts/response/WMTSGetCapabilitiesResponse.java
```diff
@@ -50,47 +48,51 @@ public class WMTSGetCapabilitiesResponse extends GetCapabilitiesResponse {
     private static final Logger LOGGER = Logging.getLogger(WMTSGetCapabilitiesResponse.class);
     private static WMTSConfiguration WMTS_CONFIGURATION = new WMTSConfiguration();
 
+    private static final int MAX_BUFFER_SIZE = 2048; // Should be enough to support the size of a ServiceException.
```
aaime, line 51: I would have merged but the PR is still marked as draft.
While I'm here, 2048 is probably enough but you never know, maybe add  a system variable for this limit so that one can tweak it, just in case.

PR #5072, opened 2025-02-06
[GEOT-7320] ServiceException from a WFS server GetFeature ends in a FactoryNotFoundException

roarbra:
[GEOT-7320](https://osgeo-org.atlassian.net/browse/GEOT-7320) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

Some WFS servers send ServiceException's like for other OGC services, even if the spec says differently.
This is a straight approach to use the same approach as other OGC services for parsing these ServiceException's and throw.

# Checklist

### DIFF
--- a/modules/unsupported/wfs-ng/src/test/java/org/geotools/data/wfs/internal/WFSResponseFactoryTest.java
+++ b/modules/unsupported/wfs-ng/src/test/java/org/geotools/data/wfs/internal/WFSResponseFactoryTest.java
@@ -1,0 +1,57 @@
+/*
+ * GeoTools - The Open Source Java GIS Toolkit http://geotools.org
+ *
+ * (C) 2025, Open Source Geospatial Foundation (OSGeo)
+ *
+ * This library is free software; you can redistribute it and/or modify it under the terms of the GNU Lesser General
+ * Public License as published by the Free Software Foundation; version 2.1 of the License.
+ *
+ * This library is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied
+ * warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU Lesser General Public License for more
+ * details.
+ */
+package org.geotools.data.wfs.internal;
+
+import static org.junit.Assert.assertEquals;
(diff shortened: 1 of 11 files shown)


modules/unsupported/wfs-ng/src/test/java/org/geotools/data/wfs/internal/WFSResponseFactoryTest.java
```diff
@@ -0,0 +1,53 @@
...
+    @Test(expected = ServiceException.class)
+    public void checkServiceExceptionIsThrown() throws Exception {
+        GetFeatureRequest request = createRequest();
+        String body = "<?xml version=\"1.0\" encoding=\"UTF-8\" standalone=\"no\"?>\n"
+                + "<ServiceExceptionReport>\n"
+                + "  <ServiceException>error occured</ServiceException>\n"
+                + "</ServiceExceptionReport>";
+        TestHttpResponse response = new TestHttpResponse(CONTENT_TYPE, "UTF-8", body);
+        WFSResponseFactory responseFactory = new ServiceExceptionResponseFactory();
+        responseFactory.createResponse(request, response);
```
aaime, line 51: I'd make this an assertThrows and then check tha the basics of the service exception are getting parsed properly (code, location, message)

aaime (changes requested):
Looks overall good, see feedback below.
Wondering about backporting vs keeping on main. The API has been broken but the module is not supported. Thoughts?

roarbra:
> Looks overall good, see feedback below. Wondering about backporting vs keeping on main. The API has been broken but the module is not supported. Thoughts?

I would very much see it back ported. The API changes will make it align with the other OGC services, and I can't see it will cause any trouble.

modules/unsupported/wfs-ng/src/test/java/org/geotools/data/wfs/internal/WFSResponseFactoryTest.java
roarbra: Rewritten as assertThrows. Location isn't set by ServiceExceptionParser.

PR #5116, opened 2025-02-26
[GEOT-7720] Сorrect determination of the number of decimal places in coordinates

kfgisbis:
[GEOT-7720](https://osgeo-org.atlassian.net/browse/GEOT-7720) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

  

# Checklist

### DIFF
--- a/modules/extension/xsd/xsd-wfs/src/main/java/org/geotools/wfs/v1_1/PropertyTypeBinding_1_1.java
+++ b/modules/extension/xsd/xsd-wfs/src/main/java/org/geotools/wfs/v1_1/PropertyTypeBinding_1_1.java
@@ -36,18 +36,14 @@ import org.xml.sax.helpers.AttributesImpl;
 import org.xml.sax.helpers.NamespaceSupport;
 
 public class PropertyTypeBinding_1_1 extends AbstractComplexEMFBinding {
-    private GMLConfiguration gml;
+    private final GMLConfiguration gml;
     private final GenericGeometryEncoder geometryEncoder;
     private static final String VALUE = "Value";
 
     public PropertyTypeBinding_1_1(WfsFactory factory, Configuration configuration) {
         super(factory);
-        gml = configuration.getDependency(GMLConfiguration.class);
 
-        if (gml == null) {
-            gml = new GMLConfiguration();
-        }
-
+        gml = initializeGMLConfiguration(configuration);
         Encoder encoder = new Encoder(gml);
         encoder.setInline(true);
         geometryEncoder = new GenericGeometryEncoder(encoder);
@@ -56,6 +52,16 @@ public class PropertyTypeBinding_1_1 extends AbstractComplexEMFBinding {
     @Deprecated
     public PropertyTypeBinding_1_1(WfsFactory factory) {
         this(factory, null);
+    }
+
+    private GMLConfiguration initializeGMLConfiguration(Configuration configuration) {
+        if (configuration == null) {
+            return new GMLConfiguration();
+        }
+
+        GMLConfiguration gml = configuration.getDependency(GMLConfiguration.class);
+
+        return gml == null ? new GMLConfiguration() : gml;
     }
 
     @Override
--- a/modules/extension/xsd/xsd-wfs/src/main/java/org/geotools/wfs/v1_1/PropertyTypeBinding_1_1.java
+++ b/modules/extension/xsd/xsd-wfs/src/main/java/org/geotools/wfs/v1_1/PropertyTypeBinding_1_1.java
@@ -36,16 +36,26 @@ import org.xml.sax.helpers.AttributesImpl;
 import org.xml.sax.helpers.NamespaceSupport;
 
 public class PropertyTypeBinding_1_1 extends AbstractComplexEMFBinding {
-    private final GMLConfiguration gml;
+    private GMLConfiguration gml;
     private final GenericGeometryEncoder geometryEncoder;
     private static final String VALUE = "Value";
 
     public PropertyTypeBinding_1_1(WfsFactory factory, Configuration configuration) {
         super(factory);
         gml = configuration.getDependency(GMLConfiguration.class);
+
+        if (gml == null) {
+            gml = new GMLConfiguration();
+        }
+
         Encoder encoder = new Encoder(gml);
         encoder.setInline(true);
         geometryEncoder = new GenericGeometryEncoder(encoder);
+    }
+
+    @Deprecated
+    public PropertyTypeBinding_1_1(WfsFactory factory) {
+        this(factory, null);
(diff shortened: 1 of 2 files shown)


modules/extension/xsd/xsd-wfs/src/main/java/org/geotools/wfs/v1_1/PropertyTypeBinding_1_1.java
```diff
@@ -67,20 +72,15 @@ public Object getProperty(final Object object, QName name) throws Exception {
...
-
-                    Encoder encoder = new Encoder(gml);
-                    encoder.setInline(true);
-
-                    GenericGeometryEncoder geometryEncoder = new GenericGeometryEncoder(encoder);
                     GMLWriter handler = new GMLWriter(
                             output,
                             new NamespaceSupport(),
-                            gml.getNumDecimals(),
+                            geometry.getPrecisionModel().getMaximumSignificantDigits(),
```
aaime, line 78: Not sure about this change, it's not in line with how the xml-xsd modules work, where the precision is picked by a GML configuration. The issue here is that the configuration is created locally, rather than being fetched from the pico container. Try adding the configuration as a constructor argument, pico works by constructor injection.
For reference, here is how GeoServer does it for its GML encoding:
https://github.com/geoserver/geoserver/blob/99ae245402fca51e306eaf9817faaa891bf89e73/src/wfs/src/main/java/org/geoserver/wfs/xml/GML3OutputFormat.java#L301

kfgisbis: @aaime I'm sorry, but I didn't quite get it. We should get the numDecimal from the GMLConfiguration, that's understandable, but where should we set this value?

If in this class, when processing the geometry value, does it make sense to set this value in GMLConfiguration, because the geometry can be different (the type of PrecisionModel is FIXED, FLOATING, FLOATING_SINGLE), and if we send a lot of elements in a transaction for updating, doesn't it make sense? Each element will retain its own value in the overall configuration. 

If we first process the values of all geometries and calculate the numDecimal (as done in geoserver in your example), at what point should we set this value in GMLConfiguration?
I'm probably confused and misunderstood, please help me figure it out.

kfgisbis:
@aaime, the test failed, I assume it is not related to my changes?

modules/extension/xsd/xsd-wfs/src/main/java/org/geotools/wfs/v1_1/PropertyTypeBinding_1_1.java
```diff
@@ -35,11 +36,16 @@
 import org.xml.sax.helpers.NamespaceSupport;
 
 public class PropertyTypeBinding_1_1 extends AbstractComplexEMFBinding {
-
+    private final GMLConfiguration gml;
+    private final GenericGeometryEncoder geometryEncoder;
     private static final String VALUE = "Value";
 
-    public PropertyTypeBinding_1_1(WfsFactory factory) {
+    public PropertyTypeBinding_1_1(WfsFactory factory, Configuration configuration) {
```
mprins, line 43: This changes the public API; that would at least require documenting in migration notes. Better, preserve the old API and deprecate that

kfgisbis: What do you mean by old API? PropertyTypeBinding_1_1 class craeted by me recently. Getting the configuration in the constructor is a recommendation from @aaime 

This edit is very important because otherwise there is no way to control the number of digits after the decimal point. If I'm wrong, tell me how to do it correctly?

mprins: I wasn't aware of that, in that case please ignore

aaime: While I did suggest to use constructor injection, that does not imply that backwards compatibility must get broken in the process. @mprins suggestion to keep also the old constructor while deprecating it should work, just  make the code safe against a null configuration?

kfgisbis: @aaime Sorry, I don't quite understand what you mean? I need to check if GMLConfiguration is null?

Something like:
```
public PropertyTypeBinding_1_1(WfsFactory factory, Configuration configuration) {
        super(factory);
        gml = configuration.getDependency(GMLConfiguration.class);

        if (gml == null) {
            gml = new GMLConfiguration();
        }
        
        Encoder encoder = new Encoder(gml);
        encoder.setInline(true);
        geometryEncoder = new GenericGeometryEncoder(encoder);
    }
```
Or do you mean something else?

modules/extension/xsd/xsd-wfs/src/main/java/org/geotools/wfs/v1_1/PropertyTypeBinding_1_1.java
```diff
@@ -35,11 +36,26 @@
...
+            gml = new GMLConfiguration();
+        }
+
+        Encoder encoder = new Encoder(gml);
+        encoder.setInline(true);
+        geometryEncoder = new GenericGeometryEncoder(encoder);
+    }
+
+    @Deprecated
+    public PropertyTypeBinding_1_1(WfsFactory factory) {
```
aaime, line 57: Yes, this is the deprecated constructor we were talking about. Now, spotbugs reports this:

```
2025-04-07T05:39:10.3554291Z [ERROR] High: Non-virtual method call in new org.geotools.wfs.v1_1.PropertyTypeBinding_1_1(WfsFactory) passes null for non-null parameter of new PropertyTypeBinding_1_1(WfsFactory, Configuration) [org.geotools.wfs.v1_1.PropertyTypeBinding_1_1] At PropertyTypeBinding_1_1.java:[line 58] NP_NULL_PARAM_DEREF_NONVIRTUAL
2025-04-07T05:39:10.3557697Z [INFO] 
```

Because indeed if invoked, it would cause a NPE, configuration is not guarded against null.

kfgisbis: Sorry, stupid mistake

mprins (approved):

aaime:
Thanks, merged and scheduled for automatic backport

PR #5130, opened 2025-03-12
[GEOT-7632]  WMS capabilities parsing should not fail if 'units' for a dimension is empty

skalesse:
[GEOT-7632](https://osgeo-org.atlassian.net/browse/GEOT-7632) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

This PR is to fix [[GEOT-7632]  WMS capabilities parsing should not fail if 'units' for a dimension is empty](https://osgeo-org.atlassian.net/browse/GEOT-7632). According to [OpenGIS Web Map Service WMS Implementation Specification](https://portal.ogc.org/files/?artifact_id=14416), for layer dimensions, `units` must **not** be missing, **but are allowed to be empty**:

> If the dimensional quantity has no units (e.g. band number in a multi-wavelength sensor), use the null string: units="".

`gt-wms` currently throws a `SAXException` in case `units.length == 0`. The two affected classes are `org.geotools.ows.wms.xml.WMSComplexTypes._DimensionType` and `org.geotools.ows.wms.xml.Dimension#Dimension`.

This PR fixes the problem by modifying the two above mentioned places to remove the length check while parsing the `units` attribute. Two unit tests have been added for some basic dimension value checks (it looks no such test-cases ever existed before). These checks include the described case. 

# Checklist

### DIFF
--- a/modules/extension/wms/src/main/java/org/geotools/ows/wms/xml/Dimension.java
+++ b/modules/extension/wms/src/main/java/org/geotools/ows/wms/xml/Dimension.java
@@ -31,6 +31,11 @@ package org.geotools.ows.wms.xml;
  * type="string"/> <attribute name="multipleValues" type="boolean"/> <attribute name="nearestValue" type="boolean"/>
  * <attribute name="current" type="boolean"/> </extension> </simpleContent> </complexType> </element>
  *
+ * <p>According to <a href="https://portal.ogc.org/files/?artifact_id=14416">OpenGIS Web Map Service WMS Implementation
+ * Specification - C.2 Declaring dimensions and their allowed value (page 52)</a>, units must not be missing, but their
+ * values are allowed to be empty: <i>"If the dimensional quantity has no units (e.g. band number in a multi-wavelength
+ * sensor), use the null string: units=""."</i>
+ *
  * @version SVN $Id$
  * @author Per Engstrom, Curalia AB, pereng@gmail.com
  */
@@ -38,6 +43,7 @@ public class Dimension {
     /** This name is often used as a lookup key */
     protected String name;
 
+    /** Must not be missing, but can be empty according to WMS spec 1.3, page 52 */
     protected String units;
 
     protected String unitSymbol;
@@ -51,7 +57,7 @@ public class Dimension {
         if (name == null || name.length() == 0) {
             throw new IllegalArgumentException("Error creating Extent: parameter name must not be null!");
         }
-        // unit must not be null, but can be empty
+        // unit must not be null, but can be empty according to WMS spec 1.3, page 52
         if (units == null) {
             throw new IllegalArgumentException("Error creating Extent: parameter units must not be null!");
         }
@@ -65,7 +71,7 @@ public class Dimension {
         if (name == null || name.length() == 0) {
             throw new IllegalArgumentException("Error creating Extent: parameter name must not be null!");
         }
-        // unit must not be null, but can be empty
+        // unit must not be null, but can be empty according to WMS spec 1.3, page 52
         if (units == null) {
             throw new IllegalArgumentException("Error creating Extent: parameter units must not be null!");
         }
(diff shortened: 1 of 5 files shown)


modules/extension/wms/src/test/java/org/geotools/ows/wms/xml/test/WMSComplexTypesTest.java
```diff
@@ -171,4 +172,127 @@ public void testEmptyOnlineResourceServiceDef() throws Exception {
...
+        // elevation dimension:
+        // GEOS-7632: units must not be missing, but can be empty
+        dimension = layer.getDimension("ELEVATION");
+        Assert.assertNotNull(dimension);
+        // unit
+        units = dimension.getUnits();
+        Assert.assertNotNull("Units must not be missing", units);
+        Assert.assertEquals("Unit should be empty .", "", units);
+        // unit symbol
+        Assert.assertEquals("Unit symbol should be 'm'", "m", dimension.getUnitSymbol());
```
aaime, line 296: The example seems a bit odd but ok, it's just a test.
According to the specification:

>  The elevation units identifier “verticalCRSid” refers to a vertical CRS as in 6.7.5. The unitSymbol attribute shall be chosen to match the units of the specified vertical CRS. 

An example of unit-less dimension is a band number, quoting the relevant section from page 52 (WMS 1.3 spec):

>   * If the dimensional quantity has no units (e.g. band number in a multi-wavelength sensor), use the null string: units="". 
>   * If the dimensional quantity has units, unit names should be taken from the Unified Code for Units of Measure (UCUM) if UCUM has an appropriate entry. When UCUM is used, the mandatory units attribute shall be an appropriate entry from the UCUM “name” column.
>  * If present, the unitSymbol shall be a 7-bit ASCII character string. When UCUM is used, the unitSymbol shall be the entry from the UCUM “c/s” column (case-sensitive abbreviation) corresponding to the UCUM name.  UCUM prefix names and symbols may be present as well.

modules/extension/wms/src/main/java/org/geotools/ows/wms/xml/Dimension.java
```diff
@@ -51,7 +51,8 @@ public Dimension(String name, String units, String unitSymbol) {
         if (name == null || name.length() == 0) {
             throw new IllegalArgumentException("Error creating Extent: parameter name must not be null!");
         }
-        if (units == null || units.length() == 0) {
+        // unit must not be null, but can be empty
+        if (units == null) {
```
aaime, line 55: This is a bizzarre choice in the spec, I would quote the relevant spec section in the class javadoc for clarification (to clarify the code is not just leveraging a loophole in the schema, it was the actual specification intention):

> * If the dimensional quantity has no units (e.g. band number in a multi-wavelength sensor), use the null string: units="".

aaime (changes requested):
Looks good, I would just clarify that it was really the specification intent in the Dimension.java javadoc.

modules/extension/wms/src/test/java/org/geotools/ows/wms/xml/test/WMSComplexTypesTest.java
skalesse: Yes you are right, the test is an odd example, but I just didn't want to make it more complicated. It still should show what we are aiming for, I guess.

skalesse:
My original intention had been to leave as much code the way it was, but given yours (and copilot's) comments, I have made the following two requested changes: 

- I have added the relevant section to the class JavaDoc of `Dimension.java`. Note that I have also cleaned up as much as possible the JavaDoc, because the code blocks and citations were formatted incorrectly. Note that I was not able to remove the empty lines in the block-quote. The code fomatter always brings them back. 
- I have fixed copilot's complains about the typos and wrong error messages. Note that while I was at it I have also removed the code duplication and fixed a minor complaint about the use of `String#length() == 0` vs. `String#isEmpty()`.

Let me know what you think. 
Thanks

aaime (approved):
Looks good to me

PR #5175, opened 2025-04-24
[GEOT-7745] PagingFeatureCollection doesn't work as expected when used at a STAC collection

roarbra:
[GEOT-7745](https://osgeo-org.atlassian.net/browse/GEOT-7745) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

When using the STACDataStore I discovered that some of the items were repeated.  I made a test with the json files in question. Tried to make them a little smaller.
The real problem lies in gt-geojson-core and the class PagingFeatureCollection.

# Checklist


modules/unsupported/geojson-core/src/main/java/org/geotools/data/geojson/GeoJSONReader.java
```diff
@@ -366,6 +366,13 @@ else if ("links".equalsIgnoreCase(name)) {
                 while (parser.nextToken() == JsonToken.START_OBJECT) {
                     ObjectNode node = mapper.readTree(parser);
                     JsonNode rel = node.get("rel");
+                    if (LOGGER.isLoggable(Level.FINE)) {
```
aaime, line 369: This looks more like a temporary debugging aid, would it help in a deployed application?

modules/unsupported/geojson-core/src/main/java/org/geotools/data/geojson/PagingFeatureCollection.java
```diff
@@ -51,23 +51,33 @@ public Integer getMatched() {
...
+    protected ObjectNode getNext() {
+        return next;
+    }
     /**
      * Reads the next feature collection, or return null if there is none. Subclasses can override if they need a
      * different logic for fetching the next page.
      */
     protected SimpleFeatureCollection readNext(ObjectNode next) throws IOException {
-        if (next == null) return null;
+        if (next == null) {
```
aaime, line 74: Could you please back out all these if related changes?

If you don't like this style, do it right:
* Ask on the list about enforcing curly braces for one liners
* Fix the entire code base (hopefully there is some tool to do it)
* Enforce the style with a QA check ([PMD can do it](https://pmd.github.io/pmd/pmd_rules_java_codestyle.html#controlstatementbraces))

Otherwise we'll end up in formatting petty wars between those that want braces everywhere and those that prefer more compact code.

modules/unsupported/geojson-core/src/main/java/org/geotools/data/geojson/PagingFeatureCollection.java
```diff
@@ -77,27 +87,30 @@ protected SimpleFeatureCollection readNext(ObjectNode next) throws IOException {
...
                     if (features instanceof PagingFeatureCollection) {
-                        this.next = ((PagingFeatureCollection) features).next;
+                        this.nextPage = ((PagingFeatureCollection) features).getNext();
+                        this.delegate = ((PagingFeatureCollection) features)
+                                .getFirstCollection()
+                                .features();
                     } else {
-                        this.next = null;
+                        this.nextPage = null;
+                        this.delegate = features.features();
```
aaime, line 113: This is the actual fix I guess?

aaime (changes requested):

modules/unsupported/geojson-core/src/main/java/org/geotools/data/geojson/PagingFeatureCollection.java
roarbra: Yes, this is the fix. The delegate should point at a ListFeatureCollection.features and not a PagingFeatureCollection.features. There could have been a clearer separation between this in the code.

modules/unsupported/geojson-core/src/main/java/org/geotools/data/geojson/PagingFeatureCollection.java
roarbra: Point taken.

PR #5176, opened 2025-04-24
[GEOT-7749] Add support for vendor parameters in WMSLayer

alan-geosolutions:
[GEOT-7749](https://osgeo-org.atlassian.net/browse/GEOT-7749) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

When making a cascading call to an external WMS ensure the vendor parameters are included in the request.
The vendor parameters are treated the same as other parameters and need to have protected characters encoded

ie 
Something like 
CQL_FILTER=LAND_KM > 200000

should be written
CQL_FILTER=LAND_KM%20%3E%20200000

Commas do not need to be encoded
CSV=5,10,20,5,1,3,1,,,,1
List=l1,l2,l3,l4,l5

Spaces need to be escaped
name=external%20map
 

# Checklist

- [☑️ ] I have read the [contribution guidelines](https://github.com/geotools/geotools/blob/main/CONTRIBUTING.md).
- [☑️] First PR targets the `main` branch (backports managed later; ignore for branch specific issues).
- [☑️] Avoid [Java 9+ split packages](http://tutorials.jenkov.com/java/modules.html#split-packages-not-allowed).

- [☑️] New unit tests have been added covering the changes.
- [] [Documentation](https://github.com/geotools/geotools/tree/main/docs) has been updated (if change is visible to end users).
- [☑️] There is an issue in [GeoTools Jira](https://osgeo-org.atlassian.net/projects/GEOT) (except for changes not visible to end users). 
- [☑️] Commit message(s) must be in the form ``[GEOT-XYZW] Title of the Jira ticket``.
- [☑️] Bug fixes and small new features are presented as a single commit.
- [☑️] The commit targets a single objective (if multiple focuses cannot be avoided, each one is in its own commit, and has a separate ticket describing it).

### DIFF
--- a/modules/extension/wms/src/main/java/org/geotools/ows/wms/map/WMSCoverageReader.java
+++ b/modules/extension/wms/src/main/java/org/geotools/ows/wms/map/WMSCoverageReader.java
@@ -16,6 +16,8 @@
  */
 package org.geotools.ows.wms.map;
 
+import static java.util.function.Predicate.not;
+
 import java.awt.Color;
 import java.awt.Rectangle;
 import java.awt.geom.AffineTransform;
@@ -28,6 +30,7 @@ import java.util.Collections;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import java.util.logging.Level;
 import java.util.logging.Logger;
@@ -402,6 +405,18 @@ public class WMSCoverageReader extends AbstractGridCoverage2DReader {
         ReferencedEnvelope requestEnvelope = gridEnvelope;
         mapRequest.setBBox(requestEnvelope);
         mapRequest.setSRS(requestSrs);
+
+        if (getLayers().size() > 1) {
+            LOGGER.warning("More than one layer found for when building Vendor parameters. Using the first one");
+        }
+        // If there are multiple layers find the first with vendor parameters to use
+        // there should only be one
+        getLayers().stream()
+                .map(Layer::getVendorParameters)
+                .filter(Objects::nonNull)
+                .filter(not(Map::isEmpty))
+                .findFirst()
+                .ifPresent(parameters -> parameters.forEach(mapRequest::setVendorSpecificParameter));
 
         this.mapRequest = mapRequest;
         this.requestedEnvelope = gridEnvelope;
--- a/modules/extension/wms/src/main/java/org/geotools/ows/wms/map/WMSCoverageReader.java
+++ b/modules/extension/wms/src/main/java/org/geotools/ows/wms/map/WMSCoverageReader.java
@@ -406,20 +406,23 @@ public class WMSCoverageReader extends AbstractGridCoverage2DReader {
         mapRequest.setBBox(requestEnvelope);
         mapRequest.setSRS(requestSrs);
 
-        var setsOfVendorParameters = getLayers().stream()
+        boolean anyVendorParameters = getLayers().stream()
+                .map(Layer::getVendorParameters)
+                .filter(Objects::nonNull)
+                .anyMatch(not(Map::isEmpty));
+
+        if (anyVendorParameters) {
+            LOGGER.warning("Vendor parameters have been found. Using the first one");
+        }
+
+        // If there are multiple layers find the first with vendor parameters to use
+        // there should only be one
+        getLayers().stream()
                 .map(Layer::getVendorParameters)
                 .filter(Objects::nonNull)
                 .filter(not(Map::isEmpty))
-                .collect(Collectors.toList());
-
-        // If there are multiple sets of vendor parameters use the first one
-        setsOfVendorParameters.stream()
                 .findFirst()
                 .ifPresent(parameters -> parameters.forEach(mapRequest::setVendorSpecificParameter));
-
-        if (setsOfVendorParameters.size() > 1) {
-            LOGGER.warning("More than one layer found for when building Vendor parameters. Using the first one");
-        }
 
         this.mapRequest = mapRequest;
         this.requestedEnvelope = gridEnvelope;
--- a/modules/extension/wms/src/main/java/org/geotools/ows/wms/map/WMSCoverageReader.java
+++ b/modules/extension/wms/src/main/java/org/geotools/ows/wms/map/WMSCoverageReader.java
@@ -406,17 +406,20 @@ public class WMSCoverageReader extends AbstractGridCoverage2DReader {
         mapRequest.setBBox(requestEnvelope);
         mapRequest.setSRS(requestSrs);
 
-        if (getLayers().size() > 1) {
-            LOGGER.warning("More than one layer found for when building Vendor parameters. Using the first one");
-        }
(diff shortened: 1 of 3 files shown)


modules/extension/wms/src/main/java/org/geotools/ows/wms/map/WMSCoverageReader.java
```diff
@@ -403,6 +406,18 @@ ReferencedEnvelope initMapRequest(ReferencedEnvelope bbox, int width, int height
         mapRequest.setBBox(requestEnvelope);
         mapRequest.setSRS(requestSrs);
 
+        if (getLayers().size() > 1) {
```
aaime, line 409: Shoulnd't this test if there are vendor parameters at all before logging this message?

aaime (changes requested):
The commit message should be "[GEOT-7749] Add support for vendor parameters in WMSLayer" too.

See a comment below about the warning log. Looks otherwise good, thanks!

modules/extension/wms/src/main/java/org/geotools/ows/wms/map/WMSCoverageReader.java
alan-geosolutions: Yes one minute

alan-geosolutions: // If there are multiple layers find the first with vendor parameters to use
        var setsOfParameters = getLayers().stream()
                .map(Layer::getVendorParameters)
                .filter(Objects::nonNull)
                .filter(not(Map::isEmpty))
                .collect(Collectors.toList());

        setsOfParameters.stream()
                .findFirst()
                .ifPresent(parameters -> parameters.forEach(mapRequest::setVendorSpecificParameter));

        // When there are more than 1 issue a warning
        if (setsOfParameters.size() > 1) {
            LOGGER.warning("More than one layer found for when building Vendor parameters. Using the first one");
        }
        
        This should do the trick

alan-geosolutions: Maybe setsOfVendorParameters

aaime: I believe you just need to check if there is at least one vendor parameter?

boolean anyVendorParameter =  getLayers().stream()
            .map(Layer::getVendorParameters)
            .filter(Objects::nonNull)
            .anyMatch(not(Map::isEmpty));

Also mind, there was a discussion during the last PSC meeting about usage of var, consensus was that for projects with as many developers as GeoTools/GeoWebCache/GeoServer usage of var should be banned.

It's not yet enforced via PMD, but it will be soon, so best not to add usages of it in the meantime.

On the other end, we agreed on the gradual integration of Lombok, whiltelisting usage of the annotations to handle beans only (e.g. @Data/@Value and individual usage of setter/getter/equals/hashcode ones too).
This is also going to be gradually introduced, as we'll need to update PMD to the latest version first (it would conflict with Lombok otherwise).

alan-geosolutions: What do we want to test for.

More than one set of vendor parameters in all of the layers
or 
There are vendor parameters and there is more than one layer.

If there are more than one set do we want to merge them or take the first set we find.

alan-geosolutions: Sorry about the var bad habit.

aaime: Thinking out loud, for multiple layers to be merged into a single GetMap request, they need to share the same vendor parameters... otherwise multiple getMap should be issued. This should be taken care of upstream, in this case in GeoServer GetMap, when the merge layer happens:
https://github.com/geoserver/geoserver/blob/main/src/wms/src/main/java/org/geoserver/wms/GetMap.java#L379

Basically, it's no longer enough to check if the two wms layers are coming from the same server, but they should also share the same set of vendor parameters (or both having them null/empty).

There are more complex cases in which the vendor parameter behaves like layers/styles, that is, it's a separated list with entries parallel to the layer list, but that's a fight for another day (there is no general rule to know if a vendor parameter is per layer or per request).

Going back to what the code should be doing here... just do a simple thing for now, warn if there are vendor parameters, but they are not equal across all layers.

aaime (changes requested):

modules/extension/wms/src/main/java/org/geotools/ows/wms/map/WMSCoverageReader.java
```diff
@@ -403,6 +406,24 @@ ReferencedEnvelope initMapRequest(ReferencedEnvelope bbox, int width, int height
...
         mapRequest.setSRS(requestSrs);
 
+        boolean anyVendorParameters = getLayers().stream()
+                .map(Layer::getVendorParameters)
+                .filter(Objects::nonNull)
+                .anyMatch(not(Map::isEmpty));
+
+        if (anyVendorParameters) {
+            LOGGER.warning("Vendor parameters have been found. Using the first one");
+        }
```
aaime, line 416: We should not have a warning every time there is a vendor parameter. I've left a comment describing the desired behavior, quoting it:

> Thinking out loud, for multiple layers to be merged into a single GetMap request, they need to share the same vendor parameters... otherwise multiple getMap should be issued. This should be taken care of upstream, in this case in GeoServer GetMap, when the merge layer happens:
https://github.com/geoserver/geoserver/blob/main/src/wms/src/main/java/org/geoserver/wms/GetMap.java#L379
>
> Basically, it's no longer enough to check if the two wms layers are coming from the same server, but they should also share the same set of vendor parameters (or both having them null/empty).
>
> There are more complex cases in which the vendor parameter behaves like layers/styles, that is, it's a separated list with entries parallel to the layer list, but that's a fight for another day (there is no general rule to know if a vendor parameter is per layer or per request).
>
> Going back to what the code should be doing here... just do a simple thing for now, warn if there are vendor parameters, but they are not equal across all layers.

modules/extension/wms/src/main/java/org/geotools/ows/wms/map/WMSCoverageReader.java
```diff
@@ -403,6 +406,25 @@ ReferencedEnvelope initMapRequest(ReferencedEnvelope bbox, int width, int height
         mapRequest.setBBox(requestEnvelope);
         mapRequest.setSRS(requestSrs);
 
+        long distinctVendorParameterSets = getLayers().stream()
+                .map(Layer::getVendorParameters)
+                .map(m -> Objects.isNull(m) ? Collections.emptyMap() : m)
```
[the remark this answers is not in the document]
aaime: Nitpick indeed, the two potential streams of vendor params are built with slightly different conditions.

aaime (approved):
Looks good, thanks!

PR #5196, opened 2025-05-19
[GEOT-7759] Adopt Maven Bill of Materials Pattern for GeoTools

groldan:
[GEOT-7759](https://osgeo-org.atlassian.net/browse/GEOT-7759) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

Implements the proposal to adopt the BOM pattern to manage GeoTools dependencies both internall and for downstream projects.

Please check this [geoserver branch](https://github.com/groldan/geoserver/commits/geoserver3/geotools-bom/) and this [geowebcache branch](https://github.com/groldan/geowebcache/commits/geoserver3/geotools-bom/) for the commits that'll use the GeoTools BOM to manage `gt-*` and their transitive dependencies. Once this PR is approved and merged, I'll be able to create pull requests for those branches.

  

# Checklist

### DIFF
--- a/docs/developer/procedures/bom.rst
+++ b/docs/developer/procedures/bom.rst
@@ -8,10 +8,12 @@ What is a BOM?
 
 A BOM (Bill of Materials) is a special POM file that manages dependency versions across a project. Rather than specifying versions for each dependency individually, you can import the BOM which provides consistent versioning.
 
-GeoTools provides two BOMs:
+GeoTools provides two BOMs that serve different purposes and are imported separately:
 
-1. ``gt-platform-dependencies`` - Manages third-party dependency versions
-2. ``gt-bom`` - Manages GeoTools module versions (also imports platform-dependencies)
+1. ``gt-platform-dependencies`` - Manages third-party dependency versions only
+2. ``gt-bom`` - Manages GeoTools module versions only
+
+**Important**: These BOMs are independent and should be imported separately based on your project's needs.
 
 BOM Structure
 -------------
@@ -28,9 +30,9 @@ The GeoTools BOM system is organized into two main components:
 2. **GeoTools Modules BOM** (``gt-bom``)
 
    * Located at: ``/bom/pom.xml``
-   * Imports the platform dependencies BOM
-   * Manages versions for all GeoTools modules
+   * Manages versions for all GeoTools modules only
    * Covers GeoTools library modules, plugin modules, extensions modules, and test artifacts
+   * Does not include third-party dependencies (these are managed separately by ``gt-platform-dependencies``)
 
 BOM Usage in GeoTools Development
 ---------------------------------
@@ -63,7 +65,9 @@ The GeoTools internal development model follows the "eat your own dog food" prin
 BOM Usage for Downstream Projects
 ---------------------------------
 
-Applications using GeoTools should import the ``gt-bom`` in their dependency management. This ensures consistent versioning of all GeoTools modules and their transitive dependencies.
+**Simple Applications using GeoTools**
+
+Applications using GeoTools should import the ``gt-bom`` in their dependency management. This ensures consistent versioning of all GeoTools modules.
 
 To use the GeoTools BOM in your project:
 
@@ -72,6 +76,34 @@ To use the GeoTools BOM in your project:
     <dependencyManagement>
       <dependencies>
         <!-- Import GeoTools BOM -->
+        <dependency>
+          <groupId>org.geotools</groupId>
+          <artifactId>gt-bom</artifactId>
+          <version>34.0</version>
+          <type>pom</type>
+          <scope>import</scope>
+        </dependency>
+      </dependencies>
+    </dependencyManagement>
+
+**Projects Heavily Integrated with GeoTools**
+
+Projects like GeoServer and GeoWebCache that are heavily integrated with GeoTools should import both BOMs to centralize management of shared third-party dependencies:
+
+.. code-block:: xml
+
+    <dependencyManagement>
+      <dependencies>
+        <!-- Import platform dependencies for third-party libraries -->
+        <dependency>
+          <groupId>org.geotools</groupId>
+          <artifactId>gt-platform-dependencies</artifactId>
+          <version>34.0</version>
+          <type>pom</type>
+          <scope>import</scope>
+        </dependency>
+
+        <!-- Import GeoTools modules BOM -->
         <dependency>
           <groupId>org.geotools</groupId>
           <artifactId>gt-bom</artifactId>
@@ -106,7 +138,7 @@ To use the GeoTools BOM in your project:
 Using Platform Dependencies Only
 --------------------------------
 
-In some cases, you might want to use only the platform dependencies BOM:
+The ``gt-platform-dependencies`` BOM can be imported separately from ``gt-bom`` when you only need third-party dependency management:
 
 .. code-block:: xml
 
@@ -122,7 +154,23 @@ In some cases, you might want to use only the platform dependencies BOM:
       </dependencies>
     </dependencyManagement>
 
-This approach is useful if you only need the third-party dependencies managed by GeoTools but not the GeoTools modules themselves.
+**When to use gt-platform-dependencies only:**
+
+* You're building a project that uses some of the same third-party libraries as GeoTools (JTS, Jackson, etc.) but don't need GeoTools modules
+* You want to ensure version compatibility with GeoTools without depending on GeoTools itself
+* You're creating a library that will be used alongside GeoTools
+
+**When to use gt-bom only:**
+
+* You're building a simple application that uses GeoTools modules (recommended for most users)
+* You're following the quickstart tutorials
+* You only need GeoTools modules and are fine with transitive dependency versions
+
+**When to use both BOMs:**
+
+* You're building a project heavily integrated with GeoTools (like GeoServer, GeoWebCache)
+* You need centralized management of both GeoTools modules and shared third-party dependencies
+* You want explicit control over third-party library versions used by both your project and GeoTools
 
 Test Artifacts
 --------------
@@ -155,7 +203,7 @@ If you need to override a specific dependency version managed by the BOM:
           <type>pom</type>
           <scope>import</scope>
         </dependency>
-        
+
         <!-- Override a specific dependency -->
         <dependency>
           <groupId>org.locationtech.jts</groupId>
@@ -165,10 +213,12 @@ If you need to override a specific dependency version managed by the BOM:
       </dependencies>
     </dependencyManagement>
 
-Multiple BOMs
--------------
+Combining Multiple BOMs
+-----------------------
 
-When using Spring Framework or other frameworks that provide their own BOMs, you can combine them with GeoTools BOM:
+You can combine GeoTools BOMs with other framework BOMs. Here are common patterns:
+
+**Simple Application with Spring Framework:**
 
 .. code-block:: xml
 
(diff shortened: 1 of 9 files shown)


bom/pom.xml
```diff
@@ -0,0 +1,801 @@
...
+  <name>GeoTools :: Bill of Materials</name>
+  <description>Provides managed dependency versions for GeoTools modules and their consistent transitive dependencies</description>
+
+  <dependencyManagement>
+    <dependencies>
+      <!-- Import platform dependencies -->
+      <dependency>
+        <groupId>org.geotools</groupId>
+        <artifactId>gt-platform-dependencies</artifactId>
+        <version>${project.version}</version>
```
mprins, line 22: this may need to be fully resolved if I recall correctly to an actual version string, eg. during the release procedure otherwise it may use the `project.version` of the actual project importing the bom

BOM-USAGE.md
```diff
@@ -0,0 +1,187 @@
...
+      <type>pom</type>
+      <scope>import</scope>
+    </dependency>
+  </dependencies>
+</dependencyManagement>
+```
+
+## Version Overrides
+
+If you need to override a specific dependency version:
```
mprins, line 135: I think in this case it would suffice to only add a property overriding the jts version:
```xml
<properties>
  <jts.version>1.19.0</jts.version>

...
```
Spring Boot works that way

mprins (commented):
Great!

It's been some time since I've set up this kind of thing, but if I recall correctly I used to run into problems using `${project.version}` all over, though that may also have been when using a shared parent pom instead of a bom

BOM-USAGE.md
groldan: When using the bom as an import in the `dependencyManagement` section (the recommended way of using the BOM), that's not enough. It's only true if you're using it as the parent pom if your downstream project root pom, which I don't recommend as mentioned above. Doing so ties you too closely to the upstream project's conventions, which might be ok for projects of the same family like the spring-boot projects themselves, but not for downstream projects.

Try it on this branch: https://github.com/groldan/geoserver/tree/geoserver3/geotools-bom (can't create a PR yet until the this one is merged)

I'm not using `jts.version` because geoserver also defines it, but let's try with another one (with and without `-Djavax.measure.version=3.0`):

```shell
cd geoserver/src
mvn -o dependency:list|grep javax.measure|sort -u
[INFO]    javax.measure:unit-api:jar:2.2:compile -- module java.measure
[INFO]    javax.measure:unit-api:jar:2.2:test -- module java.measure

mvn -o dependency:list -Djavax.measure.version=3.0 |grep javax.measure|sort -u
[INFO]    javax.measure:unit-api:jar:2.2:compile -- module java.measure
[INFO]    javax.measure:unit-api:jar:2.2:test -- module java.measure
```

bom/pom.xml
```diff
@@ -0,0 +1,837 @@
...
+  <packaging>pom</packaging>
+  <name>GeoTools :: Bill of Materials</name>
+  <description>Provides managed dependency versions for GeoTools modules and their consistent transitive dependencies</description>
+
+  <dependencyManagement>
+    <dependencies>
+      <!-- Import platform dependencies -->
+      <dependency>
+        <groupId>org.geotools</groupId>
+        <artifactId>gt-platform-dependencies</artifactId>
```
elahrvivaz: would people be open to not including the 3rd party dependencies in this bom? I love the idea of a bom for all the gt-* artifacts, but would like to be able to manage 3rd party versions independently. for those wanting both, it would then require importing `gt-platform-dependencies` separately.

mprins: I don't think it would make a difference, the bom would then need to be imported in one of the other modules and you still need to add your overrides for any transitive dependencies

.github/workflows/elasticsearch_online.yml
```diff
@@ -26,7 +26,7 @@ jobs:
         restore-keys: |
           gt-maven-
     - name: Build GeoTools dependent modules (no tests, prepare fresh artifacts)
-      run: mvn -B clean install -T1C -Dall --file pom.xml -DskipTests -pl modules/unsupported/elasticsearch -am -Dspotless.apply.skip=true
+      run: mvn -B clean install -T1C -Dall --file pom.xml -DskipTests -pl platform-dependencies,bom,modules/unsupported/elasticsearch -am -Dspotless.apply.skip=true
```
aaime: This is going to be a bit of a pain, one has to remember adding the two profiles when the bom is modified and one wants to build a specific module, right?

Is this documented anywhere? Haven't seen it, but might have just overlooked it.

BOM-USAGE.md
```diff
@@ -0,0 +1,187 @@
...
+### For GeoTools Module Development
+
+When developing GeoTools modules, you don't need to specify versions for internal dependencies:
+
+```xml
+<dependencies>
+  <!-- No version needed for GeoTools modules -->
+  <dependency>
+    <groupId>org.geotools</groupId>
+    <artifactId>gt-main</artifactId>
```
aaime, line 25: Thinking out loud, is there any tool that would enforce not having version numbers in the pom files, to be used in a QA Build? (old habits die hard, and review might not notice if the PR is large)

aaime (approved):
Looks good, have a couple of questions, see below

bom/pom.xml
elahrvivaz: I think the difference is that with the current way you get *all* the transitive dependencies used in any part of geotools, vs if it was a separate import you'd only get transitive dependencies used by the geotools modules you're importing, and maven would use it's normal dependency resolution logic to decide on a version if there are conflicts with other non-geotools dependencies.

groldan:
rebased and squash-merged, there where some dependency updates in main.
Looking at Andrea's concern about enforcing managed versions

bom/pom.xml
mprins: A BOM does not yield transitive dependencies (or any at all for that matter)  as it is part of `dependencyManagment` and not `dependencies`.

The BOM sole purpose is to provide/pin the version and scope.
Detailed in:
- https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Bill_of_Materials_.28BOM.29_POMs
- https://www.baeldung.com/maven-dependencymanagement-vs-dependencies-tags

You can use the `mvn dependency:tree` command to list your projects dependencies, it will show that you only get the transitive dependecies of any module that you have declared in the `dependencies` section of your pom

.github/workflows/elasticsearch_online.yml
aaime: Thinking out loud here, adding a dummy dependency on the bom in modules/pom.xml seems to do the trick, forces the BOM in the build without needing to explicitly adding it when targeting a set of modules with ``-pl``:

```
+    <dependency>
+        <groupId>org.geotools</groupId>
+        <artifactId>gt-bom</artifactId>
+        <version>${project.version}</version>
+        <type>pom</type>
+       <scope>compile</scope>
+       <optional>true</optional>
+      </dependency>
```

On the other end, I know this is not idiomatic. Thoughts?

bom/pom.xml
elahrvivaz: The situation I'm thinking of is as follows:

* Suppose module `gt-foo` depends on 3rd party library `foo`
* I import the `gt-bom`, and get all the 3rd party dependency versions pinned
* I use module `gt-bar` (but not `gt-foo`)
* I also use a different library (`foo-plus`) that has `foo` as a transitive dependency
* Now my `foo` is the version used by `gt-foo`, not the version used by `foo-plus`, even though I'm not actually using `gt-foo`

I haven't tested this extensively so maybe I'm misunderstanding something, but I believe that's the way it would work.

groldan: @elahrvivaz you're right.

This is the approach I'm working on right now:

* The `gt-platform-dependencies` BOM defines all third-party dependency versions used by GeoTools, except those that are only used in unsupported modules.
* The `gt-bom` BOM defines dependencies for all GeoTools artifacts.
* `gt-bom` **does not** import `gt-platform-dependencies`. In order to "eat our own dogfood" and force us to maintain the BOMs sane, `modules/pom.xml` imports both BOMs.
* Unsupported modules define their own dependencies whenever they're not already in `gt-platform-dependencies`. When/if they're promoted to supported (extension or plugin) THEN their third-party dependencies get moved to `gt-platform-dependencies`.
* If you're an occasional GeoTools user, you'll import `gt-bom` only, and then you'll get the transitive dependencies from the `gt-*` artifacts you actually use, and it's up to you to ensure compatibility across GT transitive dependencies and your own's.
* In projects that heavily rely on GeoTools and for what we want to maximize compatibility and minimize the overhead of dependency management (read, GeoServer and GeoWebCache), we'll import both `gt-platform-dependencies`  and `gt-bom`.

modules/pom.xml
```diff
@@ -61,11 +61,33 @@
   </dependencyManagement>
 
   <dependencies>
+    <!--
+     Adding gt-platform-dependencies and gt-bom as dependencies with type=pom and scope=provided here
+     does not change the dependency tree for the submodules at all, but forces the BOMs to be installed
+     in the local maven repository first, either when building the whole project or when a submodule is
+     built independently (e.g. mvn install -pl :gt-main -am), hence ensuring any change to the dependencies
+     is picked up.
+    -->
```
aaime, line 78: :+1:

bom/pom.xml
groldan: @elahrvivaz @mprins this approach is working fine now!

.github/workflows/elasticsearch_online.yml
groldan: yes it worked fine, thanks.

BOM-USAGE.md
groldan: there's not. Yet.

groldan: Ok, now there is. I've added an enforcer rule as a new module in `build/maven/enforce-managed-dependencies/` (and boy it wasn't easy even with AI help) 

Please check 25d6df89620d845087969310e9f9b7735bc5a93c

All dependencies should be "managed" now, but if you want to check add some inline dependency version to a non-test dependency and try this:

```shell
mvn install -Dqa -DskipTests
...
[INFO] BUILD FAILURE
[INFO] Total time:  0.974 s
[INFO] Finished at: 2025-06-08T12:55:36-03:00
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-enforcer-plugin:3.4.1:enforce (enforce-managed-versions) on project gt-main:
[ERROR] Rule 0: org.geotools.maven.enforcer.EnforceManagedDependencies failed with message:
[ERROR] All dependency versions must be managed in gt-platform-dependencies or gt-bom. Unsupported modules may define their own dependencyManagement sections
[ERROR]
[ERROR] EnforceManagedDependencies rule violated in project org.geotools:gt-main:
[ERROR]   - Dependency org.geotools:gt-referencing has inline version ${project.version} but is managed with version 34-SNAPSHOT
[ERROR]   - Dependency it.geosolutions.jaiext.utilities:jt-utilities has inline version 1.1.30 but is managed with version 1.1.31
[ERROR]
[ERROR] Ensure all dependencies are managed through dependencyManagement sections in this project or inherited from parent projects.
```

Tha'll build the rule. If it's already built (e.g. first run `mvn install -f build/maven/enforce-managed-dependencies`), you can also use the faster `validate` goal. E.g.:

```shell
mvn validate -Dqa
mvn validate -Dqa -pl :gt-main -am
mvn validate -Dqa -Dall -f modules/unsupported
```
etc.

With this I think we've covered all the concerns, but I won't merge it until I get approval for this new maven artifact.

----
Aside: I had to move the dependencies from `build/maven/pom.xml` to `build/maven/jjtree-javacc/` and `build/maven/xmlcodegen/` since they use the maven 2.0 API. I'd like to upgrade them to 3.8.1 (same we require as minimum in the root pom), but that'd be a separate issue.

aaime: Whoa, you had to make a plugin for it ? I was just wondering if it was possible, did not mean to ask for such an amount of development :rofl: 
But thanks!

bom/pom.xml
mprins: LGTM

mprins (approved):

bom/pom.xml
elahrvivaz: thank you, sounds great!

BOM-USAGE.md
mprins: > Aside: I had to move the dependencies from build/maven/pom.xml to build/maven/jjtree-javacc/ and build/maven/xmlcodegen/ since they use the maven 2.0 API. I'd like to upgrade them to 3.8.1 (same we require as minimum in the root pom), but that'd be a separate issue.

@groldan I had a crack at upgrading those Maven plugins in https://github.com/geotools/geotools/pull/3402; but it stranded and I had to abandon

groldan: Good to know! thanks

groldan:
Docs updated, ready to merge, waiting for the build

PR #5204, opened 2025-05-27
[GEOT-7764] Migrate GeoTools to Java 17

groldan:
[GEOT-7764](https://osgeo-org.atlassian.net/browse/GEOT-7764) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

* Upgrade build for Java 17 as base Java version
  * Configure maven-enforcer-plugin to require Java 17
  * Configure maven-enforcer-plugin to require maven 3.8+ as done in GeoServer.
  * Remove cobertura-maven-plugin (modules/extension/xsd/**), as it is not compatible with Java 11+
  * Update maven-javadoc-plugin configuration to use legacyMode=true and not fail due to module-info.java not present. 
  * Upgrade github actions workflows to use Java 17, remove the linux_jdk11.yml workflow.
* Centralize plugin versions in the root pom's pluginManagement section
* Update docs for Java 17
* Add .mvn/jvm.config with necessary --add-exports and --add-opens for internal jdk.compiler packages.
  This ensures that build-time tools and compiler plugins (e.g., Error Prone, and others interacting with javac internals) can
  function correctly under Java 17's module system.

* Plugin version upgrades:
  * jacoco 0.8.7 -> 0.8.13
  * maven-checkstyle-plugin 3.3.0 -> 3.6.0
  * maven-compiler-plugin 3.11.0 -> 3.14.0
  * maven-javadoc-plugin 3.6.3 -> 3.11.2
* Upgrade errorprone to work with Java 17:
  * add suggested compiler args `-Xep:DeadException:WARN -Xep:GuardedBy:OFF`
  * Add additional errorprone exclusions and fix trivial errors
  * Use explict charset reading streams (IntelliJ)
  * Inner class can be static, by IntelliJ
  * Removing unnecessary parenthesis (intellij)

* Fix usage of varargs in GridCoverageReader/Writer

  GridCoverageReader.read() and GridCoverageWriter.write() were declared
  to receive `GeneralParameterValue... parameters` varargs but most
  classes in their hiearchy were still using `GeneralParameterValue[]`,
  which caused errors when compiling with `-Werror` and Java 17.

  Update to replace calls with null argument like `reader.read(null)` by
  `reader.read()` and avoid the "inon-varargs call of varargs method with
  inexact argument type for last parameter" errors.

  Update ImageMosaicReader, WorldImageWriter, and GrassCoverageWriter, to
  be aware that a client call without values in the varargs parameter
  means an empty array. All other implementations didn't need it because
  the idiom `if(parameters != null){for (GeneralParameterValue param :
  parameters)...` doesn't need updating.

* Fix new -Werror compilation errors
  * Remove `strictfp` keywords: as of release 17, all floating-point expressions are evaluated strictly and 'strictfp' is not required
  * java.security.AccessController in java.security has been deprecated and marked for removal
  * non-varargs call of varargs method with inexact argument type for last parameter
  * deprecated item is not annotated with @Deprecated
  * deprecated item is not annotated with @Deprecated
  * unreachable catch clause

* Update the build/maven/jjtree-javacc plugin to add @SuppressWarnings in generated java files
  Upgrade net.java.dev.javacc:javacc to 7.0.13 and update JJTreeJavaCC.java to add @SuppressWarnings("unchecked")
  to  generated parsers so they work with javac -Werror

* Relax tolerance for GridCoverageRendererTest.testRenderRadarOnAzeq() from 20 to 50 pixels

  Update GridCoverageRendererTest.testRenderRadarOnAzeq() with a 50 pixel
  tolerance when comparing the expected and the rendered images, due to
  differences between rendering on Linux and MacOs with Java 17.
---
This is a coordinated effort across GeoTools, [GeoServer](https://github.com/geoserver/geoserver/pull/8573), and [GeoWebCache](https://github.com/GeoWebCache/geowebcache/pull/1406)

  

# Checklist


aaime:
Lots of changes indeed... wondering if you want to keep error-prone in its default configuration or just tone down some of the errors that weren't caught before.

aaime:
Had a look out of curiosity, and believe part of your predicament comes from different defaults in java linting.
When enabling -Werror all warnings become errors.

Now, I was seeing this compile issue:

```
[WARNING] COMPILATION WARNING : 
[INFO] -------------------------------------------------------------
[WARNING] /home/aaime/devel/git-gt/modules/plugin/iau-wkt/src/main/java/org/geotools/referencing/iau/IAULongitudeFirstFactory.java:[147,78] non-varargs call of varargs method with inexact argument type for last parameter;
  cast to org.geotools.api.referencing.cs.AxisDirection for a varargs call
  cast to org.geotools.api.referencing.cs.AxisDirection[] for a non-varargs call and to suppress this warning
[INFO] 1 warning
[INFO] -------------------------------------------------------------
[INFO] -------------------------------------------------------------
[ERROR] COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
[ERROR] /home/aaime/devel/git-gt/modules/plugin/iau-wkt/src/main/java/org/geotools/referencing/iau/IAULongitudeFirstFactory.java: warnings found and -Werror specified
[INFO] 1 error
```

and it was not making sense to me, this is a javac lint error, not an error-prone one.
Checking, it seems that varargs issues are emitted anyways by java 17 and cannon the controlled via -Xlint (I've tried disabling it with -Xlint:-varargs, no luck). What a pain?

groldan:
@aaime yes, it's a catch-22 because currently:

```
GridCoverageReader{
    GridCoverage read(GeneralParameterValue... parameters);
}
GridCoverage2DReader extends GridCoverageReader{
    @Override
    GridCoverage2D read(GeneralParameterValue[] parameters);
}

AbstractGridCoverage2DReader implements GridCoverage2DReader{
 @Override
 public abstract GridCoverage2D read(GeneralParameterValue[] parameters){}
}
```
And this PR fixes the varargs, but the test in GeoServer is mocking `AbstractGridCoverage2DReader`, so it has to be fixed like this:

```diff
--- a/src/main/src/test/java/org/geoserver/catalog/impl/CatalogBuilderTest.java
+++ b/src/main/src/test/java/org/geoserver/catalog/impl/CatalogBuilderTest.java
@@ -780,7 +780,10 @@ public void testSetupCoverageOnEmptyRead() throws Exception {
         expect(reader.getOriginalGridToWorld(EasyMock.anyObject(PixelInCell.class)))
                 .andReturn(new AffineTransform2D(gridToWorld))
                 .anyTimes();
-        expect(reader.read(EasyMock.anyObject(GeneralParameterValue[].class))).andReturn(null);
+        expect(reader.read(
+                        EasyMock.anyObject(GeneralParameterValue.class),
+                        EasyMock.anyObject(GeneralParameterValue.class)))
+                .andReturn(null);
```

groldan:
@aaime @jodygarnett the only remaining thing here is a problem with `GridCoverageRendererTest.testRenderRadarOnAzeq()`, which yields a 43 pixels difference in macos  with Java 17 .

The threshold used in the call to `ImageAssert.assertEquals` is 20 pixels since ever, but if I replace the sample image (see 7e6e320265bdd674c8303df334a48a015effb864) then it'll fail on linux

I'm looking for advise on how to solve this

```
java -version
openjdk version "17.0.13" 2024-10-15
OpenJDK Runtime Environment Temurin-17.0.13+11 (build 17.0.13+11)
OpenJDK 64-Bit Server VM Temurin-17.0.13+11 (build 17.0.13+11, mixed mode, sharing)
```

aaime:
The sample image looks as follows:

pacificRadar

It's 401x257, or in other words, a bit above 100k pixels. I don't see a problem relaxing the tolerance.

docs/common.py
```diff
@@ -74,7 +74,7 @@
   version = re.sub('-.*', '', version)
 
 # The branch, 8, 9, 10, etc...
-branch =  re.sub('\..*', '', re.sub('-.*', '', version))
+branch =  re.sub('\\..*', '', re.sub('-.*', '', version))
```
aaime: Are you sure this is necessary in Python? It's not java...

docs/user/build/install/jdk.rst
```diff
@@ -15,80 +15,87 @@ Java Run-time Environment:
...
+Oracle OpenJDK            GPL                 x     x              x       2029-09
+RedHat OpenJDK            GPL                 x                            2027-11
+Eclipse Adoptium          GPL                 x     x              x       2029+**
 ========================= =================== ===== ====== ======= ======= ==============
 
-.. warning:: Since the API changes from Java version to version, building a GeoTools version with a newer Java SDK is risky (you may accidentally use a new method). Pull requests are tested against Java 11 and Java 17, but we do ask you to be careful.
+.. note:: 
+   \* Oracle JDK 17 free updates ended October 2024; continued updates require paid Oracle Java SE subscription.
+   
+   \*\* Eclipse Adoptium provides free updates as long as upstream OpenJDK 17 is actively maintained (expected through 2029).
```
aaime: This page says 2027 for java 17, and 2029 for java 21

docs/user/build/install/jdk.rst
```diff
@@ -15,80 +15,87 @@ Java Run-time Environment:
...
+      sdk use java 21<tab>
 
 Windows or macOS Install
-''''''''''''''''''''''''
+^^^^^^^^^^^^^^^^^^^^^^^^
 
 1. Download an OpenJDK release for your platform:
 
-   * https://adoptium.net/temurin/releases/?version=11 Temurin 11 (LTS) - Recommended
+   * https://adoptium.net/temurin/releases/?version=17 Temurin 17 (LTS) - Recommended
```
aaime, line 98: Believe you'll need this URL:

https://adoptium.net/temurin/releases/?version=17&os=any&arch=any

modules/library/api/src/main/java/org/geotools/api/filter/sort/SortOrder.java
```diff
@@ -50,6 +50,15 @@ public final class SortOrder extends CodeList<SortOrder> {
     /** The SQL keyword for this sorting order. */
     private final String sqlKeyword;
 
+    /**
```
aaime: The constructor was moved it seems... but why?

modules/library/api/src/main/java/org/geotools/api/util/CodeList.java
```diff
@@ -64,7 +64,7 @@ public abstract class CodeList<E extends CodeList<E>> implements Comparable<E>,
      */
     @SuppressWarnings("unchecked")
     protected CodeList(String name, final Collection<E> values) {
-        this.name = (name = name.trim());
```
aaime: Why was the trim removed?

modules/library/api/src/test/java/org/geotools/api/filter/identity/VersionTest.java
```diff
@@ -61,7 +61,7 @@ public void versionDate() {
         Version version = new Version(now);
 
         assertTrue(version.isDateTime());
-        assertEquals(now, version.getDateTime());
+        assertEquals(now.getTime(), version.getDateTime().getTime());
```
aaime, line 64: Why compare the derivative getTime rather than the Date itself?

modules/library/sample-data/src/main/java/org/geotools/test/TestData.java
```diff
@@ -107,22 +107,23 @@ public class TestData implements Runnable {
      * The files to delete at shutdown time. {@link File#deleteOnExit} alone doesn't seem suffisient since it will
      * preserve any overwritten files.
      */
-    private static final LinkedList<Deletable> toDelete = new LinkedList<>();
+    private static final Deque<Deletable> toDelete = new ArrayDeque<>();
```
aaime, line 110: Interesting change, do we have a concurrency issue in tests using this class?

modules/plugin/grassraster/src/test/java/org/geotools/gce/grassraster/JGrassRegionTest.java
```diff
@@ -74,7 +74,7 @@ public void testJGrassMapEnvironment() throws IOException {
         cellhdFile = URLs.urlToFile(theUrl);
         tmpRegion = new JGrassRegion(cellhdFile.getAbsolutePath());
         Assert.assertTrue(tmpRegion.getNorth() - 0.0004 < EPSI);
-        Assert.assertTrue(tmpRegion.getSouth() - (-5.0004) < EPSI);
+        Assert.assertTrue(tmpRegion.getSouth() - -5.0004 < EPSI);
```
aaime, line 77: I actually find this one change harder to read than the original (I do get it has been done with mass change tools)

modules/library/sample-data/src/main/java/org/geotools/test/TestData.java
groldan: no, it's this bummer https://errorprone.info/bugpattern/JdkObsolete

modules/library/api/src/test/java/org/geotools/api/filter/identity/VersionTest.java
groldan: because errorProne complains https://errorprone.info/bugpattern/UndefinedEquals

docs/user/build/install/jdk.rst
groldan: leads to the same contents as far as I can see?

docs/user/build/install/jdk.rst
groldan: Thanks, changed to

```diff
-   \*\* Eclipse Adoptium provides free updates as long as upstream OpenJDK 17 is actively maintained (expected through 2029).
+   \*\* Eclipse Adoptium provides free updates as long as upstream OpenJDK is actively maintained (expected through at least Oct 2027 for Java 17 and Dec 2029 for Java 21).
```

modules/library/api/src/main/java/org/geotools/api/filter/sort/SortOrder.java
groldan: I...don't...know... reverted it, thanks

modules/library/api/src/main/java/org/geotools/api/util/CodeList.java
groldan: I know I did because some tooling complains (it does ring a bell from when I was being stricter with the checks for the core modules), but I can't remember exactly why, nor I could force it to fail after removing several errorProne exclusions... So I reverted it, now that class diff is just

```diff
diff --git a/modules/library/api/src/main/java/org/geotools/api/util/CodeList.java b/modules/library/api/src/main/java/org/geotools/api/util/CodeList.java
index 5e9094bbf3..04d48452a1 100644
--- a/modules/library/api/src/main/java/org/geotools/api/util/CodeList.java
+++ b/modules/library/api/src/main/java/org/geotools/api/util/CodeList.java
@@ -204,7 +204,7 @@ public boolean matches(String name) {
             return true;
         }
         final String identifier = identifier();
-        return (identifier != null) && name.equalsIgnoreCase(identifier);
+        return identifier != null && name.equalsIgnoreCase(identifier);
     }
```

docs/common.py
groldan: yet another good catch. Actually unrelated to the Java 17 build, we're getting `SyntaxWarning: invalid escape sequence '\.'`, but that doesn't fix the missing replacement of the `|branch|` variable in the `docs/user/tutorial/quickstart/try.txt` included file.

For example: https://docs.geotools.org/latest/userguide/tutorial/quickstart/intellij.html, shows:

image

---
since it's unrelated I've reverted it and will try to provide a fix as a separate pr

groldan: @aaime here you go: 
- https://github.com/geotools/geotools/pull/5280

modules/library/sample-data/src/main/java/org/geotools/test/TestData.java
aaime: ah ha

modules/library/api/src/test/java/org/geotools/api/filter/identity/VersionTest.java
aaime: I see

docs/user/build/install/jdk.rst
aaime: Doh indeed

PR #5230, opened 2025-06-07
Call MapContent dispose() to prevent memory leaks

BarneyWong:
zGroupedMapContent doesn't call dispose() method, when call  MapContent finalize() notice  'Call MapContent dispose() to prevent memory leaks'

  

# Checklist


modules/library/render/src/main/java/org/geotools/renderer/lite/StreamingRenderer.java
```diff
@@ -893,6 +894,7 @@ public void paint(
                         }
                     }
                 }
+               Optional.ofNullable(zGroupedMapContent).ifPresent(MapContent::dispose);
```
aaime: Optional is currently not imported by StreamingRenderer... this code change did not even go through a compiler?

Also, ZGroupLayerFactory can return the original map content if there are no z groups, do you want to dispose that one too?

Finally, what kind of memory leak do you expect from the newly built map content? It's freshly built, I don't see usage of listeners and the layers do not link back to the map content

PR #5253, opened 2025-06-23
[GEOT-7781] Add a priority to property accessors

Mike7311:
[GEOT-7781](https://osgeo-org.atlassian.net/browse/GEOT-7781) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

Currently multiple property accessor factories can be used by the GeoServer, however this can cause issues because there are no clear priority, and order of them can vary. This changes add support for a priority.

# Checklist

### DIFF
--- a/modules/extension/complex/src/main/java/org/geotools/data/complex/expression/FeaturePropertyAccessorFactory.java
+++ b/modules/extension/complex/src/main/java/org/geotools/data/complex/expression/FeaturePropertyAccessorFactory.java
@@ -287,4 +287,11 @@ public class FeaturePropertyAccessorFactory implements PropertyAccessorFactory {
             assert value == context.getValue(xpath);
         }
     }
+
+    @Override
+    public int getPriority() {
+        // this is the default and the most generic property accessor
+        // hence we give it the lowest priority
+        return LOWEST_PRIORITY;
+    }
 }
--- a/modules/library/main/src/main/java/org/geotools/filter/expression/PropertyAccessorFactory.java
+++ b/modules/library/main/src/main/java/org/geotools/filter/expression/PropertyAccessorFactory.java
@@ -26,6 +26,15 @@ import org.geotools.util.factory.Hints.Key;
  */
 public interface PropertyAccessorFactory {
 
+    /** The numeric value for highest priority. */
+    int HIGHEST_PRIORITY = 5000;
+
+    /** The numeric value for lowest priority. */
+    int LOWEST_PRIORITY = 0;
+
+    /** The numeric value for the default priority. */
+    int DEFAULT_PRIORITY = 3000;
+
     /**
      * {@link Hints} key used to pass namespace context to {@link #createPropertyAccessor(Class, String, Class, Hints)}
      * in the form of a {@link NamespaceSupport} instance with the prefix/namespaceURI mappings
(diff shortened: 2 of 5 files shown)


modules/extension/complex/src/main/java/org/geotools/data/complex/expression/FeaturePropertyAccessorFactory.java
```diff
@@ -287,4 +287,11 @@ public void set(Object object, String xpath, Object value, Class target) throws
             assert value == context.getValue(xpath);
         }
     }
+
+    @Override
+    public int getPriority() {
+        // this is the default and the most generic property accessor
+        // hence we give it the lowest priority
+        return LOWEST_PRIOIRTY;
```
mprins: all the these constant names are spelled wrong, eg. `LOWEST_PRIOIRTY` should be `LOWEST_PRIORITY`

modules/library/main/src/main/java/org/geotools/filter/expression/PropertyAccessorFactory.java
```diff
@@ -26,6 +26,15 @@
  */
 public interface PropertyAccessorFactory {
 
+    /** The numeric value for highest priority. */
+    int HIGHEST_PRIOIRTY = 0;
+
+    /** The numeric value for lowest priority. */
+    int LOWEST_PRIOIRTY = 5000;
```
mprins: it seems counter intuitive that the lowest priority has the highest value, you could change that and sort the other way around

modules/library/main/src/main/java/org/geotools/filter/expression/PropertyAccessors.java
```diff
@@ -45,6 +53,18 @@ public class PropertyAccessors {
                 cache.add(factory);
             }
         }
+        // sort the property accessors by their priority
+        cache.sort(Comparator.comparing(PropertyAccessorFactory::getPriority));
```
mprins: you could use https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/Comparator.html#reversed() instead and have the prio valus matching their name

mprins (changes requested):

modules/extension/complex/src/main/java/org/geotools/data/complex/expression/FeaturePropertyAccessorFactory.java
Mike7311: Names corrected.

modules/library/main/src/main/java/org/geotools/filter/expression/PropertyAccessorFactory.java
Mike7311: Switched the values for HIGHEST_PRIOIRTY and LOWEST_PRIOIRTY.

modules/library/main/src/main/java/org/geotools/filter/expression/PropertyAccessors.java
Mike7311: Added reveresed() comparator.

mprins (approved):
LGTM

modules/library/main/src/main/java/org/geotools/filter/expression/PropertyAccessors.java
```diff
@@ -45,6 +53,18 @@ public class PropertyAccessors {
                 cache.add(factory);
             }
         }
+        // sort the property accessors by their priority
```
aaime: The code is initially adding 3 well known property accessors. The sort should not move them, because the javadoc says:

"Sorts this list according to the order induced by the specified Comparator. The sort is stable: this method must not reorder equal elements."

but many might not remember this by heart and think the sort might ruin the initial order. I'd add a comment clarifying that.

Mike7311: Updated the comment

PR #5259, opened 2025-06-26
[GEOT-7782] Allow usage of grouping aggregation as a rendering transformation

aaime:
[GEOT-7782](https://osgeo-org.atlassian.net/browse/GEOT-7782) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

See ticket for details.
A change in a XStream mapped bean breaks the downstream build, this PR contains the fix:
https://github.com/geoserver/geoserver/pull/8630

# Checklist

### DIFF
--- a/modules/unsupported/process-feature/src/main/java/org/geotools/process/vector/AggregateProcess.java
+++ b/modules/unsupported/process-feature/src/main/java/org/geotools/process/vector/AggregateProcess.java
@@ -523,10 +523,10 @@ public class AggregateProcess implements VectorProcess {
                     case StdDev:
                     case Sum:
                     case SumArea:
+                    case Median: // in case of even number of elements, the median is actually a double
                         tb.add(function.name(), Double.class);
                         break;
                     case Max:
-                    case Median:
                     case Min:
                         tb.add(function.name(), type);
                         break;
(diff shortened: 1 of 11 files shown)


modules/unsupported/process-feature/src/main/java/org/geotools/process/vector/AggregateProcess.java
```diff
@@ -454,5 +475,103 @@ public List<Object[]> getGroupByResult() {
...
+                        tb.add(function.name(), Long.class);
+                        break;
+                    case Average:
+                    case StdDev:
+                    case Sum:
+                    case SumArea:
+                        tb.add(function.name(), Double.class);
+                        break;
+                    case Max:
+                    case Median:
```
dromagnoli: Is the median not applying any "math" on the values (just selecting either one or the other) so that it relies on the type?
(Therefore not using Double as per other aggregators)

aaime: The median does not require math on the result, no, [see Wikipedia](https://en.wikipedia.org/wiki/Median).
As such, it's of the same data type as the input values, indeed, it's one of the input values.

aaime: Hum... wait... in case of even numbers it actually requires a calculation... duh. Switching the code.

PR #5284, opened 2025-07-14
[GEOT-7787] Boolean attributes don't get stored as a true boolean in geopackages

SanderKleykens:
[GEOT-7787](https://osgeo-org.atlassian.net/browse/GEOT-7787) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

As described in [GEOT-7787](https://osgeo-org.atlassian.net/browse/GEOT-7787), it appears that right now a Boolean is mapped to a SQL INTEGER type instead of a SQL BOOLEAN when we're dealing with geopackages.
The goal of this PR is to change this to BOOLEAN, since that makes more sense when looking at features in QGIS.

# Checklist

- [x] There is an issue in [GeoTools Jira](https://osgeo-org.atlassian.net/browse/GEOT-7787) (except for changes not visible to end users). 

### DIFF
--- a/modules/plugin/geopkg/src/main/java/org/geotools/geopkg/GeoPkgDialect.java
+++ b/modules/plugin/geopkg/src/main/java/org/geotools/geopkg/GeoPkgDialect.java
@@ -251,7 +251,6 @@ public class GeoPkgDialect extends PreparedStatementSQLDialect {
         mappings.put(Integer.class, Types.INTEGER);
         mappings.put(Float.class, Types.FLOAT);
         mappings.put(Double.class, Types.DOUBLE);
-        mappings.put(Boolean.class, Types.INTEGER);
         // SQLite has no native support for UUID datatype.
         // Map Java UUID to varchar instead
         mappings.put(UUID.class, Types.VARCHAR);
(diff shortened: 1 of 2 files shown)


mprins:
Storing the boolean value as a SQLite INTEGER seems to be a requirement, see https://www.geopackage.org/spec140/index.html#table_column_data_types

aaime:
> I did not realise the geopackage spec specifies using the integer datatype for booleans. Thanks for the link.

Let's read it again: "A boolean value representing true or false. Stored as SQLite INTEGER with value 0 for false or 1 for true."

Do they mean that we need to use integer, or just remind us that sqlite will store booleans as integers regardless the fact we use the boolean storage class?

aaime:
I've confirmed the geopackage passes the CITE tests and I believe the specification is not 100% clear in that regard. 
In addition to @mprins example, there's also Table 7, which also reports a boolean_attribute with type "BOOLEAN".

Overall, the change does not seem to bring harm, so I'm merging it, let's see if that helps us collect more feedback.

PR #5285, opened 2025-07-14
[GEOT-7788] Function to parse GeoTools time notation to date

Mike7311:
[GEOT-7788](https://osgeo-org.atlassian.net/browse/GEOT-7788) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

Add a function that will allow to use GT time notations (for example 'P2M' for past 2 months) in filters.
For example time_attr >= parseTime('P2M')

# Checklist

### DIFF
--- a/docs/user/artifacts/function_list
+++ b/docs/user/artifacts/function_list
@@ -1,7 +1,7 @@
 ``DefaultFunctionFactory``
 ^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Contains 203 functions.
+Contains 204 functions.
 
 ``Area(geometry)``: returns ``area``
 ''''''''''''''''''''''''''''''''''''
@@ -2961,3 +2961,8 @@ Contains 1 functions.
 * ``tolerance`` (``double``) required
 
 * ``labelPoint`` (``Point``)
+
+``parseTime(dateString)``: returns ``date``
+'''''''''''''''''''''''''''''''''''''''''''''''''''
+
+* ``dateString`` (``String``) required. Date in ISO format, or the range in ISO duration format. Only single tokens are allowed ("P1M" or "2008-05-11T15:30:00Z" and not "P1M/P1D" or "2008-05-11T15:30:00Z/2008-05-11T18:30:00Z"). If range or list is provided only first value shall be used.
--- a/modules/library/main/src/main/java/org/geotools/filter/function/ParseTimeFunction.java
+++ b/modules/library/main/src/main/java/org/geotools/filter/function/ParseTimeFunction.java
@@ -1,0 +1,71 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ *
+ *    (C) 2025, Open Source Geospatial Foundation (OSGeo)
+ *
+ *    This library is free software; you can redistribute it and/or
+ *    modify it under the terms of the GNU Lesser General Public
+ *    License as published by the Free Software Foundation;
+ *    version 2.1 of the License.
+ *
+ *    This library is distributed in the hope that it will be useful,
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ *    Lesser General Public License for more details.
+ */
+package org.geotools.filter.function;
+
+import static org.geotools.filter.capability.FunctionNameImpl.parameter;
+import static org.geotools.util.DateTimeParser.FLAG_GET_TIME_ON_PRESENT;
+import static org.geotools.util.DateTimeParser.FLAG_IS_LENIENT;
+import static org.geotools.util.DateTimeParser.FLAG_SINGLE_DATE_AS_DATERANGE;
+
+import java.text.ParseException;
+import java.util.Collection;
+import java.util.Date;
+import org.geotools.api.filter.capability.FunctionName;
+import org.geotools.filter.FunctionExpressionImpl;
+import org.geotools.filter.LiteralExpressionImpl;
+import org.geotools.filter.capability.FunctionNameImpl;
+import org.geotools.util.DateRange;
+import org.geotools.util.DateTimeParser;
+
+/** Function parses GeoTools time notation into date. */
+public class ParseTimeFunction extends FunctionExpressionImpl {
+
+    public static FunctionName NAME = new FunctionNameImpl("parseTime", parameter("string", String.class));
+
+    DateTimeParser parser;
+
+    public ParseTimeFunction() {
+        super(NAME);
+        parser = new DateTimeParser(1, FLAG_GET_TIME_ON_PRESENT | FLAG_IS_LENIENT | FLAG_SINGLE_DATE_AS_DATERANGE);
+    }
+
+    @Override
+    public Object evaluate(Object object) {
+        Date dateTime = null;
+        try {
+            String value = (String) ((LiteralExpressionImpl) params.get(0)).getValue();
+            if (value.startsWith("P")) {
+                value = value + "/PRESENT";
+            }
+            dateTime = getDate(value, dateTime);
+        } catch (ParseException e) {
+            throw new RuntimeException(e);
+        }
+        return dateTime;
+    }
+
+    private Date getDate(String value, Date dateTime) throws ParseException {
+        Collection parsedCollection = parser.parse(value);
+        Object parsedObject = parsedCollection.iterator().next();
+        if (parsedObject instanceof DateRange) {
+            dateTime = ((DateRange) parsedObject).getMinValue();
+        } else if (parsedObject instanceof Date) {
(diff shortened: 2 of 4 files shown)


afabiani:
> Should 33.2 incorporate this change? Ref: https://discourse.osgeo.org/t/geoserver-2-27-2-release-train/148298

Possibly yes, not mandatory

afabiani (approved):
Working as expected

```
18 Jul 12:32:20 INFO   [geoserver.wfs] - 
Request: getFeature
    service = WFS
    version = 1.0.0
    baseUrl = http://vdscs-dev-geoserver.emsa.geosolutionsgroup.com:8080/geoserver/
    query[0]:
        filter = [ VDS_POSITION_TS >= parseTime([P3M]) ]
        typeName[0] = {http://www.emsa.europa.eu/frontex}FRONTEX_VDS_POSITIONS
    outputFormat = GML2
    resultType = results
    viewParams[0] = {}
18 Jul 12:32:20 DEBUG  [geotools.jdbc] - CREATE CONNECTION
18 Jul 12:32:20 DEBUG  [geotools.jdbc] - CLOSE CONNECTION
18 Jul 12:32:20 DEBUG  [geotools.jdbc] - CREATE CONNECTION
18 Jul 12:32:20 TRACE  [data.jdbc] - exporting LogicFilter
18 Jul 12:32:20 TRACE  [data.jdbc] - exporting SQL ComparisonFilter
18 Jul 12:32:20 TRACE  [data.jdbc] - exporting PropertyName
18 Jul 12:32:20 TRACE  [data.jdbc] - exporting PropertyName
18 Jul 12:32:20 DEBUG  [geotools.jdbc] - SELECT * FROM (SELECT VDS_POSITION_ID,FID,CORRELATION_ID,CORRELATION_TS,CSN_ID,SERVICE_ID,VDS_POSITION_TS,HEADING,LONGITUDE,LATITUDE,CORRELATED,OPERATIONS,STATUS,DETECTION_TYPE,LENGTH,EOD_ATTRIBUTES,CORRELATION_RANK,CORRELATION_FINAL_RANK,CORRELATED_RANKED_1ST,HIGHEST_CORRELATION_SCORE,SPAIN_BBOX,PACKAGE_IDENTIFIER,SATELLITE_NAME,OPERATIONAL_MODE,SENSOR_SPATIAL_RESOLUTION FROM (SELECT * FROM vds_position
) VTABLE WHERE (VDS_POSITION_TS >= ? AND VDS_POSITION_TS IS NOT NULL )) WHERE ROWNUM <= 1000000
18 Jul 12:32:20 DEBUG  [geotools.jdbc] - 1 = 2025-04-19 12:32:20.0
18 Jul 12:32:20 TRACE  [geotools.jdbc] - 2025-07-18 04:22:20.805 is not of type java.sql.Timestamp, value was converted
18 Jul 12:32:20 TRACE  [geotools.jdbc] - 2025-05-18 16:14:26.769 is not of type java.sql.Timestamp, value was converted
```

modules/library/main/src/main/java/org/geotools/filter/function/ParseTimeFunction.java
```diff
@@ -0,0 +1,71 @@
...
+import org.geotools.filter.capability.FunctionNameImpl;
+import org.geotools.util.DateRange;
+import org.geotools.util.DateTimeParser;
+
+/** Function parses GeoTools time notation into date. */
+public class ParseTimeFunction extends FunctionExpressionImpl {
+
+    public static FunctionName NAME = new FunctionNameImpl("parseTime", parameter("string", String.class));
+
+    DateTimeParser PARSER;
```
aaime: Not a constant, should not be all UPPERCASE.

docs/user/artifacts/function_list
```diff
@@ -2961,3 +2961,8 @@ Contains 1 functions.
 * ``tolerance`` (``double``) required
 
 * ``labelPoint`` (``Point``)
+
+``parseTime(dateString)``: returns ``date``
```
aaime, line 2965: Mind, there is already a dateParse function that parses over a user provided format.
This one instead uses the DateTimeParser and can manage a number of different syntaxes, the documentation should provide some perspective on that, especially on the extended understanding of time periods, and what happens if one provides a range instead.

modules/library/main/src/main/java/org/geotools/filter/function/ParseTimeFunction.java
```diff
@@ -0,0 +1,71 @@
...
+        super(NAME);
+        PARSER = new DateTimeParser(1, FLAG_GET_TIME_ON_PRESENT | FLAG_IS_LENIENT | FLAG_SINGLE_DATE_AS_DATERANGE);
+    }
+
+    @Override
+    public Object evaluate(Object object) {
+        Date dateTime = null;
+        try {
+            String value = (String) ((LiteralExpressionImpl) params.get(0)).getValue();
+            if (value.startsWith("P")) {
```
aaime, line 51: I see the convenience, but what if the user already provides "P1M/PRESENT"?
Maybe "if starts with P, and does not contain /" instead?

Something like "P1M" means "one month", it's a [legal ISO duration](https://en.wikipedia.org/wiki/ISO_8601#Durations) but not attached to a specific point in time. "P1M/2008-05-11T15:30:00Z" is the month before the final date, and so on. 
"P1M/PRESENT" is parsable in GT, but "PRESENT" has been added as a convenience; it's not part of the ISO spec.

Long story short, this convenience mode should be at least documented, because "P1M" alone is just a duration without any time reference (so using it to intend a relative time in the past, is a deviation from the spec).

modules/library/main/src/test/java/org/geotools/filter/function/ParseTimeFunctionTest.java
```diff
@@ -0,0 +1,90 @@
...
+import org.geotools.api.filter.expression.Function;
+import org.geotools.factory.CommonFactoryFinder;
+import org.junit.Test;
+
+public class ParseTimeFunctionTest {
+
+    public static final FilterFactory FF = CommonFactoryFinder.getFilterFactory();
+
+    @Test
+    public void testOrdinaryGTNotations() {
```
aaime: Mind this is not a GT notation, it's the [ISO duration notation](https://en.wikipedia.org/wiki/ISO_8601#Durations) (also consider it when documenting).
See also https://docs.geoserver.org/stable/en/user/services/wms/time.html#specifying-a-relative-interval

docs/user/artifacts/function_list
Mike7311: updated

modules/library/main/src/main/java/org/geotools/filter/function/ParseTimeFunction.java
Mike7311: Updated doc to mention that only single token is being used, and full range should not be used.
Allowing user to provide full range (e.g. P1M/PRESENT) can give a false impression that the second token affects processing somehow.

modules/library/main/src/test/java/org/geotools/filter/function/ParseTimeFunctionTest.java
Mike7311: Renamed test

modules/library/main/src/main/java/org/geotools/filter/function/ParseTimeFunction.java
Mike7311: updated

modules/library/main/src/main/java/org/geotools/filter/function/ParseTimeFunction.java
aaime: Yes, but... you did not say anything about a case like "2008-05-11T15:30:00Z/2008-05-11T18:30:00Z", which would be parsed in a legit way and just the first item be used.  Also, DateTimeParser can parse a comma- separated list of values, in which case it's going to return a List. If the user tries that, the code as written now will silently return null (should throw an exception, if one is using a syntax that is not mean to be supported).

Mike7311: > Yes, but... you did not say anything about a case like "2008-05-11T15:30:00Z/2008-05-11T18:30:00Z", which would be parsed in a legit way and just the first item be used. 

Will update doc to mention that.

> Also, DateTimeParser can parse a comma- separated list of values, in which case it's going to return a List

Can you provide an example of such list? I have checked the doc on parse method that mentions:
The string may contains either a single date, or a start time, end time and a period.

Also checked the DateTimeParserTest, but did not notice comma separated lists. Only ones separated by slashes (2011-10-10T10:11:12.000Z/2011-10-10T14:11:12.000Z/PT1H). This I believe can be covered by doc (saying that only single tokens are allowed).

Mike7311: Update the doc to mention that "If range or list is provided only first value shall be used."

PR #5286, opened 2025-07-17
Apply the "Upgrade to java 17" OpenRewrite recipe

groldan:
# Checklist


build/maven/xmlcodegen/src/main/java/org/geotools/maven/xmlcodegen/templates/ConfigurationTemplate.java
```diff
@@ -77,12 +77,12 @@ public String generate(Object argument)
...
       for (Object o : components) {
           XSDNamedComponent component = (XSDNamedComponent) o;
-          if (component instanceof XSDTypeDefinition) {
-              types.add((XSDTypeDefinition) component);
-          } else if (component instanceof XSDTypeDefinition) {
-              elements.add((XSDTypeDefinition) component);
-          } else if (component instanceof XSDAttributeDeclaration) {
-              attributes.add((XSDAttributeDeclaration) component);
+          if (component instanceof XSDTypeDefinition definition1) {
+              types.add(definition1);
```
groldan, line 81: this is strange but is not a change, and since I don't know how this module works I won't try to fix it:
```java
if (component instanceof XSDTypeDefinition ){
}else if (component instanceof XSDTypeDefinition ){
}
```

docs/src/main/java/org/geotools/jts/SnapToLine.java
```diff
@@ -162,7 +162,11 @@ public void visit(Feature feature) {
         }
 
         System.out.printf(
```
pmauduit, line 164: I'm surprised by `System.out.printf`, using a logger (if this is the purpose here) won't be more appropriate ? Unrelated though because the call was there before the PR, but still.

modules/extension/app-schema/app-schema/src/main/java/org/geotools/data/complex/DataAccessMappingFeatureIterator.java
```diff
@@ -516,28 +513,9 @@ protected boolean sourceFeatureIteratorHasNext() {
...
-            // </targetAttribute>
-            // <sourceExpression>
-            // <inputAttribute>mo:composition</inputAttribute>
-            // <linkElement>gsml:CompositionPart</linkElement>
-            // <linkField>gml:name</linkField>
-            // </sourceExpression>
-            // <isMultiple>true</isMultiple>
-            // </AttributeMapping>
-            // As there can be multiple nodes of mo:composition in this case, we need to retrieve
-            // all of them
```
pmauduit: consider keeping the comment ?

modules/extension/brewer/src/main/java/org/geotools/brewer/color/StyleGenerator.java
```diff
@@ -377,7 +375,7 @@ public static void modifyFTS(FeatureTypeStyle fts, int ruleIndex, String styleEx
         Rule thisRule = fts.rules().get(ruleIndex);
         Filter filter = thisRule.getFilter();
 
-        if (filter instanceof And) { // ranged expression
+        if (filter instanceof BinaryLogicOperator and) { // ranged expression
```
pmauduit: I'd have expected `instanceof And and` instead, but sounds logical with the modification below, where the object is cast onto a `BinaryLogicOperator` anyway.

modules/extension/xsd/xsd-core/src/test/java/org/geotools/xsd/FacetTest.java
```diff
@@ -89,6 +89,11 @@ public void testCDATAWhitespace() throws Exception {
         String s = (String) parser.parse();
 
         Assert.assertEquals(
-                " this is a \n" + " normal string \n" + " with some whitespace and \n" + " some new lines", s);
+                """
+                 this is a\s
+                 normal string\s
+                 with some whitespace and\s
+                 some new lines""",
```
pmauduit: this test fails, but can be easily fixed by adding `\s` before each line.

pmauduit:
Trying to build locally from the top of the branch, there is a failing test on gt-xsd-core which is quite trivial to solve.

Another test is failing into the gt-transform module afterwards (this one: `org.geotools.data.transform.TransformFeatureSourceTest.testReprojectWithSelect`), and I cannot figure out why: it fails only when `mvn verify` is fired up at the root of the repository, running it on the single module goes fine, as well as when run into my IDE.

modules/extension/xsd/xsd-core/src/test/java/org/geotools/xsd/FacetTest.java
pmauduit: modification suggested here: https://github.com/groldan/geotools/pull/2

docs/src/main/java/org/geotools/jts/SnapToLine.java
ianturton: That's in the documentation which is why it doesn't get picked up by the QA check

pmauduit:
On tests which are failing, I tried a first pass adding a `@Ignore` on them to get the compilation actually ending:

* gt-app-schema:  `FeatureChainingTest.testPropertyValueCollection`
* gt-transform: `TransformFeatureSourceTest.testReprojectWithSelect`
* gt-process-feature: `IntersectionFeatureCollectionTest.testExecute` and `testPointInPolygonReturnsPoint`

groldan:
rebased, still expecting some test failures

modules/extension/brewer/src/main/java/org/geotools/brewer/color/StyleGenerator.java
groldan: yup, one of many mistakes from openrewrite. Fixed.

modules/extension/app-schema/app-schema/src/main/java/org/geotools/data/complex/DataAccessMappingFeatureIterator.java
groldan: done

groldan:
@aaime @jodygarnett @ianturton, this is finally ready:

* `mvn clean install -Dall` works except `ImagingParametersTest.testExtensions()` failing due to ImageN 0.4-SNAPSHOT being a moving target, but it's unrelated
* `mvn clean install -Dall -Dqa -DskipTests` works

aaime:
Build failure is not actually random, see:
https://github.com/geotools/geotools/pull/5356

aaime:
Looks good to me

PR #5317, opened 2025-08-26
[GEOT-7804]: ImageMosaic SkipExternalOverviews fallback on system property

dromagnoli:
[GEOT-7804](https://osgeo-org.atlassian.net/browse/GEOT-7804) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

Supporting Java Property for external files lookup.
ImageMosaic Documentation is updated in proper GeoServer documentation section through the related PR:
https://github.com/geoserver/geoserver/pull/8743

  

# Checklist

### DIFF
--- a/modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/ImageMosaicConfigHandler.java
+++ b/modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/ImageMosaicConfigHandler.java
@@ -344,9 +344,10 @@ public class ImageMosaicConfigHandler {
             catalogConfig.setName(name);
         }
         String skipExtOverviews = runConfiguration.getParameter(Prop.SKIP_EXTERNAL_OVERVIEWS);
-        if ("true".equalsIgnoreCase(skipExtOverviews)) {
-            catalogConfig.setSkipExternalOverviews(true);
+        if (skipExtOverviews != null && !skipExtOverviews.trim().isEmpty()) {
+            catalogConfig.setSkipExternalOverviews(Boolean.parseBoolean(skipExtOverviews));
         }
+
         String wrapStore = runConfiguration.getParameter(Prop.WRAP_STORE);
         if ("true".equalsIgnoreCase(wrapStore)) {
             catalogConfig.setWrapStore(true);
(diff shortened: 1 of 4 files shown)


modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/ImageMosaicConfigHandler.java
```diff
@@ -1416,8 +1419,13 @@ public void updateConfiguration(
             } else {
                 catalogConfigurationBean.setTypeName(targetCoverageName);
             }
-            catalogConfigurationBean.setSkipExternalOverviews(
-                    IndexerUtils.getParameterAsBoolean(Prop.SKIP_EXTERNAL_OVERVIEWS, indexer));
+            String skipExternalOverviews = IndexerUtils.getParameter(Prop.SKIP_EXTERNAL_OVERVIEWS, indexer);
```
aaime, line 1420: I'm not 100% sure I understand this change... is this code running during the first mosaic harvesting?
What I'm getting at, is that I would not bake in the mosaic properties configuration file the value of a system property, but write it only if the indexer actually contained it.

modules/plugin/imagemosaic/src/test/java/org/geotools/gce/imagemosaic/ImageMosaicSkipOverviewsTest.java
```diff
@@ -0,0 +1,84 @@
+package org.geotools.gce.imagemosaic;
```
aaime, line 17: No copyright header

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/ImageMosaicConfigHandler.java
dromagnoli: The issue was in having the catalogConfigurationBean using with a primitive boolean type, so I had to set it somehow. 
I have now modified it to use a Boolean so that we don't set it if not specified and we don't set a system property to the mosaic config anymore.

modules/plugin/imagemosaic/src/test/java/org/geotools/gce/imagemosaic/ImageMosaicSkipOverviewsTest.java
dromagnoli: fixed

dromagnoli:
All the feedbacks have been addressed.

dromagnoli:
Merging (Downstream integration build was failing due to jdbc-sec failure which is not related to this change. Even restarting the job still fails).

PR #5323, opened 2025-08-30
Fix svg displacement

jyrimatti:
I had encountered some problems during rendering which looked like the displacement wasn't being applied correctly. I had some trouble figuring out why it doesn't happen always, but finally realized that rendering SVG icons takes a different code path than rendering raster icons. My current workaround of using a certain kind of AnchorPoint instead of a correct one with an appropriate Displacement works, but is so hideously ugly that I'd prefer having things work as they should.

This PR fixes the order of transformations in `org.geotools.renderer.lite.StyledShapePainter` IconStyle2D branch.
As documented in `org.geotools.api.style.Graphic.getDisplacement`, the displacement should happen _after_ rotation. The GraphicStyle2D branch seems to have it already correct.

This PR also adds some tests to the raster branch (ExternalGraphicTest) to ensure the already correct behavior there when AnchorPoint, Rotation and Displacement are all being used. Another test in DrawTest is added to reproduce the actual bug and verify the fix.

The expected images (under test-data) at least in my opinion look correct, but please do let me know if I have thoroughly misunderstood something. Also please let me know if you need anything more from this PR.
  

# Checklist


ianturton:
This fix can not be merged until a CLA is filed with OSGeo and a jira ticket is provided and referenced in the title of this PR

PR #5324, opened 2025-08-30
Fix json nested property access

jyrimatti:
This PR fixes `org.geotools.filter.expression.geojson.JSONNodePropertyAccessorFactory.JSONNodePropertyAccessor` to allow getting nested property values from GEOJson data with `someProperty/someNestedProperty`. This is much less verbose syntax than using the jsonPointer function. The earlier implementation couldn't handle FeatureType objects, which some code paths passed to it, so I'm guessing this class didn't work at all.

This PR also changes how `org.geotools.feature.simple.SimpleFeatureTypeBuilder.retype` handles copying attribute descriptors. Now it
1) doesn't produce duplicates even if the caller requests for them,
2) doesn't produce nulls if attribute is not found,
3) includes descriptors whose local name is a prefix of a requested attribute, if no exact match exists.
Since getting nulls in the list (from non-existent attributes) fails in a later code path, I'm guessing that no piece of code has ever called this with a nonexistent attribute and thus accepting prefixes shouldn't break anything.

Please let me know if this change seems like unwanted behavior, or if you'd prefer some alternative approach.
  

# Checklist

### DIFF
--- a/modules/library/main/src/main/java/org/geotools/feature/simple/SimpleFeatureTypeBuilder.java
+++ b/modules/library/main/src/main/java/org/geotools/feature/simple/SimpleFeatureTypeBuilder.java
@@ -25,6 +25,7 @@ import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
@@ -949,6 +950,14 @@ public class SimpleFeatureTypeBuilder {
         return retype(original, Arrays.asList(attributes));
     }
 
+    /** Returns first AttributeDescriptor whose localName is a prefix of the provided name. */
+    private static final Optional<AttributeDescriptor> findFirstWithPrefixingName(
+            List<AttributeDescriptor> descriptors, String name) {
+        return descriptors.stream()
+                .filter(ad -> name.startsWith(ad.getLocalName()))
+                .findFirst();
+    }
+
     /**
      * Create a SimpleFeatureType containing just the attribute descriptors indicated.
      *
@@ -965,10 +974,14 @@ public class SimpleFeatureTypeBuilder {
         // clear the attributes
         b.attributes().clear();
 
-        // add attributes in order
-        for (String type : attributes) {
-            b.add(original.getDescriptor(type));
-        }
+        // add recognized attributes in order and only once
+        List<AttributeDescriptor> ads = attributes.stream()
+                .flatMap(attr -> Optional.ofNullable(original.getDescriptor(attr))
+                        .or(() -> findFirstWithPrefixingName(original.getAttributeDescriptors(), attr))
+                        .stream())
+                .distinct()
+                .toList();
+        b.addAll(ads);
 
         // handle default geometry
         GeometryDescriptor defaultGeometry = original.getGeometryDescriptor();
--- a/modules/unsupported/geojson-core/src/main/java/org/geotools/filter/expression/geojson/JSONNodePropertyAccessorFactory.java
+++ b/modules/unsupported/geojson-core/src/main/java/org/geotools/filter/expression/geojson/JSONNodePropertyAccessorFactory.java
@@ -25,6 +25,7 @@ import java.util.Date;
 import java.util.Map;
 import org.geotools.api.feature.simple.SimpleFeature;
 import org.geotools.api.feature.simple.SimpleFeatureType;
+import org.geotools.api.feature.type.AttributeDescriptor;
 import org.geotools.data.geojson.GeoJSONReader;
 import org.geotools.filter.expression.PropertyAccessor;
 import org.geotools.filter.expression.PropertyAccessorFactory;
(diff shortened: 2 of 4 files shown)


modules/library/main/src/main/java/org/geotools/feature/simple/SimpleFeatureTypeBuilder.java
```diff
@@ -965,10 +974,14 @@ public static SimpleFeatureType retype(SimpleFeatureType original, List<String>
         // clear the attributes
         b.attributes().clear();
 
-        // add attributes in order
-        for (String type : attributes) {
-            b.add(original.getDescriptor(type));
-        }
+        // add recognized attributes in order and only once
```
aaime, line 980: If the attribute is added twice with the same exact name, an error should be thrown. We don't do lenient attribute matching by prefix, a simple feature is simple, but not sloppy.

modules/unsupported/geojson-core/src/main/java/org/geotools/filter/expression/geojson/JSONNodePropertyAccessorFactory.java
```diff
@@ -54,19 +55,25 @@ static class JSONNodePropertyAccessor implements PropertyAccessor {
...
+                            } else {
+                                JsonPointer pointer = JsonPointer.compile(parts[1]);
+                                JsonNode jsonNode = ((JsonNode) value).at(pointer);
+                                return !jsonNode.isMissingNode();
+                            }
                         }
+                    } else if (object instanceof SimpleFeatureType) {
+                        SimpleFeatureType type = (SimpleFeatureType) object;
+                        AttributeDescriptor ad = type.getDescriptor(parts[0]);
+                        return ad != null && ad.getType().getBinding().equals(Object.class);
```
aaime: So here you're making the pair with the behavior of the GeoJSON reader, which assigns Object.class as the binding for JSON object properties. And that's why here you're checking specifically for biding being equal to Object.class. An explicit comment would help clarify that.

The big downside I see with this extension is that the code will trigger with features types not genreated by the geojson reader, e.g., a property accessor is completely generic, if another module creates a feature type with Object.class as the binding, and then "foo/bar" is used to access the foo property, then this property accessor will be used and it will fail to behave correctly.

We'd need to mark the properties generated by the GeoJSON reader in a clearer way. Thinking out loud, either make them JSONNode rather than object, or stick some property in the userData map, to clearly indicate that's a JSON property.

jyrimatti: Ah, I see, thank you for the explanation! I picked Object.class at random, since I didn't know better.

JsonNode.class sounds like a sensible approach, since in GeoJSONReader:
```
else if (binding == Object.class) {
    builder.set(n.getKey(), n.getValue());
}
```
n.getValue() is JsonNode anyway. Would you happen to know if changing
```
// a complex object, we don't know what it is going to be
typeBuilder.add(n.getKey(), Object.class);
```
to
```
// a complex object, we don't know what it is going to be
typeBuilder.add(n.getKey(), JsonNode.class);
```
would bring any problems? This is what you had in mind?

modules/library/main/src/main/java/org/geotools/feature/simple/SimpleFeatureTypeBuilder.java
jyrimatti: `If the attribute is added twice with the same exact name, an error should be thrown.`
I agree, just figured I should keep the non-erroring behavior, since this is not really relevant to this PR. I can add a check if that's the preferred way.

`We don't do lenient attribute matching by prefix, a simple feature is simple, but not sloppy.`
I'm not in any way an expert with the codebase, but it seems to me that the extension cannot work without some kind of leniency somewhere. Instead of prefix matching I could of course match anything before the first '/', but I guess that would be just as lenient.

In my use case I use a `FeatureLayer` which internally builds a `CollectionFeatureSource` (from a `FeatureCollection`) which later invokes `retype`:
```
if (query.getPropertyNames() != Query.ALL_NAMES) {
    // rebuild the type and wrap the reader
    SimpleFeatureType schema = features.getSchema();
    SimpleFeatureType target = SimpleFeatureTypeBuilder.retype(schema, query.getPropertyNames());
```
and I don't seem to be able to affect the `query` so that this block would be skipped. To my understanding `query.getPropertyNames()` is right to contain xpaths like `person/firstName`, and I though `retype` should in that case recognize `person` to be retained.

Or do you think that this block in `CollectionFeatureSource` would be a better place for modification? To strip all '/...' suffixes from all propertyNames passed to `retype`? That's also leniency, but maybe in a less central place?

modules/unsupported/geojson-core/src/main/java/org/geotools/filter/expression/geojson/JSONNodePropertyAccessorFactory.java
aaime: That sounds sensible indeed. I don't know if it's going to cause trouble, but we should find one quickly... one of the Github actions running after commits is a downstream integration build, which will be geowebcache and geosever too, on top of the change. That should tell us if there are problems.

Be mindful we're about to make a very big change in the codebase (switch from JAI to ImageN), if that lands you'll need to rebase this PR or the downstream build wont' even compile.

modules/library/main/src/main/java/org/geotools/feature/simple/SimpleFeatureTypeBuilder.java
jyrimatti: I changed this to check if unrecognized attribute name contains `/`, in which case the attribute corresponding to its top-level name is added. Would this be clean enough, or should I try moving the leniency to `CollectionFeatureSource` instead?

PR #5337, opened 2025-09-03
[GEOT-7807] SHP determin correct JTS CoordinateSequence Dimension and Measure

jodygarnett:
[GEOT-7807](https://osgeo-org.atlassian.net/browse/GEOT-7807) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

When reading from Shapefile the resulting GeometryAttribute Descriptor does not contain Dimension / Measure information needed for the creation of CoordinateSequence with the correct Coordinate subclass.

# Checklist

### DIFF
--- a/modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/shp/PointHandler.java
+++ b/modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/shp/PointHandler.java
@@ -80,17 +80,8 @@ public class PointHandler implements ShapeHandler {
         }
 
         Coordinate c;
-        if (shapeType == ShapeType.POINTZ && !flatGeometry) {
-
-            c = new CoordinateXYZM();
-        } else if (shapeType == ShapeType.POINTM && !flatGeometry) {
-            c = new CoordinateXYM();
-        } else {
-            c = new CoordinateXY();
-        }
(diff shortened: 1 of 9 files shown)


modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/shp/PointHandler.java
```diff
@@ -80,32 +80,33 @@ public Object read(ByteBuffer buffer, ShapeType type, boolean flatGeometry) {
...
+                return geometryFactory.createPoint(c);
+            }
+            else {
+                c = new Coordinate(x, y, z);
+                return geometryFactory.createPoint(c);
+            }
         }
 
+        // Fallthrough to POINT or flatGeometry
+        c = new CoordinateXY(x,y);
```
jodygarnett, line 109: The only way to tell if optional M is present is by reading XYZ and seeing if buffer.hasRemaining()!

With this in mind I changed the logic to ready each value into a variable, and then create the correct coordinate subclass when we know for sure we have everything.  This avoids the ambiguity of returning a CoordinateXYZM by default, when only Coordinate may be needed.

A similar change is now needed for the other handlers.

PR #5338, opened 2025-09-04
GeoTools switch to ImageN

aaime:
Switches to ImageN (0.4-SNAPSHOT for the moment) and a ImageIO-EXT 2.0-SNAPSHOT.

# Checklist


modules/library/coverage/src/main/java/org/geotools/image/io/ImageIOExt.java
```diff
@@ -239,16 +226,6 @@ public static final ImageInputStreamSpi getImageInputStreamSPI(
         return spi;
     }
 
-    /**
-     * Tells me whether or not the native libraries for JAI/ImageIO are active or not.
-     *
-     * @return <code>false</code> in case the JAI/ImageIO native libs are not in the path, <code>
-     * true</code> otherwise.
-     */
-    public static boolean isCLibAvailable() {
```
dromagnoli, line 248: Should we deprecate it or is it fine to just get rid of it?
(I think that nobody is using it anyway)

modules/library/main/pom.xml
```diff
@@ -259,8 +259,12 @@
...
     </dependency>
     <dependency>
-      <groupId>it.geosolutions.jaiext.utilities</groupId>
-      <artifactId>jt-utilities</artifactId>
+      <groupId>org.eclipse.imagen</groupId>
+      <artifactId>utilities</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.eclipse.imagen</groupId>
+      <artifactId>rendered-image-browser</artifactId>
```
dromagnoli, line 267: Does main depends on the RenderedImageBrowser?
(I guess it was already like that with jai-ext but now it has its own dedicated module in ImageN). 
Just asking

modules/library/referencing/src/main/java/org/geotools/parameter/ImagingParameter.java
```diff
@@ -94,11 +94,11 @@ public double doubleValue() throws InvalidParameterTypeException {
...
-            if (type.equals(Float.class)) parameters.getFloatParameter(name);
-            if (type.equals(Long.class)) parameters.getLongParameter(name);
-            if (type.equals(Integer.class)) parameters.getIntParameter(name);
-            if (type.equals(Short.class)) parameters.getShortParameter(name);
-            if (type.equals(Byte.class)) parameters.getByteParameter(name);
+            if (type.equals(Float.class)) return parameters.getFloatParameter(name);
+            if (type.equals(Long.class)) return parameters.getLongParameter(name);
+            if (type.equals(Integer.class)) return parameters.getIntParameter(name);
+            if (type.equals(Short.class)) return parameters.getShortParameter(name);
+            if (type.equals(Byte.class)) return parameters.getByteParameter(name);
```
dromagnoli, line 101: Nice. how it worked like that for ages? :)
(Probably always returning the parameter as double.)

modules/library/referencing/src/test/java/org/geotools/parameter/ImagingParametersTest.java
```diff
@@ -51,17 +50,17 @@ public final class ImagingParametersTest {
...
     @Test
     public void testDescriptors() {
-        final String author = Citations.JAI.getTitle().toString();
-        final String vendor = "com.sun.media.jai";
+        final String author = Citations.IMAGEN.getTitle().toString();
+        final String vendor = "org.eclipse.imagen.media";
         final String mode = RenderedRegistryMode.MODE_NAME;
         final RegistryElementDescriptor descriptor =
-                JAI.getDefaultInstance().getOperationRegistry().getDescriptor(mode, "AddConst");
+                JAI.getDefaultInstance().getOperationRegistry().getDescriptor(mode, "scale");
```
dromagnoli: is lower case "scale" fine here?

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/ImageMosaicFormatFactory.java
```diff
@@ -40,267 +35,11 @@
...
-    private static final String GDAL_SPI = "it.geosolutions.imageio.gdalframework.GDALImageReaderSpi";
-
-    private static final String KAKADU_SPI = "it.geosolutions.imageio.plugins.jp2k.JP2KKakaduImageReaderSpi";
-
     /** Logger. */
     private static final Logger LOGGER = org.geotools.util.logging.Logging.getLogger(ImageMosaicFormatFactory.class);
 
     static {
         replaceTIFF();
-
```
dromagnoli, line 60: Do we need to update any documentation for that?
(Probably not, I'm not sure if this JP2K drivers overriding has ever been documented somewhere)

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/RemappingFilteringObjectInputStream.java
```diff
@@ -0,0 +1,288 @@
...
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.function.Predicate;
+import java.util.regex.Pattern;
+
+public final class RemappingFilteringObjectInputStream extends ObjectInputStream {
```
dromagnoli, line 57: I honestly didn't understand what is this class.
Can you please add some javadoc/comment?

EDIT:
I went through the Utils class and I think I have better understood.
However, I would add some javadoc anyway.

modules/library/coverage/src/main/java/org/geotools/coverage/processing/OperationJAI.java
```diff
@@ -1110,28 +1108,24 @@ protected static void handleROINoDataInternal(
         // Getting the internal ROI property
         ROI innerROI = CoverageUtilities.getROIProperty(sourceCoverage);
         // Checking if it canbe set as parameter
-        if (JAIExt.isJAIExtOperation(operationName) && roiIndex >= 0) {
```
dromagnoli, line 1113: Should we preserve the roiIndex check?

modules/library/coverage/src/main/java/org/geotools/coverage/processing/OperationJAI.java
```diff
@@ -1110,28 +1108,24 @@ protected static void handleROINoDataInternal(
...
+            newROI = roiParam != null ? innerROI.intersect(roiParam) : innerROI;
         }
+        // Setting of the new ROI
+        parameters.set(newROI, roiIndex);
 
         // Getting NoData propery
         NoDataContainer nodataProp = CoverageUtilities.getNoDataProperty(sourceCoverage);
         Range innerNodata = nodataProp != null ? nodataProp.getAsRange() : null;
         // Setting the NoData Range parameter if not present
-        if (JAIExt.isJAIExtOperation(operationName) && noDataIndex >= 0) {
```
dromagnoli, line 1130: Should we preserve the noDataIndex >=0 check ?

modules/library/coverage/src/main/java/org/geotools/image/io/ImageIOExt.java
aaime: I want to just get rid of this one, it's very specific to the code in GT and the answers has been "false" for a number of years now.

modules/library/main/pom.xml
aaime: There are tests dependingon it, as a replacement for JAI ScrolingImagePanel, used to display images in interactive debugging. That said, well spotted, it should be a test dependency, and main is not using it directly (I'll add it to the other modules needing it for testing).

modules/library/referencing/src/main/java/org/geotools/parameter/ImagingParameter.java
aaime: yeah, always returned double... crazy indeed

modules/library/referencing/src/test/java/org/geotools/parameter/ImagingParametersTest.java
aaime: oddly enough the test passes anyawys, but I've updated it to "Scale" anyways.

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/ImageMosaicFormatFactory.java
aaime: Documentation updates are not part of this PR, but yeah, not ever documented, the jai_imageio reader has never been useful, and it was systematically replaced with one available from GDAL (if it was there). 
But now it's gone, so no replacement needed any longer.

modules/plugin/imagemosaic/src/main/java/org/geotools/gce/imagemosaic/RemappingFilteringObjectInputStream.java
aaime: Extensive javadoc being added in the incoming update

modules/library/coverage/src/main/java/org/geotools/coverage/processing/OperationJAI.java
aaime: Well spotted

modules/library/coverage/src/main/java/org/geotools/coverage/processing/OperationJAI.java
aaime: Yep

modules/library/main/pom.xml
aaime: Actually main is using it directly, but should still be a test dependency (it's using ImageViewer, also provided by the same R.I.B. module).

jodygarnett:
Collecting feedback that can be merged now or later here: https://github.com/geotools/geotools/pull/5343

modules/library/coverage/src/main/java/org/geotools/image/jai/Registry.java
```diff
@@ -77,7 +75,7 @@ private Registry() {}
      */
     public static synchronized void setNativeAccelerationAllowed(
```
jodygarnett, line 58: Is this method useful or can it be deprecated? I do not think we have any hard where acceleration presently ...

dromagnoli: Hi Jody. Good point, I think that we can deprecate it. 
I'm going to update the branch with that deprecated.

PR #5358, opened 2025-09-15
[GEOT-7816] - Quality Degradation in Scanned Map Rendering (GeoServer 2.25.3 onward)

jemacchi:
[GEOT-7816](https://osgeo-org.atlassian.net/browse/GEOT-7816) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

This PR includes a fix for calculation of subsampling factors (keeping the refactor already done) in order to resolve the quality loss mentioned in the related ticket (GEOS-11823) while ensuring that the original issue (whitelines in rendered images) is also working as expected.

Please check this PR in GeoServer (since some expected test images rollback to the original ones)

* https://github.com/geoserver/geoserver/pull/8787

Update: created related ticket in GEOT, as required by conventions/procedures. 
https://osgeo-org.atlassian.net/browse/GEOT-7816
  

# Checklist


modules/library/coverage/src/main/java/org/geotools/image/util/ImageUtilities.java
```diff
@@ -1361,18 +1361,74 @@ public static void setSubsamplingFactors(
...
-        int subSamplingFactorX = (int) Math.round(requestedRes[0] / selectedRes[0]);
-        subSamplingFactorX = subSamplingFactorX == 0 ? 1 : subSamplingFactorX;
 
-        while (subSamplingFactorX > 0 && rasterWidth / subSamplingFactorX <= 0) subSamplingFactorX--;
-        subSamplingFactorX = subSamplingFactorX <= 0 ? 1 : subSamplingFactorX;
-
-        int subSamplingFactorY = (int) Math.round(requestedRes[1] / selectedRes[1]);
-        subSamplingFactorY = subSamplingFactorY == 0 ? 1 : subSamplingFactorY;
+        // Control parameters
+        final double resEpsilon = 1e-9; // Small epsilon to avoid floating point errors when using floor
```
aaime: These should become user-controllable via System.getProperty.

modules/library/coverage/src/main/java/org/geotools/image/util/ImageUtilities.java
```diff
@@ -1361,18 +1361,74 @@ public static void setSubsamplingFactors(
...
+
+        subSamplingFactorX = Math.max(1, subSamplingFactorX);
+        subSamplingFactorY = Math.max(1, subSamplingFactorY);
+
+        // Explicit clamp per dimension (replaces integer division)
+        // Ensures we never request a factor larger than the raster size
+        if (subSamplingFactorX > rasterWidth) subSamplingFactorX = rasterWidth;
+        if (subSamplingFactorY > rasterHeight) subSamplingFactorY = rasterHeight;
+
+        // Upward adjustment ONLY if it falls within tolerance
```
aaime: This needs to be explained better. Something like:

> The previous code used floor to compute an integer subsampling factor. Now testing if switching to the next integer (ceil) is within tolerance, and in that case, use it to get better performance. 

(and was it for better performance? I think so, reading less pixels, but want to double check)

modules/library/coverage/src/main/java/org/geotools/image/util/ImageUtilities.java
```diff
@@ -1361,18 +1361,74 @@ public static void setSubsamplingFactors(
...
+            double relErrFloor = Math.abs(ratioY - subSamplingFactorY) / Math.max(ratioY, 1e-9);
+            double relErrUp = Math.abs(ratioY - candidateY) / Math.max(ratioY, 1e-9);
+            if (relErrUp <= overviewTolerance && relErrUp < relErrFloor) {
+                subSamplingFactorY = candidateY;
+            }
+        }
 
-        while (subSamplingFactorY > 0 && rasterHeight / subSamplingFactorY <= 0) subSamplingFactorY--;
-        subSamplingFactorY = subSamplingFactorY <= 0 ? 1 : subSamplingFactorY;
+        // Align SourceRegion to subsampling factors to avoid seams
```
aaime, line 1434: Ah, this is new, interesting. (or did you pick it from some older code? I don't remember seeing region adjustment too).

The javadoc of the method should be updated accordingly.

modules/library/coverage/src/main/java/org/geotools/image/util/ImageUtilities.java
```diff
@@ -1361,18 +1361,74 @@ public static void setSubsamplingFactors(
...
+            int alignedX = region.x - Math.floorMod(region.x, subSamplingFactorX);
+            int alignedY = region.y - Math.floorMod(region.y, subSamplingFactorY);
+
+            // Align width/height to the lower multiple of the factors (avoids half rows/columns)
+            int alignedW = region.width - Math.floorMod(region.width, subSamplingFactorX);
+            int alignedH = region.height - Math.floorMod(region.height, subSamplingFactorY);
+
+            // Ensure positive, non-zero sizes (at least one subsampling block)
+            if (alignedW <= 0) alignedW = subSamplingFactorX;
+            if (alignedH <= 0) alignedH = subSamplingFactorY;
```
aaime, line 1448: @dromagnoli can we have a reader with negative starting points?

modules/library/coverage/src/main/java/org/geotools/image/util/ImageUtilities.java
```diff
@@ -1361,18 +1361,74 @@ public static void setSubsamplingFactors(
...
+
+            // Conservative clipping if realignment exceeds raster bounds
+            // (assuming (0,0)-(rasterWidth,rasterHeight) as the full raster bounds)
+            if (alignedX < 0) alignedX = 0;
+            if (alignedY < 0) alignedY = 0;
+            if (alignedX + alignedW > rasterWidth) alignedW = Math.max(subSamplingFactorX, rasterWidth - alignedX);
+            if (alignedY + alignedH > rasterHeight) alignedH = Math.max(subSamplingFactorY, rasterHeight - alignedY);
```
aaime, line 1455: This still does not guarantee that we go above the raster width/height, could a reader throw an exception if we do, @dromagnoli ?

aaime (commented):
The change is not trivial. For the moment, laying down some comments and asking @dromagnoli to give it a second look. I'm also going to try it out with some dataset (will take some time).

aaime:
But there is one think that you can start working on right away... a test data sample weighing 4MB is unacceptable. Please make it work with a test data set that is a few 10s of KB, 100KB at the most.

modules/library/coverage/src/main/java/org/geotools/image/util/ImageUtilities.java
dromagnoli: Hi @aaime, I'm not sure but I don't think so. 
I have seen some negative minx and miny on the image properties along the JAI chain but I don't think they were at the very beginning (so at the first image read). I think they get negative after some transformations in the middle (scale, affine, mosaic,... ). Moreover, the IIOParam (parent class of ImageReadParam) will throw an exception when setting a source region with negative x and/or y

modules/library/coverage/src/main/java/org/geotools/image/util/ImageUtilities.java
dromagnoli: @aaime 
Some readers may invoke the ImageReader.computeRegions method that internally does some maths and region intersections and that will throw an exception in case of empty regions. So I think that the region intersections will prevent to have a request going above the raster width/height. 
However, I think that it would be safer to prevent that situation so that we are covered for cases where the reader is not involving the computeRegions method.

modules/library/coverage/src/main/java/org/geotools/image/util/ImageUtilities.java
jemacchi: Yes. Reading fewer pixels reduces the amount of data decoded and transferred from disk into memory, so it speeds up image reads. The tolerance check ensures this doesn’t cause noticeable quality degradation.

modules/library/coverage/src/main/java/org/geotools/image/util/ImageUtilities.java
jemacchi: Yes, this part is new. It was not in the old code. The idea is to improve the accuracy of subsampling when the floor factor is close to the requested ratio  (what I found during the regression is that we need to keep use of floor, but also needed to maintain the fix you applied about whitelines).

Specifically, it tests whether moving from the floor factor to the next integer (+1) would actually reduce the relative error, while still staying within the allowed tolerance (3% in our case). If so, we adopt the larger factor. This avoids unnecessary quality loss and ensures better alignment of the subsampling to the requested resolution.

Since this logic is non-trivial, the Javadoc of the method should be updated to describe both steps:

- Compute the base subsampling factor with floor
- Optionally adjust upward by +1 if that gives a closer match to the requested resolution and is within tolerance.

modules/library/coverage/src/main/java/org/geotools/image/util/ImageUtilities.java
jemacchi: my 2 cents, negative x/y should not occur at the first image read, since ImageReadParam.setSourceRegion() rejects them. The check is defensive, to prevent failures in case alignment math produces out-of-bounds values.

modules/library/coverage/src/main/java/org/geotools/image/util/ImageUtilities.java
jemacchi: This block is specifically what -from what I tested/reviewed - fixes -as an alternative solution- the “whiteline” problem: by aligning the source region to multiples of the subsampling factors, we avoid partial rows/columns being skipped, which would otherwise show as gaps between tiles.

The clipping to >= 0 is defensive; ImageReadParam itself does not allow negative x/y at the first read, so in practice they should not occur. 

The main purpose of this section is seam prevention, while the earlier tolerance fix addresses quality (that is the purpose of the PR).

PR #5364, opened 2025-09-29
[GEOT-7819] GeoTools update for JAI to ImageN Rename of classes and constants

jodygarnett:
[GEOT-7819](https://osgeo-org.atlassian.net/browse/GEOT-7819) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

Concurrent merge of jai-rename branches:

* https://github.com/eclipse-imagen/imagen/pull/115
* https://github.com/geosolutions-it/imageio-ext/pull/334
* https://github.com/geotools/geotools/pull/5364
* https://github.com/mapfish/mapfish-print-v2/pull/39
* https://github.com/GeoWebCache/geowebcache/pull/1434
* https://github.com/geoserver/geoserver/pull/8821

Reviewers: This change requires these branches to be merged at the same time, see upstream [pull request](https://github.com/eclipse-imagen/imagen/pull/115) for a complete integration test of the jai-rename branches.
 
# Checklist


modules/library/metadata/src/main/java/org/geotools/util/factory/Hints.java
```diff
@@ -640,7 +640,7 @@ public class Hints extends RenderingHints {
      *
      * @since 2.4
      */
-    public static final Key USE_JAI_IMAGEREAD = new Key(Boolean.class);
+    public static final Key USE_IMAGEN_IMAGEREAD = new Key(Boolean.class);
```
aaime, line 643: This introduces a backwards incompatible change, as Key.getName() uses reflection to lookup the static constant name, so whenever we have, in a stored configuration like GeoServer coverage.xml, the string "USE_JAI_IMAGEREAD", that won't match any longer.

Looking at the GeoServer code, it's instantiating the read parameters using such stored strings directly:
https://github.com/geoserver/geoserver/blob/e72819d33929fc729f2d76664d15d64383d1a71b/src/main/src/main/java/org/geoserver/data/util/CoverageUtils.java#L137

The GUI is affected too, because it's also based on string matches:

image


So... we'll definitely need some sort of translation there (at the config level at least, maybe translate on read), but it would probably be prudent to also add a compatibility layer here, in the various readers.

Going to start look at what can be done in GeoServer once I'm one with the review of the changes happened in the last few days.

aaime:
Btw, I also removed ComponentSampleModelImageN, we're on jdk 17, it's just not useful anymore. ComponentSampleModelJAI is preserved for serialization sake.

PR #5514, opened 2026-01-29
[GEOT-7866] Update GeoTools to Jackson 3

aaime:
[GEOT-7866](https://osgeo-org.atlassian.net/browse/GEOT-7866) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)

Updating to Jackson 3 as part of the GeoServer 3 effort (Spring 7 uses Jackson 3, deprecated usage of version 2).
Downstream changes that need to go with the merge of this one:
* https://github.com/GeoWebCache/geowebcache/pull/1467
* https://github.com/geoserver/geoserver/pull/9074

# Checklist


modules/library/cql2-json/pom.xml
```diff
@@ -10,16 +10,18 @@
   <artifactId>gt-cql2-json</artifactId>
   <dependencies>
     <dependency>
-      <groupId>com.fasterxml.jackson.core</groupId>
+      <groupId>tools.jackson.core</groupId>
       <artifactId>jackson-core</artifactId>
+      <version>3.0.2</version>
```
mprins: I'm curious as to why this is not using `${jackson.version}` which is defined at 3.0.4. 
I think the version can be omitted altogether since the `tools.jackson:jackson-bom` is imported.

mprins (commented):
in several places I see a downgrade/override from 3.0.4 specified in platform-dependencies to 3.0.2; I think in most or all of those cases the `version` should actually not be specified and it should rely on the imported BOM

modules/library/cql2-json/pom.xml
aaime: Good catch, it's likely a leftover of the OpenRewrite migration script (it likes to sprinkle the version number all over the place)

modules/unsupported/geoparquet/src/main/java/org/geotools/jackson/datatype/projjson/CoordinateReferenceSystemDeserializer.java
```diff
@@ -39,42 +37,32 @@
...
 
     @Override
-    public CoordinateReferenceSystem deserialize(JsonParser p, DeserializationContext ctxt)
-            throws IOException, JsonProcessingException {
+    public CoordinateReferenceSystem deserialize(JsonParser p, DeserializationContext ctxt) throws JacksonException {
 
         // Parse the JSON tree
-        ObjectMapper mapper = (ObjectMapper) p.getCodec();
+        ObjectMapper mapper = (ObjectMapper) p.objectReadContext();
         JsonNode root = mapper.readTree(p);
```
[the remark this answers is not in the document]
aaime: @groldan I have a question for you. Copilot correctly pointed out a cast that would not work here.
And yet all tests are passing. While working on setting up a better jackson3 switch, I've noticed that no single test is going through this deserializer, and yet all of them pass. 
Wondering, can it be removed, or are we just lacking test coverage here?

modules/unsupported/geojson-store/src/main/java/org/geotools/data/geojson/store/GeoJSONDataStoreFactory.java
```diff
@@ -145,10 +144,8 @@ public DataStore createNewDataStore(Map<String, ?> params) throws IOException {
     public synchronized boolean isAvailable() {
         if (isAvailable == null) {
             try {
-                @SuppressWarnings("PMD.UnusedLocalVariable")
-                Class geoJSONReaderType = Class.forName("com.fasterxml.jackson.databind.ObjectMapper");
-                ObjectMapper mapper =
-                        JsonMapper.builder().addModule(new JtsModule()).build();
+                Class.forName("com.fasterxml.jackson.databind.ObjectMapper");
+                JsonMapper.builder().addModule(new JtsModule()).build();
```
mprins: > `Class.forName("com.fasterxml.jackson.databind.ObjectMapper");`

this is a Jackson 2 package/class; likely this will then return `false`; I think it should be `tools.jackson.databind.ObjectMapper` instead

modules/library/cql2-json/pom.xml
groldan: this can be resolved already right? I got confused and tried to fix from the wrong branch (my bad)

aaime: Yes, resolved already, try pulling from https://github.com/aaime/geotools/tree/jackson3_clean

modules/unsupported/geojson-store/src/main/java/org/geotools/data/geojson/store/GeoJSONDataStoreFactory.java
groldan: yep this one seems to deserve some looking into

aaime: Also checked... you can see it's marked as "outdated"

PR #5552, opened 2026-03-12
[GEOT-7881] Add inter-schema support for JDBC joinning feature chaining and nested filtering

fernandor777:
[GEOT-7881](https://osgeo-org.atlassian.net/browse/GEOT-7881) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)  

JIRA issue:
https://osgeo-org.atlassian.net/browse/GEOT-7881

This PR introduce the described inter-schemas support functionality for app-schema jonning.
I tested the online integration tests with Postgis, all passing.
  

# Checklist

### DIFF
--- a/modules/extension/app-schema/app-schema/src/main/java/org/geotools/appschema/jdbc/JoiningJDBCFeatureSource.java
+++ b/modules/extension/app-schema/app-schema/src/main/java/org/geotools/appschema/jdbc/JoiningJDBCFeatureSource.java
@@ -153,7 +153,14 @@ public class JoiningJDBCFeatureSource extends JDBCFeatureSource {
     }
 
     /** Create order by field for specific table name */
-    protected void sort(String typeName, String alias, SortBy[] sort, Set<String> orderByFields, StringBuffer sql)
+    protected void sort(
+            JoiningQuery query,
+            String typeName,
+            String schemaName,
+            String alias,
+            SortBy[] sort,
+            Set<String> orderByFields,
+            StringBuffer sql)
             throws IOException, SQLException {
         for (SortBy sortBy : sort) {
             if (SortBy.NATURAL_ORDER.equals(sortBy) || SortBy.REVERSE_ORDER.equals(sortBy)) {
@@ -163,7 +170,7 @@ public class JoiningJDBCFeatureSource extends JDBCFeatureSource {
                 if (alias != null) {
                     encodeColumnName2(sortBy.getPropertyName().getPropertyName(), alias, mySql, null);
                 } else {
-                    encodeColumnName(sortBy.getPropertyName().getPropertyName(), typeName, mySql, null);
+                    encodeColumnName(sortBy.getPropertyName().getPropertyName(), typeName, schemaName, mySql, null);
                 }
                 if (!mySql.toString().isEmpty() && orderByFields.add(mySql.toString())) {
                     // if it's not already in ORDER BY (because you can't have duplicate column
@@ -185,9 +192,10 @@ public class JoiningJDBCFeatureSource extends JDBCFeatureSource {
         // GEOT-4554: sort by PK if idExpression is not there
         if (sort.length == 0) {
             PrimaryKey joinKey = null;
-            SimpleFeatureType joinFeatureType = getDataStore().getSchema(typeName);
+            SimpleFeatureType joinFeatureType = resolveFeatureType(query, typeName, schemaName);
             try {
-                joinKey = getDataStore().getPrimaryKey(joinFeatureType);
+                JDBCDataStore store = resolveDataStore(query, typeName, schemaName);
+                joinKey = store.getPrimaryKey(joinFeatureType);
             } catch (IOException e) {
                 throw new RuntimeException(e);
             }
@@ -196,7 +204,7 @@ public class JoiningJDBCFeatureSource extends JDBCFeatureSource {
                 if (alias != null) {
                     encodeColumnName2(col.getName(), alias, mySql, null);
                 } else {
-                    encodeColumnName(col.getName(), typeName, mySql, null);
+                    encodeColumnName(col.getName(), typeName, schemaName, mySql, null);
                 }
                 if (!mySql.toString().isEmpty() && orderByFields.add(mySql.toString())) {
                     // if it's not already in ORDER BY (because you can't have duplicate column
@@ -215,14 +223,24 @@ public class JoiningJDBCFeatureSource extends JDBCFeatureSource {
     }
 
     protected void addMultiValuedSort(
-            String tableName, Set<String> orderByFields, JoiningQuery.QueryJoin join, StringBuffer sql)
+            JoiningQuery query,
+            String tableName,
+            Set<String> orderByFields,
+            JoiningQuery.QueryJoin join,
+            StringBuffer sql)
             throws IOException, FilterToSQLException, SQLException {
 
         StringBuffer field2 = new StringBuffer();
-        encodeColumnName(join.getForeignKeyName().toString(), join.getJoiningTypeName(), field2, null);
+        String joinSchema = resolveSchemaForJoin(query, join);
+        JDBCDataStore joinStore = resolveDataStore(query, join.getJoiningTypeName(), joinSchema);
+        joinSchema = effectiveSchema(joinStore, joinSchema);
+        encodeColumnName(join.getForeignKeyName().toString(), join.getJoiningTypeName(), joinSchema, field2, null);
 
         StringBuffer field1 = new StringBuffer();
-        encodeColumnName(join.getJoiningKeyName().toString(), tableName, field1, null);
+        String tableSchema = resolveSchemaForTypeName(query, tableName);
+        JDBCDataStore tableStore = resolveDataStore(query, tableName, tableSchema);
+        tableSchema = effectiveSchema(tableStore, tableSchema);
+        encodeColumnName(join.getJoiningKeyName().toString(), tableName, tableSchema, field1, null);
 
         if (orderByFields.add(field1.toString()) && orderByFields.add(field2.toString())) {
             // check that they don't already exists in ORDER BY because duplicate column names
@@ -249,10 +267,14 @@ public class JoiningJDBCFeatureSource extends JDBCFeatureSource {
 
             if (sort != null) {
                 if (j < 0) {
-                    sort(query.getTypeName(), null, sort, orderByFields, joinOrders);
+                    String rootSchema = resolveSchemaForTypeName(query, query.getTypeName());
+                    JDBCDataStore rootStore = resolveDataStore(query, query.getTypeName(), rootSchema);
+                    rootSchema = effectiveSchema(rootStore, rootSchema);
+                    sort(query, query.getTypeName(), rootSchema, null, sort, orderByFields, joinOrders);
 
                     if (query.getQueryJoins() != null && !query.getQueryJoins().isEmpty()) {
                         addMultiValuedSort(
+                                query,
                                 query.getTypeName(),
                                 orderByFields,
                                 query.getQueryJoins().get(0),
@@ -268,7 +290,7 @@ public class JoiningJDBCFeatureSource extends JDBCFeatureSource {
                         for (String pk : pkColumnNames) {
 
                             StringBuffer pkSql = new StringBuffer();
-                            encodeColumnName(pk, query.getTypeName(), pkSql, null);
+                            encodeColumnName(pk, query.getTypeName(), rootSchema, pkSql, null);
 
                             if (!pkSql.toString().isEmpty() && orderByFields.add(pkSql.toString())) {
                                 sql.append(", ");
@@ -277,13 +299,17 @@ public class JoiningJDBCFeatureSource extends JDBCFeatureSource {
                         }
                     }
                 } else {
+                    String joinSchema = resolveSchemaForJoin(query, join);
+                    JDBCDataStore joinStore = resolveDataStore(query, join.getJoiningTypeName(), joinSchema);
+                    joinSchema = effectiveSchema(joinStore, joinSchema);
                     if (aliases != null && aliases[j] != null) {
-                        sort(join.getJoiningTypeName(), aliases[j], sort, orderByFields, joinOrders);
+                        sort(query, join.getJoiningTypeName(), joinSchema, aliases[j], sort, orderByFields, joinOrders);
                     } else {
-                        sort(join.getJoiningTypeName(), null, sort, orderByFields, joinOrders);
+                        sort(query, join.getJoiningTypeName(), joinSchema, null, sort, orderByFields, joinOrders);
                     }
                     if (query.getQueryJoins().size() > j + 1) {
                         addMultiValuedSort(
+                                query,
                                 join.getJoiningTypeName(),
                                 orderByFields,
                                 query.getQueryJoins().get(j + 1),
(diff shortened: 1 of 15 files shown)


modules/extension/app-schema/app-schema/src/main/java/org/geotools/appschema/jdbc/JoiningJDBCFeatureSource.java
```diff
@@ -310,20 +342,256 @@ public void encodeColumnName2(String colName, String typeName, StringBuffer sql,
...
+
+    private String resolveSchemaForTypeName(JoiningQuery query, String typeName) {
+        if (!AppSchemaDataAccessConfigurator.isCrossSchemaJoiningEnabled() || query == null || typeName == null) {
+            return null;
+        }
+        FeatureTypeMapping rootMapping = query.getRootMapping();
+        if (matchesTypeName(rootMapping, typeName)) {
+            return rootMapping.getSourceDatabaseSchema();
+        }
+        for (QueryJoin join : query.getQueryJoins()) {
```
aaime: In others parts of the code getQueryJoins() is guarded against null, but here it is not. Maybe it's impossible for the query joins to be null in this context, I'm just pointing out the difference.

modules/extension/app-schema/app-schema/src/main/java/org/geotools/appschema/jdbc/NestedFilterToSQL.java
```diff
@@ -272,14 +274,16 @@ private void encodeJoinCondition(FeatureChainedAttributeDescriptor nestedAttrDes
         String parentTableColumn = parentToSQL.encodeToString(parentExpression);
 
         String nestedTableAlias = nestedStep.getAlias();
-        WrappedFilterToSql nestedFilterToSQL = createFilterToSQL(parentFeature);
+        WrappedFilterToSql nestedFilterToSQL = createFilterToSQL(nestedFeature);
```
aaime, line 277: Ok, I'm lost here. Why was the call changed from parent to nested feature?

modules/extension/app-schema/app-schema/src/main/java/org/geotools/appschema/jdbc/NestedFilterToSQL.java
```diff
@@ -494,6 +530,9 @@ protected Object visit(BinaryLogicOperator filter, Object extraData) {
                         // selectClause will carry the parent SELECT FROM clauses, so we use it to
                         // build UNION
                         out.write(" UNION " + selectClause + " ");
+                        if (!selectClause.trim().toUpperCase().endsWith("WHERE")) {
```
aaime, line 534: This seems to come out of the blue. What is the relationship between "let's support cross-schema encoding" and this change?

modules/extension/app-schema/app-schema/src/main/java/org/geotools/data/complex/config/AppSchemaDataAccessConfigurator.java
```diff
@@ -120,6 +121,10 @@ public class AppSchemaDataAccessConfigurator {
 
     public static final String PROPERTY_REPLACE_OR_UNION = "app-schema.orUnionReplace";
 
+    public static final String PROPERTY_CROSS_SCHEMA_JOINING = "app-schema.crossSchemaJoining";
+
+    private static final String SCHEMA_STORE_KEY_SEPARATOR = "::schema::";
```
aaime, line 131: So this is a separator but it's not meant to be used to parse back the components, it's just to generate a unique identifier? I would clarify it in the javadoc

modules/extension/app-schema/app-schema/src/main/java/org/geotools/data/complex/config/AppSchemaDataAccessConfigurator.java
```diff
@@ -136,17 +141,51 @@ public class AppSchemaDataAccessConfigurator {
...
+        }
+        return value != null && value.equalsIgnoreCase("true");
+    }
+
+    private static String getJoiningOverride() {
+        String value = System.getProperty(PROPERTY_JOINING);
+        if (value != null) {
+            return value;
+        }
+        return System.getenv("APP_SCHEMA_JOINING");
```
aaime: Is this new constant documented anywhere?

modules/extension/app-schema/app-schema/src/main/java/org/geotools/data/complex/config/AppSchemaDataAccessConfigurator.java
```diff
@@ -136,17 +141,51 @@ public class AppSchemaDataAccessConfigurator {
...
+    }
+
+    private static String getCrossSchemaJoiningOverride() {
+        String value = System.getProperty(PROPERTY_CROSS_SCHEMA_JOINING);
+        if (value != null) {
+            return value;
+        }
+        return System.getenv("APP_SCHEMA_CROSS_SCHEMA_JOINING");
```
aaime: Is this new constant documented anywhere?

modules/extension/app-schema/app-schema/src/main/java/org/geotools/data/joining/JoiningNestedAttributeMapping.java
```diff
@@ -205,10 +223,16 @@ public DataAccessMappingFeatureIterator initSourceFeatures(
 
         DataAccessMappingFeatureIterator daFeatureIterator = (DataAccessMappingFeatureIterator) featureIterator;
 
-        List<Expression> foreignIds = new ArrayList<>();
-        for (int i = 0; i < query.getQueryJoins().size(); i++) {
-            for (int j = 0; j < query.getQueryJoins().get(i).getIds().size(); j++) {
-                foreignIds.add(filterFac.property(JoiningJDBCFeatureSource.FOREIGN_ID + "_" + i + "_" + j));
+        // Ensure row-level metadata is initialized before deriving FOREIGN_ID_* expressions.
+        boolean hasSourceRows = daFeatureIterator.hasNext();
```
aaime: This call alone is odd, normally the pattern is a loop of hasNext/next... are you trying to get some side-effect out of it?

modules/extension/app-schema/app-schema/src/main/java/org/geotools/appschema/jdbc/JoiningJDBCFeatureSource.java
```diff
@@ -310,20 +342,256 @@ public void encodeColumnName2(String colName, String typeName, StringBuffer sql,
...
+        }
+        return typeName.equals(mapping.getSource().getSchema().getName().getLocalPart());
+    }
+
+    private SimpleFeatureType resolveFeatureType(JoiningQuery query, String typeName, String schema)
+            throws IOException {
+        FeatureTypeMapping mapping = resolveMappingForTypeName(query, typeName, schema);
+        if (mapping == null && schema != null) {
+            mapping = resolveMappingForTypeName(query, typeName, null);
+        }
```
aaime: This little block pop ups here and there, is there any way to centralize it and document it?

modules/extension/app-schema/app-schema/src/main/java/org/geotools/appschema/jdbc/JoiningJDBCFeatureSource.java
```diff
@@ -631,6 +957,48 @@ public String selectSQL(
         return sql.toString();
     }
 
+    private void sanitizeJoinIds(JoiningQuery query) throws IOException {
```
aaime: Perhaps this should return a new instance of JoiningQuery (or the same one, if no change has been done)? Also, some javadoc explaining what is going on would not hurt

modules/extension/app-schema/app-schema/src/main/java/org/geotools/appschema/jdbc/JoiningJDBCFeatureSource.java
```diff
@@ -1450,6 +1973,66 @@ private String createJoiningCountQuery(
...
+        }
+    }
+
+    /**
+     * Joining fallback for count generation when no id column can be resolved.
+     *
+     * <p>We use {@code DISTINCT *} over the joining subquery to preserve de-duplication without emitting an empty
+     * DISTINCT projection.
+     */
+    private String createJoiningCountQueryWithoutIdColumns(
```
aaime, line 2070: This makes sense but I'm wondering if this is sort of an expected code path, or more like tolerance for an unexpected case. If it's the latter, should there be some logging?

aaime (commented):
My familiarity with app-schema is insufficient to do a good review, so take what I wrote with a grain of salt. 
One thing that is making me wonder, I don't see integration tests that actually use a database, do you plan to write one downstream in GeoServer?

modules/extension/app-schema/app-schema/src/main/java/org/geotools/appschema/jdbc/JoiningJDBCFeatureSource.java
```diff
@@ -310,20 +342,256 @@ public void encodeColumnName2(String colName, String typeName, StringBuffer sql,
...
+            throws SQLException {
+        if (schemaName == null) {
+            getDataStore().encodeTableName(typeName, sql, hints);
+            return;
+        }
+        SQLDialect dialect = getDataStore().getSQLDialect();
+        dialect.encodeSchemaName(schemaName, sql);
+        sql.append(".");
+        dialect.encodeTableName(typeName, sql);
+    }
```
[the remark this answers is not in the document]
fernandor777: Thanks pretrained transformer, good catch actually, I'm applying a fix for this.

modules/extension/app-schema/app-schema/src/main/java/org/geotools/appschema/jdbc/JoiningJDBCFeatureSource.java
```diff
@@ -310,20 +342,256 @@ public void encodeColumnName2(String colName, String typeName, StringBuffer sql,
...
+            return;
+        }
+        SQLDialect dialect = getDataStore().getSQLDialect();
+        dialect.encodeSchemaName(schemaName, sql);
+        sql.append(".");
+        dialect.encodeTableName(typeName, sql);
+        if (alias != null) {
+            dialect.encodeTableAlias(alias, sql);
+        }
+    }
```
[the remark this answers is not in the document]
fernandor777: Thanks, fixing.

modules/extension/app-schema/app-schema/src/main/java/org/geotools/appschema/jdbc/NestedFilterToSQL.java
```diff
@@ -356,14 +360,46 @@ private WrappedFilterToSql createFilterToSQL(FeatureTypeMapping featureMapping)
...
+        if (schemaName == null) {
+            store.encodeTableName(typeName, sql, hints);
+            return;
+        }
+        SQLDialect dialect = store.getSQLDialect();
+        dialect.encodeSchemaName(schemaName, sql);
         sql.append(".");
-        store.dialect.encodeColumnName(null, colName, sql);
+        dialect.encodeTableName(typeName, sql);
+    }
```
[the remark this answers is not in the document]
fernandor777: Thanks, fix applied.

modules/extension/app-schema/app-schema/src/main/java/org/geotools/appschema/jdbc/NestedFilterToSQL.java
```diff
@@ -356,14 +360,46 @@ private WrappedFilterToSql createFilterToSQL(FeatureTypeMapping featureMapping)
...
+            return;
+        }
+        SQLDialect dialect = store.getSQLDialect();
+        dialect.encodeSchemaName(schemaName, sql);
+        sql.append(".");
+        dialect.encodeTableName(typeName, sql);
+        if (alias != null) {
+            dialect.encodeTableAlias(alias, sql);
+        }
     }
```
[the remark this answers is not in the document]
fernandor777: Thanks, addresed.

modules/extension/app-schema/app-schema/src/main/java/org/geotools/appschema/jdbc/JoiningJDBCFeatureSource.java
fernandor777: Makes sense, thanks, applied.

modules/extension/app-schema/app-schema/src/main/java/org/geotools/appschema/jdbc/NestedFilterToSQL.java
fernandor777: This was a latent bug that got triggered once a different schema table got referenced as a nested feature, making a wrong schema sql expression built.  Only fixing this made the manual tests and new integration tests pass.

modules/extension/app-schema/app-schema/src/main/java/org/geotools/appschema/jdbc/NestedFilterToSQL.java
fernandor777: Indeed it is related since in some flows (especiually the ones touched during inter-schema work), selectClause can be just SELECT ... FROM ... (no WHERE).
Then branch SQL can start with predicates that include AND (from chained filter construction, liike encodeChainedAttribute around line 246), producing invalid SQL like:
```... UNION SELECT ... FROM ... AND ...```
So this is a safety net to avoid missing predicate start clause, basicly.

modules/extension/app-schema/app-schema/src/main/java/org/geotools/data/complex/config/AppSchemaDataAccessConfigurator.java
fernandor777: Right, javadoc added.

modules/extension/app-schema/app-schema/src/main/java/org/geotools/data/complex/config/AppSchemaDataAccessConfigurator.java
fernandor777: Mmm, Initially I thought it was a good idea to "normalize" the sys props into env vars counterparts.  But now it feels too forced.  I'm removing the added env var support and keep the sys prop original support only.

modules/extension/app-schema/app-schema/src/main/java/org/geotools/data/complex/config/AppSchemaDataAccessConfigurator.java
fernandor777: Right, this is being added on GeoServer app-schema Joining documentation section.

modules/extension/app-schema/app-schema/src/main/java/org/geotools/data/joining/JoiningNestedAttributeMapping.java
fernandor777: Right, it is a simplified initialization effect, but not clear to read, I'm refactoring to be clean on the intention.

modules/extension/app-schema/app-schema/src/main/java/org/geotools/appschema/jdbc/JoiningJDBCFeatureSource.java
fernandor777: Right, code smell, refactoring.

modules/extension/app-schema/app-schema/src/main/java/org/geotools/appschema/jdbc/JoiningJDBCFeatureSource.java
fernandor777: Yes, this is the right way, refactoring.

modules/extension/app-schema/app-schema/src/main/java/org/geotools/appschema/jdbc/JoiningJDBCFeatureSource.java
fernandor777: It is a supported use case: createJoiningCountQueryWithoutIdColumns(...) is used when the query has no usable id columns after normalization (views support), so the code has to fall back to a DISTINCT * count strategy to preserve de-duplication.

PR #5631, opened 2026-04-28
macOS CI: cache GDAL build between runs

groldan:
The "*Build GDAL*" step rebuilds GDAL `3.10.1` from source on every PR, taking ~30 min.

Split the step for the brew dependency install (still required at runtime for libproj/libgeos/etc. dylibs) to run always, but the cmake build only runs on cache miss.

The cache key includes the runner image identity (`ImageOS` + `ImageVersion`) to invalidate when GitHub bumps the macOS image, since brew bottle ABI changes can break the cached gdal binary. A smoke test (`gdalinfo --version`) runs after restore; on failure, the build step re-runs the job and self-heals instead of producing a broken run.

Also makes the gdal uninstall tolerant of brew not having installed it.

  

# Checklist

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -26,9 +26,18 @@ jobs:
         distribution: 'temurin'
         architecture: 'arm64'
     - name: Install GDAL build/runtime dependencies via Homebrew
+      # Skip 'brew update' and 'brew upgrade' (which would upgrade every
+      # preinstalled formula on the runner image, ~5+ min for packages
+      # unrelated to gdal). We rely on whatever versions of proj/geos/etc.
+      # ship in the runner image; GitHub rotates the macOS image roughly
+      # every 2 weeks, which gives us fresh-enough libraries. The gdal
+      # cache key below includes ImageVersion, so any image refresh
+      # invalidates the cache and rebuilds gdal against the new bottles.
+      env:
+        HOMEBREW_NO_AUTO_UPDATE: 1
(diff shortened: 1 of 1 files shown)


groldan:
### Timings

| Run | Total | Brew install | Build GDAL | Maven |
|---|---|---|---|---|
| Baseline (current `main`) | ~40m | ~7m | ~24m | ~8m |
| New — cold cache | **~33m** | ~4m | ~21m| ~7m|
| New — warm cache | **~16m** | ~6m | skipped | ~9m |

> - [First run, cold cache](https://github.com/geotools/geotools/actions/runs/25032853198/job/73318143377)
> - [Second run, warm cache](https://github.com/geotools/geotools/actions/runs/25032853198/job/73321575868)

Warm runs are ~60% faster than baseline. Cold runs (after image rotation, GDAL version bump, or workflow edit) save ~7 min from the brew skip alone.

### Notes

- The cache key includes `ImageOS` + `ImageVersion`. GitHub rotates the macOS image roughly every 2 weeks; any rotation invalidates the cache and rebuilds GDAL against the new bottles. A `gdalinfo --version` smoke test runs after restore. If the cached binary fails (e.g., dylib drift slipped past the key), the build step re-runs and the job self-heals.
- Brew install and Maven build show ~2 min variance, which seems normal depending load?
- Cached GDAL artifact reports as ~11 MB, smaller than expected. Functionally OK (warm runs build successfully), but worth a follow-up to confirm the install layout.

aaime:
Nice thanks!

PR #5639, opened 2026-05-08
[GEOT-7898] Setup XMLUtils to assist in use of DefaultEntityResolver

jodygarnett:

To be merged at the same time:

* https://github.com/GeoWebCache/geowebcache/pull/1534
* https://github.com/geoserver/geoserver/pull/9502

This is a QA activity to double check that `GeoTools.getEntityResolver(hints)` is used.

To aid in this goal:
* PMD rules put in place to check for direct construction of xml objects that require entity resolver be supplied
* XMLUtils has been updated cover direct use of xml objects, ensuring entity resolver is supplied via `GeoTools.getEntityResolver(hints)`
* Tests that require local file access have been updated to use `NullEntityResolver.INSTANCE` (which is great confirmation the approach is working)

A request has come in for the following:

* https://github.com/geotools/geotools/pull/5662

## XMLUtils API change enforced by PMD Rules

Direct use of DocumentBuilderFactory.newDocumentBuilder() replaced with:

```java
 DocumentBuilder builder = XMLUtils.newDocumentBuilder();
 Document document = builder.parse(source)
```

Or if factory configuration is required direct use of DocumentBuilderFactory.newInstance() replaced with:

```java
 DocumentBuilderFactory factory = XMLUtils.newDocumentBuilderFactory();
 factory.setValidating(true);
 factory.setNamespaceAware(true);
 DocumentBuilder builder = XMLUtils.newDocumentBuilder(factory);
 Document document = builder.parse(source)
```

Test cases that need access to local file system can configure via ENTITY_RESOLVER Hint:
```java
 Hints hints = GeoTools.getDefaultHints();
 hints.put(Hints.ENTITY_RESOLVER, NullEntityResolver.INSTANCE);

 DocumentBuilder builder = XMLUtils.newDocumentBuilder(hints);
 Document document = builder.parse(source);
```

While this is a QA activity ahead of release, I would like to backport *just* new `XMLUtil` methods so they are available for use.

## GeoTools Hints

I found some problems with GeoTools initialization and Hints notifications during this work, assigned separate ticket:

* https://osgeo-org.atlassian.net/projects/GEOT-7906

## ⚠️ Integration workflow change

This PR changes `integration.yml` and `integration-qa.yml`to look for a branch of the same name when running integration tests:
```
env:
   BRANCH_NAME: ${{ github.head_ref || github.ref_name }}
```

It is my hope that this reduces the need for manual testing of downstream PRs making API changes. The integration branch just needs to be defined, it does not already need to be a pull-request:

```
echo "Checking out GeoServer"
mkdir geoserver
git clone --depth 1 -q https://github.com/geoserver/geoserver.git geoserver
cd geoserver
GS_BRANCH_EXISTS=$(git ls-remote --heads origin "$BRANCH_NAME" | grep -q "$BRANCH_NAME"; echo $?)
if [ "$GS_BRANCH_EXISTS" -eq 0 ]; then
  echo "Branch found on geoserver. Switching and tracking: $BRANCH_NAME"
  git switch "$BRANCH_NAME"
fi
```

As a result This PR is tested against downstream branches:

* https://github.com/geoserver/geoserver/tree/default-entity-resolver-check
* https://github.com/GeoWebCache/geowebcache/tree/default-entity-resolver-check

# Checklist

### DIFF
--- a/build/qa/pmd-ruleset.xml
+++ b/build/qa/pmd-ruleset.xml
@@ -157,24 +157,101 @@ GeoTools ruleset. See https://pmd.github.io/pmd/pmd_userdocs_making_rulesets.htm
       </property>
     </properties>
   </rule>
-  <rule name="AvoidTransform"
+  <rule name="AvoidTransformFactory"
         language="java"
-        message="Do not use TransformerFactory directly; recommend XMLUtils.newTransformer()"
+        message="Do not use TransformerFactory directly; recommend XMLUtils.newInstance()"
         class="net.sourceforge.pmd.lang.rule.xpath.XPathRule">
     <description>
       Notice use of TransformerFactory.newInstance() which should be avoided in favor of
-      XMLUtils.newInstance(Hints) to respect GeoTools init configuration.
+      XMLUtils.newTransformerFactory(Hints) to respect GeoTools init configuration.
     </description>
     <priority>3</priority>
     <properties>
       <property name="xpath">
           <value>
               <![CDATA[
-              //MethodCall[pmd-java:matchesSig("javax.xml.transform.TransformerFactory#newTransformer()")]
+              //MethodCall[pmd-java:matchesSig("javax.xml.transform.TransformerFactory#newInstance()")]
               ]]>
           </value>
       </property>
     </properties>
   </rule>
+  <rule name="AvoidTransformer"
+        language="java"
+        message="Do not use TransformerFactory.newTransformer() directly; recommend XMLUtils.newTransformer()"
+        class="net.sourceforge.pmd.lang.rule.xpath.XPathRule">
+    <description>
+      Notice use of TransformerFactory.newTransformer() which should be avoided in favor of
+      XMLUtils.newTransformer(Hints) to respect GeoTools init configuration.
+    </description>
+    <priority>3</priority>
+    <properties>
+      <property name="xpath">
+        <value>
+          <![CDATA[
+              //MethodCall[pmd-java:matchesSig("javax.xml.transform.TransformerFactory#newTransformer()")]
+              ]]>
+        </value>
+      </property>
+    </properties>
+  </rule>
+  <rule name="AvoidTransformerSource"
+        language="java"
+        message="Do not use TransformerFactory.newTransformer(source) directly; recommend XMLUtils.newTransformer(source)"
+        class="net.sourceforge.pmd.lang.rule.xpath.XPathRule">
+    <description>
+      Notice use of TransformerFactory.newTransformer(Source) which should be avoided in favor of
+      XMLUtils.newTransformer(Source,Hints) to respect GeoTools init configuration.
+    </description>
+    <priority>3</priority>
+    <properties>
+      <property name="xpath">
+        <value>
+          <![CDATA[
+              //MethodCall[pmd-java:matchesSig("javax.xml.transform.TransformerFactory#newTransformer(javax.xml.transform.Source)")]
+              ]]>
+        </value>
+      </property>
+    </properties>
+  </rule>
+  <rule name="AvoidSAXParserFactory"
+        language="java"
+        message="Do not use ParserFactory.newInstance() directly; recommend XMLUtils.newSAXParserFactory()"
+        class="net.sourceforge.pmd.lang.rule.xpath.XPathRule">
+    <description>
+      Notice use of SAXParserFactory.newInstance() which should be avoided in favor of
+      XMLUtils.newSAXParserFactory(Hints) to respect GeoTools init configuration.
+    </description>
+    <priority>3</priority>
+    <properties>
+      <property name="xpath">
+        <value>
+          <![CDATA[
+              //MethodCall[pmd-java:matchesSig("javax.xml.parsers.SAXParserFactory#newInstance()")]
+              ]]>
+        </value>
+      </property>
+    </properties>
+  </rule>
+  <rule name="AvoidSaxParser"
+        language="java"
+        message="Do not use SAXParserFactory.newSAXParser() directly; recommend XMLUtils.newSAXParser()"
+        class="net.sourceforge.pmd.lang.rule.xpath.XPathRule">
+    <description>
+      Notice use of SAXParserFactory.newSAXParser() which should be avoided in favor of
+      XMLUtils.newSAXParser(Hints) to respect GeoTools init configuration.
+    </description>
+    <priority>3</priority>
+    <properties>
+      <property name="xpath">
+        <value>
+          <![CDATA[
+              //MethodCall[pmd-java:matchesSig("javax.xml.parsers.SAXParserFactory#newInstance()")]
+              ]]>
+        </value>
+      </property>
+    </properties>
+  </rule>
+
 
 </ruleset>
--- a/build/qa/pmd-ruleset.xml
+++ b/build/qa/pmd-ruleset.xml
@@ -96,7 +96,10 @@ GeoTools ruleset. See https://pmd.github.io/pmd/pmd_userdocs_making_rulesets.htm
   <rule ref="category/java/performance.xml/UseArrayListInsteadOfVector" />
   <rule ref="category/java/performance.xml/BigIntegerInstantiation"/>
   <rule ref="category/java/performance.xml/StringInstantiation"/>
-  <rule name="wildcards" language="java" message="No Wildcard Imports" class="net.sourceforge.pmd.lang.rule.xpath.XPathRule">
+  <rule name="wildcards"
+        language="java"
+        message="No Wildcard Imports"
+        class="net.sourceforge.pmd.lang.rule.xpath.XPathRule">
     <description>
       Don't use wildcard imports
     </description>
@@ -119,7 +122,7 @@ GeoTools ruleset. See https://pmd.github.io/pmd/pmd_userdocs_making_rulesets.htm
   <rule name="AvoidDocumentBuilder"
         language="java"
         message="Do not use DocumentBuilder directly; recommend XMLUtils.newDocumentBuilder()"
-        class="net.sourceforge.pmd.lang.rule.XPathRule">
+        class="net.sourceforge.pmd.lang.rule.xpath.XPathRule">
     <description>
       Notice use of DocumentBuilderFactory.newDocumentBuilder() which should be avoided in favor of
       XMLUtils.newDocumentBuilder(Hints) to respect GeoTools init configuration.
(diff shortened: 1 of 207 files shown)


ianturton:
sounds like a useful thing to backport to me

aaime:
Several downstream build failures, do we need a PR for GeoServer as well to make a good "pair"? Os is it a case of having to change the implementation here?

jodygarnett:
Okay I am really happy that this builds, the downstream integration build is failing on GeoServer. Ideally I would like this to work "out of the box", but expect we will need a branch to ask GeoServer to use `XMLUtils` on occasion.

Initial research shows:

* EntityResolverProvider.ENTITY_RESOLUTION_UNRESTRICTED will need to return NullEntityResolver.INSTANCE (rather than `null`) for tests to avoid being configured DefaultEntityResolver 
* XMLUtils.transform is failing for CSW, not sure why yet

build/qa/pmd-ruleset.xml
```diff
@@ -114,5 +117,64 @@ GeoTools ruleset. See https://pmd.github.io/pmd/pmd_userdocs_making_rulesets.htm
...
+          </value>
+      </property>
+    </properties>
+  </rule>
+  <rule name="AvoidTransform"
+        language="java"
+        message="Do not use TransformerFactory directly; recommend XMLUtils.newTransformer()"
+        class="net.sourceforge.pmd.lang.rule.xpath.XPathRule">
+    <description>
+      Notice use of TransformerFactory.newInstance() which should be avoided in favor of
```
aaime, line 165: Says here it catches newInstance, but the implementation below catches newTransformer instead. I see in the diff that several "TransformerFactory.newInstance()" were replaced, you might want to check that too... unless there is a reason I don't see?

modules/library/xml/src/main/java/org/geotools/xml/XMLUtils.java
```diff
@@ -32,12 +64,322 @@
...
+     * @return GTTransformerFactory or GTSAXTransformFactory respecting GeoTools configuration.
+     */
+    protected static TransformerFactory toTransformerFactory(TransformerFactory factory, Hints hints) {
+        if (factory == null) {
+            return new GTTransformerFactory(hints);
+        } else if (factory instanceof GTSAXTransformerFactory saxFactory) {
+            if (saxFactory.hints == hints) {
+                return factory;
+            } else {
+                return new GTSAXTransformerFactory(saxFactory, hints);
```
aaime: So here, you're basically double-wrapping. Why not merge the hints and replace instead?

modules/library/xml/src/main/java/org/geotools/xml/XMLUtils.java
```diff
@@ -125,4 +467,529 @@ private static boolean isXMLValidChar(char c) {
...
+            this.entityResolver = entityResolver;
+            this.uriResolver = uriResolver;
+        }
+
+        @Override
+        public Source resolve(String href, String base) throws TransformerException {
+            try {
+                // step 1: check with entity resolver, will return null (external), source (internal), or exception
+                // (forbidden)
+                entityResolver.resolveEntity(href, base);
```
aaime: Can we ever have a resolver that replaces the entity maybe with a local one (e.g. to allow usage in sealed environments, no network access) rather than just checking the source? If so we should probably return the resolved input source no? (in other words, use the value returned by resolveEntity).

modules/library/xml/src/main/java/org/geotools/xml/XMLUtils.java
```diff
@@ -32,12 +64,322 @@
...
+     *
+     * @param factory Transformer factory, or {@code null}
+     * @param hints Factory configuration
+     * @return GTTransformerFactory or GTSAXTransformFactory respecting GeoTools configuration.
+     */
+    protected static TransformerFactory toTransformerFactory(TransformerFactory factory, Hints hints) {
+        if (factory == null) {
+            return new GTTransformerFactory(hints);
+        } else if (factory instanceof GTSAXTransformerFactory saxFactory) {
+            if (saxFactory.hints == hints) {
```
aaime: Exactly the same hints? Equality would be better. 
To make it null safe, use java.utils.Objects.equals(a,b)

build/qa/pmd-ruleset.xml
```diff
@@ -114,5 +117,64 @@ GeoTools ruleset. See https://pmd.github.io/pmd/pmd_userdocs_making_rulesets.htm
...
+      <property name="xpath">
+          <value>
+              <![CDATA[
+              //MethodCall[pmd-java:matchesSig("javax.xml.transform.TransformerFactory#newTransformer()")]
+              ]]>
+          </value>
+      </property>
+    </properties>
+  </rule>
+
```
aaime, line 275: I see a number of SAXParserFactory.newInstance() being removed in the diff (good) but there is no rule to block them here.

modules/library/xml/src/main/java/org/geotools/xml/XMLUtils.java
```diff
@@ -125,4 +467,529 @@ private static boolean isXMLValidChar(char c) {
...
+        protected final Hints hints;
+        protected final TransformerFactory factory;
+
+        public GTTransformerFactory(Hints hints) {
+            this(TransformerFactory.newInstance(), hints);
+        }
+
+        public GTTransformerFactory(TransformerFactory factory, Hints hints) {
+            this.hints = hints != null ? hints : GeoTools.getDefaultHints();
+            this.factory = factory;
```
aaime: I see that the GTDocumentBuilderFactory and GTSAXParserFactory set the secure processing feature, but here it's not set. Is it not applicable here as well?

modules/library/xml/src/main/java/org/geotools/xml/XMLUtils.java
```diff
@@ -125,4 +467,529 @@ private static boolean isXMLValidChar(char c) {
...
+    private static class GTSAXTransformerFactory extends SAXTransformerFactory {
+        protected final Hints hints;
+        protected final SAXTransformerFactory factory;
+
+        public GTSAXTransformerFactory(Hints hints) {
+            this((SAXTransformerFactory) SAXTransformerFactory.newInstance(), hints);
+        }
+
+        public GTSAXTransformerFactory(SAXTransformerFactory factory, Hints hints) {
+            this.factory = factory;
```
aaime, line 1053: Same here, not enabling secure processing (again, there might be a reason I don't see).

modules/library/xml/src/main/java/org/geotools/xml/XMLUtils.java
```diff
@@ -125,4 +467,529 @@ private static boolean isXMLValidChar(char c) {
...
+            return config(factory.newTransformer()); // NOPMD AvoidTransform
+        }
+
+        /**
+         * Apply GeoTools configuration to the transformer, including {@code GeoTools.getEntityResolver(hints)}.
+         *
+         * @param transformer XML Transformer
+         * @return transformer
+         */
+        protected Transformer config(Transformer transformer) {
```
aaime: This method seems to be duplicated later in the same class (I guess in a inner class), it should be centralized.

modules/library/xml/src/main/java/org/geotools/xml/XMLUtils.java
```diff
@@ -125,4 +467,529 @@ private static boolean isXMLValidChar(char c) {
...
+        }
+
+        @Override
+        public TransformerHandler newTransformerHandler() throws TransformerConfigurationException {
+            return config(factory.newTransformerHandler());
+        }
+
+        @Override
+        public TemplatesHandler newTemplatesHandler() throws TransformerConfigurationException {
+            return factory.newTemplatesHandler();
```
aaime, line 1138: No configuration around this one? It is returned as-is

aaime (dismissed):
Overall looks good, see comments below

build/qa/pmd-ruleset.xml
jodygarnett: Thanks, added some more rules to cover TransformerFactory.newTransformer() and TransformerFactory.newTransformer()

modules/library/xml/src/main/java/org/geotools/xml/XMLUtils.java
jodygarnett: Yes, but ... it is a different API: sax InputSource vs xml Source.

I have a converter, that reads the that was required to insert EntityResolver in the chain for parsing sax events. Let me use that now and see if anything fails.

build/qa/pmd-ruleset.xml
jodygarnett: Let me setup rules for those, yawn so much code.

modules/library/xml/src/main/java/org/geotools/xml/XMLUtils.java
jodygarnett: Ah yes, the secure processing flags are not available here - the secure processing applies to parsing sax events which are input to the transform.

modules/library/xml/src/main/java/org/geotools/xml/XMLUtils.java
jodygarnett: Right those are both parsing content and have settings for secure processing.

The transform process happens later (on already parsed input). It has a little bit of entity resolution for filtering, but most of the work is already done by now.

modules/library/xml/src/main/java/org/geotools/xml/XMLUtils.java
jodygarnett: I will pull all the config methods up, they may be useful.

modules/library/xml/src/main/java/org/geotools/xml/XMLUtils.java
jodygarnett: Yes, I did not see anything useful to configure in TemplatesHandler / ContentHandler?

modules/library/xml/src/main/java/org/geotools/xml/XMLUtils.java
jodygarnett: Thanks, now check for both GTTransformerFactory and GTSAXTransformerFactory wrappers to avoid double wrapping.

jodygarnett:
https://github.com/geoserver/geoserver/pull/9502 feedback from @pmauduit:

> there could also be the following to catch:
> 
> * javax.xml.stream.XMLInputFactory.newFactory()
> * org.jdom2.input.SAXBuilder()


I think `XMLInputFactory.newFactory()` would be useful to catch. jdom is not used often enough to matter (perhaps it could be be phased out)

jodygarnett:
Downstream GeoWebCache branch is setup, integration.yml is testing now:
```
Checking out GeoWebCache
Branch found on geowebcache. Switching and tracking: default-entity-resolver-check
Switched to a new branch 'default-entity-resolver-check'
branch 'default-entity-resolver-check' set up to track 'origin/default-entity-resolver-check'.
Checking out GeoServer
Branch found on geoserver. Switching and tracking: default-entity-resolver-check
Switched to a new branch 'default-entity-resolver-check'
branch 'default-entity-resolver-check' set up to track 'origin/default-entity-resolver-check'.
```

jodygarnett:
I have the downstream GS3 PR testing locally, re-running the integration tests above to confirm it builds successfully here also.

jodygarnett:
@pmauduit I have introduced `XMLUtils.newSchemaFactory(string)`, but I notice that GeoServer has a slightly better implementation in the form of an `EntityResolverToLSResourceResolver` adapter which is more capable than the one included here.

PR #5662, opened 2026-05-25
[GEOT-7898] Introduce PMD rules to avoid direct use of XMLInputFactory

jodygarnett:

Addition to:

* https://github.com/geotools/geotools/pull/5639

I am trying to introduce utility methods for XMLInputFactory, but have not got it stable yet.

# Checklist

### DIFF
--- a/modules/extension/xsd/xsd-core/src/main/java/org/geotools/xsd/PullParser.java
+++ b/modules/extension/xsd/xsd-core/src/main/java/org/geotools/xsd/PullParser.java
@@ -141,7 +141,7 @@ public class PullParser {
     }
 
     XMLStreamReader createPullParser(InputStream input) {
-        XMLInputFactory factory = XMLInputFactory.newInstance();//XMLUtils.newXMLInputFactory();
+        XMLInputFactory factory = XMLUtils.newXMLInputFactory();
         factory.setProperty(XMLInputFactory.IS_NAMESPACE_AWARE, true);
         factory.setProperty(XMLInputFactory.IS_VALIDATING, false);
         factory.setProperty(XMLInputFactory.SUPPORT_DTD, false);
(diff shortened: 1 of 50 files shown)


modules/extension/xsd/xsd-core/src/main/java/org/geotools/xsd/PullParser.java
```diff
@@ -60,6 +62,7 @@ public PullParser(Configuration config, InputStream input, Object... handlerSpec
 
     public PullParser(Configuration config, InputStream input, PullParserHandler handler) {
         this.handler = handler;
+        this.handler.setEntityResolver(GeoTools.getEntityResolver(null));
```
jodygarnett, line 65: @pmauduit This is the addition that is causing problems for wfs-ng `testParseGeoServer_ArchSites_XXE`. With the change in place the xsd reference is ignored, and the parsing succeeds.

This is a regression as parsing was expected to fail for some reason.

pmauduit: The test is making use of a NullEntityResolver, I have read that in case the resolveEntity() returns null, then it is possible - depending on the implementation - that the XML parser will pass the resolution to its internal resolver and actually resolves it. Why not using the DefaultEntityResolver provided by GT here instead ?

pmauduit: The test also fails with DefaultEntityResolver.INSTANCE though, trying to figure out why

pmauduit: The GT XMLUtils class is setting the `javax.xml.stream.isReplacingEntityReferences` to `false` in the method `public static XMLInputFactory newXMLInputFactory(Hints hints)`. If we want the test to pass, if I understand the intent of the test correctly, we still need to have this property to be set to `true`.

It seems that setting it into the XmlSimpleFeatureParser class around line 123, we get back the original behaviour of the test, whether `XMLInputFactory.IS_SUPPORTING_EXTERNAL_ENTITIES` being true or false.

jodygarnett: Oh nice work! I guess we just need to decide which setting is better.

PR #5681, opened 2026-05-29
[GEOT-7898] XMLUtils interaction with AppSchemaEntityResolver

jodygarnett:

Follow on to XMLUtils integration:

* https://github.com/geotools/geotools/pull/5639

This PR is aimed at addressing regressions with app-schema GeoServer integration testing.

* https://github.com/geoserver/geoserver/pull/9502

The general problem is that app-schema is picking up the more restrictive EntityResolver when parsing local configuration files, even when parsing with a provided NullEntityResolver.INSTANCE.

# Checklist

### DIFF
--- a/modules/extension/xsd/xsd-core/src/main/java/org/geotools/xsd/DOMParser.java
+++ b/modules/extension/xsd/xsd-core/src/main/java/org/geotools/xsd/DOMParser.java
@@ -24,9 +24,7 @@ import javax.xml.transform.TransformerConfigurationException;
 import javax.xml.transform.TransformerException;
 import javax.xml.transform.dom.DOMSource;
 import javax.xml.transform.sax.SAXResult;
-import org.geotools.util.NullEntityResolver;
 import org.geotools.util.factory.GeoTools;
-import org.geotools.util.factory.Hints;
 import org.geotools.xml.XMLUtils;
 import org.geotools.xsd.impl.ParserHandler;
 import org.w3c.dom.Document;
@@ -83,9 +81,7 @@ public class DOMParser {
         // "Parser" traverses Document issuing SAX events (fake as document has already been parsed)
         Parser fake = new Parser(configuration);
         if (entityResolver != null) {
-            Hints hints = new Hints(Hints.ENTITY_RESOLVER, NullEntityResolver.INSTANCE);
-
-            fake.setEntityResolver(GeoTools.getEntityResolver(hints));
+            fake.setEntityResolver(entityResolver);
         } else {
             fake.setEntityResolver(GeoTools.getEntityResolver(null));
         }
--- a/modules/extension/xsd/xsd-core/src/main/java/org/geotools/xsd/Parser.java
+++ b/modules/extension/xsd/xsd-core/src/main/java/org/geotools/xsd/Parser.java
@@ -178,7 +178,11 @@ public class Parser {
      */
     public Object parse(InputSource source) throws IOException, SAXException, ParserConfigurationException {
         SAXParser parser = parser();
-
+        if (handler.getEntityResolver() == null) {
+            LOGGER.warning("Parsing with no entity resolver configured");
+        } else {
+            LOGGER.info("Parsing with entity resolver: " + handler.getEntityResolver());
+        }
         parser.parse(source, handler);
 
         return handler.getValue();
--- a/modules/extension/xsd/xsd-core/src/main/java/org/geotools/xsd/SchemaLocator.java
+++ b/modules/extension/xsd/xsd-core/src/main/java/org/geotools/xsd/SchemaLocator.java
@@ -21,6 +21,7 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 import org.eclipse.xsd.XSDSchema;
 import org.eclipse.xsd.util.XSDSchemaLocator;
+import org.xml.sax.EntityResolver;
 import org.xml.sax.SAXException;
 
 /**
@@ -54,6 +55,14 @@ public class SchemaLocator implements XSDSchemaLocator {
      */
     public SchemaLocator(XSD xsd) {
         this.xsd = xsd;
+    }
+
+    public void setEntityResolver(EntityResolver entityResolver) {
+        xsd.entityResolver = entityResolver;
+    }
+
+    public EntityResolver getEntityResolver() {
+        return xsd.entityResolver;
     }
 
     /**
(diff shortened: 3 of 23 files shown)


petersmythe:
Hi Jody,

I've been investigating the test failures on this PR and the downstream GeoServer breakage (CSW `AbstractCatalogStoreTest`, `RecordsTest`, `WfsRemoteStoreTest`, and the `appschema-postgis` suite). Here's what I've found:

## Root Cause

The PR propagates the parser's `entityResolver` to `XSD.entityResolver` via `SchemaLocator.setEntityResolver()`. When `XSD.buildSchema()` then calls `Schemas.parse(getSchemaLocation(), ..., entityResolver)`, the entity resolver check at line ~335 of `Schemas.parse()`:

```java
if (entityResolver != null) {
    entityResolver.resolveEntity(null, uri.toString());
}
```

...blocks `file:` URIs for internal classpath schemas. For example:

```
Entity resolution disallowed for file:/Users/runner/work/geotools/geotools/modules/extension/xsd/xsd-wfs/target/classes/org/geotools/wfs/v1_0/WFS-transaction.xsd
```

The `DefaultEntityResolver` only allows `http(s)` to OGC/W3C/INSPIRE domains and internal `jar:file`/`vfs` references — it does NOT allow plain `file:` protocol. But when running from an IDE or Maven `target/classes`, the schema locations resolve to `file:` URIs rather than `jar:file:` URIs.

## Why it affects GeoServer

In GeoServer, `CSWRecordDescriptor` has a static initializer that calls `EmfComplexFeatureReader.parse(new URL("http://schemas.opengis.net/csw/2.0.2/record.xsd"))`. The `SchemaResolver` resolves this to a local cached file. Then `Schemas.findSchemas()` checks the entity resolver against the resolved local path → blocked → `ExceptionInInitializerError`.

## Proposed Fix

The issue is that `XSD.entityResolver` was previously always `null`, and `Schemas.parse()` skipped the entity resolver check when `null`. The PR now sets it, but internal/classpath schemas don't need this check — they're trusted by definition.

**Option A (minimal):** In `XSD.buildSchema()`, don't pass the entity resolver to `Schemas.parse()` for internal schemas. The schema location comes from `getSchemaLocation()` which is a known trusted location:

```java
// Internal schemas from getSchemaLocation() are trusted - no entity resolution check needed
return Schemas.parse(getSchemaLocation(), locators, resolvers, Collections.emptyList(), null);
```

And remove the `schemaLocator.setEntityResolver(entityResolver)` calls from `ParserHandler.loadSchemas()` and `XSD.buildSchema()`, since they propagate a restrictive resolver to trusted internal schema loading.

**Option B (more targeted):** Keep the propagation but make `DefaultEntityResolver` recognize `file:` URIs ending in `.xsd` from known classpath locations (matching the `INTERNAL_URIS` pattern for `jar:file`). This would mean extending the pattern to also allow `file:...target/classes/...` paths. However this is fragile and environment-dependent.

**Option C (recommended):** The entity resolver check in `Schemas.parse()` should only apply when the schema is being loaded from an external/untrusted source. Since `XSD.buildSchema()` always loads from `getSchemaLocation()` (which is a classpath resource), it should pass `null` as the entity resolver. The entity resolver should only be applied in `Schemas.findSchemas()` for the permission-check use case, and in the `Parser` for actual XML parsing.

In short: revert the `locator.setEntityResolver(entityResolver)` additions in `ParserHandler` and `XSD.buildSchema()`, and leave `XSD.entityResolver` as `null` (its previous default). The entity resolver is already correctly applied at the SAX parser level for actual XML entity resolution during parsing.

## Additional note

I also found that the JAXP `accessExternalDTD` error message is **localized** by the JVM — the Italian locale CI build produces the message in Italian. Any assertions checking for specific English error text will fail on locale builds. I've addressed this in GeoServer PR [#9538](https://github.com/geoserver/geoserver/pull/9538) using locale-independent substrings.

Let me know if you'd like me to prepare a patch for any of the above options.

petersmythe:
## Additional issue: `XmlStreamGeometryReader` test failures on `main`

There's another GEOT-7898 regression affecting tests on `main` (visible in PRs like [#5682](https://github.com/geotools/geotools/pull/5682) which don't touch any related code):

```
java.lang.IllegalStateException: XMLStreamReader allows DTDs but XmlStreamGeometryReader is not configured to allow unsafe XML
    at org.geotools.gml.stream.XmlStreamGeometryReader.checkUnsafeXML(XmlStreamGeometryReader.java:147)
```

**Root cause:** In `XMLUtils.newXMLInputFactory()`, the access string parsing uses `access.contains("file")` to determine if internal file access is allowed. `DefaultEntityResolver.getAccess()` returns `"http,jar:file,jar:nested,vfs"` — and `"http,jar:file,jar:nested,vfs".contains("file")` is **true** because "jar:**file**" contains the substring "file".

This causes `SUPPORT_DTD` to be set to `true`:

```java
boolean internal = access.contains("all")
        || access.contains("file")   // ← matches "jar:file" substring!
        || access.contains("jar")
        || access.contains("vfs");
// ...
factory.setProperty(XMLInputFactory.SUPPORT_DTD, internal || external);  // true
```

Then `XmlStreamGeometryReader.checkUnsafeXML()` sees `SUPPORT_DTD=true` and throws because `isUnsafeXMLAllowed()` defaults to false.

**Fix:** The access string parsing should use word-boundary matching rather than substring contains. For example:

```java
// Split on comma and check for exact protocol matches
Set<String> protocols = Set.of(access.split(","));
boolean internal = protocols.contains("all")
        || protocols.contains("file")
        || protocols.stream().anyMatch(p -> p.startsWith("jar"))
        || protocols.contains("vfs");
```

Or alternatively, `DefaultEntityResolver.getAccess()` should not include "file" as a substring of "jar:file" — perhaps use a different separator or format like `"http|jar:file|jar:nested|vfs"` with pipe-delimited parsing.

jodygarnett:
Thanks Peter that agrees with my own assessment, I had tried asking XSD to use NullEntityResolver.INSTANCE locally (since the comments indicated it was just working against local files) however this resulted in a few XXE tests failing in app-schema.

I am going to go back to using the passed in configured for XSD, and will need to ask the GeoServer app schema tests to use DevModeEntityResolver.INSTANCE in order to pass.

Right now this is very labour intensive getting things to work for app-schema tests in geoserver, and then doubling back to find what is broken in geotools, and then cycle through again. 

Longer turn one thing would help is if I can determine which standards need DTD, and which do not. Presently I think just GML2 needs needs DTD.

PR #5682, opened 2026-05-30
[GEOT-7913] Sort namespace declarations for deterministic XML output

petersmythe:
[GEOT-7913](https://osgeo-org.atlassian.net/browse/GEOT-7913) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)  

https://osgeo-org.atlassian.net/browse/GEOT-7913

Sort xmlns:prefix attributes alphabetically in TransformerBase to ensure deterministic XML output. Previously, namespace declarations were emitted in HashMap iteration order which varies across JVM restarts, causing spurious diffs when SLD files are re-serialized.

The fix collects all namespace prefix declarations into a TreeMap before adding them to the output attributes, and sorts startPrefixMapping calls in ContentHandlerFilter for consistent JAXP Transformer output.

Note: The JAXP Transformer always places the element's own namespace prefix first (e.g. 'sld' for sld:StyledLayerDescriptor), which is unavoidable. The remaining prefixes are now guaranteed to be in alphabetical order.

# Checklist

### DIFF
--- a/modules/library/xml/src/main/java/org/geotools/xml/transform/TransformerBase.java
+++ b/modules/library/xml/src/main/java/org/geotools/xml/transform/TransformerBase.java
@@ -22,12 +22,14 @@ import java.nio.charset.StandardCharsets;
 import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Deque;
+import java.util.Enumeration;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.TreeMap;
 import javax.xml.XMLConstants;
 import javax.xml.transform.OutputKeys;
 import javax.xml.transform.Source;
@@ -415,14 +417,17 @@ public abstract class TransformerBase {
                 throws SAXException {
             if (namespaceDecls != null) {
 
-                // call startPrefixMapping for any namespace prefixes
+                // Collect namespace prefixes into a TreeMap for deterministic output
+                Map<String, String> nsPrefixes = new TreeMap<>();
                 for (int i = 0; i < namespaceDecls.getLength(); i++) {
                     final String nsQName = namespaceDecls.getQName(i);
                     if (nsQName != null && nsQName.startsWith("xmlns:")) {
-                        final String nsPrefix = nsQName.substring(6);
-                        final String nsUri = namespaceDecls.getValue(i);
-                        startPrefixMapping(nsPrefix, nsUri);
+                        nsPrefixes.put(nsQName.substring(6), namespaceDecls.getValue(i));
                     }
+                }
+                // call startPrefixMapping in sorted order
+                for (Map.Entry<String, String> entry : nsPrefixes.entrySet()) {
+                    startPrefixMapping(entry.getKey(), entry.getValue());
                 }
 
                 for (int i = 0, ii = atts.getLength(); i < ii; i++) {
@@ -786,7 +791,7 @@ public abstract class TransformerBase {
         /** Utility method to copy namespace declarations from "sub" translators into this ns support... */
         protected void addNamespaceDeclarations(TranslatorSupport trans) {
             NamespaceSupport additional = trans.getNamespaceSupport();
-            java.util.Enumeration declared = additional.getDeclaredPrefixes();
+            Enumeration declared = additional.getDeclaredPrefixes();
             while (declared.hasMoreElements()) {
                 String prefix1 = declared.nextElement().toString();
                 nsSupport.declarePrefix(prefix1, additional.getURI(prefix1));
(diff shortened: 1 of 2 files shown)


petersmythe:
Waiting on https://github.com/geotools/geotools/pull/5681 and https://github.com/geoserver/geoserver/pull/9538 to fix the broken tests

aaime:
I do understand wanting a deterministic order for XML files saved in version control, to avoid spurious changes, but this seems to affect stuff like capabilities document generation... is order of any relevance in that case? What's the actual problem you're trying to address?

petersmythe:
> deterministic order for XML files saved in version control, to avoid spurious changes

Exactly this. See also geoserver/geoserver#9537 which sorts Set elements at write time — e.g. 2 styles in a layer are always written in alphabetical order for clean git diffs, while for consumers of the XML the order of set siblings is irrelevant (within XSD schema constraints).

The `xmlns` namespace declaration order is controlled entirely within GeoTools' TransformerBase.java, but manifests in the same GeoServer config files. Hence the need for this GeoTools PR to fix it at the source.

```
+++ b/data/styles/test.sld
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="UTF-8"?><sld:StyledLayerDescriptor xmlns="http://www.opengis.net/sld" xmlns:sld="http://www.opengis.net/sld" xmlns:gml="http://www.opengis.net/gml" xmlns:ogc="http://www.opengis.net/ogc" version="1.0.0">
+<?xml version="1.0" encoding="UTF-8"?><sld:StyledLayerDescriptor xmlns:sld="http://www.opengis.net/sld" xmlns="http://www.opengis.net/sld" xmlns:gml="http://www.opengis.net/gml" xmlns:ogc="http://www.opengis.net/ogc" version="1.0.0">
```

Is there a better approach you'd suggest?

petersmythe:
> Is there a better approach you'd suggest?

Researching a bit myself, can we do `LinkedHashMap` similar to https://github.com/geoserver/geoserver/pull/9537:

> The `NamespaceSupport` class is from the JDK (`org.xml.sax.helpers.NamespaceSupport`). It's not a GeoTools class — it's part of the Java standard library. You can't change its internal implementation.
> 
> That's the key reason: **`NamespaceSupport` is a JDK class** (`org.xml.sax.helpers.NamespaceSupport`). Its `getPrefixes()` method returns an `Enumeration` backed by a `Hashtable` internally. GeoTools doesn't control this class — it's part of `java.xml`. You can't swap its internals to `LinkedHashMap`.
> 
> Even if GeoTools wrapped it with a custom implementation, the insertion order still wouldn't be meaningful here. Namespaces are declared by various `Translator` subclasses in arbitrary order depending on which elements they encounter first. The "first declared" order would vary depending on the document content, not just the schema. Alphabetical sorting gives a truly stable, content-independent order.
> 
> So the answer to "why not fix NamespaceSupport to use LinkedHashMap" is:
> 1. It's a JDK class — can't modify it
> 2. Even if you could, insertion order isn't stable (depends on document content and code path execution order)
> 3. Alphabetical sort at write time is simpler, more predictable, and produces the cleanest diffs

modules/library/xml/src/main/java/org/geotools/xml/transform/TransformerBase.java
```diff
@@ -415,15 +417,18 @@ public void startElement(String namespaceURI, String localName, String qName, At
                 throws SAXException {
             if (namespaceDecls != null) {
 
-                // call startPrefixMapping for any namespace prefixes
+                // Collect namespace prefixes into a TreeMap for deterministic output
+                Map<String, String> nsPrefixes = new TreeMap<>();
```
jodygarnett, line 421: You could use `java.util.LinkedHashMap` if you just wish to preserve order rather than sort?

petersmythe: @jodygarnett TreeMap is used at L421 because the input source (namespaceDecls - an Attributes object) has no guaranteed order. It comes from the SAX Attributes interface which is populated by the XML transformer, backed by a Hashtable inside NamespaceSupport (which we can't modify - see  https://github.com/geotools/geotools/pull/5682#issuecomment-4592897977 above. Using LinkedHashMap would just preserve the non-deterministic iteration order of that input - which is exactly the bug/improvement I'm trying to fix/make.

TreeMap sorts by key (the prefix), giving alphabetical output regardless of what order the attributes arrived in. LinkedHashMap preserves insertion order, but insertion order here IS the random order - so it would solve nothing.

So yes, this PR does introduce the possibility of existing/hand-written XML files changing.  I don't see a way around that.

pmauduit (approved):
Changes are looking good to me, but I don't get where/why the output should be deterministic, I presume comparing generated documents as strings (e.g. in tests) ? If the produced XML documents are equivalent, it should not matter ?

Having the library generating the same output sounds as a good move anyway.

Ok, after having read some linked tickets on JIRA, I have a bit more context, and I agree on the fact that it could also make diffing (by a human reader) easier.

petersmythe:
Thank you for your feedback @pmauduit 

> Changes are looking good to me, but I don't get where/why the output should be deterministic, I presume comparing generated documents as strings (e.g. in tests) ? 

Yes, comparing "documents" (in GeoServer's case, the XML config files in the data directory), and specifically using Git to compare the config files.  That is precisely what https://github.com/geoserver/geoserver/pull/9537 does (no longer sorted alphabetically, but using a LinkedHashMap to retain the previous order) - except that GeoServer does not have any way to control the order of the `xmlns` namespace declarations, thus the change has to be done inside GeoTools (this PR).

For general GeoTools use, where the XML generated is parsed by a machine, this deterministic ordering is over-constrained, yes, but that does not matter.  For document comparison, on the other hand, it does matter:

Taking the example from above (and adding clarifying whitespace):

```
+++ b/data/styles/test.sld
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="UTF-8"?><sld:StyledLayerDescriptor 
xmlns="http://www.opengis.net/sld"        <---
xmlns:sld="http://www.opengis.net/sld"    <---
xmlns:gml="http://www.opengis.net/gml" 
xmlns:ogc="http://www.opengis.net/ogc" version="1.0.0">

+<?xml version="1.0" encoding="UTF-8"?><sld:StyledLayerDescriptor 
xmlns:sld="http://www.opengis.net/sld"   <---- was #2
xmlns="http://www.opengis.net/sld"       <---- was #1
xmlns:gml="http://www.opengis.net/gml" 
xmlns:ogc="http://www.opengis.net/ogc" version="1.0.0">
```

Without a defined order, the XML files will show random "changes" when in fact there are none (other than order).

PR #5695, opened 2026-06-08
NIOUtilities add a MemorySegement based cleanup for JDK22+

gwalbran:
The existing cleanupAfterJdk8 outputs warnings when run on JDKs22+.

It looks like MemorySegment is the preferred way to do this kind of cleanup on more modern JDKs.

Related to [GEOT-7914](https://osgeo-org.atlassian.net/browse/GEOT-7914)

  

# Checklist

### DIFF
--- a/modules/library/metadata/src/main/java/org/geotools/util/NIOUtilities.java
+++ b/modules/library/metadata/src/main/java/org/geotools/util/NIOUtilities.java
@@ -18,13 +18,16 @@ package org.geotools.util;
 
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
+import java.nio.Buffer;
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Queue;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Function;
 import java.util.logging.Level;
 import org.geotools.util.logging.Logging;
 
@@ -58,9 +61,32 @@ public final class NIOUtilities {
     /** Wheter direct buffers usage is enabled, or not */
     private static boolean directBuffersEnabled = true;
 
+    /** A function to implement {@link #clean(java.nio.ByteBuffer) }. */
+    private static final Function<ByteBuffer, Boolean> CLEANUP_FUNCTION;
+
     static {
         String directBuffers = System.getProperty("geotools.nioutilities.direct", "true");
         directBuffersEnabled = "TRUE".equalsIgnoreCase(directBuffers);
+
+        // set the cleanup function based on the system property.
+        String cleanup = System.getProperty("geotools.nioutilities.cleanup", "");
+        if (cleanup.equalsIgnoreCase("memorysegment")) {
+            CLEANUP_FUNCTION = memorySegmentCleanup().orElseThrow();
+        } else if (cleanup.equalsIgnoreCase("unsafe")) {
+            CLEANUP_FUNCTION = cleanupAfterJdk8(true).orElseThrow();
+        } else if (cleanup.equalsIgnoreCase("noop")) {
+            CLEANUP_FUNCTION = NIOUtilities::noopClean;
+        } else {
+            // user hasn't made a choice (or an unknown choice) so we decide what's best
+            // cleanupAfterJdk8 is the best clean up available currently but may trigger deprecated API warnings
+            // on newer JVMs. If you need to stop these warnings more than you need the better cleanup then
+            // setting skipUnsafeVersionCheck to false will prevent cleanupAfterJdk8 being used on these JVMs
+            CLEANUP_FUNCTION = cleanupAfterJdk8("TRUE"
+                            .equalsIgnoreCase(
+                                    System.getProperty("geotools.nioutilities.skipUnsafeVersionCheck", "true")))
+                    .or(() -> memorySegmentCleanup())
+                    .orElse(NIOUtilities::noopClean);
+        }
     }
 
     /** Wheter direct buffers are used, or not (defaults to true) */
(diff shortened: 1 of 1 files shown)


modules/library/metadata/src/main/java/org/geotools/util/NIOUtilities.java
mprins, line 1: there is a formatting error:

```
Error:  Failed to execute goal com.diffplug.spotless:spotless-maven-plugin:3.0.0:check (default) on project gt-metadata: The following files had format violations:
Error:      src/main/java/org/geotools/util/NIOUtilities.java
Error:          @@ -180,14 +180,12 @@
Error:           ········if·(buffer·==·null·||·!buffer.isDirect())·{
Error:           ············return·true;
Error:           ········}
Error:          -········return·memorySegmentmentCleanup()
Error:          -············.orElse((buf)·->·cleanupAfterJdk8(buf))
Error:          -············.apply(buffer);
Error:          +········return·memorySegmentmentCleanup().orElse((buf)·->·cleanupAfterJdk8(buf)).apply(buffer);
Error:           ····}
Error:           
Error:           ····/**
Error:          -·····*·Returns·a·cleanup·function·that·uses·<code>java.lang.foreign.MemorySegment</code>·which·was·added
Error:          -·····*·in·JDK22.·This·is·the·preferred·way·to·signal·to·the·JVM·that·memory·can·be·freed.
Error:          +·····*·Returns·a·cleanup·function·that·uses·<code>java.lang.foreign.MemorySegment</code>·which·was·added·in·JDK22.·This
Error:          +·····*·is·the·preferred·way·to·signal·to·the·JVM·that·memory·can·be·freed.
Error:           ·····*/
Error:           ····private·static·Optional<Function<ByteBuffer,·Boolean>>·memorySegmentmentCleanup()·{
Error:           ········try·{
Error:  Run 'mvn spotless:apply' to fix these violations.
Error:  -> [Help 1]
Error:  
Error:  To see the full stack trace of the errors, re-run Maven with the -e switch.
Error:  Re-run Maven using the -X switch to enable full debug logging.
Error:  
Error:  For more information about the errors and possible solutions, please read the following articles:
Error:  [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoExecutionException
Error:  
Error:  After correcting the problems, you can resume the build with the command
Error:    mvn <args> -rf :gt-metadata
```

mprins (changes requested):
Note that the current baseline for GeoTools 35.x / `main` branch is Java 17: https://github.com/geotools/geotools/blob/eed497f8c6f91beed43b408f0c994eef75768286/pom.xml#L951-L962

So these warnings? may be expected.

Please add a Jira ticket with the warnings

modules/library/metadata/src/main/java/org/geotools/util/NIOUtilities.java
```diff
@@ -177,7 +180,44 @@ public static boolean clean(final ByteBuffer buffer) {
         if (buffer == null || !buffer.isDirect()) {
             return true;
         }
-        return cleanupAfterJdk8(buffer);
+        return memorySegmentmentCleanup().orElse((buf) -> cleanupAfterJdk8(buf)).apply(buffer);
```
aaime: I think you tried to fall back on jdk8 if the segment approach returned false here, but FALSE is not null, won't work. Why is it important? Because a buffer can be direct yet not mapped, and the memory segment approach won't clean it in that case.

modules/library/metadata/src/main/java/org/geotools/util/NIOUtilities.java
```diff
@@ -177,7 +180,44 @@ public static boolean clean(final ByteBuffer buffer) {
...
+
+    /**
+     * Returns a cleanup function that uses <code>java.lang.foreign.MemorySegment</code> which was added in JDK22. This
+     * is the preferred way to signal to the JVM that memory can be freed.
+     */
+    private static Optional<Function<ByteBuffer, Boolean>> memorySegmentmentCleanup() {
+        try {
+            // MemorySegment is a new API added in JDK22. If this class cannot be found then the class
+            // loader will throw an exception which will be caught below and the empty optional returned.
+            final Class<?> memorySegmentClass = Class.forName("java.lang.foreign.MemorySegment");
```
aaime, line 222: All this reflection business is repeated over and over... can you do it in the class initialization instead?
The same happens in the jdk8+ path.. very old code, could use improvement for sure.

aaime (changes requested):
The code as written will leave out direct yet non-mapped buffers, not cleaning them. Performance can be optimzied a bit.

modules/library/metadata/src/main/java/org/geotools/util/NIOUtilities.java
gwalbran: So you're saying that cleaning up direct non-mapped buffers would be needed as well?

It looks like this can be done with an [Arena](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/foreign/Arena.html) in modern JDKs but that only let's you release all the buffers allocated from the arena at once. So maybe that doesn't fit with the use case here.

In what conditions is this cleanup actually needed though? I guess a server type system could need this. But what about on a single user system.

modules/library/metadata/src/main/java/org/geotools/util/NIOUtilities.java
```diff
@@ -177,28 +197,97 @@ public static boolean clean(final ByteBuffer buffer) {
...
+    /**
+     * Returns a cleanup function that uses <code>java.lang.foreign.MemorySegment</code> which was added in JDK22.
+     *
+     * <p>Note: memory segment clean up only works for direct mapped buffers. Other buffers are not cleaned at all.
+     *
+     * <p>This cleanup is better than nothing but not as good as {@link #cleanupAfterJdk8(boolean) }. In the future it
+     * might be possible to allocate direct non-mapped buffers via an Arena which can be cleaned up by closing the Arena
+     * that the buffers were allocated from.
+     */
+    private static Optional<Function<ByteBuffer, Boolean>> memorySegmentmentCleanup() {
```
aaime: segmentMent? 😆

modules/library/metadata/src/main/java/org/geotools/util/NIOUtilities.java
aaime: A single-user system is of secondary interest, for a simple practical reason: today we have GeoTools perform frequent releases and get this many changes/fixes/improvements only because GeoServer is feeding into it everything that can be considered library code. So, a code change targeting single user/batch systems is welcomed, but must not badly affect a server system... this class exists specifically to support better GeoServer memory management under heavy load.

Currently the shapefile datastore uses a lot of direct, non-mapped allocation, so if one were to choose memory segment cleanup, that would cause memory issues under pressure. Ideally, the code should perform arena allocation when on jdk 24+, but given the codebase is built for java 17, we'd have to use reflection for allocation too.

aaime (changes requested):
Sorry for the late reply, we have had FOSS4G Europe in the middle and the two weeks before and after it have been pretty intense.

aaime:
Merged, marked for backport on 35.x too.

PR #5713, opened 2026-06-18
[GEOT-7919]: duckdb upgrade to 1.5.3.0

dromagnoli:
[GEOT-7919](https://osgeo-org.atlassian.net/browse/GEOT-7919) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)  

Update to newer DuckDB to fix an issue related to a false positive on ETAG check on DuckDB Driver when reading from geoparquet on S3.

# Checklist

### DIFF
--- a/modules/unsupported/duckdb/src/main/java/org/geotools/data/duckdb/DuckDBDialect.java
+++ b/modules/unsupported/duckdb/src/main/java/org/geotools/data/duckdb/DuckDBDialect.java
@@ -131,7 +131,10 @@ public class DuckDBDialect extends BasicSQLDialect {
     //    }
 
     public List<String> getDatabaseInitSql() {
-        return List.of("install spatial", "load spatial");
+        // DuckDB 1.5 warns unless users explicitly choose the old or new axis-order behavior for affected
+        // spatial functions. The dialect does not currently emit those functions, but opt into X/Y order
+        // now to align with common GIS longitude/latitude behavior and DuckDB's planned future default.
+        return List.of("install spatial", "load spatial", "SET geometry_always_xy = true");
     }
 
     @Override
--- a/modules/unsupported/duckdb/src/main/java/org/geotools/data/duckdb/DuckDBDialect.java
+++ b/modules/unsupported/duckdb/src/main/java/org/geotools/data/duckdb/DuckDBDialect.java
@@ -131,7 +131,10 @@ public class DuckDBDialect extends BasicSQLDialect {
     //    }
 
     public List<String> getDatabaseInitSql() {
-        return List.of("install spatial", "load spatial");
+        // DuckDB 1.5 warns unless users explicitly choose the old or new axis-order behavior for affected
+        // spatial functions. The dialect does not currently emit those functions, but opt into X/Y order
+        // now to align with common GIS longitude/latitude behavior and DuckDB's planned future default.
+        return List.of("install spatial", "load spatial", "SET geometry_always_xy = true");
     }
 
     @Override
@@ -198,7 +201,7 @@ public class DuckDBDialect extends BasicSQLDialect {
         }
 
         // Check if it's a geometry column
-        if ("GEOMETRY".equalsIgnoreCase(typeName)) {
+        if (typeName != null && typeName.toUpperCase().startsWith("GEOMETRY")) {
             return Geometry.class;
         }
 
--- a/modules/unsupported/duckdb/src/main/java/org/geotools/data/duckdb/datasource/DuckdbConnectionFactory.java
+++ b/modules/unsupported/duckdb/src/main/java/org/geotools/data/duckdb/datasource/DuckdbConnectionFactory.java
@@ -20,6 +20,8 @@ import static java.util.Objects.requireNonNull;
 
 import java.sql.Connection;
 import java.sql.Driver;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
 import java.util.List;
@@ -153,12 +155,35 @@ class DuckdbConnectionFactory extends DriverConnectionFactory {
                 if (!allowInstall && isInstallStatement(sql)) {
                     continue;
                 }
-                stmt.execute(sql);
+                executeInitSql(conn, stmt, sql);
             }
         }
     }
 
+    private void executeInitSql(Connection conn, Statement stmt, String sql) throws SQLException {
+        if (isGeometryAlwaysXySetStatement(sql) && !isDuckDbSettingAvailable(conn, "geometry_always_xy")) {
+            return;
+        }
+        stmt.execute(sql);
+    }
+
     private boolean isInstallStatement(String sql) {
         return sql != null && sql.trim().regionMatches(true, 0, "install", 0, "install".length());
+    }
+
+    private boolean isDuckDbSettingAvailable(Connection conn, String settingName) throws SQLException {
+        try (PreparedStatement ps =
+                conn.prepareStatement("SELECT 1 FROM duckdb_settings() WHERE lower(name) = lower(?) LIMIT 1")) {
+            ps.setString(1, settingName);
+            try (ResultSet rs = ps.executeQuery()) {
+                return rs.next();
+            }
+        }
+    }
+
+    static boolean isGeometryAlwaysXySetStatement(String sql) {
+        String trimmed = sql == null ? "" : sql.trim();
+        return trimmed.toLowerCase(java.util.Locale.ROOT)
+                .matches("set\\s+geometry_always_xy\\s*=\\s*(true|false)\\s*;?");
     }
 }
--- a/modules/unsupported/geoparquet/src/main/java/org/geotools/data/geoparquet/GeoParquetDialect.java
+++ b/modules/unsupported/geoparquet/src/main/java/org/geotools/data/geoparquet/GeoParquetDialect.java
@@ -33,6 +33,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Function;
 import java.util.logging.Level;
@@ -431,7 +432,9 @@ public class GeoParquetDialect extends DuckDBDialect {
             if (!rs.next()) throw new RuntimeException("Could not compute bounds from bbox column, no result found");
             Geometry fullBounds = parseWKB(rs.getBlob(1));
             GeometryDescriptor geometryDescriptor = featureType.getGeometryDescriptor();
-            CoordinateReferenceSystem crs = geometryDescriptor.getCoordinateReferenceSystem();
+            CoordinateReferenceSystem crs = geometryDescriptor != null
+                    ? geometryDescriptor.getCoordinateReferenceSystem()
+                    : getGeoparquetMetadata(featureType.getTypeName(), cx).getCrs();
             return new ReferencedEnvelope(fullBounds.getEnvelopeInternal(), crs);
         }
     }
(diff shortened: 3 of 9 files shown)


modules/unsupported/duckdb/src/main/java/org/geotools/data/duckdb/datasource/DuckdbConnectionFactory.java
```diff
@@ -153,12 +154,46 @@ private void initDatabase(DuckDBConnection conn, boolean allowInstall) throws SQ
...
+    }
+
+    private void executeInitSql(Statement stmt, String sql) throws SQLException {
+        try {
+            if (isGeometryAlwaysXySetStatement(sql) && !isDuckDbSettingAvailable(stmt, "geometry_always_xy")) {
+                return;
+            }
+            stmt.execute(sql);
+        } catch (SQLException e) {
+            if (!isUnsupportedGeometryAlwaysXy(sql, e)) {
```
aaime: What if this returned true instead, we would just eat the exception? That does not seem ok.

modules/unsupported/duckdb/src/main/java/org/geotools/data/duckdb/datasource/DuckdbConnectionFactory.java
```diff
@@ -153,12 +154,46 @@ private void initDatabase(DuckDBConnection conn, boolean allowInstall) throws SQ
...
             }
         }
     }
 
     private boolean isInstallStatement(String sql) {
         return sql != null && sql.trim().regionMatches(true, 0, "install", 0, "install".length());
     }
+
+    private boolean isDuckDbSettingAvailable(Statement stmt, String settingName) throws SQLException {
+        try (ResultSet rs = stmt.executeQuery(
```
aaime: I would use prepared statements if possible here.

modules/unsupported/geoparquet/src/main/java/org/geotools/data/geoparquet/GeoParquetDialect.java
```diff
@@ -633,17 +646,29 @@ private SimpleFeatureType buildTypeNarrowedGeometryFeatureType(SimpleFeatureType
      * @param orig The original geometry descriptor with generic type
      * @return A new geometry descriptor with the specific geometry type, or the original if unchanged
      */
-    private GeometryDescriptor buildGeometryDescriptorOverride(String typeName, GeometryDescriptor orig) {
-        String geometryAttirbute = orig.getLocalName();
+    private AttributeDescriptor buildGeometryDescriptorOverride(
```
aaime: If it's building a GeometryDescriptor, why go back to AttributeDescriptor?

modules/unsupported/geoparquet/src/main/java/org/geotools/data/geoparquet/GeoParquetDialect.java
```diff
@@ -633,17 +646,29 @@ private SimpleFeatureType buildTypeNarrowedGeometryFeatureType(SimpleFeatureType
...
+                && narrowedGeomType.equals(geometryOrig.getType().getBinding())) {
+            return geometryOrig;
         }
 
         AttributeTypeBuilder builder = new AttributeTypeBuilder();
-        builder.init(orig);
+        if (orig != null) {
+            builder.init(orig);
+        } else {
+            builder.setName(geometryAttirbute);
```
aaime: Attirbute?

modules/unsupported/geoparquet/src/main/java/org/geotools/data/geoparquet/GeoParquetDialect.java
```diff
@@ -605,11 +607,22 @@ private SimpleFeatureType buildTypeNarrowedGeometryFeatureType(SimpleFeatureType
...
-                .forEach(overriding -> builder.set(overriding.getLocalName(), overriding));
+        GeoparquetDatasetMetadata metadata;
+        try {
+            metadata = getGeoparquetMetadata(schema.getTypeName());
+        } catch (IOException e) {
+            throw new UncheckedIOException(e);
+        }
+        metadata.getColumnNames().forEach(geometryAttribute -> {
+            AttributeDescriptor descriptor = schema.getDescriptor(geometryAttribute);
+            AttributeDescriptor overriding =
```
aaime: So there is always a geometry... what if the parquet file has none?

modules/unsupported/geoparquet/src/main/java/org/geotools/data/geoparquet/GeoParquetDialect.java
```diff
@@ -605,11 +607,22 @@ private SimpleFeatureType buildTypeNarrowedGeometryFeatureType(SimpleFeatureType
...
         builder.init(schema);
 
-        schema.getAttributeDescriptors().stream()
-                .filter(GeometryDescriptor.class::isInstance)
-                .map(GeometryDescriptor.class::cast)
-                .map(d -> buildGeometryDescriptorOverride(schema.getTypeName(), d))
-                .forEach(overriding -> builder.set(overriding.getLocalName(), overriding));
+        GeoparquetDatasetMetadata metadata;
+        try {
+            metadata = getGeoparquetMetadata(schema.getTypeName());
```
aaime: Is the metadata always available?

aaime (commented):
The change appears to be doing more than just an upgrade, can you describe that somewhere (Jira title/description).
I have some doubts about the code changes, see below.

modules/unsupported/duckdb/src/main/java/org/geotools/data/duckdb/datasource/DuckdbConnectionFactory.java
dromagnoli: removing that and keeping it part of the duckdb_settings check.

modules/unsupported/geoparquet/src/main/java/org/geotools/data/geoparquet/GeoParquetDialect.java
dromagnoli: I'll add a check so that if metadata is empty it will return the schema unchanged.

dromagnoli:
Feedbacks addressed and JIRA Description updated about the geometry topic

modules/unsupported/geoparquet/src/main/java/org/geotools/data/geoparquet/GeoParquetDialect.java
dromagnoli: We are returning the schema unchanged

modules/unsupported/duckdb/src/main/java/org/geotools/data/duckdb/datasource/DuckdbConnectionFactory.java
```diff
@@ -153,12 +155,35 @@ private void initDatabase(DuckDBConnection conn, boolean allowInstall) throws SQ
...
         }
     }
 
+    private void executeInitSql(Connection conn, Statement stmt, String sql) throws SQLException {
+        if (isGeometryAlwaysXySetStatement(sql) && !isDuckDbSettingAvailable(conn, "geometry_always_xy")) {
+            return;
+        }
+        stmt.execute(sql);
+    }
+
```
aaime, line 169: What if this returned true instead, we would just eat the exception? That does not seem ok.

aaime:
@groldan do you have some time to look at this one? It's been sitting here for a month. It's a dependency upgrade, but it's a tricky one because DuckDb behavior changed in a backwards incompatible way.

modules/unsupported/duckdb/src/main/java/org/geotools/data/duckdb/datasource/DuckdbConnectionFactory.java
dromagnoli: what do you mean?

dromagnoli:
Hi @groldan. Sorry to bother. Did you manage to have some time to check that?

modules/unsupported/duckdb/src/main/java/org/geotools/data/duckdb/datasource/DuckdbConnectionFactory.java
```diff
@@ -109,6 +112,7 @@ private DuckDBConnection initializeSentinelIfNeeded() throws SQLException {
     private DuckDBConnection createAndInitializeSentinel() throws SQLException {
         DuckDBConnection created = (DuckDBConnection) super.createConnection();
         try {
+            geometryAlwaysXyAvailable = isDuckDbSettingAvailable(created, "geometry_always_xy");
```
groldan: looks like this is always `false`. Runs before `initDatabase` invokes `load spatial` and `geometry_always_xy` is contributed by the spatial extension, hence `initDatabase()->executeInitSql()` will never see `geometryAlwaysXyAvailable == true` ?

modules/unsupported/duckdb/src/main/java/org/geotools/data/duckdb/datasource/DuckdbConnectionFactory.java
```diff
@@ -153,12 +157,35 @@ private void initDatabase(DuckDBConnection conn, boolean allowInstall) throws SQ
                 if (!allowInstall && isInstallStatement(sql)) {
                     continue;
                 }
-                stmt.execute(sql);
+                executeInitSql(stmt, sql);
             }
         }
     }
 
+    private void executeInitSql(Statement stmt, String sql) throws SQLException {
```
groldan: it seems to me that if you change this method to something like the following you could remove the `geometryAlwaysXyAvailable` instance variable altogether and would automatically fix the bug with geometryAlwaysXyAvailable being always false (comment above)

```java
try {
 stmt.execute(sql);
} catch (SQLException e) {
   if (!isGeometryAlwaysXySetStatement(sql)) {
       throw e;
    }
    LOGGER.log(Level.FINE, "geometry_always_xy not supported by this DuckDB version, skipping", e);
}
```

groldan (changes requested):
commented on what I found on `DuckdbConnectionFactory`, will need a second round to check the geoparquet changes.

verified the claim with this quick test that prints
```
probe on a fresh connection, before any init SQL runs (what createAndInitializeSentinel() sees): available=false
DuckDB version: v1.5.3
probe after the init SQL ran 'load spatial': available=true
duckdb_settings() value for geometry_always_xy = null
```
and shows SET geometry_always_xy = true init statement is silently skipped

```java
    @Test
    public void testGeometryAlwaysXyIsAppliedWhenSpatialIsLoaded() throws Exception {
        try (Connection fresh = new DuckDBDriver().connect("jdbc:duckdb:", new Properties())) {
            System.out.println("probe on a fresh connection, before any init SQL runs "
                    + "(what createAndInitializeSentinel() sees): available="
                    + isGeometryAlwaysXyInDuckdbSettings(fresh));
        }

        Path database = createDatabasePath();
        DuckdbConnectionFactory factory = new DuckdbConnectionFactory(
                new DuckDBDriver(),
                "jdbc:duckdb:" + database.toAbsolutePath(),
                new Properties(),
                List.of("install spatial", "load spatial", "SET geometry_always_xy = true"));

        try (Connection connection = factory.createConnection()) {
            System.out.println("DuckDB version: " + connection.getMetaData().getDatabaseProductVersion());
            System.out.println("probe after the init SQL ran 'load spatial': available="
                    + isGeometryAlwaysXyInDuckdbSettings(connection));
            String value = geometryAlwaysXyValue(connection);
            System.out.println("duckdb_settings() value for geometry_always_xy = " + value);
            assertEquals(
                    "SET geometry_always_xy = true was silently skipped: the availability probe ran "
                            + "before 'load spatial', when duckdb_settings() does not list the setting yet",
                    "true",
                    value);
        } finally {
            factory.close();
        }
    }

    private boolean isGeometryAlwaysXyInDuckdbSettings(Connection connection) throws SQLException {
        String probeSql = "SELECT 1 FROM duckdb_settings() WHERE lower(name) = lower(?) LIMIT 1";
        try (PreparedStatement ps = connection.prepareStatement(probeSql)) {
            ps.setString(1, "geometry_always_xy");
            try (ResultSet rs = ps.executeQuery()) {
                return rs.next();
            }
        }
    }

    private String geometryAlwaysXyValue(Connection connection) throws SQLException {
        String query = "SELECT value FROM duckdb_settings() WHERE name = 'geometry_always_xy'";
        try (Statement statement = connection.createStatement();
                ResultSet rs = statement.executeQuery(query)) {
            if (rs.next()) {
                return rs.getString(1);
            }
            return "<setting does not exist in this DuckDB version>";
        }
    }
```

dromagnoli:
@groldan you are right, 
In the last commit of wednesday I was adding that variable. It's indeed preventing proper initialization so I have reverted that and I have also added your test to verify that now it's properly initialized.

dromagnoli:
Hi @groldan. Please, let us know once you are done with the 2nd round on GeoParquet and if/when this can be merged.

dromagnoli:
Hi @groldan. Any update on this?
Do you think we can merge it?

modules/unsupported/duckdb/src/main/java/org/geotools/data/duckdb/DuckDBDialect.java
groldan: I found that duckdb now returns the CRS as part of the type name, for example: `GEOMETRY('OGC:CRS84')` or `GEOMETRY('EPSG:4326')`, you might want to check that since it bites both the duckdb and the geoparquet store. `typeName != null && typeName.toUpperCase().startsWith("GEOMETRY"))` would do

modules/unsupported/geoparquet/src/main/java/org/geotools/data/geoparquet/GeoParquetDialect.java
groldan, line 1: With the `DuckDBDialect.getMapping()` fix the changes to this file are not necessary at all. With the prefix check, JDBCDataStore builds a proper GeometryDescriptor again and getGeometrySRID() / createCRS() run as they did before

modules/unsupported/geoparquet/src/main/java/org/geotools/data/geoparquet/GeoparquetDatasetMetadata.java
groldan, line 1: same as for GeoParquetDialect, with the prefix check in getMapping this is unnecessary

groldan (changes requested):
@dromagnoli sorry for the late reply. I kept coming back to the geoparquet side trying to ask for as few changes as possible, since how duckdb reports geometries keeps on changing.

The main problem is we could have mixed up versions of parquet and geoparquet features, but the test data is generated with whatever duckdb version is pinned as a dependency. I think the best would be to have vendored .parquet files instead of generating them, but I would do that myself as a follow up.

I've commented on what the minimum changes would be inline.

dromagnoli:
Hi @groldan. I have fixed the geometry typename check as per your comment and restored the geoparquet dialect and datasetmetadata.

Please, let me know if that can be merged.

groldan (approved):
LGTM

PR #5741, opened 2026-07-09
[GEOT-7928] PMTiles datastore breaks with filters/functions that need Feature/SimpleFeature implementations

aaime:
[GEOT-7928](https://osgeo-org.atlassian.net/browse/GEOT-7928) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)  

See ticket for a detailed description. @groldan can you have a look?

# Checklist

### DIFF
--- a/modules/unsupported/pmtiles/src/main/java/org/geotools/vectortiles/store/VectorTilesFeatureSource.java
+++ b/modules/unsupported/pmtiles/src/main/java/org/geotools/vectortiles/store/VectorTilesFeatureSource.java
@@ -54,6 +54,7 @@ import org.geotools.api.feature.type.AttributeDescriptor;
 import org.geotools.api.feature.type.FeatureTypeFactory;
 import org.geotools.api.filter.Filter;
 import org.geotools.api.filter.FilterFactory;
+import org.geotools.api.filter.expression.Function;
 import org.geotools.api.filter.sort.SortBy;
 import org.geotools.api.referencing.FactoryException;
 import org.geotools.api.referencing.crs.CoordinateReferenceSystem;
@@ -64,6 +65,7 @@ import org.geotools.data.store.ContentFeatureSource;
 import org.geotools.feature.simple.SimpleFeatureTypeBuilder;
 import org.geotools.filter.spatial.DefaultCRSFilterVisitor;
 import org.geotools.filter.spatial.ReprojectingFilterVisitor;
+import org.geotools.filter.visitor.DefaultFilterVisitor;
 import org.geotools.filter.visitor.SimplifyingFilterVisitor;
 import org.geotools.filter.visitor.SpatialFilterVisitor;
 import org.geotools.geometry.jts.GeometryCoordinateSequenceTransformer;
@@ -124,6 +126,22 @@ import org.locationtech.jts.geom.util.AffineTransformation;
 public class VectorTilesFeatureSource extends ContentFeatureSource {
 
     private static final Logger LOGGER = Logging.getLogger(VectorTilesFeatureSource.class);
+
+    /** Detects presence of a function call in a filter */
+    private static class FunctionDetector extends DefaultFilterVisitor {
+
+        boolean foundFunction = false;
+
+        @Override
+        public Object visit(Function expression, Object data) {
+            foundFunction = true;
+            return data;
+        }
+
+        public boolean foundFunction() {
+            return foundFunction;
+        }
+    }
 
     /** TileJSON layer metadata */
     protected final VectorLayer layerMetadata;
@@ -480,7 +498,23 @@ public class VectorTilesFeatureSource extends ContentFeatureSource {
      */
     private Predicate<Feature> preFilter(Query query) {
         Filter filter = query.getFilter();
+        if (Filter.INCLUDE.equals(filter)) {
+            return f -> true;
+        }
+        // Functions like geometry() instanceof/cast to Feature/SimpleFeature and return null against a plain
+        // MvtFeature, silently dropping every feature. When the filter uses a function, evaluate it against a
+        // temporary SimpleFeature view; otherwise the property accessor handles PropertyName access directly.
+        if (containsFunction(filter)) {
+            SimpleFeatureType schema = getSchema();
+            return mvtFeature -> filter.evaluate(new VectorTilesSimpleFeature(schema, mvtFeature));
+        }
         return filter::evaluate;
+    }
+
+    private static boolean containsFunction(Filter filter) {
+        FunctionDetector detector = new FunctionDetector();
+        filter.accept(detector, null);
+        return detector.foundFunction();
     }
 
     private Filter postFilter(Query query) {
--- a/modules/unsupported/pmtiles/src/test/java/org/geotools/pmtiles/store/PMTilesDataStoreTest.java
+++ b/modules/unsupported/pmtiles/src/test/java/org/geotools/pmtiles/store/PMTilesDataStoreTest.java
@@ -46,6 +46,8 @@ import org.geotools.api.feature.type.AttributeDescriptor;
 import org.geotools.api.feature.type.GeometryDescriptor;
 import org.geotools.api.feature.type.Name;
 import org.geotools.api.filter.Filter;
+import org.geotools.api.filter.FilterFactory;
+import org.geotools.api.filter.expression.Function;
 import org.geotools.api.referencing.FactoryException;
 import org.geotools.api.referencing.crs.CoordinateReferenceSystem;
 import org.geotools.data.simple.SimpleFeatureCollection;
@@ -204,5 +206,26 @@ public class PMTilesDataStoreTest {
                 assertThat(defaultGeometry, instanceOf(Geometry.class));
             }
         }
+    }
+
+    /**
+     * A filter using a function that casts to Feature (here dimension(geometry())) must be evaluated against a
+     * SimpleFeature view of each vector tile feature, not the raw MvtFeature, which would return null and drop
+     * everything. The buildings layer is all polygons, so the polygon predicate keeps them all and the line predicate
+     * keeps none.
+     */
+    @Test
+    public void getFeaturesWithGeometryFunctionFilter() throws IOException {
+        FilterFactory ff = store.getFilterFactory();
+        SimpleFeatureSource buildings = store.getFeatureSource("buildings");
+        int total = buildings.getFeatures().size();
+        assertThat(total, greaterThan(0));
+
+        Function dimension = ff.function("dimension", ff.function("geometry"));
+        assertEquals(
+                total,
+                buildings.getFeatures(ff.equals(dimension, ff.literal(2))).size());
+        assertEquals(
+                0, buildings.getFeatures(ff.equals(dimension, ff.literal(1))).size());
     }
 }


groldan:
The fix works, but two things should change before merge: the ticket's cause attribution, and the `preFilter`/`postFilter` design.

## 1. The bug rationale misattributes the cause

The ticket says the store "builds internally, outside of GeoTools reach, a MvtFeature ... and then passes it for evaluation to the functions, which will fail to evaluate." Two things are off:

The store isn't "outside GeoTools reach." It registers `VectorTilesFeaturePropertyAccessorFactory` through the `PropertyAccessorFactory` SPI, the hook for making a non-`Feature` object filterable. It resolves attributes and the default geometry through the `""` property name, which is the GeoTools-wide default-geometry convention defined in `SimpleFeaturePropertyAccessorFactory`. The pmtiles factory mirrors it and every `PropertyName`-based filter, default geometry included, already evaluates fine against an `MvtFeature`.

Then, the failure is in the function, not the feature. `GeometryFunction` (the `geometry()` function) resolves its operand like this:

```java
@Override
public Object evaluate(Object object) {
    if (object instanceof Feature feature) {
        return evaluate(feature);
    } else {
        return null;
    }
}
```

`io.tileverse.vectortile.model.VectorTile.Layer.Feature` is a tileverse interface and doesn't extend `org.geotools.api.feature.Feature`. Against it, `geometry()` returns `null`, `dimension(null)` returns `null`.

Worth conceding, though: the accessor SPI only intercepts `PropertyName` evaluation, and `geometry()` is "the default geometry of a Feature" but breaks the `Object evaluate(Object object)` contract. The real fix is upstream in `GeometryFunction`: when the object isn't a `Feature`, fall back to the accessor (e.g. `new AttributeExpressionImpl("").evaluate(object, Geometry.class)`).

That'd make `geometry()` work against anything with a registered accessor, but touches a core function and is out of scope here. Evaluating function filters against a real `SimpleFeature` in the store is a fine stopgap in the meantime.

Please just reword the ticket: the accessor SPI is honored, the gap is `geometry()`'s `instanceof Feature`, and the real fix is upstream.

## 2. `postFilter` is left as dead code with a landmine

The store has two hooks: `preFilter` (a `Predicate` on the raw `MvtFeature`, inside tileverse, before the `SimpleFeature` is built) and `postFilter` (a `Filter` on the built `SimpleFeature` in the reader, `VectorTilesFeatureSource.java:332`). The PR puts the whole filter in `preFilter` and leaves:

```java
private Filter postFilter(Query query) {
    return Filter.INCLUDE; // query.getFilter();
}
```

That commented `// query.getFilter()` is a trap now that the pre-filter doesn't handle all filters. Uncomment it to "turn on post-filtering" and you double-apply the filter, since `preFilter` already ran all of it. 

The clean version is a real split, the same shape the JDBC stores use with `PostPreProcessFilterSplittingVisitor` + `FilterCapabilities`:

- pre: the accessor-evaluable part (no `Feature`-requiring functions), on the `MvtFeature`, for the cheap early drop. `MvtFeature` decodes geometry lazily on `getGeometry()`, so an attribute-only predicate rejects a feature without ever decoding its geometry.
- post: the residual (function-bearing sub-filters), on the `SimpleFeature` the reader already builds.

This PR's `containsFunction()` drops all functions from the fast path, even `abs(prop) > 5` that the accessor handles fine, pushes the whole filter onto the `SimpleFeature` "slow path" avoiding evaluation against the `MvtFeature`. A per-sub-filter split (see below) keeps the fast path for the accessor-friendly parts.

On a function filter today, `preFilter` builds a throwaway `VectorTilesSimpleFeature` once per stream feature to evaluate the filter, then discards it. The reader building its own wrapper for the survivors isn't waste, it has to emit them regardless. The wrapper is cheap and geometry decode is memoized on the shared `MvtFeature`, which leaves two real costs: the per-feature throwaway, and running the full filter against a `SimpleFeature` for every candidate, including ones a cheap attribute pre-filter would have dropped. A proper split builds no throwaways: pre drops on attributes without a wrapper, post reuses the reader's `SimpleFeature`.

This matters because the store owns the whole filter. `canFilter()` returns `true` unconditionally, and the framework wraps no residual `FilteringFeatureReader`. There's no safety net above the store, which is why getting the internal split right is the PR's job, not the framework's.

To be clear, this is design and efficiency, not a correctness bug. The throwaway pre-stage `VectorTilesSimpleFeature` does see the fully transformed geometry, because tileverse decodes and reprojects lazily through the `MvtFeature`'s decoder before the predicate runs (`VectorTileLayerReader.java`, `MvtLayer.getFeatures`). The PR is functionally correct but kills the fast path.

aaime:
Regarding point 1, I think you are out of touch with the last years of GeoTools development. It is true that at one point there was a big push to have filters and functions work against any object, property accessors were introduced and so on. However:
* I cannot find any javadoc or documentation stating that a properly implemented function should always use property accessors (if I missed it, please share a pointer)
* The javadoc for the geometry function is quite explicit in this regard, "Function that returns the default geometry of a feature, or null if there is none, or it's not a JTS geometry". It says "feature", not random object.
* There are different functions, written by different people at different times, that assume Feature is what stores work with:
   
  | Function | Author |
  |---|---|
  | [GeometryFunction](https://github.com/geotools/geotools/blob/main/modules/library/main/src/main/java/org/geotools/filter/function/GeometryFunction.java#L69) | Andrea Aime |
  | [IDFunction](https://github.com/geotools/geotools/blob/main/modules/library/main/src/main/java/org/geotools/filter/function/IDFunction.java#L48) | Jody Garnett |
  | [BoundedByFunction](https://github.com/geotools/geotools/blob/main/modules/library/main/src/main/java/org/geotools/filter/function/BoundedByFunction.java#L40) | Andrea Aime |
  | [PropertyExistsFunction](https://github.com/geotools/geotools/blob/main/modules/library/main/src/main/java/org/geotools/filter/function/PropertyExistsFunction.java#L84) |  unclear, goes back to the svn history |
  | [AbstractQuantityClassificationFunction](https://github.com/geotools/geotools/blob/main/modules/library/main/src/main/java/org/geotools/filter/function/AbstractQuantityClassificationFunction.java#L156) |  extracted from QuantileFunction, whose history goes back to svn |
  | [EqualIntervalFunction](https://github.com/geotools/geotools/blob/main/modules/library/main/src/main/java/org/geotools/filter/function/EqualIntervalFunction.java#L196) | James Macgill; Cory Horner |
  | [JenksNaturalBreaksFunction](https://github.com/geotools/geotools/blob/main/modules/library/main/src/main/java/org/geotools/filter/function/JenksNaturalBreaksFunction.java#L64) | Ian Turton |
  | [StandardDeviationFunction](https://github.com/geotools/geotools/blob/main/modules/library/main/src/main/java/org/geotools/filter/function/StandardDeviationFunction.java#L95) | Cory Horner |
  | [UniqueIntervalFunction](https://github.com/geotools/geotools/blob/main/modules/library/main/src/main/java/org/geotools/filter/function/UniqueIntervalFunction.java#L142) | Cory Horner |

So, the reality is that the actual GeoTools implementation drifted away from the "we can work against any object" ideal little by little, and the store is written against an assumption that does not match reality any longer (that, and the documentation gap). So I propose to reword stating that it's not safe to assume functions use property accessors, as the implementation do not honor it, and the documentation does not require it (which also means functions in downstream codebase could do the same).

Point number 2. looks like a deep AI hallucination, none of the arguments really stand. The change does not open any problem, it's just wrapping the MVTFeature into an actual GeoTools feature so that functions assuming Feature can work properly. The whole original filter is run against the wrapper (and only when needed, the logic is there to prevent unecessary wrapping when functions are not used). The visitor is not dropping anything, it's not changing the filter at all in fact, it's just inspecting it. So, I don't see anything left to evaluate in a POST filter.
Am I missing something? Do we need better javadoc/comments to clarify what the code is doing?

jodygarnett:
I am just joining the chat - evaluating expressions should always get value via properties accessor.  Like just calling expression eval should do the trick. 

For examples like default geometry and identifier - what would you like to see done if the object is a non feature? Could have property accessor implementations for those ideas if needed.

Look forward to your thoughts Gabe.

groldan:
*"I think you are out of touch with the last years of GeoTools development"* is dismissive. As a non native English speaker I double checked and I wasn't wrong. Same as in #5743, please keep it civil. This sort of things discourage contributors with a thinner skin than me.

Reading back I see sometimes a statement of mine (e.g. "Then, the failure is in the function, not the feature. GeometryFunction") may read as overconfident and be wrong, but it's never a personal attack, and I'm always glad to be proven wrong.

Now on what matters more:

I concede my interpretation of the long-standing geotools convention may be out of touch since it's not an API contract. The project may have abandoned the idea of resolving propery names against foreign object models and didn't adapt the API accordingly. FWIW, the convention  is basically using `""` as the property name to resolve the default geometry buries somewhere in 20+ years of commit history.

With that in mind, this PR, as it stands, does represent a correctness fix, and I'm willing to accept it as a temporary measure, because it would unblock or ease the testing and further development I understand Andrea is performing, with the following caveats:
 
I'll ask for a follow-up pull request, before the next release, that restores the evaluation of the functions that do can work directly against `VectorTile.Layer.Feature` to the "fast path".

By the "fast path" I mean filters evaluated directly against `VectorTile.Layer.Feature`. The only implementation currently is `MvtFeature`, which works for MVT, but MLT is also in the library's horizon. This is important because `MvtFeature` (and plausibly a future `MltFeature`) lazily decodes the geometry and the feature attribute index, meaning a filter that doesn't touch at least one of them can run without unnecessary memory and cpu overhead.

For instance:
```java
class MvtFeature implements VectorTile.Layer.Feature
...
    @Override
    public Geometry getGeometry() {
        if (geometry == null) {
            geometry = decoder.decode(this);
        }
        return geometry;
    }

    @Override
    public Object getAttribute(String attributeName) {
        return getAttributes().getOrDefault(attributeName, null);
    }

    @Override
    public Map<String, Object> getAttributes() {
        if (valueIndex == null) {
            valueIndex = buildValues();
        }
        return valueIndex;
    }
```

So, my understanding of the PR is that the moment *any* function is used in a filter, it drops the fast path through `containsFunction(Filter)`, and creates unnecessary memory churn with a throw-away `VectorTilesSimpleFeature` instantiation per filter evaluation to then create another one for survivor features, whereas the "fast path" would create a `VectorTilesSimpleFeature` only for survivors.

Since `preFilter()` before this PR is supposed to handle all filters, `postFilter()` returns `Filter.INCLUDE`. With the PR's all-or-nothing approach, we gain correctness for the functions that I didn't encounter during development, but lose the fast path, which proved to be important during several profiling sessions.

So the request for a follow-up pr. The approach can be the typical `FilterCapabilities` based pre-post filter splitting, and for safety it'd include the functions known to work against `VectorTile.Layer.Feature`, of special interest, besides the common ones, the ones defined in `gt-mbstyle`'s `MBFunctionFactory.java`. 

I hope that makes sense.

We've traded fixes like this before (your OOM report led to https://github.com/tileverse-io/tileverse/pull/18 and https://github.com/geotools/geotools/pull/5481 )
In case you can't find the time for it just let me know and I'll see if I can do it myself.

aaime:
> I am just joining the chat - evaluating expressions should always get value via properties accessor. Like just calling expression eval should do the trick.

Jody during the meeting you confirmed that the idea of always using property accessors has been dropped and the project moved in another direction. Here you say the opposite. Rather than working by personal memories/feelings, can anyone point me at some documentation stating that this is the intended contract?
We cannot just focus on what is inside the GeoTools library, there is a number of out of libraries implementation out there written by people that don't have access to the project history.

Also, for a more practical standpoint, how would you turn IDFunction to use property accessors? The empty string is already taken for the default geometry; any other property name might conflict with a simple feature implementation.

Perhaps, considering the whole context of the discussion, we could have a marker interface FeatureFunction that indicates a function implementation works only against Feature, that would allow to support a filter splitter without using a fixed list of function.

aaime:
> _"I think you are out of touch with the last years of GeoTools development"_ is dismissive. As a non native English speaker I double checked and I wasn't wrong. Same as in #5743, please keep it civil. This sort of things discourage contributors with a thinner skin than me.

Gabriel, you have not been directly involved in GeoTools and GeoServer development for a bunch of years while focusing on geogig, my statement is not a personal attack, it's a simple observation, a lot happened in the meantime and Jody confirmed it during the PSC meeting (although he flipped to the opposite narrative when commenting here, which has me quite perplexed). 

> I concede my interpretation of the long-standing geotools convention may be out of touch since it's not an API contract. 
The project may have abandoned the idea of resolving property names against foreign object models and didn't adapt the API accordingly. FWIW, the convention is basically using `""` as the property name to resolve the default geometry buries somewhere in 20+ years of commit history.

Ok, we are on the same page here.

> 
> With that in mind, this PR, as it stands, does represent a correctness fix, and I'm willing to accept it as a temporary measure, because it would unblock or ease the testing and further development I understand Andrea is performing, with the following caveats:
> 
> I'll ask for a follow-up pull request, before the next release, that restores the evaluation of the functions that do can work directly against `VectorTile.Layer.Feature` to the "fast path".

I'm ok with a follow-up pull request, but who's going to make it? The people who broke the undocumented contract in those functions implementations, during the various years? Are they to blame for something that was known by a few people by word of mouth?
I think we can also go and document the contract in 35.x, no strong opposition there, but it will also have to be noted in the upgrade guide. 


> By the "fast path" I mean filters evaluated directly against `VectorTile.Layer.Feature`. The only implementation currently is `MvtFeature`, which works for MVT, but MLT is also in the library's horizon. This is important because `MvtFeature` (and plausibly a future `MltFeature`) lazily decodes the geometry and the feature attribute index, meaning a filter that doesn't touch at least one of them can run without unnecessary memory and cpu overhead.

I can see that, I'm just not sure how much that would matter in the context of a GeoServer OGC call... did you measure and find a significant difference? You know I love benchmarks and numbers, did you run any?

> So, my understanding of the PR is that the moment _any_ function is used in a filter, it drops the fast path through `containsFunction(Filter)`, and creates unnecessary memory churn with a throw-away `VectorTilesSimpleFeature` instantiation per filter evaluation to then create another one for survivor features, whereas the "fast path" would create a `VectorTilesSimpleFeature` only for survivors.

True, I was tempted to go for a function exclusion list, but then I noticed that the contract is not documented, so any function out there might have been implemented against Feature directly. As an unsupported module pmtiles is not widely used yet, but as popularity grows, it will eventually hit customized systems.
See also my answer to Jody's comment on how a filter splitter could be implemented in a general way.

> Since `preFilter()` before this PR is supposed to handle all filters, `postFilter()` returns `Filter.INCLUDE`. With the PR's all-or-nothing approach, we gain correctness for the functions that I didn't encounter during development, but lose the fast path, which proved to be important during several profiling sessions.

Ah ha, "proved to be important during several profiling sessions", right, so you do have benchmarks and numbers?
Possibly a full OGC benchmark (a microbenchmark can show a big improvement in the specific section and have no visible effect on actual requests).

Speaking of pre/post filter, I see no mention of the "postFilter is left as dead code with a landmine" bit, the "That commented // query.getFilter() is a trap now that the pre-filter doesn't handle all filters." statement.... did I get it wrong and is that actually a problem, is there anything to actually fix there? 

> So the request for a follow-up pr. The approach can be the typical `FilterCapabilities` based pre-post filter splitting, and for safety it'd include the functions known to work against `VectorTile.Layer.Feature`, of special interest, besides the common ones, the ones defined in `gt-mbstyle`'s `MBFunctionFactory.java`.
> 
> I hope that makes sense.
> 
> We've traded fixes like this before (your OOM report led to [tileverse-io/tileverse#18](https://github.com/tileverse-io/tileverse/pull/18) and #5481 ) In case you can't find the time for it just let me know and I'll see if I can do it myself.

Ok, so that seems to confirm you're asking me to fix all the functions and handle the documentation changes and the backwards-incompatible change (the new contract will make existing functions using Feature directly "broken", while today they seem to be a match with what's documented).
Please help me jog my memory a bit... I remember I've reported a OOM, but I cannot find where anymore... did I put you in a position where I was asking you to perform that development? 
Maybe the wording got misunderstood, can you give me pointers to that conversation? I cannot find it....

aaime:
Btw, I have updated the Jira ticket description.

groldan:
Let me be cristal clear on the follow-up scope, because my earlier blaming on the GeometryFunction was the wrong call on the assumption that the `""` property name convention was to be honored, which isn't, and I corrected afterwards: I'm **not** proposing a follow up pr to fix/change the functions. I'm proposing a focused pr for the pmtiles module to restore the lost functionality. A black list of functions wouldn't work, I've proposed a whitelist approach, of special interest the mbstyles ones.

> Ah ha, "proved to be important during several profiling sessions", right, so you do have benchmarks and numbers?

by "profiling sessions" I mean YJP during the OOM round trips, not benchmarking. The bounded-memory cache (below) caps memory so losing the fast path likely won't OOM again, but it'd cause GC/throughput pressure under concurrency. If you want that quantified, it can be done during the follow up.

`VectorTilesSimpleFeature` implements `SimpleFeature` and prevents creating an `Object[]` backing from `SimpleFeatureImpl`, wrapping a `VectorTile.Layer.Feature` which lazily instantiates a Map of pointers to attribute values the `VectorTile.Layer` holds as a whole.
And IIRC, that was the first cut of optimizations, which reduced memory consumption on the hot path and sped up evaluation of heavily function-based filters as it happens with an mbstyle. Since the store logic was originally inspired in the mbtiles one, and I verified the same OOM happened with it too when doing reprojection, the fix was to derive the zoom level from the wms request instead of from the generalization distance provided by the renderer, which over-fetches across tiles. As that wasn't enough I learnt the usual `softValues()` approach used in several caches leads to high GC pressure and implemented the bounded-memory cache manager linked in the tileverse pr above.

> Speaking of pre/post filter, I see no mention of the "postFilter is left as dead code with a landmine" bit, the "That commented // query.getFilter() is a trap now that the pre-filter doesn't handle all filters." statement....

It comes attached to the idea, again, to use a `PostPreProcessFilterSplittingVisitor` + `FilterCapabilities` to restore the lost functionality. Without it reads as you're interpreting it: an AI hallucination. And it was made under the assumption/expectation we wouldn't lose functionality from this PR.

Note the `PostPreProcessFilterSplittingVisitor` call was off the top of my head, looking at the codebase what let us implement the function whitelist is `CapabilitiesFilterSplitter` through `Capabilities.addType/addName`

>  did I get it wrong and is that actually a problem, is there anything to actually fix there?

we're losing functionality that was hard to come up with. When using mbstyle, almost always the filter is function bearing.

> Ok, so that seems to confirm you're asking me to fix all the functions and handle the documentation changes and the backwards-incompatible change (the new contract will make existing functions using Feature directly "broken", while today they seem to be a match with what's documented).

I'm asking if you **can**, and offerred to do it myself if you cannot. But again, not a backwards-incompatible change, not touching any function, but a confined change to the pmtiles store to restore the fast path. I'd rather not ship something that drops a hard earned perf improvement.

> Please help me jog my memory a bit... I remember I've reported a OOM, but I cannot find where anymore... did I put you in a position where I was asking you to perform that development?

GS3 code sprint, in person, you mentioned a customer of yours reported it to you, I welcomed the notice and implemented the fix the moment I had a chance. You didn't ask me to perform that development. And I'm not asking you in the sense of requiring, I'm asking if you would (is there a better word for it?)

aside, as mentioned in my first comment, the "convention" is in `SimpleFeaturePropertyAccessorFactory`:

```java
    /**
     * Conventional property name used to indicate the "efault gedometry" of a feature, that is, the one returned by
     * {@link Feature#getDefaultGeometryProperty()} or {@link SimpleFeature.getDefaultGeometry()}.
     */
    public static final String DEFAULT_GEOMETRY_NAME = "";
```
and implemented for `Feature` in `FeaturePropertyAccessorFactory.DefaultGeometryFeaturePropertyAccessor`.

and AFAIK, it's how geoserver handles bbox queries without a concrete geometry property name, so it's usage is pervasive?

+1 to document the convention in 35.x and noting it in the upgrade guide if that holds.

aaime:
> . I'm proposing a focused pr for the pmtiles module to restore the lost functionality. A black list of functions wouldn't work, I've proposed a whitelist approach, of special interest the mbstyles ones.

See my previous proposal of marking functions, this would allow SPI contributed functions to play in the system as well, one could mark a function as "needing Feature" freeing the store from a static list, or a list backed by a modifiable system property.
But the module is under your maintenance, if you believe it's better to have a whitelist, you're king in here. 
I'm personally not interested in implementing it, though.

A better documentation of the  ``""`` convention would indeed be a good idea.

jodygarnett:
Some notes:

* The initial design is here: https://github.com/geotools/geotools/wiki/Expression-Improvements
* There is some text about expression expecting to work with Feautre, but also java beans and maps here: https://docs.geotools.org/latest/userguide/library/api/filter.html when introcucing expressions as a concept.

* I am *also* out of touch with years of development, since enforcing this goal will have often escaped QA review leading to degrations such as we find in this thread.
  
 My comment on expecting it just to work is based on AttribtueExpressionImpl delegating
 to PropertyAccessor [here](https://github.com/geotools/geotools/blob/main/modules/library/main/src/main/java/org/geotools/filter/AttributeExpressionImpl.java#L189).
 
 The functions should be evaluating their expressions in order to get a value.
 
 If the expression is a PropertyAttribute then it should use AttribtueExpressionImpl and be cheerful...

* The history of default geometry working wiht `""` is ... actually a optimization shortcut.

  The initial WMS / Styling standards assumed that an XPath selecting a feature to draw (rather than a specific feature attribute ... would draw *all* the geometry in the feature, similar to drawing a geometry collection).
  
  To avoid checking all the attributes, the idea of default geometry was introduced as an optimization. So `""` just indicates the XPath stopped at the Feature. Presumably you could have an xpath that went further and drilled into a an attribute (as we do when drawing features with multiple geometry attributes), or goes further to drill into part of a MultiPolygon (not sure we support that), etc...

* Having a marker interface to indicate if a function on works with a Feature is ... fine?
  It may be smarter to have a fallback value (either documented, or supplied by user).
  
  I wish fallback values were more smoothly integrated into the data model.

groldan:
King does the dishes it is. 

Merge it as the correctness fix and I'm taking the follow up myself.

About the marker interface, it'd be a separate discussion. For this module in particular, I'd still go with a whitelist to be safe against external implementations that do not apply it.

groldan (approved):

jodygarnett:
> Also, for a more practical standpoint, how would you turn IDFunction to use property accessors? The empty string is already taken for the default geometry; any other property name might conflict with a simple feature implementation.

I see I am the author of IDFunction in GeoTools 2.2 :P

The implementaitons seems clear: returning an identifier if one is available, or empty `""` if it is being called on something that does not have an identifier. I would be happy to update the javadocs.

Now that I know more about XPath, we probably have an alternative to using 
GML2 has `gml2:fid`, GML3 has `gml3:id`:

* `//@gml2:fid`
* `//@gml3:id`

Also filter is kind of annoying with ID based selection a bit distinct from the more general expression approach. We try our best to smooth out this difference but is never going to be perfect as it does not match the design of OGC Filters which has two workflows.

It is presented as such in our user manual: https://docs.geotools.org/latest/userguide/library/main/filter.html

* Selection: Using FeatureId with a set of strings (see [FidFilter](https://github.com/geotools/geotools/blob/main/modules/library/main/src/main/java/org/geotools/filter/FidFilterImpl.java), the IDFunction is the expression equivalent of this filter.

So yeah we try and close the gap, and IDFunction is never going to be perfect as a result. 

I note that FidFilterImpl amazingly uses:
```
        // NC - updated, using attributeexpressionimpl will be easiest, don't have to copy and paste lots of code
        FilterFactory ff = CommonFactoryFinder.getFilterFactory(null);
        String evaluate = ff.property("@id").evaluate(feature, String.class);
```

In order to delegate to PropertyAccessors! Wow... and it seems to be gml3 specific assumption.

Going to just sit with this a bit to see if it makes sense.

jodygarnett:
On reflection I am somewhat surprised that IDFunction uses direct access (when I would expect expression), and then FidFilterImpl goes out of its way to use expressions (when I would assume direct access).

Let me float the following suggestion:

* Harmonize the two implementations IDFunction and FidFilterImpl in terms of use of `@id` XPath, and see if we can clarify the javadocs as needed to match.
* As a result both have an oppertunity to use ProperyAccessor ... with `@id`.
* What happens in XML when using XPath against an attribute that is not defined?  GetAttribute returns {@code ""}, while getAttributeNode returns {@code null}.
  
  As a result IDFunction returning {@code ""} when `@id` is not available is expected.

  FidFilter not matching anything when `@id` is not available is expected.

aaime:
@jodygarnett that sort of makes sense but we need to address the lack of javadoc/documentation here. A random wiki page won't do, I am familiar with the project and didn't know we had docs in the wiki, let alone them being authoritative (edit, I see now it's a proposal... nice, but the proposal did not land proper docs).

The other bit is that I understood all along that Gabriel wanted to me to take on the whole GeoTools feature vs function cleanup, but I see now that he wants a local approach instead. As a result, we are left with nobody interested in the global cleanup approach.
 
@groldan I can do the dishes but I consider a static whitelist problematic, it will bite me down the line.... am I game if we can mark the functions as FeatureFuntion instead (at the very least I can go in the downstream custom codebases and add the markers as we upgrade, without bothering the deploy team with adding a system variable with fully qualified names of the functions I need to be dodged).

groldan:
> one could mark a function as "needing Feature" freeing the store from a static list, or a list backed by a modifiable system property.

The whitelist idea is to keep the change confined to the module as a capability (`CapabilitiesFilterSplitter` driven split),  not as a system prop. Similar to how jdbc stores do for filter push down.

The marker interface can work of course. I tried to keep it as a separate discussion for a correctness fix and perf follow-up to an unsupported module not driving a core API change. We need to be conscious though that with the marker interface only (or whatever the mechanism is to signify "requires Feature"), there might be function implementations out there that require feature but don't upgrade, and that we're officially acknowledging the filter model is designed to work with foreign object models, which is fine for me.

Jody identified the tension with `@id` filter/function needing consolidation. The `""` for default geometry convention falls in the same bucket. Next question is whether `""` becomes a first-class citizen, and hence functions that currently would be marked as `RequiresFeature`, like `GeometryFunction`, can be generalized to resolve using the attribute expression when a foreign object model is evaluated.

In any case, I support the core API consolidation/documentation idea. I'm just saying it should be moved to the devel list, and it's quite orthogonal to the pmtiles follow-up. 

PMTiles will require pre-post filter splitting anyways, either dumb with the list of supported functions in the splitter or dynamically with the marker. It can start dumb while the core API conventions settle.

jodygarnett:
@groldan I created https://osgeo-org.atlassian.net/browse/GEOT-7941 to address `@id` and `` filters in consistencies. 

As you are correct we can move this to development list.

PR #5743, opened 2026-07-10
[GEOT-7930] PMTiles reads a zoom level too deep compared to on screen display

aaime:
[GEOT-7930](https://osgeo-org.atlassian.net/browse/GEOT-7930) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)  

@groldan another one to get protomaps to render so that it matches the in-browser version (that, and some 20-ish other fixes in mbstyles, for now I'm collecting the issues, the actual fixes will follow).

Btw, this one will definitely conflict with https://github.com/geotools/geotools/pull/5741, once either is merged I'll amend the other to match.

# Checklist

### DIFF
--- a/modules/unsupported/pmtiles/src/main/java/org/geotools/pmtiles/store/PMTilesDataStoreFactory.java
+++ b/modules/unsupported/pmtiles/src/main/java/org/geotools/pmtiles/store/PMTilesDataStoreFactory.java
@@ -46,6 +46,7 @@ import org.geotools.tileverse.rangereader.ParamBuilder;
 import org.geotools.tileverse.rangereader.RangeReaderParams;
 import org.geotools.util.Converters;
 import org.geotools.util.logging.Logging;
+import org.geotools.vectortiles.store.VectorTilesDataStore;
 import org.jspecify.annotations.NonNull;
 import org.jspecify.annotations.NullMarked;
 
@@ -72,6 +73,8 @@ import org.jspecify.annotations.NullMarked;
  * <ul>
  *   <li>{@linkplain #URIP pmtiles} - PMTiles archive URI (required)
  *   <li>{@linkplain #NAMESPACEP namespace} - Feature type namespace (optional)
+ *   <li>{@linkplain #DISPLAY_TILE_SIZE displayTileSize} - Display tile size in pixels rendering clients use, defaults
+ *       to {@value VectorTilesDataStore#DEFAULT_DISPLAY_TILE_SIZE}
  *   <li>{@linkplain #RANGEREADER_PROVIDER_ID io.tileverse.rangereader.provider} - Optionally force a specific
  *       {@code RangeReaderProvider} by its {@link RangeReaderProvider#getId() id}, defaults to automatic detection by
  *       {@link RangeReaderFactory#findBestProvider(io.tileverse.rangereader.spi.RangeReaderConfig)}
@@ -156,6 +159,22 @@ public class PMTilesDataStoreFactory implements DataStoreFactorySpi {
             .title("Namespace prefix")
             .type(String.class)
             .optional()
+            .advancedLevel()
+            .build();
+
+    /**
+     * Display tile size (in pixels) the archive's rendering clients are assumed to use, overriding
+     * {@link VectorTilesDataStore#DEFAULT_DISPLAY_TILE_SIZE}.
+     */
+    public static final Param DISPLAY_TILE_SIZE = ParamBuilder.builder()
+            .key("displayTileSize")
+            .title("Display tile size")
+            .description("Tile size in pixels rendering clients of this archive are assumed to use when picking a "
+                    + "zoom level. Protomaps and the MapLibre/Mapbox GL family target 512px; override only if the "
+                    + "archive's clients use a different convention.")
+            .type(Integer.class)
+            .optional()
+            .defaultValue(VectorTilesDataStore.DEFAULT_DISPLAY_TILE_SIZE)
             .advancedLevel()
             .build();
 
@@ -309,7 +328,7 @@ public class PMTilesDataStoreFactory implements DataStoreFactorySpi {
                 .toList();
 
         Predicate<Param> filter = p -> !ignore.contains(p.key);
-        return RangeReaderParams.appendAfter(filter, NAMESPACEP, URIP);
+        return RangeReaderParams.appendAfter(filter, NAMESPACEP, URIP, DISPLAY_TILE_SIZE);
     }
 
     @Override
@@ -347,6 +366,9 @@ public class PMTilesDataStoreFactory implements DataStoreFactorySpi {
 
         String namespaceURI = lookup(NAMESPACEP, params, String.class);
         store.setNamespaceURI(namespaceURI);
+
+        Integer displayTileSize = lookup(DISPLAY_TILE_SIZE, params, Integer.class);
+        store.setDisplayTileSize(displayTileSize);
 
         return store;
     }
--- a/modules/unsupported/pmtiles/src/main/java/org/geotools/vectortiles/store/VectorTilesDataStore.java
+++ b/modules/unsupported/pmtiles/src/main/java/org/geotools/vectortiles/store/VectorTilesDataStore.java
@@ -75,7 +75,12 @@ public abstract class VectorTilesDataStore extends ContentDataStore {
     public static final GeometryFactory DEFAULT_GEOMETRY_FACTORY =
             new GeometryFactory(new PackedCoordinateSequenceFactory());
 
+    /** Default value for {@link #setDisplayTileSize(int)}, matching the Protomaps/MapLibre/Mapbox GL convention. */
+    public static final int DEFAULT_DISPLAY_TILE_SIZE = 512;
+
     private VectorTileStore tileStore;
+
+    private int displayTileSize = DEFAULT_DISPLAY_TILE_SIZE;
 
     /**
      * Creates a new vector tiles datastore.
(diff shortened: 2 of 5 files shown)


groldan (changes requested):
Thanks for tracking this down. The bug is real and the correction is in the right direction:
the pyramid resolutions are computed for 256px `tileWidth`, but most of the time the data is authored for a 512px display, and scaling the requested resolution by `512/tileWidth` to pick one level shallower is the fix.

Three things I'd like to see addressed before merge.

## 1. Configuration mechanism: datastore parameter, not a system property

```java
private static final int VECTOR_TILE_DISPLAY_SIZE =
        Integer.getInteger("org.geotools.vectortiles.displayTileSize", 512);
```

A `static final` read once at class load is the wrong knob for something the javadoc calls "the rare data authored for a different target":

JVM-global, in GeoServer it applies to **every** PMTiles datastore, which is the opposite of a per-dataset override. It can't even be changed without a JVM restart, and two stores in the same instance can't differ.

This should be a **datastore parameter**.  A `displayTileSize` `Param` (default `512`) threaded down to the feature source is per-store, discoverable, and configurable from the GeoServer UI.

Since the display authoring size is **not derivable from the data**: a dataset authored for 256px display sits on the same 256px WebMercatorQuad matrix as one authored for 512px, so `tileWidth` (and everything else in the pyramid/TileJSON) reports 256 in both cases. The `assertEquals(256, ms.tileWidth())` in the test documents the matrix grid, not the authoring target. Since the target can't be auto-detected, explicit per-store config is the only mechanism that can express a non-512 value.

## 2. Wording of the 512px justification

The field javadoc states:

> Display tile size vector tiles are authored against (Mapbox/MapLibre/protomaps use 512px, which is universal for vector tiles)

**"universal for vector tiles" overstates it.** 512px is the Mapbox/MapLibre GL rendering convention (protomaps builds on MapLibre), not a spec property of vector tiles. OGC WMTS / TileMatrixSet vector tiles and other stacks can be authored against 256px. Please frame it as "the display size the dominant GL clients render at," which is the actual basis for the `512/tileWidth` ratio.

## 3. Test placement and style

`VectorTilesZoomSelectionTest` is a new class named after the behavior rather than the class under test (pervasive pattern in claude code). The module convention is `<ClassUnderTest>Test`, and the logic under test lives in `VectorTilesFeatureSource`.

- Rename to `org.geotools.vectortiles.store.VectorTilesFeatureSourceTest`

aaime:
I've read the structure of your answer, and I'm having a strong "reading a message written by Claude" vibe.  I believe it's important you join us on the PSC meeting on Tuesday; we are about to set up an AI policy for GT/GS (new GSIP). While it's not going to be anywhere as strict as the [GDAL one](https://gdal.org/en/stable/community/ai_tool_policy.html), we are going to set up guardrails on several things (including banning the usage of AI as a communication proxy in issues). The topic was also a hot one in a BOF at FOSS4G Europe, you might want to read [the outcomes of it](https://www.osgeo.org/foundation-news/ai-in-osgeo-projects-notes-from-the-timisoara-bof/). 

Answering bit by bit:

1. I had considered using a store parameter and the first implementation actually had one, then I decided not to overcomplicate things, early/eager generalization is the kind of decision an AI would make and a pragmatic, practical human would skip.
My reasoning was simple: the MVT ecosystem pervasively assumes 512px per tile on display, and protomaps (the creators of pmtiles) do so in particular (this fix came out while trying to render protomaps pmtiles with their own styles). 
Your call about OGC standards is theoretical, and the 256px display assumption is often broken on practical terms: there's a reason why the GeoServer "well-known gridsets" do not contain only the base one, but also x2 variants with 512px... because using 256px tiles on today's screen generates too many requests and overloads the server. At GeoSolutions, we stopped using 256px tiles as the default years ago.
So, I consider a system that's using 256px vector tiles in 2026 a rare occurrence, and one that's using both conventions at the same time quite unlikely... but given the pushback, I'm guessing you actually have one handy?
Anyways, I'll add the store parameter back, no problem; you're the maintainer. 

2. I'm a practical person, and I approach engineering as a practical activity, rather than a theoretical science (if the vast majority of what I see does one way, I simplify the statement to "universal"). 
That said, while I don't share your view, I understand the point, and I'm amending the comments accordingly.

3. I take issue with the implication that the naming convention comes from AI. [These are tests](https://github.com/geotools/geotools/tree/main/modules/library/render/src/test/java/org/geotools/renderer/lite) that I wrote over many years in the renderer module, before AI-driven coding tools even existed. The JDBC tests are another [vivid example](https://github.com/geotools/geotools/tree/main/modules/library/jdbc/src/test/java/org/geotools/jdbc), at the beginning, I was not even involved in them (Justin created the structure of those tests). It seems I'm following a common GeoTools convention here.
Anyhow, I'll follow the "module convention" since you're the maintainer.

groldan:
Hi Andrea,

TL;DR: thanks for the changes, look good to me.

I think the whole picture is roughly:
- standard renderers (GL) do use 512x512 target rendered tiles
- vector tiles/pmtiles are agnostic of target rendering size by spec
- you can use 256x256 tile matrices in geoserver (standard gridsets)
- hence the datastore parameter is a better middle ground than a per request tuning and a system property

---

> I've read the structure of your answer, and I'm having a strong "reading a message written by Claude" vibe. I believe it's important you join us on the PSC meeting on Tuesday; we are about to set up an AI policy for GT/GS (new GSIP). While it's not going to be anywhere as strict as the GDAL one, we are going to set up guardrails on several things (including banning the usage of AI as a communication proxy in issues). The topic was also a hot one in a BOF at FOSS4G Europe, you might want to read the outcomes of it.

I've spent too much of my Saturday reviewing your pull requests as to feel apologetic about using a tool to help me fact check (AI web searches faster than I do) and to help me structure a reply message. The reasoning is mine. There's a recurring "an AI would do X, a practical human wouldn't" pattern in your replies. You may have noticed by now my version of being practical is picking my battles and I don't care much about flame wars, try to accompany every bug report with a pull request, and be available to help whenever I can. So I would appreciate dropping the passive aggressiveness.

> Answering bit by bit:

> I had considered using a store parameter and the first implementation actually had one, then I decided not to overcomplicate things, early/eager generalization is the kind of decision an AI would make and a pragmatic, practical human would skip.

Your initial fix (system property) would break all standard authored stores in a single geoserver instance the moment you configure it for the rare case where the target rendering tile size is not 512.

> My reasoning was simple: the MVT ecosystem pervasively assumes 512px per tile on display, and protomaps (the creators of pmtiles) do so in particular (this fix came out while trying to render protomaps pmtiles with their own styles).

nonetheless AFAIK both the pmtiles and VT specs are agnostic of target tile size, that's a consumer/renderer decision.

> Your call about OGC standards is theoretical, and the 256px display assumption is often broken on practical terms: there's a reason why the GeoServer "well-known gridsets" do not contain only the base one, but also x2 variants with 512px... because using 256px tiles on today's screen generates too many requests and overloads the server. At GeoSolutions, we stopped using 256px tiles as the default years ago.
> So, I consider a system that's using 256px vector tiles in 2026 a rare occurrence, and one that's using both conventions at the same time quite unlikely... 

there are lots of users out there, and I'd rather not judge whether they can or cannot use a store for a standard target size and another for a custom/legacy one.

> but given the pushback, I'm guessing you actually have one handy?

I honestly think it was a reasonable pushback, 256 is standard in geoserver/gwc, isn't it? besides using it for a layergroup with mbstyle, the layers from pmtiles store are regular layers to be queried through WMS with regular SLDs. Hence that's the standard path, no need for a specifically authored dataset.

> Anyways, I'll add the store parameter back, no problem; you're the maintainer.
> I'm a practical person, and I approach engineering as a practical activity, rather than a theoretical science (if the vast majority of what I see does one way, I simplify the statement to "universal").
> That said, while I don't share your view, I understand the point, and I'm amending the comments accordingly.
> I take issue with the implication that the naming convention comes from AI.

It read as AI to me, and didn't mean it in a demeaning way. I take it back.

groldan (approved):

aaime:
> I've spent too much of my Saturday reviewing your pull requests as to feel apologetic about using a tool to help me fact check (AI web searches faster than I do) and to help me structure a reply message. The reasoning is mine. There's a recurring "an AI would do X, a practical human wouldn't" pattern in your replies. You may have noticed by now my version of being practical is picking my battles and I don't care much about flame wars, try to accompany every bug report with a pull request, and be available to help whenever I can. So I would appreciate dropping the passive aggressiveness.

At FOSS4G Europe I have seen and heard things that probably re-framed a bit what I consider aggressive, regarding AI evaluation. I assure I meant no personal offense, the AI policy that I'm drafting will contain rules on what is acceptable and what is not, a long list of "AI would do this while humans would do that", and we're going to set up an automated review to flag classic AI smells (using what... an AI, of course, with no extra funds to fight AI slop manually, the only way seems to be to fight fire with fire). 

Let me give you some more context. Picture yourself doing a manual review... what does one do? The reviewer goes through the code and adds comments over the lines of code that seem problematic, repeat that until done, and then close with a short overall PR review message to wrap it up. Instead, your review is a long-winded, well-structured, overly formatted block of text. It's evidently AI-written, and you confirmed it. The immediate reaction is "he did not even take the time to write the answer". The effort that went into its preparation does not show.

Mind, I do believe that you spent a lot of time doing research, because I did the same to write the rebuttal above (yes, I also use AI all the time). In particular, I spent well over an hour doing the research, and Claude gave me the exact same points you listed. When it was time to write the reply, I wrote it by hand, and then had Claude point out the issues in it, which I also corrected by hand until the AI had no qualms with it any longer. Then I went over the other PR and spent another hour in the same fashion (and yes, that was a Sunday).

aaime:
Ah, one last point about the idea that vector tiles production is independent of viewer choices, which I think is the crux of the problem... while they are seemingly technically independent and physically separate in execution, it is not true that the producer has no "on-screen" target. 

When doing a classic rendering of PNG tiles we set up a SLD style and then the rendering engine does the following:
1) Determine filters, scale dependencies, and target (one screen) generalization distance, and then query the data source.
2) In memory, clip the data to the viewing area + gutter
3) Find the symbolizers that match, and render them. 

Vector tiles production does points 1. and 2. in pre-processing and bakes the results in the tiles. That does not make the tiles screen independent; the choice of how much data to put in a tile and the generalization level is still informed by the on-screen desired results: "I don't put main roads in this tile yet, because they would be too crowded, but at the next zoom level it's fine". The generalization going into the tiles production also assumes a certain target screen resolution, so that the simplifications performed are not evident on screen. My point is that vector tiles are produced with a specific screen target; they are not really agnostic about it because of the decisions that go into their production.

groldan:
"agnostic" was a bad choice, I mean you can't derive the producer's intent from the file. We're saying the same thing. It doesn't matter if the VT is authored with 4096x4096 or other local coords extent, or how many total vertices are in a tile. There's no target tile size to query in either the pmtiles or indivitual vt's. That sharpens the point for the per store config param. In any case the pr is approved, feel free to merge.

PR #5775, opened 2026-07-21
[CI] Handle errors when checking out integration build PR in workflow

mprins:
Add error handling to PR checkout command.

Because the PR body has an instructional  comment that triggers this script, a GeoTools PR will fail unless you are "lucky" that it happens to exist.

see eg. https://github.com/geotools/geotools/actions/runs/29824458995/job/88652900437?pr=5773

  

# Checklist


mprins:
the integration build still fails because it uses the workflow from the main branch

aaime:
How about we just edit the PR template so that it would not trigger the magic? Something like this:

```
  -Downstream GeoWebCache PR: 1234
  -Downstream GeoServer PR: 5678
  -Downstream mapfish-print-v2 PR: 9012
  +Downstream GeoWebCache PR: <number>
  +Downstream GeoServer PR: <number>
  +Downstream mapfish-print-v2 PR: <number>
```

<number> does not start with a digit, it should be ignored by the grep as a consequence.

mprins:
That should work too, pushed caeb020ce405d5ffe59e695e11f6772736348798

PR #5781, opened 2026-07-22
[GEOT-7938] Improved redirect-following checks in PreventLocalEntityResolver

axl8713:
[GEOT-7938](https://osgeo-org.atlassian.net/browse/GEOT-7938) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)  

These changes improve the redirect checks in `PreventLocalEntityResolver`.

  

# Checklist


jodygarnett:
Initial feedback is -- do we need something similar in DefaultEntityResolver?

docs/user/library/metadata/geotools.rst
```diff
@@ -115,6 +115,8 @@ GeoTools also includes several ``EntityResolver3`` implementations:
      - ``"http,jar,jar:file,vfs,file"``
      - Supports IDE development where ``target/classe`` and ``target/test-classes`` may be used during testing.
 
+When resolving an allow-listed ``http(s)`` reference, ``PreventLocalEntityResolver`` follows HTTP redirects itself, validating each intermediate location against the same allow-list before following it.
```
jodygarnett: If this is specific for `PreventLocalEntityResolver` only, then the commend could move to the `PreventLocalEntityResolver` bullet point further up the page.

modules/library/metadata/src/main/java/org/geotools/util/PreventLocalEntityResolver.java
```diff
@@ -99,27 +111,81 @@ public InputSource getExternalSubset(String name, String baseURI) throws SAXExce
...
+     * Fetches the content of an http(s) URI already validated against {@link #ALLOWED_URIS}, following redirects
+     * manually and re-validating each redirect target against the same allow-list before following it.
+     */
+    private static InputSource fetchAllowedHttpUri(String uri) throws IOException, SAXException {
+
+        String currentUri = uri;
+
+        for (int redirects = 0; redirects <= MAX_REDIRECTS; redirects++) {
+
+            HttpURLConnection connection = (HttpURLConnection) new URL(currentUri).openConnection();
```
jodygarnett, line 170: I expect we need some try/finally block here to disconnect / close.

We also have an [http-client](https://docs.geotools.org/latest/userguide/library/http/index.html) factory in GeoTools to allow the surrounding application to configure which library to use for connections - not sure if it can be used here?

modules/library/metadata/src/test/java/org/geotools/util/PreventLocalEntityResolverTest.java
```diff
@@ -19,23 +19,75 @@
 import static org.geotools.util.PreventLocalEntityResolver.ERROR_MESSAGE_BASE;
 import static org.geotools.util.PreventLocalEntityResolver.INSTANCE;
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
+import com.sun.net.httpserver.HttpServer;
```
jodygarnett, line 27: I had to look up and confirm that this is a supported part of java now - usually `com.sun.*` classes are internal. In this case it is built-in as part of `jdk.httpserver` module.

It still makes me a little uncomfortable, as we have had problems with LDAP tests conflicting on the same build server (forcing us to randomize port used).

Is it possible to use a mock object here rather than open up a port...

modules/library/metadata/src/test/java/org/geotools/util/PreventLocalEntityResolverTest.java
```diff
@@ -19,23 +19,75 @@
...
 
 @SuppressWarnings("BadImport")
 public class PreventLocalEntityResolverTest {
 
+    private static HttpServer server;
+    private static String baseUrl;
+
+    @BeforeClass
+    public static void startServer() throws IOException {
+        server = HttpServer.create(new InetSocketAddress("localhost", 0), 0);
```
jodygarnett, line 46: Okay this is allocating against an operating system provided port; so it should be fine for concurrent builds on the same build server.

jodygarnett (changes requested):
I would like a second opinion, but I am +1 on this PR (I just wish to be cautious about anything to do with EntityResolvers and get a further review).

Concrete feedback:

* I really like the redirect check and wonder if we can make this functionality easier to use by other downstream implementations also. I know this is out-of-scope for your targeted fix for PreventLocalEntityResolver but it would be good to consider the library as a whole.

 Discussion: Can this method be factored out, if we can find a good location.

 Action: The implementation needs a few try/finally safeties to to close connections.

* GeoTools tries not use open `http` connections directly, allowing the surrounding application to provide the http client library used.

  Action: Can you check if the GeoTools HTTPClient wrapper is sufficient to implement the redirect check? It does have the ability to check header.

* It looks like use of built-in `HttpServer.create` is fine but I would like a second opinion from another GeoTools developer on if this is fine, or we need to copy one of the mock examples in the codebase.

Thanks so much for this very welcome improvement.

modules/library/metadata/src/main/java/org/geotools/util/PreventLocalEntityResolver.java
mprins, line 170: What happens when eg. a WFS requires authentication?

modules/library/metadata/src/main/java/org/geotools/util/PreventLocalEntityResolver.java
```diff
@@ -62,6 +63,17 @@ public class PreventLocalEntityResolver implements EntityResolver3, Serializable
      */
     private static final Pattern ALLOWED_URIS = Pattern.compile("(?i)(jar:file|jar:nested|http|vfs)[^?#;]*\\.xsd");
 
+    /** Max number of redirects followed when fetching an allow-listed http(s) URI, to avoid a redirect loop */
+    private static final int MAX_REDIRECTS =
```
mprins: making this private means it does not show up in the API docs, which may lead to surprises

modules/library/metadata/src/main/java/org/geotools/util/PreventLocalEntityResolver.java
```diff
@@ -62,6 +63,17 @@ public class PreventLocalEntityResolver implements EntityResolver3, Serializable
...
+
+    /**
+     * Connect/read timeout (seconds) when fetching an allow-listed http(s) URI, same default as
+     * {@code SimpleHttpClient}
+     */
+    private static final int TIMEOUT_SECONDS =
+            Integer.getInteger("org.geotools.util.preventLocalEntityResolver.timeout", 30);
+
     /** Singleton instance of PreventLocalEntityResolver */
     public static final PreventLocalEntityResolver INSTANCE = new PreventLocalEntityResolver();
```
mprins, line 79: making this private means it does not show up in the API docs, which may lead to surprises.
If it supposed to be the same as SimpleHttpClient, it should actuallty use that value

modules/library/metadata/src/main/java/org/geotools/util/PreventLocalEntityResolver.java
mprins: Not using GeoTools http-client will cause surprises, especially when one is set through a system property like `org.geotools.http.client`

mprins (changes requested):
I don't think setting up http connection for each and every http resource is a good idea; this will slow down entity resolution considerably especially since no caching is implemented.

The `PreventLocalEntityResolver` is a naive class as it is, assuming any XSD is a file that ends with ".xsd" (see eg GEOT-7922) this implementation does not make thing better, as a redirect from a https uri does not have to be another http implementation - that assumes this resolver should mimic a webbrowser-like behaviour.

I feel this functionality is better implemented in a new EntityResolver3 implementation (or as part of your application and not the library)

docs/user/library/metadata/geotools.rst
axl8713: Done. Thanks!

modules/library/metadata/src/main/java/org/geotools/util/PreventLocalEntityResolver.java
axl8713: Right, fixed. Thanks.

modules/library/metadata/src/main/java/org/geotools/util/PreventLocalEntityResolver.java
axl8713: Fixed, thanks!
I've also dropped the "same default as SimpleHttpClient" wording from the javadoc since it's just a coincidentally equal literal, not a synced reference.

modules/library/metadata/src/main/java/org/geotools/util/PreventLocalEntityResolver.java
axl8713: Starting here the discussion on the usage of the GeoTools http-client.

`gt-metadata` doesn't depend on `gt-http`, but conversely, `gt-http` depends on `gt-metadata`.
Hence, importing `SimpleHttpClient/HTTPClient` here would create a circular module dependency. `SimpleHttpClient.DEFAULT_TIMEOUT` is also private, so it isn't even accessible as-is.

Using the client is then not possible without restructuring the module boundaries, which is out of scope for this fix.

axl8713: > I expect we need some try/finally block here to disconnect / close.

Every exit path that doesn't hand the stream to the caller now closes explicitly.
Only the success path stays open on purpose, since the InputStream has to survive for the SAX parser to consume; a wrapping try/finally would close it too early.

modules/library/metadata/src/test/java/org/geotools/util/PreventLocalEntityResolverTest.java
axl8713: Resolving, due to the following comment. Thanks for checking.

modules/library/metadata/src/main/java/org/geotools/util/PreventLocalEntityResolver.java
axl8713: It simply gets an authentication error response code, which is now rejected cleanly with a SAXException.

mprins (commented):
I stand by my earlier comment/opionion:
> I don't think setting up http connection for each and every http resource is a good idea; this will slow down entity resolution considerably especially since no caching is implemented.
>
> The PreventLocalEntityResolver is a naive class as it is, assuming any XSD is a file that ends with ".xsd" (see eg GEOT-7922) this implementation does not make thing better, as a redirect from a https uri does not have to be another http implementation - that assumes this resolver should mimic a webbrowser-like behaviour.
>
> I feel this functionality is better implemented in a new EntityResolver3 implementation (or as part of your application and not the library)

Next to that: validating remote URI's is not what I would expect from a class that says to "prevent local entity resolution" and not using a pre-configured HTTP client will result in brittle behaviour. 
Also, in my experience, the entity resolver does not follow redirects; eg. when loading https://service.pdok.nl/hwh/luchtfotorgb/wmts/v1_0?request=getcapabilities getting the schema currently fails because the redirect from http://inspire.ec.europa.eu/schemas/inspire_vs_ows11/1.0/inspire_vs_ows_11.xsd to https://inspire.ec.europa.eu/schemas/inspire_vs_ows11/1.0/inspire_vs_ows_11.xsd is not followed

axl8713:
> Also, in my experience, the entity resolver does not follow redirects; eg. when loading https://service.pdok.nl/hwh/luchtfotorgb/wmts/v1_0?request=getcapabilities getting the schema currently fails because the redirect from http://inspire.ec.europa.eu/schemas/inspire_vs_ows11/1.0/inspire_vs_ows_11.xsd to https://inspire.ec.europa.eu/schemas/inspire_vs_ows11/1.0/inspire_vs_ows_11.xsd is not followed

I checked this specifically and can't reproduce it on this branch's code.

Running `PreventLocalEntityResolver.resolveEntity()` from this branch against that exact URL follows the redirect and fetches the schema successfully.

This happens because `ALLOWED_URIS` matches "http" as a literal prefix `+ [^?#;]*\.xsd`, and the wildcard already absorbs the s in https, so both schemes match. 

Added `testHttpRedirectToAllowedHttpsUriIsAccepted` to cover this case.

If you're still hitting a failure, can you share version/error details?

PR #5804, opened 2026-08-03
[GEOT-7942] [GEOT-7943] [GEOT-7944] Reprojection can use the wrong datum shift signs and a single operation factory

aaime:
[GEOT-7942](https://osgeo-org.atlassian.net/browse/GEOT-7942) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)  

Three related fixes to raster and vector reprojection, all found while investigating a Belgian Lambert 72 to Belgian Lambert 2008 coverage reprojection that came out about 100 m off, and found it was caused by a mix of three different issues, each one ifxed in a separate commit.

First, the datum shift built from a datum to WGS84 negated the wrong parameter for Coordinate Frame Rotation (EPSG method 9607): ey was negated twice and ez never, which moves points by several meters. After fixing the code (very old bug!) I've noticed that the generated epsg.properties in epsg-wkt baked the sign errors, and had to regenate here. 
New tests compare a bunch datums against PROJ golden values, including Position Vector and
Geocentric Translation controls to prove the sign conversion does not reach where it does not belong (golden values taken from cs2cs).

Second, PropertyCoordinateOperationFactory replaced the default operation factory instead of augmenting it, so a CRS pair missing from the property file lost its EPSG predefined operation and fell straight to analytic
synthesis. It now consults the EPSG operations before synthesizing, with the property file definitions still winning.
In most operations that did not cause issues, but see the next issue.

Finally, coverage and some feature reprojection asked for a single coordinate operation factory, so a user-configured factory at maximum priority could hide the EPSG operations from every other factory. Resampler2D, the two
reprojecting iterators and TransformedPosition now resolve operations through all registered factories by using the CRS facade.

Ticket references:
- https://osgeo-org.atlassian.net/browse/GEOT-7942
- https://osgeo-org.atlassian.net/browse/GEOT-7943
- https://osgeo-org.atlassian.net/browse/GEOT-7944

# Checklist


dromagnoli (approved):
Looks good to me

PR #5810, opened 2026-08-06
Optimizations for raster reprojection

aaime:
Assorted speedups found profiling raster reprojection under load, specifically under hard-to-reproject cases (geostationary satellite). One commit each:

* [GEOT-7945](https://osgeo-org.atlassian.net/browse/GEOT-7945) Read resolution
  computation gave up when some test points fell outside the projection valid
  area (common with geostationary), so the reader read the granule at full
  resolution. Now bad points are skipped, and there is a last ditch fallback before
  giving up.
* [GEOT-7946](https://osgeo-org.atlassian.net/browse/GEOT-7946) The
  geostationary valid area cache built a WKT string and ran two regexes on every
  lookup. Uses the CRS itself as key now.
* [GEOT-7947](https://osgeo-org.atlassian.net/browse/GEOT-7947) Every coverage
  processor operation lookup took a lock. Reads are lock-free now.
* [GEOT-7948](https://osgeo-org.atlassian.net/browse/GEOT-7948) A mosaic on top
  of a crop with nodata/roi copies all the pixels twice (because the Crop is implemented as a second mosaic).   
  When it is safe, the two are merged into one. Disable with `-Dorg.geotools.image.reduceMosaicCrop=false`.

# Checklist

### DIFF
--- a/modules/library/coverage/src/main/java/org/geotools/image/ImageWorker.java
+++ b/modules/library/coverage/src/main/java/org/geotools/image/ImageWorker.java
@@ -49,6 +49,7 @@ import java.util.List;
 import java.util.Locale;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.stream.DoubleStream;
 import javax.imageio.IIOException;
 import javax.imageio.IIOImage;
 import javax.imageio.ImageIO;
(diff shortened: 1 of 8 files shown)


modules/library/coverage/src/main/java/org/geotools/image/ImageWorker.java
```diff
@@ -4381,6 +4410,68 @@ public ImageWorker mosaic(
...
 
+    /**
+     * The nodata range of a single source operation that crops/mosaic by filling with nodata, {@code null} when it does
+     * anything more, or when the fill is not a {@link #harmlessFill harmless} one. This "crop" is a mosaic: either a
+     * Mosaic operation, which is what {@link org.geotools.coverage.processing.operation.Crop} builds when nodata or a
+     * ROI are involved, or a Crop operation, which image processing renders as a mosaic too once it sees the nodata
+     * range. A Crop without nodata is instead a pure bounds change, its tiles are the source ones, so there is nothing
+     * to fuse and this method returns {@code null} for it.
+     */
+    private static Range singleSourceNoData(RenderedOp op, double[] background) {
```
dromagnoli, line 4425: Not sure I have fully understood this part and the harmless fill meaning.
Could you please describe it with an example?

dromagnoli (approved):
Looks good to me.
I have asked for a clarification about the singleSourceNodata method and the related harmlessFill

modules/library/coverage/src/main/java/org/geotools/image/ImageWorker.java
aaime: I've updated the javadoc with two examples, so that it stays there for future readers too. Is it any clearer now?

PR #5825, opened 2026-08-11
[GEOT-7955]: Merging adjacent zones before producing the BETWEEN filter

dromagnoli:
[GEOT-7955](https://osgeo-org.atlassian.net/browse/GEOT-7955) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)  

  

# Checklist

### DIFF
--- a/modules/unsupported/dggs/dggs-h3/src/main/java/org/geotools/dggs/h3/H3DGGSInstance.java
+++ b/modules/unsupported/dggs/dggs-h3/src/main/java/org/geotools/dggs/h3/H3DGGSInstance.java
@@ -347,8 +347,60 @@ public class H3DGGSInstance implements DGGSInstance<Long> {
     @Override
     public Filter getChildFilter(
             FilterFactory ff, Long zoneId, int resolution, boolean upTo, AttributeDescriptor zoneAttribute) {
+        // upTo is not honored: the resolution bits outrank the digit ones, so a range only ever covers
+        // the target resolution. No caller passes upTo = true today.
         H3Index idx = new H3Index(zoneId);
         return buildChildRangeFilter(ff, zoneAttribute, idx, resolution);
+    }
+
+    @Override
+    public Filter getChildrenFilter(
+            FilterFactory ff, List<Long> zoneIds, int resolution, boolean upTo, AttributeDescriptor zoneAttribute) {
+        // each zone's range is the full contiguous block of its subtree id space, so blocks that touch
+        // can be merged without matching any id that was not already matched
+        List<long[]> ranges = new ArrayList<>();
+        for (Long zoneId : zoneIds) {
+            H3Index idx = new H3Index(zoneId);
+            ranges.add(new long[] {idx.lowestIdChild(resolution), idx.highestIdChild(resolution)});
+        }
+        List<long[]> merged = mergeRanges(ranges);
+
+        Expression prop = ff.property(zoneAttribute.getLocalName());
+        boolean numeric = Number.class.isAssignableFrom(zoneAttribute.getType().getBinding());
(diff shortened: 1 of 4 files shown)


modules/unsupported/dggs/dggs-h3/src/main/java/org/geotools/dggs/h3/H3DGGSInstance.java
```diff
@@ -351,6 +351,50 @@ public Filter getChildFilter(
         return buildChildRangeFilter(ff, zoneAttribute, idx, resolution);
     }
 
+    @Override
+    public Filter getChildrenFilter(
+            FilterFactory ff, List<Long> zoneIds, int resolution, boolean upTo, AttributeDescriptor zoneAttribute) {
+        Class<?> binding = zoneAttribute.getType().getBinding();
+        // range merge relies on numeric ordering matching id ordering, not true for the hex string encoding
```
aaime: For H3 with hex encoding this is, I believe, not actually true because they are always encoded in 15 lowercase chars, so they would compare lexicographically just like as with numbers. I'd recommend to double check with the H3 javadocs/code and if it is confirmed to be true, get the speed improvement on strings too.

The code at the end could look like this:

```java
  List<long[]> merged = mergeRanges(ranges);
  Expression prop = ff.property(zoneAttribute.getLocalName());
  boolean numeric = Number.class.isAssignableFrom(binding);
  List<Filter> filters = merged.stream()
          .map(r -> numeric
                  ? ff.between(prop, ff.literal(r[0]), ff.literal(r[1]))
                  : ff.between(prop, ff.literal(h3.h3ToString(r[0])), ff.literal(h3.h3ToString(r[1]))))
          .collect(Collectors.toList());
```

modules/unsupported/dggs/dggs-h3/src/main/java/org/geotools/dggs/h3/H3DGGSInstance.java
```diff
@@ -351,6 +351,50 @@ public Filter getChildFilter(
...
+        if (!Number.class.isAssignableFrom(binding)) {
+            return DGGSInstance.super.getChildrenFilter(ff, zoneIds, resolution, upTo, zoneAttribute);
+        }
+
+        // one range per zone, computed from that zone alone (not the whole batch's span), so trailing
+        // unused digits stay filled with each zone's own sentinel bits and the range stays as tight
+        // as a single getChildFilter call would produce
```
aaime: I find this one difficult to grok... maybe:

```
  // each zone's range is the full contiguous block of its subtree ID space, so blocks that touch
  // can be merged without matching any id that was not already matched
```

modules/unsupported/dggs/dggs-h3/src/main/java/org/geotools/dggs/h3/H3DGGSInstance.java
```diff
@@ -351,6 +351,50 @@ public Filter getChildFilter(
         return buildChildRangeFilter(ff, zoneAttribute, idx, resolution);
     }
 
+    @Override
+    public Filter getChildrenFilter(
+            FilterFactory ff, List<Long> zoneIds, int resolution, boolean upTo, AttributeDescriptor zoneAttribute) {
```
aaime, line 358: upTo is ignored... maybe if upTo is true, then call the base class method

aaime (changes requested):
Overall looks good, see comments below.

dromagnoli:
Feedbacks addressed.

aaime:
Looks good to me, but before merging please address the QA failure (looks legit)

PR #5835, opened 2026-08-19
[GEOT-7954] - interoperability issue with gt-geojson-core 35.0 and Ja…

roarbra:
[GEOT-7954](https://osgeo-org.atlassian.net/browse/GEOT-7954) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)  

A response to a reported failure at GeoTools mailing list from Bas Duineveld.

# Checklist

### DIFF
--- a/modules/unsupported/geojson-core/src/main/java/org/geotools/data/geojson/ObjectMapperFactory.java
+++ b/modules/unsupported/geojson-core/src/main/java/org/geotools/data/geojson/ObjectMapperFactory.java
@@ -23,7 +23,7 @@ import tools.jackson.databind.json.JsonMapper;
 /**
  * Support class providing {@link tools.jackson.databind.ObjectMapper} instances configured with the desired settings.
  */
-class ObjectMapperFactory {
+public class ObjectMapperFactory {
 
     private static final ObjectMapper DEFAULT_MAPPER;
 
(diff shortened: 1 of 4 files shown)


modules/unsupported/geojson-core/src/main/java/org/geotools/data/geojson/ObjectMapperFactory.java
```diff
@@ -23,7 +23,7 @@
 /**
  * Support class providing {@link tools.jackson.databind.ObjectMapper} instances configured with the desired settings.
  */
-class ObjectMapperFactory {
+public class ObjectMapperFactory {
```
aaime, line 26: Why was this class made public?

aaime (commented):
Overall looks good, the module is not using autodiscovery anyways. But I have a question below.

modules/unsupported/geojson-core/src/main/java/org/geotools/data/geojson/ObjectMapperFactory.java
roarbra: I have seen it before when I was working with Json and wanted an easy way of combining JtsModule with ObjectMapper. There might be better ways of doing this, but this would be the fastest.

PR #5878, opened 2026-09-02
[GEOT-7963]: Measured geometry types faulty GML output

dromagnoli:
[GEOT-7963](https://osgeo-org.atlassian.net/browse/GEOT-7963) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)  

There is also a GS PR depending on this:
https://github.com/geoserver/geoserver/pull/9917

  

# Checklist

### DIFF
--- a/modules/extension/xsd/xsd-gml2/src/main/java/org/geotools/gml2/bindings/GML2EncodingUtils.java
+++ b/modules/extension/xsd/xsd-gml2/src/main/java/org/geotools/gml2/bindings/GML2EncodingUtils.java
@@ -31,6 +31,7 @@ import org.geotools.api.feature.type.FeatureType;
 import org.geotools.api.metadata.Identifier;
 import org.geotools.api.referencing.ReferenceIdentifier;
 import org.geotools.api.referencing.crs.CoordinateReferenceSystem;
+import org.geotools.geometry.jts.CurvedGeometry;
 import org.geotools.geometry.jts.JTS;
 import org.geotools.geometry.jts.ReferencedEnvelope;
 import org.geotools.geometry.jts.coordinatesequence.CoordinateSequences;
@@ -44,6 +45,7 @@ import org.geotools.util.factory.Hints;
 import org.geotools.util.logging.Logging;
 import org.geotools.xsd.Configuration;
 import org.geotools.xsd.SchemaIndex;
+import org.locationtech.jts.geom.CoordinateSequence;
 import org.locationtech.jts.geom.Geometry;
 import org.w3c.dom.Document;
 import org.w3c.dom.Element;
@@ -281,11 +283,36 @@ public class GML2EncodingUtils {
             return null;
         }
 
-        /**
-         * For the dimension, use the actual dimension of the geometry. Using the dimension of the CRS is not
-         * sufficient, since currently CRSes don't support 3D.
-         */
-        return CoordinateSequences.coordinateDimension(geometry);
+        return getCoordinateDimension(geometry, config);
+    }
+
+    /**
+     * Number of ordinates written per coordinate: the natural dimension of the geometry, minus its measures when
+     * measure encoding is off. The declaration and every encoder path derive their count from here.
+     */
+    public static int getCoordinateDimension(Geometry geometry, Configuration config) {
+        // the dimension comes from the data, not from the CRS: 3D CRSs do exist, but 2+1D data is commonly
+        // published against a 2D one, so the CRS would under report
+        if (geometry instanceof CurvedGeometry) {
+            // curves are 2D by construction, CircularString rejects anything else, and they keep control points
+            // rather than a sequence
+            return CoordinateSequences.coordinateDimension(geometry);
+        }
+        CoordinateSequence sequence = CoordinateSequences.firstSequence(geometry);
+        if (sequence == null) {
+            return CoordinateSequences.coordinateDimension(geometry);
+        }
+        return getCoordinateDimension(sequence, encodeMeasures(config));
+    }
+
+    /** Same count as {@link #getCoordinateDimension(Geometry, Configuration)}, for a single sequence. */
+    public static int getCoordinateDimension(CoordinateSequence sequence, boolean encodeMeasures) {
+        return CoordinateSequences.ordinateIndices(sequence, encodeMeasures).length;
+    }
+
+    /** Whether coordinate measures should be encoded; off when no configuration is available, matching the default. */
+    public static boolean encodeMeasures(Configuration config) {
+        return GMLConfiguration.encodeMeasures(config);
     }
 
     public static Integer getEnvelopeDimension(ReferencedEnvelope e, Configuration configuration) {
--- a/modules/extension/xsd/xsd-gml3/src/main/java/org/geotools/gml3/bindings/DirectPositionListTypeBinding.java
+++ b/modules/extension/xsd/xsd-gml3/src/main/java/org/geotools/gml3/bindings/DirectPositionListTypeBinding.java
@@ -25,8 +25,10 @@ import org.geotools.geometry.Position2D;
 import org.geotools.geometry.Position3D;
 import org.geotools.geometry.jts.coordinatesequence.CoordinateSequences;
 import org.geotools.gml.producer.CoordinateFormatter;
+import org.geotools.gml2.bindings.GML2EncodingUtils;
 import org.geotools.gml3.GML;
 import org.geotools.xsd.AbstractComplexBinding;
+import org.geotools.xsd.Configuration;
 import org.geotools.xsd.ElementInstance;
 import org.geotools.xsd.Node;
 import org.locationtech.jts.geom.CoordinateSequence;
@@ -65,6 +67,13 @@ import org.w3c.dom.Element;
 public class DirectPositionListTypeBinding extends AbstractComplexBinding {
 
     CoordinateFormatter formatter;
+
+    Configuration config;
+
+    public DirectPositionListTypeBinding(CoordinateFormatter formatter, Configuration config) {
+        this.formatter = formatter;
+        this.config = config;
+    }
 
     public DirectPositionListTypeBinding(CoordinateFormatter formatter) {
         this.formatter = formatter;
@@ -171,12 +180,12 @@ public class DirectPositionListTypeBinding extends AbstractComplexBinding {
         CoordinateSequence cs = (CoordinateSequence) object;
         StringBuffer sb = new StringBuffer();
 
-        int dim = CoordinateSequences.coordinateDimension(cs);
+        int[] ordinates = CoordinateSequences.ordinateIndices(cs, GML2EncodingUtils.encodeMeasures(config));
         int size = cs.size();
-        int nOrdWithSpace = size * dim - 1;
+        int nOrdWithSpace = size * ordinates.length - 1;
         int count = 0;
         for (int i = 0; i < size; i++) {
-            for (int d = 0; d < dim; d++) {
+            for (int d : ordinates) {
                 double ordinate = cs.getOrdinate(i, d);
                 if (formatter != null) {
                     formatter.format(ordinate, sb);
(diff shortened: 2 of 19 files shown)


modules/extension/xsd/xsd-gml2/src/main/java/org/geotools/gml2/bindings/GML2EncodingUtils.java
```diff
@@ -281,11 +283,62 @@ public static Integer getGeometryDimension(Geometry geometry, Configuration conf
...
+    }
+
+    /**
+     * Number of ordinates written per coordinate: the natural dimension of the geometry, minus its measures when
+     * measure encoding is off. The declaration and every encoder path derive their count from here.
+     */
+    public static int getCoordinateDimension(Geometry geometry, Configuration config) {
+        // the geometry dimension is used rather than the CRS one, since CRSes don't support 3D yet
+        // a curved geometry keeps control points, not a sequence: it states its own dimension and reading a
+        // sequence out of it would linearize it first
```
aaime: The two statements are following toghether, they look like a single sentence while they are not. Also, it's not true that CRS cannot represent 3D, we have 3D CRSs, but often times in practice a 2D CRS is used for 2+1D data.

modules/extension/xsd/xsd-gml2/src/main/java/org/geotools/gml2/bindings/GML2EncodingUtils.java
```diff
@@ -281,11 +283,62 @@ public static Integer getGeometryDimension(Geometry geometry, Configuration conf
...
+        CoordinateSequence sequence =
+                geometry instanceof CurvedGeometry ? null : CoordinateSequences.firstSequence(geometry);
+        if (sequence == null) {
```
aaime, line 302: I'm lost here, so if the geometry is curved we provide no Z nor M support? But I had to read the code to understand it, the comment above just states curves handle their own dimensions, a different idea.

modules/extension/xsd/xsd-gml3/src/main/java/org/geotools/gml3/bindings/DirectPositionTypeBinding.java
```diff
@@ -130,25 +140,29 @@ public Object parse(ElementInstance instance, Node node, Object value) throws Ex
...
+                if (!Double.isFinite(v) && d > 1) {
                     continue;
                 }
-
                 formatAndAppend(sb, v);
+                written++;
+            }
+            // the count decides how many measures follow, none of them when measures are not encoded
+            for (int d = firstMeasure; d < cs.getDimension() && written < ordinates; d++) {
+                formatAndAppend(sb, cs.getOrdinate(0, d));
```
aaime, line 150: What if there is a NaN in the measures too? (drawing a parallel with the ordinate handling above)

modules/library/xml/src/main/java/org/geotools/gml/producer/CoordinateWriter.java
```diff
@@ -259,13 +272,29 @@ public void writeCoordinates(CoordinateSequence c, ContentHandler output) throws
...
+                }
+            } else {
+                boolean zAvailable = coordSeqDimension > 2
+                        && !Double.isNaN(c.getOrdinate(i, 2))
+                        && !Double.isInfinite(c.getOrdinate(i, 2));
+                if (D == 3 || zAvailable || useDummyZ) {
+                    z = zAvailable ? c.getOrdinate(i, 2) : dummyZ;
+                    coordBuff.append(coordinateDelimiter);
+                    coordFormatter.format(z, coordBuff);
+                }
```
aaime: Given the code repetition I would use a single sequential code with a couple of ifs to handle measures

modules/extension/xsd/xsd-gml3/src/main/java/org/geotools/gml3/bindings/DirectPositionListTypeBinding.java
aaime, line 187: We have a problem here... getCoordianteDimensions above checks if Z is all NaN and then reports 2 (or 3 if there is one dimension), but this loop never checks if the Z needs to be skipped, and will end up encoding NaN for the Z and skip the M entirely.

The issue happens in the other coordinate writers too, not just here.

A way to address it might be to get a list of indices to use from GML2EncodingUtils:

```java
/** Ordinate indices to write, in order: X, Y, the height when the sequence fills one, then the measures. */
  public static int[] ordinateIndices(CoordinateSequence sequence, boolean encodeMeasures) {
```

and then in DirectPositionListTypeBinding.encode:

```java
  int[] ordinates = GML2EncodingUtils.ordinateIndices(cs, GML2EncodingUtils.encodeMeasures(config));
  int nOrdWithSpace = size * ordinates.length - 1;
  for (int i = 0; i < size; i++) {
      for (int d : ordinates) {
          double ordinate = cs.getOrdinate(i, d);
          ...
```

    int[] ordinates = GML2EncodingUtils.ordinateIndices(cs, GML2EncodingUtils.encodeMeasures(config));
  int nOrdWithSpace = size * ordinates.length - 1;
  for (int i = 0; i < size; i++) {
      for (int d : ordinates) {
          double ordinate = cs.getOrdinate(i, d);
          ...

GMLWriter.coordinates takes the same treatment, and DirectPositionTypeBinding.encode collapses onto it too.

modules/extension/xsd/xsd-gml3/src/test/java/org/geotools/gml3/simple/MeasuredGeometryDimensionTest.java
```diff
@@ -0,0 +1,223 @@
...
+        LineStringEncoder encoder = new LineStringEncoder(gtEncoder, "gml", namespace);
+        Document dom = encode(encoder, line, encodeMeasures, "line", 6, false, false, namespace);
+        String posList = dom.getElementsByTagNameNS(namespace, "posList")
+                .item(0)
+                .getTextContent()
+                .trim();
+        assertEquals(expectedPosList, posList);
+
+        setEncodeMeasures(configuration, encodeMeasures);
+        Integer declared = GML2EncodingUtils.getGeometryDimension(line, configuration);
```
aaime: Rather than testing the utility directly, make sure that the emitted GML contains the right dimension, if possible?

modules/extension/xsd/xsd-gml3/src/test/java/org/geotools/gml3/simple/MeasuredGeometryDimensionTest.java
```diff
@@ -0,0 +1,223 @@
...
+        } else {
+            ((org.geotools.gml3.v3_2.GMLConfiguration) configuration).setEncodeMeasures(encodeMeasures);
+        }
+    }
+
+    /** The bug being guarded: the declared dimension has to be the ordinates written per coordinate. */
+    private void assertDeclarationMatches(String srsDimension, String posList, int coordinates) {
+        int written = posList.split("\\s+").length;
+        int declared = Integer.parseInt(srsDimension);
+        assertEquals("declared dimension differs from the ordinates written", declared * coordinates, written);
```
aaime: Seems a bit weak... we could have the right declared * coordiantes value by accident (less coordinates and more ordinates for example)

modules/library/main/src/main/java/org/geotools/geometry/jts/coordinatesequence/CoordinateSequences.java
```diff
@@ -138,6 +138,14 @@ public static int computeOrientation(CoordinateSequence cs, int p1, int p2, int
      * @param g a Geometry
      * @return the dimension of the coordinates in the Geometry
      */
+    /**
```
aaime, line 134: Two javadocs in a row here, something got misplaced.

aaime (changes requested):
Overall looks good, comments attached per line.

One thing that it's hard to point to, because it's a bit distributed:

The new flag now lives in a Configuration property, and the PR reads it two ways that do not answer the same question.

```java
  // gml3.GMLConfiguration.getEncodeMeasures() — this configuration's own property set
  return getProperties().contains(ENCODE_MEASURES);

  // GML2EncodingUtils.encodeMeasures(config) — the whole dependency tree
  return config != null && config.hasProperty(GMLConfiguration.ENCODE_MEASURES);
```

The difference is that ``hasProperty`` walks ``allDependencies()``, while getProperties().contains does not. Given how GeoServer sets up the configuration, it does not really matter, but a direct GeoTools user might be caught up in the difference. I would recomment to centralize the lookup and use a single uniform approach instead.

Also, gml2 GMLconfiguration did not get a getter/setter for the measures, like GML3Configuration does. One can access the properties and put it in there, but it's inconsistent with the pre-existing config classes. Add the getter/setter and use it in calling points instead of direct access to the properties.

Final note, NaN handling is done twice in incosistent ways:

```java
  // CoordinateWriter.java:257
  // same rule the GML3 encoders apply, duplicated because gt-xml sits below gt-xsd-gml2: a shapefile
  // PolyLineZ declares a measure and leaves it NaN, so an unfilled M does not count as one
  final int measures = c.getMeasures() > 0 && coordCount > 0 && !Double.isNaN(c.getM(0)) ? c.getMeasures() : 0;

  // GML2EncodingUtils.carriedMeasures
  if (sequence.getMeasures() == 0 || sequence.size() == 0 || Double.isNaN(sequence.getM(0))) {
      return 0;
  }
  return sequence.getMeasures();
```

Best to centralize for consistency and use in all calling points

modules/extension/xsd/xsd-gml2/src/main/java/org/geotools/gml2/bindings/GML2EncodingUtils.java
dromagnoli: curves are 2D by construction, CircularString rejects anything else.
I'm Improving the comment.

modules/extension/xsd/xsd-gml2/src/main/java/org/geotools/gml2/bindings/GML2EncodingUtils.java
dromagnoli: Fixed

modules/library/xml/src/main/java/org/geotools/gml/producer/CoordinateWriter.java
dromagnoli: Fixed

modules/extension/xsd/xsd-gml3/src/main/java/org/geotools/gml3/bindings/DirectPositionListTypeBinding.java
dromagnoli: Good catch, fixed as suggested. 
ordinateIndices (moved in gt-main, relying on CoordinateSequence) now returns the indices to write and all the writers iterate it, so the count and the values come from the same place and cannot disagree.

CoordinateWriter is converted too, except the height, which stays decided per coordinate: the index list is built from the first coordinate alone, so driving the height off it would drop real Z values from every later coordinate  of a sequence whose first Z is NaN. That is not a new choice, the check was already per coordinate before (!isNaN && !isInfinite).

modules/library/main/src/main/java/org/geotools/geometry/jts/coordinatesequence/CoordinateSequences.java
dromagnoli: Fixed

modules/extension/xsd/xsd-gml3/src/test/java/org/geotools/gml3/simple/MeasuredGeometryDimensionTest.java
dromagnoli: Fixed

modules/extension/xsd/xsd-gml3/src/main/java/org/geotools/gml3/bindings/DirectPositionTypeBinding.java
dromagnoli: Covered in the javadoc.

dromagnoli:
Requested changes applied and feedbacks addressed.

aaime:
Looks good to me, merging.

PR #5895, opened 2026-09-14
[GEOT-7968] MultithreadedHttpClient "Connect timeout" and "Read timeout" neglected

roarbra:
[GEOT-7968](https://osgeo-org.atlassian.net/browse/GEOT-7968) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)  

Turned it into a lazy initialization of the PoolingHttpClientConnectionManager and HttpClient.

The testConnectTimeout might not work in certain networks, but it shouldn't fail. It just won't check if the connect timeout works.
 
The Timeout-attributes is extra precaution.

# Checklist


mprins (approved):
lgtm

modules/plugin/http-commons/src/main/java/org/geotools/http/commons/MultithreadedHttpClient.java
```diff
@@ -85,36 +85,36 @@ public class MultithreadedHttpClient extends AbstractHttpClient implements HTTPC
 
     private static final Logger LOGGER = Logging.getLogger(MultithreadedHttpClient.class);
 
-    private final PoolingHttpClientConnectionManager connectionManager;
+    private PoolingHttpClientConnectionManager connectionManager = null;
 
-    private HttpClient client;
+    private boolean renewManager = true;
```
aaime: The whole approach here is not thread-safe... I do recognize that GeoTools own HTTP client does not promise to be thread-safe, but this class is called MultithreadedHttpClient, which hints at it being fine under concurrent load. But see below, this whole thing might not be required to start with.

modules/plugin/http-commons/src/main/java/org/geotools/http/commons/MultithreadedHttpClient.java
```diff
@@ -294,6 +307,7 @@ public void setConnectTimeout(int connectTimeout) {
         connectionConfig = ConnectionConfig.copy(connectionConfig)
                 .setConnectTimeout(Timeout.ofSeconds(connectTimeout))
                 .build();
+        renewManager = true;
```
aaime: The connection manager can actually take a new config after being built, it should be safe and less complex:

```java
  @Override
  public void setConnectTimeout(int connectTimeout) {
      connectionConfig = ConnectionConfig.copy(connectionConfig)
              .setConnectTimeout(Timeout.ofSeconds(connectTimeout))
              .build();
      connectionManager.setDefaultConnectionConfig(connectionConfig);
  }
```

I've tried it out and the test pass

modules/plugin/http-commons/src/main/java/org/geotools/http/commons/MultithreadedHttpClient.java
```diff
@@ -306,22 +320,25 @@ public void setReadTimeout(int readTimeout) {
...
     public int getMaxConnections() {
-        return connectionManager.getDefaultMaxPerRoute();
+        return maxConnections;
     }
 
     @Override
     public void setMaxConnections(final int maxConnections) {
-        connectionManager.setDefaultMaxPerRoute(maxConnections);
-        connectionManager.setMaxTotal(maxConnections);
+        this.maxConnections = maxConnections;
```
aaime: The same could be done here, just rebuild the config and set it.

modules/plugin/http-commons/src/test/java/org/geotools/http/commons/MultithreadedHttpClientTest.java
```diff
@@ -330,4 +330,45 @@ public void testGzipEncodedResponse() throws IOException {
             }
         }
     }
+
+    /** Verifies that a response arriving slower than the configured readTimeout fails fast instead of hanging. */
+    @Test(timeout = 5000)
+    public void testReadTimeout() {
```
aaime, line 336: This test seems to pass even without the fixes, can you double check?

Also, I'm worried about the timeout=5000 when running on github actions... they can be sitting there for 5 seconds doing nothing, because the CPUs have been allocated to something else. 
I would increase the timeout to 60 seconds, assuming the test pass quickly under normal conditions

modules/plugin/http-commons/src/test/java/org/geotools/http/commons/MultithreadedHttpClientTest.java
```diff
@@ -330,4 +330,45 @@ public void testGzipEncodedResponse() throws IOException {
...
+                "Expected read timeout to fire around the configured 1s, but took " + elapsed + "ms", elapsed < 2000);
+    }
+
+    /**
+     * Verifies that a connection attempt taking longer than the configured connectTimeout fails fast instead of
+     * hanging. WireMock cannot delay the TCP handshake itself, so a non-routable address (RFC 5737 TEST-NET-1) is used
+     * to simulate a connection that never completes.
+     */
+    // Guards against the test hanging the build if connectTimeout isn't honored in some environment.
+    @Test(timeout = 5000)
```
aaime: Same timeout worry as above.

aaime (changes requested):

modules/plugin/http-commons/src/test/java/org/geotools/http/commons/MultithreadedHttpClientTest.java
roarbra: The Issue-title is misleading. It was the connect timeout that was the problem. Still a test for read timeout was missing, so I added that as well.

modules/plugin/http-commons/src/main/java/org/geotools/http/commons/MultithreadedHttpClient.java
roarbra: I'm aware about the problem with thread-safety. I've rewritten to avoid recreating the connectionManager. Still it should perhaps been said in the documentation that the properties should be set before any call, and that the credentials will be used for all hosts.

PR #5906, opened 2026-09-18
[GEOT-7970] Clean up loose exceptions in CQL

ianturton:
[GEOT-7970](https://osgeo-org.atlassian.net/browse/GEOT-7970) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)  

  

# Checklist

CQL can throw unwrapped exceptions that can cause issues for downstream users

+  0CQL-001 — a raw java.lang.Error ("Invalid escape character") escapes uncaught from JavaCharStream.readChar on a malformed \u escape (e.g. \uB!) in a CQL_FILTER value.
+   CQL-002 — NumberFormatException in AbstractFilterBuilder.buildLiteralInteger on an oversized integer literal in CQL_FILTER.
+ CQL-003 — ClassCastException in AbstractFilterBuilder.getFilterList when an expression appears where a filter is expected (e.g. floor(1)).
+ CQL-004 — EmptyStackException in BuildResultStack.popResult (ecql PointBuilder) on malformed geometry text.


modules/library/cql/src/main/java/org/geotools/filter/text/cql2/CQLCompiler.java
```diff
@@ -78,10 +78,8 @@ public void compileFilter() throws CQLException {
             super.FilterCompilationUnit();
         } catch (TokenMgrError tokenError) {
             throw new CQLException(tokenError.getMessage(), getTokenInPosition(0), this.source);
-        } catch (CQLException e) {
-            throw e;
-        } catch (ParseException e) {
-            throw new CQLException(e.getMessage(), getTokenInPosition(0), e.getCause(), this.source);
+        } catch (java.lang.Error | ParseException er) {
```
aaime, line 83: CQLException is already a ParseException, rebuilding it might lose information.
Catching and wrapping Error is normally problematic, as Error is thrown by the JVM under severe issues like an OOM. What was the issue that motivated it? Stack overflow?

Something like this would be preferable (I did not include a part about handling Error as that needs clarification. 

```
          } catch (CQLException e) {
              throw e;
          } catch (ParseException e) {
              throw new CQLException(e.getMessage(), getTokenInPosition(0), e.getCause(), this.source);
          } 
```

modules/library/cql/src/main/java/org/geotools/filter/text/cql2/CQLCompiler.java
```diff
@@ -78,10 +78,8 @@ public void compileFilter() throws CQLException {
...
+            throw new CQLException(er.getMessage(), getTokenInPosition(0), er.getCause(), this.source);
```
aaime: Careful about this rebuilding... the CQLException constructor has asserts inside the code:

```
    public CQLException(final String message, final IToken token, final Throwable cause, final String cqlSource) {
        super(message);

        assert message != null : "message can not be null";
        assert cqlSource != null : "cqlSource can not be null";

        this.currentToken = token;
        this.cause = cause;
        this.cqlSource = cqlSource;
    }
```

Now all exceptions have a message; for example, StackOverflowError often does not.
Now, it's true that an assert triggers only when asserts are enabled, but still...

modules/library/cql/src/test/java/org/geotools/filter/text/ecql/ECQLTest.java
```diff
@@ -367,6 +367,21 @@ public void filterListToECQLWithRefencedGeometries() throws Exception {
...
+    @Test
+    public void unexpectedFunction() throws CQLException {
+        String expectedECQL =
+                "INTERSECTS(the_geom, SRID=4326;POINT (1 2)); FLOOR(1); INTERSECTS(abcd, SRID=4962;POINT (0 0))";
+
+        Exception exception = Assert.assertThrows(CQLException.class, () -> {
+            ECQL.toFilterList(expectedECQL);
+        });
+
+        String expected =
```
aaime: The format of this message could change based on the JDK, there is no strict contract here. 
This might be a viable alternative; check the cause and check something in the message that is unlikely to change in the future:

```java
  CQLException e = assertThrows(CQLException.class, () -> ECQL.toFilterList(ecql));
  assertThat(e.getMessage(), containsString("floor"));
  assertThat(e.getCause(), instanceOf(ClassCastException.class));
```

aaime (commented):
Overal looks good, see feedback below for some feedback on catch/rebuild of CQLException

modules/library/cql/src/main/java/org/geotools/filter/text/cql2/CQLCompiler.java
ianturton: A raw java.lang.Error ("Invalid escape character") escapes uncaught from JavaCharStream.readChar on a malformed \u escape (e.g. \uB!) in a CQL_FILTER value.

Makes sense not to remake the CQLexception

ianturton:
@aaime - any objection to merging this with out a windows CI pass, I can't even get it to fail in the same place twice but I'm confident the fails have nothing to do with this PR

PR #5908, opened 2026-09-18
[GEOT-7319] Change flow of ComplexFeatureIteratorImpl

roarbra:
[GEOT-7319](https://osgeo-org.atlassian.net/browse/GEOT-7319) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geotools&utm_campaign=badge_info)  

The fix have been ready for a long time. Got help from AI with the unit test.

  

# Checklist


modules/unsupported/wfs-ng/src/test/java/org/geotools/data/wfs/internal/ComplexFeatureIteratorImplTest.java
```diff
@@ -0,0 +1,130 @@
...
+
+            verify(parser, times(1)).close();
+            assertFalse(iterator.hasNext());
+        }
+    }
+
+    @Test
+    public void close_whenParserThrowsIOException_doesNotPropagate() throws IOException {
+        XmlComplexFeatureParser parser = mock(XmlComplexFeatureParser.class);
+        org.mockito.Mockito.doThrow(new IOException("boom")).when(parser).close();
```
aaime: Please avoid fully qualified references; prefer imports.

modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/internal/ComplexFeatureIteratorImpl.java
```diff
@@ -49,36 +49,33 @@ public class ComplexFeatureIteratorImpl implements FeatureIterator<Feature> {
...
+        }
+        Feature next = parsedFeature;
+        parseNext();
+        return next;
+    }
+
+    private void parseNext() {
+        try {
+            parsedFeature = parser.parse();
+        } catch (IOException e) {
```
aaime, line 76: Not a new error, but still problematic... if the parse throws an IOException, to the caller it looks like a legit iteration, just shorter than normal.
Mind, if we throw, at least the parser should be closed here.

modules/unsupported/wfs-ng/src/test/java/org/geotools/data/wfs/internal/ComplexFeatureIteratorImplTest.java
```diff
@@ -0,0 +1,130 @@
...
+import static org.mockito.Mockito.when;
+
+import java.io.IOException;
+import java.util.NoSuchElementException;
+import org.geotools.api.feature.Feature;
+import org.geotools.data.wfs.internal.parsers.XmlComplexFeatureParser;
+import org.junit.Test;
+
+/** Test class for {@link ComplexFeatureIteratorImpl}. */
+public class ComplexFeatureIteratorImplTest {
```
aaime, line 37: Test here is legit but also very unit-testing oriented. It would be useful to also have an integration test that actually parses something, where hasNext is called twice in a row.

aaime (commented):
Looks good and the PR could go in as-is, but I've noticed something pre-existing that is worrysome, see below.

modules/unsupported/wfs-ng/src/main/java/org/geotools/data/wfs/internal/ComplexFeatureIteratorImpl.java
roarbra: Setting the parsedFeature=null means the iteration will stop, not just skipping a feature, but I will not mind throwing an exception. Removing the close was deliberately because I've been in a situation where the closing was confusing. Since it's a try-with-resources I recon the close will be called anyway.

modules/unsupported/wfs-ng/src/test/java/org/geotools/data/wfs/internal/ComplexFeatureIteratorImplTest.java
roarbra: I fixed the name.

modules/unsupported/wfs-ng/src/test/java/org/geotools/data/wfs/internal/ComplexFeatureIteratorImplTest.java
roarbra: I tried, but the EntityResolution makes it so hard to write integration tests with complex FeatureType.

